PR #1, opened 2015-12-30
Add very basic framework of classifier and bundler.

Osmose:
@mythmon This is a very basic pass at very simple bundler/classifier functionality. It's missing a lot of things. And NO TESTS WHATSOEVER.

But I figured it was enough to have you look at when you get back rather than just to commit blindly.


normandy/classifier/views.py
```diff
@@ -0,0 +1,52 @@
...
+    def __init__(self, recipes=None):
+        self.recipes = recipes or []
+
+    @property
+    def ids(self):
+        return [recipe.id for recipe in self.recipes]
+
+    @property
+    def hash(self):
+        sorted_recipes = sorted(self.recipes, key=lambda r: r.filename)
```
mythmon: The recipes should be sorted when they are sorted, instead of sorting them every time.

normandy/classifier/views.py
```diff
@@ -0,0 +1,52 @@
...
+    """A bundle of recipes to be sent to the client."""
+    def __init__(self, recipes=None):
+        self.recipes = recipes or []
+
+    @property
+    def ids(self):
+        return [recipe.id for recipe in self.recipes]
+
+    @property
+    def hash(self):
```
mythmon: Should this be the `__hash__` magic method instead?

normandy/bundler/views.py
```diff
@@ -0,0 +1,13 @@
+from django.http import HttpResponse, HttpResponseBadRequest
+
+from normandy.recipes.models import Recipe
+
+
+def bundle(request):
```
mythmon: If this was a CBV, it could have a `classmethod` like `Bundler.url_for(bundle)`, so you don't have to do things like `urlparams(reverse('normandy.bundler'), hash=bundle.hash, ids=bundle.ids)`. Or rather, so you can do that complex function call in one centralized place.

mythmon:
I think the idea of using classes for things like Bundle and Client seems like a nice way to namespace related functions. I think we should stick with it. We can also use them as convenient parameters to other functions in the future.

normandy/classifier/views.py
Osmose: Nope, because [__hash__ return values are ints and truncated to 4 or 8 bytes](https://docs.python.org/3/reference/datamodel.html#object.__hash__). SHA-1 returns a 20 byte string and we're going to shove the hash in a URL anyway, so converting it to an int and back to a hex string seems silly.

Osmose:
Updated based on feedback, and also added a LICENSE and readme.

normandy/base/utils.py
```diff
@@ -0,0 +1,23 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
```
mythmon: The MPLv2.0 says 

> If it is not possible or desirable to put the notice in a particular file, then You may include the notice in a location (such as a LICENSE file in a relevant directory) where a recipient would be likely to look for such a notice.
([MPLv2.0 Exhibit A](https://www.mozilla.org/en-US/MPL/2.0/#exhibit-a---source-code-form-license-notice))

Personally, I find it always undesirable to include these header blocks unless the individual file has a different license than the surrounding file. I'd vote to not include them.

Osmose: I usually do too, but Bedrock was specifically asked to add them, so I assumed it'd be a thing to do. But I'll remove them.

PR #2, opened 2016-01-08
Add docs and a rough draft of the final SHIELD protocol.

Osmose:
This contains a very early rough draft of the protocol we'd like to use between the SHIELD system add-on and Normandy. Since this protocol has implications on our admin UI and data models I figured now would be a good time to at least think about the shape of it, even if our first task is to replace self-repair as it is now.

@mythmon Feedback?


docs/dev/architecture.rst
```diff
@@ -0,0 +1,127 @@
+Architecture
+============
+Normandy is a component in a larger system called SHIELD. The end goal of SHIELD
+is to allow Firefox to perform a variety of actions to aid the user and gather
+feedback to guide us in developer a better browser for them. To do this, we need
```
mythmon: s/developer/developing/

docs/dev/architecture.rst
```diff
@@ -0,0 +1,127 @@
...
+
+Normandy
+   Service that acts as the source of actions that we wish to perform, and
+   determines which users receive which recipes.
+Self-Repair System Addon
+   Ships with Firefox and handles fetching recipes and action implementations
+   from Normandy and executing them within Firefox.
+
+.. note:: Normandy provides both the names of actions and, separately,
+   implmentations for them to be used in the self-repair system addon. Other
```
mythmon: s/implmentations/implementations/

docs/dev/architecture.rst
```diff
@@ -0,0 +1,127 @@
...
+
+   {
+      "recipes": [
+         {
+            "name": "heartbeat.australis.1",
+            "actions": [
+               {
+                  "name": "show_survey",
+                  "impl_hash": "4011cf569f3639b9069fc2c50117d3f8597e83c61be2014081ad06ee0fe427c2",
+                  "arguments": {
```
mythmon: Although not all clients will use the Normandy-provided implementations, would it be good to provide a URL to the API that provides the implementation here? Something like

```yaml
recipes: [
    {
        name: "heartbeat.australis.1"
        actions: [
            name: "show_survey"
            impl_hash: "4011cf569f3639b9069fc2c50117d3f8597e83c61be2014081ad06ee0fe427c2"
            url: "https://shield.mozilla.org/api/v1/actions/show_survey/"
        ]
    }
]
```

mythmon: Oh. I guess you have a different url for getting implementations. I'm not too picky about the url (though I think mine is better), my main point is to provide the url for where to download it in the response.

docs/dev/architecture.rst
```diff
@@ -0,0 +1,127 @@
...
+.. code-block:: json
+
+   {
+      "impl": "function action(selfRepairAPI, args) { /* Do something */ }"
+   }
+
+Legacy Self-Repair
+------------------
+Normandy also hosts an endpoint so that it can replace the predecessor
+Self-Repair server, which delivers a single JavaScript file for Firefox to
```
mythmon: The old Self-Repair server serves a single HTML file that contains inline JS. We will have to do the same. (We might be able to not make it inline, but I don't see what that would gain us.)

docs/dev/architecture.rst
```diff
@@ -0,0 +1,127 @@
...
+
+   <!DOCTYPE html>
+   <html lang="en">
+   <head>
+   <meta charset="utf-8">
+   </head>
+   <body>
+   <script>/* Do something */</script>
+   </body>
+   </html>
```
mythmon: Maybe indent this to make it read easier?

mythmon:
I ignored all the Sphinx stuff (config and scripts and what not). I read through the document, and it looks pretty good. A few typos/nits, and a recommendation, but that is all.

docs/dev/architecture.rst
Osmose: I like that. Perhaps this:

```js
recipes: [
    {
        name: "heartbeat.australis.1"
        actions: [
            name: "show_survey",
            selfRepairImpl: {
                hash: "4011cf569f3639b9069fc2c50117d3f8597e83c61be2014081ad06ee0fe427c2",
                url: "https://shield.mozilla.org/api/v1/actions/show_survey/"
            }
        ]
    }
]
```

mythmon: +1. Though the url you have here is a generic url (it wouldn't link to this specific action). Did you mean that? I think the url should be to this particular action.

Osmose: > Though the url you have here is a generic url (it wouldn't link to this specific action).

I don't understand. It links to the `show_survey` action, unless you mean it doesn't link to the version of the `show_survey` action with the given hash?

mythmon: Oh. Oh I'm just dumb. I was reading "show_action" as a api-action, ie something like "/api/do_the_thing". We're all good here.

PR #3, opened 2016-01-13
Replace recipe content with actions and add self-repair.

Osmose:
@mythmon r?


normandy/bundler/models.py
```diff
@@ -14,3 +16,14 @@ def ids(self):
     def hash(self):
         recipe_hashes = '_'.join([recipe.content_hash for recipe in self.recipes])
         return hashlib.sha1(recipe_hashes.encode('utf-8')).hexdigest()
+
+    def __iter__(self):
+        for recipe in self.recipes:
+            yield recipe
```
mythmon: I think this would be better as `yield from self.recipes`. (yay Python 3!) Or maybe `return iter(self.recipes)`.

normandy/bundler/models.py
```diff
@@ -14,3 +16,14 @@ def ids(self):
...
+
+    @classmethod
+    def for_client(cls, client):
```
mythmon: A minor quibble: I think it would be better to put constructor-a-like methods like this right after `__init__`, order-wise.

mythmon:
With this patch, the `/classify/` view gives me this:

```
Traceback (most recent call last):
  File "/home/mythmon/src/normandy/venv/lib/python3.5/site-packages/django/core/handlers/base.py", line 149, in get_response
    response = self.process_exception_by_middleware(e, request)
  File "/home/mythmon/src/normandy/venv/lib/python3.5/site-packages/django/core/handlers/base.py", line 147, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/mythmon/src/normandy/normandy/classifier/views.py", line 15, in classify
    return HttpResponseRedirect(BundlerView.url_for(bundle))
  File "/home/mythmon/src/normandy/normandy/bundler/views.py", line 29, in url_for
    return urlparams(reverse('normandy.bundler'), hash=bundle.hash, ids=bundle.ids)
  File "/home/mythmon/src/normandy/normandy/bundler/models.py", line 17, in hash
    recipe_hashes = '_'.join([recipe.content_hash for recipe in self.recipes])
  File "/home/mythmon/src/normandy/normandy/bundler/models.py", line 17, in <listcomp>
    recipe_hashes = '_'.join([recipe.content_hash for recipe in self.recipes])
AttributeError: 'Recipe' object has no attribute 'content_hash'
```

Osmose:
> With this patch, the /classify/ view gives me this:

This is expected, classify can be fixed in a future PR.

Otherwise I've fixed all the problems you pointed out.

mythmon:
GitHub won't let me merge this. And it isn't only because I merged #4, that happened before I merged it.

PR #6, opened 2016-01-15
Hash based sampling

mythmon:
@Osmose r?

### DIFF
--- a/normandy/bundler/models.py
+++ b/normandy/bundler/models.py
@@ -21,7 +21,10 @@ class Bundle(object):
 
     @property
     def hash(self):
-        recipe_hashes = '_'.join([recipe.content_hash for recipe in self.recipes])
+        hash_parts = []
+        for recipe in self.recipes:
+            hash_parts.extend(action.implementation_hash for action in recipe.actions.all())
+        recipe_hashes = '_'.join(hash_parts)
         return hashlib.sha1(recipe_hashes.encode('utf-8')).hexdigest()
 
     def __repr__(self):
--- a/normandy/classifier/tests/test_utils.py
+++ b/normandy/classifier/tests/test_utils.py
@@ -1,0 +1,16 @@
+from fractions import Fraction
+
+from normandy.classifier.utils import fraction_to_key
+
+
+def test_fraction_to_key_exact():
+    assert fraction_to_key(Fraction(0, 4)) == '0' * 64
+    assert fraction_to_key(Fraction(1, 4)) == '3' + 'f' * 63
+    assert fraction_to_key(Fraction(2, 4)) == '7' + 'f' * 63
+    assert fraction_to_key(Fraction(3, 4)) == 'b' + 'f' * 63
+    assert fraction_to_key(Fraction(4, 4)) == 'f' * 64
+
+
+def test_fraction_to_key_floats():
+    assert fraction_to_key(0.0) == '0' * 64
+    assert fraction_to_key(1.0) == 'f' * 64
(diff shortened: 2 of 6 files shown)


normandy/bundler/models.py
```diff
@@ -21,7 +21,10 @@ def ids(self):
 
     @property
     def hash(self):
-        recipe_hashes = '_'.join([recipe.content_hash for recipe in self.recipes])
+        hash_parts = []
+        for recipe in self.recipes:
+            hash_parts.extend(action.implementation_hash for action in recipe.actions.all())
```
Osmose: Maybe add a property to the ``Recipe`` class that returns a list or concatenated string of all the hashes for the actions in the recipe?

normandy/recipes/models.py
```diff
@@ -28,35 +31,54 @@ class Recipe(models.Model):
     def implementation(self):
         return ';'.join(action.implementation for action in self.actions.all())
 
+    def log_rejection(self, client=None, field=None, msg=None, op='!='):
```
Osmose: I think instead of having flexible optional arguments you should make this just take a message, and then have another method that wraps it for the special case of locale and country.

Also, `op` isn't used anywhere.

normandy/recipes/utils.py
```diff
@@ -0,0 +1,48 @@
+import hashlib
+
+
+def fraction_to_key(frac):
+    """Map from the range [0, 1] to [0, max(sha256)]. The result is a string."""
+    # SHA 256 hashes are 64 hexadecimal numbers. The largest possible SHA 256
+    # hash is 2^64 - 1
```
Osmose: You mean `2^256 - 1`, right?

normandy/recipes/utils.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def deterministic_sample(rate, inputs):
+    """
+    Deterministically choose True or False based for a set of inputs.
+
+    Internally, converts `rate` into a point in the sha256 hash space. If the
+    hash of `inputs` is less than that point, it returns True.
+
+    @parameter rate The probability of returning True
```
Osmose: What docstring style is this, again?

mythmon: Oops. This is a bastardization of Javadoc that I was using in JS. I'll switch this over to [http://www.sphinx-doc.org/en/stable/domains.html#info-field-lists](Sphinx style).

normandy/bundler/models.py
mythmon: This is all going away soon, I guess. This was mostly just getting it to stop erroring out. Probably better to leave it as is an make it better once we implement the newer API.

normandy/recipes/models.py
mythmon: I originally expected this to be used more, and then it wasn't, and I didn't go back and fix it. Now I did.

normandy/classifier/tests/test_utils.py
```diff
@@ -0,0 +1,35 @@
...
+        assert fraction_to_key(Fraction(1, 4)) == '3' + 'f' * 63
+        assert fraction_to_key(Fraction(2, 4)) == '7' + 'f' * 63
+        assert fraction_to_key(Fraction(3, 4)) == 'b' + 'f' * 63
+        assert fraction_to_key(Fraction(4, 4)) == 'f' * 64
+
+    def test_floats(self):
+        assert fraction_to_key(0.0) == '0' * 64
+        assert fraction_to_key(1.0) == 'f' * 64
+
+    @pytest.mark.parametrize('bad_val', [-1, -0.5, 1.5, 2])
```
Osmose: Faaaaaancy

normandy/bundler/models.py
Osmose: Oh yeah, we don't really need it anymore, forgot about that.

Osmose:
r+, nice work!

PR #9, opened 2016-01-20
Store product_details locales in the database.

Osmose:
This depends on a pending PR for product_details that fixes it in
Python 3 and up.

PR is at https://github.com/mozilla/django-product-details/pull/54. I got this working locally by just `setup.py install`ing my local copy of product-details.

@mythmon r? Might want to wait for the product_details change to land. This also wipes out our migrations, but I think that's fine at this stage. Otherwise I kept hitting a migration issue that made no sense that I couldn't figure out.


mythmon:
I'm uncomfortable with needing to wipe out all the migrations on a matter of principle. It isn't really a problem at this point, but if we had a production DB this would be a big problem, since it seems to require trashing the database too. Or did I get that wrong?

I installed the new version of product_details, ran the migrations, and ran `./manage.py update_product_details`. It ran without errors, but the `recipes_locale` table is empty. That seems wrong.

Also, you should remove the `PROD_DETAILS_STORAGE` from the Test configuration.

Osmose:
> I'm uncomfortable with needing to wipe out all the migrations on a matter of principle. It isn't really a problem at this point, but if we had a production DB this would be a big problem, since it seems to require trashing the database too. Or did I get that wrong?

It would be if we had a prod DB, the "fine at this stage" comment was specifically referring to the fact that it's fine since we don't have one.

However, I've gone back and figure out the issue was converting what was a string field to an integer field in the database. My solution was to add two migrations; one to remove the existing locale field, and another to add the new ForeignKey one.

Until we squash the migrations we'll have to keep `LocaleField` around, which sucks.

> I installed the new version of product_details, ran the migrations, and ran ./manage.py update_product_details. It ran without errors, but the recipes_locale table is empty. That seems wrong.

Make sure you run it with `--force` in case you've already got up-to-date locale info; the storage classes don't get called unless there's new data.

> Also, you should remove the PROD_DETAILS_STORAGE from the Test configuration.

Done!

normandy/recipes/storage.py
```diff
@@ -0,0 +1,27 @@
...
+        super().update(name, content, last_modified)
+
+        # If we are updating firefox versions, update the table.
+        if name == 'languages.json':
+            languages = json.loads(content)
+            for locale_code, names in languages.items():
+                Locale.objects.update_or_create(code=locale_code, defaults={
+                    'english_name': names['English'],
+                    'native_name': names['native'],
+                })
```
mythmon: Should we worry about deleting locales that get removed from product_details?

Osmose: We should!

mythmon:
I like it, and so does Travis. r+

Should we wait until product_details updates to merge this?

Osmose:
> Should we wait until product_details updates to merge this?

I think so.

PR #12, opened 2016-01-22
[Bug 1240965] Add fetch_bundle api endpoint.

mythmon:
This is fairly large, mainly because I built out a lot of infrastructure in the form of model factories and serializers for a bunch of stuff.

r?

### DIFF
--- a/normandy/base/utils.py
+++ b/normandy/base/utils.py
@@ -18,3 +18,12 @@ def urlparams(url, fragment=None, **kwargs):
         urlencode(query, doseq=True),
         fragment if fragment is not None else parsed.fragment
     ))
+
+
+class Whatever(object):
+
+    def __init__(self, test=lambda x: True):
+        self.test = test
+
+    def __eq__(self, other):
+        return self.test(other)
--- a/normandy/base/utils.py
+++ b/normandy/base/utils.py
@@ -18,12 +18,3 @@ def urlparams(url, fragment=None, **kwargs):
         urlencode(query, doseq=True),
         fragment if fragment is not None else parsed.fragment
     ))
-
-
-class Whatever(object):
-
-    def __init__(self, test=lambda x: True):
-        self.test = test
-
-    def __eq__(self, other):
-        return self.test(other)
--- a/normandy/bundler/tests/test_serializers.py
+++ b/normandy/bundler/tests/test_serializers.py
@@ -1,0 +1,15 @@
+import pytest
+
+from normandy.bundler.tests import BundleFactory
+from normandy.bundler.serializers import BundleSerializer
+from normandy.recipes.tests import RecipeFactory
+
+
+@pytest.mark.django_db()
+def test_bundle_serializer():
+    recipe = RecipeFactory()
+    bundle = BundleFactory(recipes=[recipe])
+    serializer = BundleSerializer(bundle)
+
+    assert len(serializer.data['recipes']) == 1
+    assert serializer.data['recipes'][0] == {'name': recipe.name, 'actions': []}
--- a/normandy/bundler/tests/test_serializers.py
+++ b/normandy/bundler/tests/test_serializers.py
@@ -11,5 +11,4 @@ def test_bundle_serializer():
     bundle = BundleFactory(recipes=[recipe])
     serializer = BundleSerializer(bundle)
 
-    assert len(serializer.data['recipes']) == 1
-    assert serializer.data['recipes'][0] == {'name': recipe.name, 'actions': []}
+    assert serializer.data['recipes'] == [{'name': recipe.name, 'actions': []}]
--- a/normandy/classifier/api.py
+++ b/normandy/classifier/api.py
@@ -1,0 +1,27 @@
+from rest_framework import serializers, views, status
+from rest_framework.response import Response
+
+from normandy.bundler.serializers import BundleSerializer
+from normandy.classifier.models import Client
+from normandy.bundler.models import Bundle
+
+
+class FetchBundle(views.APIView):
+
+    serializer_class = BundleSerializer
+
+    class Parameters(serializers.Serializer):
+        locale = serializers.CharField(default=None)
+
+    def post(self, request, format=None):
+        """
+        Determine the recipes that matches the requesting client.
+        """
+        params = self.Parameters(data=request.POST)
+        if not params.is_valid():
+            return Response(params.errors, status=status.HTTP_400_BAD_REQUEST)
+
+        client = Client(request, locale=params.data['locale'])
+        bundle = Bundle.for_client(client)
+        serializer = self.serializer_class(bundle)
+        return Response(serializer.data)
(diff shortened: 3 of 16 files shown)


normandy/base/utils.py
```diff
@@ -18,3 +18,12 @@ def urlparams(url, fragment=None, **kwargs):
         urlencode(query, doseq=True),
         fragment if fragment is not None else parsed.fragment
     ))
+
+
+class Whatever(object):
```
mythmon: Input has something like this. It is useful for tests. Input put it in `fjord/base/tests/__init__.py`, but we don't have a `normandy/base/tests/` yet, so I put it here.

normandy/recipes/tests/__init__.py
```diff
@@ -0,0 +1,36 @@
+import factory
+from factory import fuzzy
+
+from normandy.recipes.models import Action, Recipe, RecipeAction
+
+
+class FuzzyUnicode(fuzzy.FuzzyText):
```
mythmon: This should probably go elsewhere, but I wasn't sure where. Any ideas?

normandy/settings.py
```diff
@@ -120,5 +120,4 @@ class Production(Base):
 class Test(Base):
     SECRET_KEY = values.Value('not a secret')
     PROD_DETAILS_STORAGE = values.Value('product_details.storage.PDFileStorage')
-    DATABASES = values.DatabaseURLValue('sqlite://:memory:')
```
mythmon: py.test correctly creates test database(s) for us like `test_normandy`. Plus, we can't use sqlite since we have tied ourselves to Postgres.

normandy/base/utils.py
Osmose: Might as well create that `__init__.py` file then.

normandy/classifier/api.py
```diff
@@ -0,0 +1,27 @@
...
+from normandy.bundler.serializers import BundleSerializer
+from normandy.classifier.models import Client
+from normandy.bundler.models import Bundle
+
+
+class FetchBundle(views.APIView):
+
+    serializer_class = BundleSerializer
+
+    class Parameters(serializers.Serializer):
```
Osmose: what is this madness its not in the docs

where did you learn such things

and where can i learn too

normandy/bundler/tests/test_serializers.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+@pytest.mark.django_db()
+def test_bundle_serializer():
+    recipe = RecipeFactory()
+    bundle = BundleFactory(recipes=[recipe])
+    serializer = BundleSerializer(bundle)
+
+    assert len(serializer.data['recipes']) == 1
+    assert serializer.data['recipes'][0] == {'name': recipe.name, 'actions': []}
```
Osmose: Any reason you're not just asserting that it's a list directly?

normandy/recipes/tests/__init__.py
Osmose: `normandy.base.tests`?

normandy/classifier/tests/test_api.py
```diff
@@ -0,0 +1,49 @@
+import pytest
+
+from normandy.base.utils import Whatever
+from normandy.recipes.tests import RecipeActionFactory, RecipeFactory
+
```
Osmose: Two spaces below imports `</nit>`

normandy/classifier/tests/test_api.py
```diff
@@ -0,0 +1,49 @@
...
+@pytest.mark.django_db
+class TestFetchBundleAPI(object):
+
+    def test_it_works(self, client):
+        res = client.post('/api/v1/fetch_bundle/')
+        assert res.status_code == 200
+        assert res.data == {'recipes': []}
+
+    def test_it_serves_recipes(self, client):
+        recipe_action = RecipeActionFactory(arguments={'foo': 'bar'})
```
Osmose: Probably wanna be explicit and `create` this in the database.

normandy/classifier/tests/test_api.py
```diff
@@ -0,0 +1,49 @@
+import pytest
+
+from normandy.base.utils import Whatever
+from normandy.recipes.tests import RecipeActionFactory, RecipeFactory
+
+@pytest.mark.django_db
+class TestFetchBundleAPI(object):
+
+    def test_it_works(self, client):
```
Osmose: Does the client just magically get passed in by py.test?

Osmose:
Tests failed due to a missing `jsonb` field type. I believe we need to be explicit about using Postgres 9.4 in our `.travis.yml`: https://docs.travis-ci.com/user/using-postgresql/#Selecting-a-PostgreSQL-Version

normandy/classifier/tests/test_api.py
mythmon: Hmm? FactoryBoy factories call `.save()` by default, if that's what you mean.

normandy/classifier/api.py
mythmon: This isn't a standard DRF feature. I just made this up today. Well, I've used serializers for API parameters before, but this nested class is new. I used it by hand in the `post` method below, but in the future it is probably worth making a mixin that does it automatically.

normandy/classifier/tests/test_api.py
mythmon: Yes. This is what py.test calls fixtures. It depends on the *name* of the parameter. If I had called this `foobar`, I would get something entirely different. Probably nothing. [Docs here](http://pytest-django.readthedocs.org/en/latest/helpers.html#fixtures)

normandy/bundler/tests/test_serializers.py
mythmon: I see now that it would be smaller that way. I used to have `assert serializer.data == ...`, which was really verbose, and this was simpler. Asserting the list directly would be better now.

normandy/classifier/tests/test_api.py
Osmose: Huh, I swear it used to used the [build strategy](http://factoryboy.readthedocs.org/en/latest/reference.html#factory.BUILD_STRATEGY) by default. 

What I was trying to convey was a suggestion to call `RecipeActionFactory.create` instead of just `RecipeActionFactory`, so that it's obvious whether you're creating it in the database or not regardless of what you set the default strategy to be.

normandy/classifier/tests/test_api.py
Osmose: Man, I ___hate___ dependency injection. This is fine because it's how they do it, but jeez.

normandy/classifier/tests/test_api.py
mythmon: Hmm. I think I'd prefer to assume that the default strategy is create, instead of putting `.create()` everywhere. I didn't realize the default used to be the build strategy, which is why I was confused.

mythmon:
Updated ^

Osmose:
arrrrr pluuuusssssss

PR #13, opened 2016-01-25
Bug 1240969: Store action implementation as a file.

Osmose:
Action implementations are now files, and the recipe API links to them properly now. This doesn't fully fix the bug because the production storage bits haven't really been tested / figured out.

@mythmon r?


normandy/settings.py
```diff
@@ -69,6 +70,10 @@ class Core(Configuration):
     STATIC_URL = '/static/'
     STATIC_ROOT = os.path.join(BASE_DIR, 'static')
 
+    # User-uploaded Media
+    MEDIA_URL = '/media/'
```
Osmose: This will probably get promoted to Base in the future.

normandy/settings.py
```diff
@@ -116,6 +124,11 @@ class Production(Base):
     """Settings for the production environment."""
     STATICFILES_STORAGE = 'whitenoise.django.GzipManifestStaticFilesStorage'
 
+    DEFAULT_FILE_STORAGE = 'storages.backends.s3boto.S3BotoStorage'
+    AWS_ACCESS_KEY_ID = values.SecretValue()
```
mythmon: `SecretValue` throws an error if it doesn't have a configuration value, right? Is that what we want?

normandy/settings.py
```diff
@@ -15,6 +15,7 @@ class Core(Configuration):
         'normandy.selfrepair',
 
         'adminsortable',
+        'rest_framework',
```
mythmon: Oops, thanks. The docs for django-storages say you should add `'storages'` here too. http://django-storages.readthedocs.org/en/latest/#installation

normandy/settings.py
mythmon: I read this more closely, and I realize now this is the Production config. Making these required sounds like a good thing in that case.

normandy/settings.py
```diff
@@ -116,6 +124,11 @@ class Production(Base):
     """Settings for the production environment."""
     STATICFILES_STORAGE = 'whitenoise.django.GzipManifestStaticFilesStorage'
 
+    DEFAULT_FILE_STORAGE = 'storages.backends.s3boto.S3BotoStorage'
```
mythmon: [The docs say](http://django-storages.readthedocs.org/en/latest/backends/amazon-S3.html#usage) that the boto backend requires the `python-boto` package to be installed, and it doesn't do it automatically.

Osmose:
Updated, totally forgot that I should add commits instead of rebasing, sorry. :(

mythmon:
No worries. This looks good to me now.

PR #14, opened 2016-01-25
[Bug 1240973] Changes to run on Heroku

mythmon:
The app is at https://dashboard.heroku.com/apps/normandy-dev/

I haven't put it through its paces, but it seems to be working. We can fix any problems we run into on the way.


runtime.txt
```diff
@@ -0,0 +1 @@
+python-3.5.1
```
Osmose: I mean I guess it matches `requirements.txt` but it feels weird. `package.json` gets this right, I think. Oh well.

Osmose:
r+, nice work!

PR #15, opened 2016-01-26
Remove bundler app and move remnants to classifier.

Osmose:
Since we no longer will have any bundler views, the only things worth saving are the model and serializer, and those don't really warrant their own app without views.

@mythmon r?


mythmon:
r+

PR #16, opened 2016-01-26
Run CI in Python 3.4 and 3.5

mythmon:
@relud has informed me that Python 3.4.3 will be a bit easier to get into production, so lets make sure we can support it.


Osmose:
r+

mythmon:
Well, as fun as that was, I have since learned we don't actually need this. I'll revert the merge.

PR #18, opened 2016-01-26
add Dockerfile

relud:


mythmon:
After reading a bit more about [Docker networking][1] to figure out how this would talk to external services, this looks good.

[1]: https://docs.docker.com/engine/userguide/networking/dockernetworks/

PR #19, opened 2016-01-27
Self repair driver

Osmose:
A basic first pass at the self-repair driver.

@mythmon r?

### DIFF
--- a/normandy/classifier/api.py
+++ b/normandy/classifier/api.py
@@ -6,7 +6,8 @@ from normandy.classifier.models import Bundle, Client
 
 
 class FetchBundle(views.APIView):
-
+    authentication_classes = []
+    permission_classes = []
     serializer_class = BundleSerializer
 
     class Parameters(serializers.Serializer):
--- a/normandy/selfrepair/static/js/self_repair_driver.js
+++ b/normandy/selfrepair/static/js/self_repair_driver.js
@@ -15,7 +15,7 @@ function xhr(method, url, options={}) {
     return new Promise((resolve, reject) => {
         let req = new XMLHttpRequest();
         req.addEventListener('load', e => {
-            if (req.status !== 200) {
+            if (req.status >= 400) {
                 reject(req);
             } else {
                 resolve(req);
@@ -54,7 +54,7 @@ window.Normandy = {
  * Download the implementation of the given action from the server.
  *
  * @param {Action} action  - Action taken from a Recipe.
- * @promise Resolves once the action has been loaded and registered.
+ * @promise {Function} The implementation function for the given action.
  * @rejects {Error} Rejects if the action could not be loaded or did not
  *     register itself.
  */
@@ -67,12 +67,12 @@ function loadAction(action) {
                 if (!registeredActions[action.name]) {
                     reject(new Error(`Could not find action with name ${action.name}.`));
                 } else {
-                    resolve();
+                    resolve(registeredActions[action.name]);
                 }
             };
             document.head.appendChild(script);
         } else {
-            resolve();
+            resolve(registeredActions[action.name]);
         }
     });
 }
@@ -83,9 +83,9 @@ function loadAction(action) {
  * @promise {Array<Recipe>} List of recipes.
  */
 function fetchRecipes() {
-    let {recipe_url, locale} = document.documentElement.dataset;
+    let {recipeUrl, locale} = document.documentElement.dataset;
 
-    return xhr.post(recipe_url, {
+    return xhr.post(recipeUrl, {
         data: {locale: locale},
         headers: {Accept: 'application/json'}
     }).then(request => {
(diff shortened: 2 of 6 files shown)


normandy/classifier/api.py
```diff
@@ -6,7 +6,8 @@
 
 
 class FetchBundle(views.APIView):
-
+    authentication_classes = []
+    permission_classes = []
```
mythmon: Why did you add these? Do we need them?

normandy/selfrepair/static/js/self_repair_driver.js
```diff
@@ -0,0 +1,132 @@
...
+ * @param {Object} options.headers - Headers and values to set on the
+ *     request.
+ * @promise {XMLHttpRequest} The request after it has succeeded.
+ * @rejects {XMLHttpRequest} The request after it has failed.
+ */
+function xhr(method, url, options={}) {
+    return new Promise((resolve, reject) => {
+        let req = new XMLHttpRequest();
+        req.addEventListener('load', e => {
+            if (req.status !== 200) {
```
mythmon: This should be something like `if (req.status >= 400) {`. [201][] and [204][] are totally useful status codes.

[201]: http://http.cat/201
[204]: http://http.cat/204

normandy/selfrepair/static/js/self_repair_driver.js
```diff
@@ -0,0 +1,132 @@
...
+ * @promise Resolves once all actions are complete.
+ */
+function runRecipe(recipe) {
+    return Promise.all(recipe.actions.map(loadAction)).then(() => {
+        let recipePromise = Promise.resolve();
+
+        for (let action of recipe.actions) {
+            let func = registeredActions[action.name];
+            recipePromise = recipePromise.then(() => {
+                return func(null, action.arguments);
```
mythmon: At some point we'll need a better stub instead of `null` here.

normandy/selfrepair/static/js/self_repair_driver.js
```diff
@@ -0,0 +1,132 @@
...
+fetchRecipes().then((recipes) => {
+    let chain = Promise.resolve();
+
+    for (let recipe of recipes) {
+        chain.then(() => {
+            return runRecipe(recipe);
+        });
+    }
+
+    return chain;
```
mythmon: A silly one liner that does the same thing, if you're into that.

```javascript
recipes.reduce((chain, recipe) => chain.then(() => runRecipe(recipe)), Promise.resolve());
```

normandy/classifier/api.py
Osmose: We do, because otherwise the view requires a CSRF token due to the default auth/permissions.

normandy/selfrepair/static/js/self_repair_driver.js
Osmose: Yes, that will be in a later PR

normandy/selfrepair/static/js/self_repair_driver.js
```diff
@@ -0,0 +1,132 @@
...
+function loadAction(action) {
+    return new Promise((resolve, reject) => {
+        if (!registeredActions[action.name]) {
+            let script = document.createElement('script');
+            script.src = action.implementation.url;
+            script.onload = () => {
+                if (!registeredActions[action.name]) {
+                    reject(new Error(`Could not find action with name ${action.name}.`));
+                } else {
+                    resolve();
```
mythmon: I think this function should resolve with the action that was loaded. That way other things, like `runRecipe` can be simplified a little to not have to deal with the global registry.

normandy/selfrepair/static/js/self_repair_driver.js
Osmose: I feel like that's super-unreadable. Maybe a helper method for a promise chain would be better, though. Passing that callback to `then` seems like unnecessary noise.

mythmon: That's fair. I'm not sure what you mean by unnecessary noise you mean though, since `chain.then(() => runRecipe(recipe))` is the same as

```javascript
chain.then(() => {
    return runRecipe(recipe);
});
```

just on fewer lines. Is that the part of the code you meant?

Osmose: I meant like maybe we make a helper function that removes it:

```js
promiseChain(list_of_promises);
```

Initially I don't think that'd work because by the time you've generated the promises, they're already executing, but maybe we can figure something out. Maybe it's not worth the time. I'll think about it.

mythmon: Yeah, it would have to be `promiseChain(list_of_things_that_generate_promises)`. That would probably be `promiseChain(recipes.map(recipe => runRecipe.bind(null, recipe)))`, which isn't very nice.

Maybe something like `promiseSerialMap(runRecipe, recipes)` would be nicer?

mythmon: This isn't really an important point. The rest of the patch is good, and we can merge this as is.

normandy/classifier/api.py
mythmon: Does it make sense to change the defaults? I don't think our API will ever want CSRF.

Osmose: I think it's good to make a lack of CSRF the exception rather than the default. It's better to have a view that shouldn't have CSRF accidentally have it then to have the opposite.

Osmose:
Updated!

normandy/classifier/api.py
mythmon: To be clear, we can make the API specifically CSRF-free, without making that the default for all views.

Osmose: I think it's a good thing to do even for the API, but I can change it anyway if you feel strongly about it.

PR #20, opened 2016-01-27
Pin package hashes

mythmon:
I got excited.

Plus, it would be good to have this in place before we get serious about deploying, so that we don't have to figure it out with our hit points later.

One problem here is that pip will use wheels where it can. That will result in different hashes on different systems for some packages (those with platform specific wheels). I think we are going to have to find these by trial and error (at least, until @peterbe finishes *Peepin 2: Hashin Boogaloo*).

r?


requirements.txt
```diff
@@ -1,18 +1,68 @@
...
+snowballstemmer==1.2.1 \
+    --hash=sha256:9f3bcd3c401c3e862ec0ebe6d2c069ebc012ce142cce209c098ccb5b09136e89
+sphinx_rtd_theme==0.1.9 \
+    --hash=sha256:46c2ca8b482b0398b9621bb5a8ec7c9c5c7226926e5363e18d744f5f0d39d78f
+Sphinx==1.3.3 \
+    --hash=sha256:3ad4cb89ab4baa5f9bb99548a9a1fb5127b3ee83a5213b83da4de7578afdc891
+wheel==0.26.0 \
+    --hash=sha256:c92ed3a2dd87c54a9e20024fb0a206fe591c352c745fff21e8f8c6cdac2086ea
+whitenoise==2.0.6 \
+    --hash=sha256:826ffe5d608c9dc8daebef1b0b43d01f7958f17c2fce36e75c80e26160172c4f
```
peterbe: This file lacks the alternative hashes you might get. For example, for psycopg2 you'll need all hashes so people on other systems will be able to run `pip install -r requirments.txt` without getting hash mismatch errors. 

I used `hashin.py` and generated this: https://gist.github.com/peterbe/6e8211432de96863c6e3

mythmon: I know. I said this in comment 0:

> One problem here is that pip will use wheels where it can. That will result in different hashes on different systems for some packages (those with platform specific wheels). I think we are going to have to find these by trial and error (at least, until @peterbe finishes Peepin 2: Hashin Boogaloo).

Is `hashin.py` ready for use? I thought it was still a work in progress. Got a link?

mythmon: Oh, now I feel bad. I didn't realize your link was a copy of *our* `requirements.txt` file. This still isn't quite what we need since it doesn't include the subdependencies. Thanks for the help though!

peterbe:
I wonder what it would take to make `hashin` be able to automatically figure out the dependencies. 
Perhaps there's some dry-run thing inside pip we can (ab)use. 

Yes, it's "ready" http://www.peterbe.com/plog/hashin by the way.

Osmose:
r+

PR #21, opened 2016-01-27
Generate SHA-256 cert for SSL dev server.

Osmose:
The certificate included with runsslserver is SHA1 and will stop
working with Firefox soon, and having all the devs trust a single cert
instead of generating their own personal ones is a bad idea anyway.

@mythmon r?


mythmon:
WFM.

PR #22, opened 2016-01-28
Add setup and workflow docs.

Osmose:
@mythmon r?

### DIFF
--- a/docs/dev/install.rst
+++ b/docs/dev/install.rst
@@ -1,0 +1,135 @@
+Developer Setup
+===============
+The following describes how to set up an instance of the site on your
+computer for development.
+
+Prerequisites
+-------------
+This guide assumes you have already installed and set up the following:
+
+1. Git_
+2. `Python 3.5`_, `pip 8`_ or higher, and virtualenv_
+3. `Postgres 9.4`_
+4. ``openssl``
+
+These docs assume a Unix-like operating system, although the site should, in
+theory, run on Windows as well. All the example commands given below are
+intended to be run in a terminal.
+
+.. _Git: https://git-scm.com/
+.. _Python 2.7: https://www.python.org/
+.. _pip 8: https://pip.pypa.io/en/stable/
+.. _virtualenv: https://virtualenv.pypa.io/en/latest/
+.. _Postgres 9.4: http://www.postgresql.org/
+
+Installation
+------------
+1. Clone this repository or your fork_:
+
+   .. code-block:: bash
+
+      git clone --recursive https://github.com/mozilla/normandy.git
+      cd normandy
+
+2. Create a virtualenv for Normandy and activate it:
+
+   .. code-block:: bash
+
+      virtualenv venv
+      source ./venv/bin/activate
+
+   .. note::
+
+      Whenever you want to work on Normandy in a new terminal you'll have to
+      re-activate the virtualenv. Read the virtualenv_ documentation to learn
+      more about how virtualenv works.
+
+3. Install the dependencies using pip:
+
+   .. code-block:: bash
+
+      pip install -r requirements.txt
+
+   .. note::
+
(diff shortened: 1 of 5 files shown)


docs/dev/install.rst
```diff
@@ -0,0 +1,135 @@
...
+===============
+The following describes how to set up an instance of the site on your
+computer for development.
+
+Prerequisites
+-------------
+This guide assumes you have already installed and set up the following:
+
+1. Git_
+2. `Python 3.5`_, `pip 8`_ or higher, and virtualenv_
```
mythmon: The Python 3.5 link is broken.

docs/dev/install.rst
```diff
@@ -0,0 +1,135 @@
...
+.. _virtualenv: https://virtualenv.pypa.io/en/latest/
+.. _Postgres 9.4: http://www.postgresql.org/
+
+Installation
+------------
+1. Clone this repository or your fork_:
+
+   .. code-block:: bash
+
+      git clone --recursive https://github.com/mozilla/normandy.git
```
mythmon: We don't need `--recursive` (and we probably shouldn't have it, since it implies things that are wrong).

docs/dev/install.rst
```diff
@@ -0,0 +1,135 @@
...
+
+3. Install the dependencies using pip:
+
+   .. code-block:: bash
+
+      pip install -r requirements.txt
+
+   .. note::
+
+      If you get an error like `pip: error: no such option: --hash`, you are
```
mythmon: I think you want double ticks here. It renders as just plain text now.

mythmon: Also, all the pip error handling is pretty big. Maybe we should have a separate troubleshooting section for it?

docs/dev/install.rst
```diff
@@ -0,0 +1,135 @@
...
+
+   .. note::
+
+      If you use a different name for the database, or otherwise want to
+      customize how you connect to the database, you may specify the database
+      URL by adding the following to a ``.env`` file at the root of the repo:
+
+      .. code-block:: ini
+
+         DATABASE_URL=postgres://username:password@server_addr/database_name
```
mythmon: I think that most people will need a grant command here? Or maybe that's just me. Or should we just assume that people can do their own DBA?

docs/dev/install.rst
```diff
@@ -0,0 +1,135 @@
...
+   .. code-block:: bash
+
+      python manage.py createsuperuser
+
+7. Pull the latest data on Firefox releases and supported locales with the
+   ``update_product_details`` command:
+
+   .. code-block:: bash
+
+      python manage.py update_product_details
```
mythmon: If this is here, I suspect we should have them run the script to get the geoip database too.

docs/dev/install.rst
Osmose: I figure that if they need this note they're complicated enough to do their own DBA. Since we use the `postgres` user by default it should normally work.

mythmon: Oh sorry. I meant to refer to the entire section about the database, not this line specifically.

Osmose: Ah, okay. My point was that they don't need a grant by default since we're using the root `postgres` account, so the only time they'd need one is if they're using a custom `DATABASE_URL` and doing advanced stuff, or if they somehow gave the `postgres` user reduced privileges, which I'm not sure is even possible.

Osmose:
Updated!

docs/dev/install.rst
mythmon: Oh right, because most postgres setups includes a no-password auth for the root postgres user. I guess most devs would probably have that. I definitely fall into the second camp that needs to do my own DBA, so I forget how the normal way works sometimes.

PR #23, opened 2016-01-29
Bug 1240968: Add API for submitting actions.

Osmose:
@mythmon r?


normandy/recipes/api.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class ActionViewSet(viewsets.ModelViewSet):
+    """Viewset for viewing and uploading recipe actions."""
+    queryset = Action.objects.all()
+    serializer_class = ActionSerializer
+    permission_classes = [permissions.DjangoModelPermissions]
+
+    lookup_field = 'name'
+    lookup_value_regex = r'[_\w]+'
```
mythmon: I think the default for this value is `r'[^/\.]'`, which is a superset of this regex. Is there a reason to restrict this down?

normandy/recipes/api.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def get_object(self, queryset=None):
+        if queryset is None:
+            queryset = self.get_queryset()
+
+        return queryset.get(name=self.kwargs['name'])
```
mythmon: Is this not automatic based on the args like `queryset` and `lookup_field`?

If it isn't, this should probably use `self.kwargs[self.lookup_field]` instead of `'name'`. Also it is weird that it isn't automatic.

normandy/recipes/api.py
Osmose: Because slugs are only alphanumeric plus `_`?

normandy/recipes/api.py
Osmose: It is, I forgot to remove this after messing with the lookup stuff.

normandy/recipes/urls.py
```diff
@@ -0,0 +1,14 @@
+from django.conf.urls import url, include
+
+from rest_framework.routers import DefaultRouter
+
+from normandy.recipes.api import ActionViewSet
+
+# API Router
+router = DefaultRouter()
+router.register(r'actions', ActionViewSet)
```
mythmon: Names of resources that occur in the URL should be singular. So this should be `/api/v1/action/` instead of `/api/v1/actions/`.

normandy/recipes/api.py
mythmon: True. I've just never bothered changing it (except this one time we needed to makes `.` legal). This is fine as is then.

mythmon:
In the current form, this requires form encoded file uploads for the implementation field, which is going to be pretty annoying for clients and scripting. I think we'll need to do some extra work with DRF so that we can use this API as a JSON api, and include the contents of the file as just another JSON field. I'm not totally sure how to do that.

mythmon:
When submitting a new action (`POST /api/v1/actions/`), if the contents of the file are exactly the same as another file, the API gives an error. That might be ok? Or maybe we want to silently de-dupe to the existing file?

Either way, a larger issue is that the error is a 500 caused by a server error. The error is `Key (implementation_hash)=(d77ebfb1abdf63aa3ce992ab601c9507941ec865) already exists`. [Here is a full traceback](https://gist.github.com/mythmon/11f285f19eb04db4d319). We should at least give a nicer error message that includes the details of what went wrong, instead of a server error.

mythmon:
I wish there were some tests for this. I think writing the tests might have revealed the first issue above, and *maybe* the second one, if you were lazy with the file contents.

Osmose:
Updated based on feedback, will work on adding some tests as this particular part of the site isn't likely to change much.

Osmose:
Updated with tests, this is ready for another look!

normandy/recipes/api/fields.py
```diff
@@ -0,0 +1,20 @@
+from contextlib import closing
+
+from django.core.files.base import ContentFile
+
+from rest_framework.fields import CharField
+
+
+class ContentFileField(CharField):
+    """Serilizer field that deserializes text into a ContentFile."""
```
mythmon: Typo: Serilizer

mythmon:
Looks good. I wonder if the `ContentFileField` could have tests. I don't even know what those tests would really do though.

After you fix that typo, r+. Unless you have some clever idea about `ContentFileField` tests.

PR #24, opened 2016-02-03
Use in-memory file storage for tests.

Osmose:
Avoids making tons of files in /media during test runs.

@mythmon r?


mythmon:
r+

PR #26, opened 2016-02-05
Bug 1240968: Add argument schemas.

Osmose:
Includes some admin customizations for seeing what recipes an action is being used by, and prevents actions from being edited in the admin or via API when it's in use to prevent mismatching schemas.

@mythmon r?


Osmose:
Whoops, just noticed this is missing tests for submitting a schema via the API. Will see about fixing that.

normandy/recipes/admin.py
```diff
@@ -45,11 +43,30 @@ class RecipeAdmin(NonSortableParentAdmin):
...
+
+    def recipe_list(self, action):
+        """List all recipes that the action is being used by."""
+        return render_to_string('admin/field_recipe_list.html', {
+            'recipes': action.recipes_used_by.order_by('name'),
+        })
+    recipe_list.short_description = 'Used in Recipes'
+
+    def in_use(self, action):
+        """Callable on the ModelAdmin so we can set boolean=True."""
```
mythmon: I don't know what this docstring means.

Osmose: It means that we can't just use the `in_use` property on the model because it won't show as the nice boolean checkmark or X on the admin with the `boolean` property set on it, and we can't set it on the property itself.

I'll try and be clearer about that.

Osmose:
Found a few bugs in validation, will post up patches for them on Wednesday hopefully.

Osmose:
Alright, this should be ready for real now. Upload seems to be working with https://github.com/mozilla/normandy-actions.

normandy/recipes/templates/admin/field_recipe_list.html
```diff
@@ -0,0 +1,9 @@
+{% if recipes %}
+  <ul>
+    {% for recipe in recipes %}
+      <li><a href="{% url 'admin:recipes_recipe_change' recipe.id %}">{{ recipe.name }}</a></li>
+    {% endfor %}
+  </ul>
+{% else %}
+  <p>Not used in any recipes.</p>
```
mythmon: I think this should be "Not used in any active recipes". Either that, or this section should include both active and inactive recipes, and indicate that.

mythmon: And by "indiciate that" I mean it should show that the recipe is active or inactive with :white_check_mark: / :x:.

normandy/recipes/forms.py
```diff
@@ -0,0 +1,38 @@
...
+        if self.instance.in_use:
+            self.add_error(
+                None,
+                ValidationError('Action is currently in use and cannot be edited.')
+            )
+
+
+class RecipeActionInlineForm(forms.ModelForm):
+    def clean(self):
+        """Validate the arguments against their schema."""
```
mythmon: I'm a little nervous about making this specific to the admin interface (and other things that use forms). Although right now this is the only way that recipes are created and edited, I can certainly image other ways in the future. Namely I'm thinking of API access to this, which I imagine we will want eventually.

One way we could handle this is to re-implement the schema validation logic in the API. I'm not excited about this. Another option would be to put this validation into the model somehow. That makes nice error messages harder though, I imagine.

Is this something you want to worry about now? Or should we just kick it down the road until we actually have another editor of these models?

Also, tests for validation would be nice.

mythmon:
Nice. This works well. I tested the schemas, and they behave the way we talked about. When a recipe has invalid arguments, it gives nice error messages (though only one error at a time, which is sad but a fault of the library).

Lets talk about the json schema validation thing, but besides that, r+

normandy/recipes/templates/admin/field_recipe_list.html
Osmose: I like the checkmark suggestion, but I think we can add that later. I'll go with the easy fix for now.

normandy/recipes/forms.py
Osmose: I think kicking it down the road is a good option for now, but you're right that in the long run we want this to be tied to the model rather than the form. I can't think of the reason I chose to go with a form specifically but I do remember trying to do it with model validation at first.

mythmon: This is definitely an open question when mixing forms and DRF. Kicking it down the road it is!

PR #28, opened 2016-02-09
build and test with docker on travis

relud:


relud:
this will push docker images to hub.docker.com when secrets are available. I think that secrets are available on pull requests originating from this repo, and on commits to branches in this repo. otherwise it will just build the docker image and run pytest.

Dockerfile
```diff
@@ -2,11 +2,12 @@ FROM python:3.5.1
 WORKDIR /app
 RUN groupadd --gid 1001 app && useradd -g app --uid 1001 --shell /usr/sbin/nologin app
 COPY ./requirements.txt /app/requirements.txt
-RUN pip install --upgrade --no-cache-dir -r requirements.txt
+RUN pip install -U 'pip>=0.8' && pip install --upgrade --no-cache-dir -r requirements.txt
```
mythmon: This needs to be `pip>=8` not `0.8`.

.travis.yml
```diff
@@ -1,11 +1,27 @@
...
+after_success:
+  - '[ -z "$DOCKER_EMAIL" ] ||
+    docker login -e="$DOCKER_EMAIL" -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD"'
+  - '[ -z "$DOCKER_EMAIL" ] ||
+    ([ "$TRAVIS_BRANCH" == master ] && TAG=latest || TAG="$TRAVIS_BRANCH" &&
+    docker tag normandy:build "mozilla/normandy:$TAG" &&
+    docker push "mozilla/normandy:$TAG")'
+  - '[ -z "$DOCKER_EMAIL" -o -z "$TRAVIS_TAG" ] ||
+    (docker tag normandy:build "mozilla/normandy:$TRAVIS_TAG" &&
+    docker push "mozilla/normandy:$TRAVIS_TAG")'
```
mythmon: How would you feel about putting all of this `after_success` hook into something like `bin/deploy/dockerhub.sh`?

Dockerfile
relud: doh >_<

.travis.yml
relud: sure

.travis.yml
```diff
@@ -1,11 +1,18 @@
-sudo: false
+sudo: required
```
Osmose: I'll miss you, fast builds

relud: same

mythmon:
Looks pretty good. One more suggestion: Travis has a `script` deploy provider. By using that instead of `after_success` (which is also recommended in the docs), the deploy script won't even both running on PRs, and we can control which branches it runs on.

https://docs.travis-ci.com/user/deployment/script

```yaml
deploy:
  provider: script
  script: bin/deploy/dockerhub.sh
  on:
  	branch: master
```

relud:
sounds good.

relud:
actually, I think it would be better to do:
<pre>
deploy:
  provider: script
  script: bin/deploy/dockerhub.sh
</pre>
so that it will build any branch and deploy it, but not PRs since deploy doesn't happen on PRs

mythmon:
+1

PR #30, opened 2016-02-12
Very simple implementation of Heartbeat for the self-repair driver.

Osmose:
This is not final at all, it just makes https://github.com/mozilla/normandy-actions/blob/master/actions/show-heartbeat/index.js function correctly. We still need to decide on a better interface.

Also, I considered adding django-pipeline for bundling the UITour stuff, but I figured that it was overkill at this point in time. Maybe once we add a second external lib we can look at compiling the JS for self-repair together?

@mythmon r?

### DIFF
--- a/normandy/selfrepair/static/js/self_repair_driver.js
+++ b/normandy/selfrepair/static/js/self_repair_driver.js
@@ -46,8 +46,40 @@ let registeredActions = {};
 window.Normandy = {
     registerAction(name, func) {
         registeredActions[name] = func;
-    }
+    },
+
+    heartbeatCallbacks: [],
+    showHeartbeat(options) {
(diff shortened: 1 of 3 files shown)


normandy/selfrepair/static/js/self_repair_driver.js
```diff
@@ -46,10 +46,42 @@ let registeredActions = {};
 window.Normandy = {
     registerAction(name, func) {
         registeredActions[name] = func;
-    }
+    },
+
+    heartbeatCallbacks: [],
```
mythmon: The curmudgeon in me points out that putting this hear makes it available to actions, which probably isn't what we want to do.

PR #33, opened 2016-02-13
Add setting to disable admin interface

mythmon:
I couldn't figure out how to test this. Changing the setting during the test doesn't cause urls.py to reload. I think this will have to be a pre-deployment test, once we have those. And we definitely want to test it there, because of the security considerations.

Another thought I had was a proxy view that intercepted all the `/admin/.*` urls and either forwarded them to the real admin app or 404ed them, depending on the setting. That seemed too complex though.

r?

### DIFF
--- a/normandy/settings.py
+++ b/normandy/settings.py
@@ -125,6 +125,7 @@ class Base(Core):
     }
 
     CAN_EDIT_ACTIONS_IN_USE = values.BooleanValue(False)
+    ADMIN_ENABLED = values.BooleanValue(True)
 
 
 class Development(Base):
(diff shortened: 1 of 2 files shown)


willkg:
I'm pretty sure I had tests in richard that enabled/disabled urls. I'll write a todo item to hunt for whether I did do that and how I did it.

willkg:
This is what I did in richard:

* https://github.com/pyvideo/richard/blob/master/src/richard/videos/urls.py#L72
* https://github.com/pyvideo/richard/blob/master/tests/test_videos/test_api.py#L33

That's with Django 1.7, I think. I don't know if:

1. that's a good idea
2. there are better ways to do it
3. if it'd work for your case or not

Osmose:
r+, I think a pre-deployment test is the way to go here.

PR #34, opened 2016-02-15
Allow PUT requests to create objects in the action API.

Osmose:
Fixes bug 1248305.

@mythmon r?


mythmon:
Ugh. This is kind of gross, but I think it is the best way to do it.

PR #35, opened 2016-02-16
Disable action API when admin is disabled

mythmon:
Fixes bug 1248630.

### DIFF
--- a/normandy/recipes/tests/test_api.py
+++ b/normandy/recipes/tests/test_api.py
@@ -101,9 +101,15 @@ class TestActionAPI(object):
 
         res = api_client.patch('/api/v1/action/active/', {'implementation': 'foobar'})
         assert res.status_code == 403
+        assert res.data == {
+            'detail': 'Cannot edit this object while it is in use'
+        }
 
         res = api_client.delete('/api/v1/action/active/')
         assert res.status_code == 403
+        assert res.data == {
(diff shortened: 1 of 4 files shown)


normandy/recipes/tests/test_api.py
```diff
@@ -101,9 +101,15 @@ def test_it_cant_edit_actions_in_use(self, api_client, settings):
 
         res = api_client.patch('/api/v1/action/active/', {'implementation': 'foobar'})
         assert res.status_code == 403
+        assert res.data == {
+            'detail': 'Cannot edit this object while it is in use'
```
Osmose: I've been on the fence about testing specific messages being returned by the API, but it seems like you fall on the "yes, do that" side of that question?

Osmose:
r+, nice work!

normandy/recipes/tests/test_api.py
mythmon: I want to assert that the right kind of error got thrown. I want to make sure this is an "object in use" error, and not an "admin disabled" error. We could also this with something like `assert res.data['detail'] == NotInUse.message`, if changing error strings ends up being annoying.

Osmose: I like that suggestion and will use it in future PRs when I can. :D

PR #39, opened 2016-02-18
Make Recipe locale and country selection many-to-many

mythmon:
Fixes bug 1248263.

@Osmose r?


normandy/recipes/migrations/0017_countries.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def remove_countries(apps, schema_editor):
+    Country = apps.get_model('recipes', 'Country')
+    Country.objects.all().delete()
+
+
+def set_locale_sort_order(apps, schema_editor):
+    Locale = apps.get_model('recipes', 'Locale')
+    english = Locale.objects.get(code='en-US')
```
Osmose: This fails if `en-US` is missing, make it a `filter` instead.

Osmose:
Tests failllllll

normandy/recipes/migrations/0017_countries.py
mythmon: Gah. I had a try/catch here, but removed it for easier debugging

mythmon:
Bleh. I just missed an entire part of this patch. I'm going to close this and make it a part of #41, since they share a lot in common.

PR #41, opened 2016-02-19
Release channels and M2M for country and locale

mythmon:
@Osmose r?


mythmon:
The release channel change and the M2M changes for country and locale share a lot in common. It was much easier to combine them into one branch, so I did that.

normandy/classifier/models.py
```diff
@@ -2,22 +2,24 @@
...
 
 
 class Client(object):
     """A client attempting to fetch a set of recipes."""
-    def __init__(self, request, locale=None):
+    def __init__(self, request, locale=None, release_channel=None, country=None):
         self.request = request
         self.locale = locale
+        self.release_channel = release_channel
+        self._country = country
```
Osmose: You can actually just set `self.country = country` and it will use that instead of the cached property. So setting it if country isn't `None` will let you retain the decorator while allowing it to be set manually.

normandy/classifier/tests/__init__.py
```diff
@@ -1,8 +1,17 @@
...
 class BundleFactory(factory.Factory):
     class Meta:
         model = Bundle
+
+
+class ClientFactory(factory.Factory):
+    class Meta:
+        model = Client
+
+    request = factory.LazyAttribute(lambda o: RequestFactory().get('/'))
```
Osmose: Oh man this is awesome

normandy/recipes/models.py
```diff
@@ -21,13 +20,48 @@ class Locale(models.Model):
...
+    slug = models.SlugField(max_length=255, unique=True)
+    name = models.CharField(max_length=255)
+
+    class Meta:
+        ordering = ['slug']
+
+    def __str__(self):
+        return self.name
+
+    def matches(self, other):
```
Osmose: It's a bit vague what this is matching against; perhaps we pass in the client and it extracts from the client the fields it's matching against?

normandy/recipes/models.py
```diff
@@ -36,33 +70,41 @@ class Recipe(models.Model):
...
     sample_rate = PercentField(default=100)
+    release_channels = models.ManyToManyField(ReleaseChannel, blank=True)
 
     def log_rejection(self, msg):
         logger.debug('{} rejected: {}'.format(self, msg))
 
+    def check_many(self, name, field_qs, client_val):
+        field_vals = list(field_qs.all())
+
+        if field_vals == []:
```
Osmose: Just check the `len`, Django will make it a list and cache it so no need to manually convert it.

normandy/recipes/models.py
```diff
@@ -36,33 +70,41 @@ class Recipe(models.Model):
...
+    countries = models.ManyToManyField(Country, blank=True)
     start_time = models.DateTimeField(blank=True, null=True, default=None)
     end_time = models.DateTimeField(blank=True, null=True, default=None)
     sample_rate = PercentField(default=100)
+    release_channels = models.ManyToManyField(ReleaseChannel, blank=True)
 
     def log_rejection(self, msg):
         logger.debug('{} rejected: {}'.format(self, msg))
 
+    def check_many(self, name, field_qs, client_val):
```
Osmose: This should be named `matches_something`. With something being a better word that I cannot come up with currently. :P

Also a docstring would be nice.

normandy/recipes/models.py
mythmon: I'm going to change this with the other matching refactors we talked about, since changing this sort of breaks `check_many.

Osmose:
r+wc, nice work!

PR #44, opened 2016-02-25
Refactor recipe matching to be many registered handlers.

mythmon:
Maybe this is better than making that match function giant and terrible. Maybe not.

@Osmose r?

### DIFF
--- a/normandy/recipes/models.py
+++ b/normandy/recipes/models.py
@@ -130,11 +130,17 @@ def match_sample_rate(recipe, client):
 
 
 def multivalue_matcher(field_getter):
-    """Make a matcher that accepts the client if any of the recipe's values accept the client."""
+    """
+    Makes matcher that accepts if any of the recipe's values accept.
+
+    :param field_getter: A function that returns an object with a
+        `match` method which will be passed the client to decide if the
+        object accepts the client.
+    """
     def matcher(recipe, client):
         field_vals = list(field_getter(recipe))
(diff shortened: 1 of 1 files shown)


normandy/recipes/models.py
```diff
@@ -77,64 +80,84 @@ class Recipe(models.Model):
...
-        return '<Recipe "{name}">'.format(name=self.name)
 
-    def __str__(self):
-        return self.name
+def multivalue_matcher(field_getter):
+    """Make a matcher that accepts the client if any of the recipe's values accept the client."""
+    def matcher(recipe, client):
+        field_vals = list(field_getter(recipe))
+
+        if field_vals == []:
```
Osmose: Or `not field_vals`

normandy/recipes/models.py
```diff
@@ -77,64 +80,84 @@ class Recipe(models.Model):
...
-        return True
+    return True
 
-    def __repr__(self):
-        return '<Recipe "{name}">'.format(name=self.name)
 
-    def __str__(self):
-        return self.name
+def multivalue_matcher(field_getter):
+    """Make a matcher that accepts the client if any of the recipe's values accept the client."""
```
Osmose: I think this docstring should document that we expect a `matches` function on the values we're comparing and what arguments it should take.

Osmose:
r+wc, nice work!

PR #46, opened 2016-02-26
Use consistent STATICFILES_STORAGE across environments.

Osmose:
collectstatic only adds hashes to static file filenames when
STATICFILES_STORAGE is set correctly. We want the Dockerfile and
other things to have the hashed filenames, but the Dockerfile runs
collectstatic under the Development environment. Also, there's no
problem using the manifest storage during development.

So we move the setting.

@mythmon r?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ _build
 /.cache/
 /etc/ssl
 /__version__/
+/tmp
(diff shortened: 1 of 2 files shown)


mythmon:
I still wonder if we should build the Docker image with as a Production environment instead of Dev, but this is a good work around. r+ when Travis finishes.

PR #49, opened 2016-02-29
Add circle.yml config

mythmon:
This is probably going to be a lot faster. Trying it out. No review needed right now.


mythmon:
Note: This doesn't do deploys yet. I think @relud needs to set that up.

@Osmose r?

Osmose:
arrplus

PR #52, opened 2016-02-29
Store test artifacts for CI processing

mythmon:
@relud r?

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -17,7 +17,10 @@ dependencies:
 
 test:
   override:
-    - docker run --net host -e DJANGO_CONFIGURATION=Test normandy:build py.test
+    - >
+      docker run --net host -e DJANGO_CONFIGURATION=Test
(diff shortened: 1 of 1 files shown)


relud:
you might need -u root

mythmon:
Wait, that's an option? That would have saved me a lot of debugging...

mythmon:
Though, on the other hand running the tests as root might hide errors that we will hit when the servers run as not-root. I'm going to try and work around it for now.

relud:
you could work around it by creating the reports dir and chowning it as app in the dockerfile before the requirements.txt is copied in.

mythmon:
Sort of, except the directory already exists, and is created by CircleCI. I chowned it as a part of the CI set up, and it worked.

relud:
ah, okay, cool

mythmon:
This worked and we got data in the last test run. Merging.

PR #54, opened 2016-03-01
Add docs about how to run Normandy for QA

mythmon:
@Osmose r?

### DIFF
--- a/docs/qa/docker.rst
+++ b/docs/qa/docker.rst
@@ -17,12 +17,11 @@ Preparing an environment file
 -----------------------------
 Normandy gets all of its configuration via environment variables. Docker
 has built in support for this pattern. For this set up, we will be using
-and environment file to specifiy this configuration, and having Docker
+an environment file to specifiy this configuration, and having Docker
 read that file.
 
-The file can be named anything, but canonically is named ``.env``. The
-only thing that is required is a database config, which looks something
-like the below. We'll go over how to fill in the blanks next.
+Make a file named ``.env``, and put the following in it. We'll go over
+how to fill in the blanks next.
 
 .. code-block:: ini
 
@@ -33,13 +32,18 @@ Postgres. ``<db_name>`` is the name of the database to connect to, and
 is usually ``normandy``.
 
 ``<host>`` is more complicated. This needs to be something that the
-Docker container can use to connect to Postgres. There are a few options here:
+Docker container can use to connect to Postgres. This needs to be the IP
+address of the machine running Postgres, on a network that the Docker
+can access. This can be figured out by looking at the IP address of the
+host on the bridge network device.
+
+The defaults for this are *usually* one of these two values:
 
 Linux with local Postgres
-  Use ``172.17.0.1``
+  ``172.17.0.1``
 
 _`Docker Toolbox` with local Postgres (OSX or Windows)
-  Use ``192.168.99.1``
+  ``192.168.99.1``
 
 .. _the Docker Toolbox: https://docs.docker.com/engine/installation/mac/
 
--- a/docs/qa/docker.rst
+++ b/docs/qa/docker.rst
@@ -93,7 +93,7 @@ Running Normandy
      # Create a super user
      docker run -t --env-file=.env mozilla/normandy:latest ./manage.py createsuperuser
      # Run the web server
-     docker run -t -p 8000 --env-file=.env mozilla/normandy:latest
+     docker run -t -p 8000:8000 --env-file=.env mozilla/normandy:latest
 
    This process will run until it is manually stopped with Ctrl-c. Open another
    terminal for the remaining steps.
--- a/docs/qa/docker.rst
+++ b/docs/qa/docker.rst
@@ -88,33 +88,40 @@ Running Normandy
 
    .. code-block:: bash
 
-     CID=$(docker run -d -t mozilla/normandy:latest --env-file=.env)
+     # Initialize the DB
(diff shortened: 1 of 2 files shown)


docs/qa/docker.rst
```diff
@@ -0,0 +1,116 @@
...
+2. `Postgres 9.4`_
+
+.. _Docker: https://docs.docker.com/engine/installation/
+.. _Postgres 9.4: http://www.postgresql.org/
+
+Preparing an environment file
+-----------------------------
+Normandy gets all of its configuration via environment variables. Docker
+has built in support for this pattern. For this set up, we will be using
+and environment file to specifiy this configuration, and having Docker
```
Osmose: "an"

The "and having Docker read that file" part is probably unnecessary.

docs/qa/docker.rst
```diff
@@ -0,0 +1,116 @@
...
+read that file.
+
+The file can be named anything, but canonically is named ``.env``. The
```
Osmose: We should probably just have them name it `.env` rather than bother pointing out that the name is optional.

docs/qa/docker.rst
```diff
@@ -0,0 +1,116 @@
...
+
+``<username>`` and ``<password>`` are the details to authenticate with
+Postgres. ``<db_name>`` is the name of the database to connect to, and
+is usually ``normandy``.
+
+``<host>`` is more complicated. This needs to be something that the
+Docker container can use to connect to Postgres. There are a few options here:
+
+Linux with local Postgres
+  Use ``172.17.0.1``
```
Osmose: If there are commands that can generate these IP addresses we should have them here (and maybe suggest the IP they probably are set to).

Osmose: That looks fine, although there is a docker-machine command for finding the IP, right? Probably good to mention that.

mythmon: The docker-machine command you showed me is the command the find the IP of the *virtual machine*. Here we need the IP of the *host machine* that is accessible to the container.

mythmon:
^ Feedback

docs/qa/docker.rst
```diff
@@ -0,0 +1,116 @@
...
+   same place as ``postgresql.conf``. Add a line referencing the docker
+   network, like this:
+
+   .. code-block:: ini
+
+     host    all             all             172.17.0.0/16           trust
+
+   For Docker Toolbox, use ``192.168.99.0/24``.
+
+Running Normandy
```
Osmose: Can we write a script that does this and outputs the IP, but doesn't run it in the background, so that the person running it can halt it manually? The more like "runserver" this whole process is the better.

Osmose:
r+wc, nice work!

docs/qa/docker.rst
mythmon: We could probably write that script. Though I'm actually unsure if the "get the IP address" instructions work here. On a mac you might have to use the docker-machine command to get the IP and use that instead. Can you test this part?

Most guides about this subject leave off the `-d` flag in the `docker run` command, and have "now open another terminal and run...". That way you can halt it manually. Maybe that is a better approach?

I imagine that most QA people are going to want the more detailed explanation, because they will want to be able to tweak things. Maybe I'm wrong about that.

docs/qa/docker.rst
```diff
@@ -0,0 +1,120 @@
...
+----------------
+1. Download and run the Normandy Docker image:
+
+   .. code-block:: bash
+
+     CID=$(docker run -d -t mozilla/normandy:latest --env-file=.env)
```
Osmose: On OSX the `--env-file` flag cannot come after the `-t` flag. I dunno if it's like that on linux too.

Osmose: Also this doesn't work for me with `docker-machine` without `-p 8000:8000`.

mythmon: /me shakes his fist at Docker Machine

docs/qa/docker.rst
Osmose: > We could probably write that script. Though I'm actually unsure if the "get the IP address" instructions work here. On a mac you might have to use the docker-machine command to get the IP and use that instead. Can you test this part?

You're right, the `docker-machine ip` command gives me the correct IP, the current instructions do not.

> Most guides about this subject leave off the -d flag in the docker run command, and have "now open another terminal and run...". That way you can halt it manually. Maybe that is a better approach?

I like that over encouraging people to leave stuff running in the background.

docs/qa/docker.rst
mythmon: I learned I was wrong about the `-t` flag anyways. It isn't the tag to use. It is "allocate a TTY". The tag to use is a positional argument.

mythmon:
I realized this needed some more info about migrations and super users. I added that in. Please re-r?

After that, I think everything is ready to go.

docs/qa/docker.rst
```diff
@@ -0,0 +1,127 @@
...
+1. Download and run the Normandy Docker image:
+
+   .. code-block:: bash
+
+     # Initialize the DB
+     docker run -it --env-file=.env mozilla/normandy:latest ./manage.py migrate
+     # Create a super user
+     docker run -t --env-file=.env mozilla/normandy:latest ./manage.py createsuperuser
+     # Run the web server
+     docker run -t -p 8000 --env-file=.env mozilla/normandy:latest
```
Osmose: I think you need `-p 8000:8000` here.

Osmose:
r+wc again

PR #55, opened 2016-03-01
Add classifier preview to debug which recipes are sent to a client.

Osmose:
Fixes bug 1240974.

@mythmon r?


normandy/classifier/admin/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Import views so that admin autodiscover picks them up.
+import normandy.classifier.admin.views
```
mythmon: grosssss

normandy/recipes/tests/test_models.py
```diff
@@ -119,3 +120,27 @@ def test_filter_by_country_many(self):
...
+    def test_filter_exclude(self):
+        recipe = RecipeFactory(enabled=False)
+        client = ClientFactory()
+
+        assert not recipe.matches(client)
+        assert recipe.matches(client, exclude=[match_enabled])
+
+    def test_filter_exclude_many(self):
+        locale1, locale2 = LocaleFactory.create_batch(2)
+        locale2 = LocaleFactory()
```
mythmon: You crate `locale2` twice.

normandy/recipes/tests/test_models.py
```diff
@@ -119,3 +120,27 @@ def test_filter_by_country_many(self):
...
+        assert not never_match.matches(client)
+
+    def test_filter_exclude(self):
+        recipe = RecipeFactory(enabled=False)
+        client = ClientFactory()
+
+        assert not recipe.matches(client)
+        assert recipe.matches(client, exclude=[match_enabled])
+
+    def test_filter_exclude_many(self):
```
mythmon: I think this name is a victim of copy/paste, since I don't see anything "many" about this test.

normandy/classifier/admin/views.py
```diff
@@ -0,0 +1,25 @@
...
+    if form.is_valid():
+        client = form.save()
+        bundle = Bundle.for_client(client, exclude=[match_sample_rate])
+    else:
+        bundle = None
+
+    ctx = admin_site.each_context(request)
+    ctx.update({
+        'form': form,
+        'client': client,
```
mythmon: `client` is undefined in the case that `form.is_valid() == False`.

normandy/recipes/tests/test_models.py
Osmose: It's meant to test the exclusion on a multivalue matcher field, but doesn't actually use multiple values. I'll write a better name or make it actually test multiple values.

normandy/classifier/admin/views.py
Osmose: Looks like I should've written a test case for the invalid form. Whoopsie.

Osmose:
Also gotta add the initial_data command to the QA docs that are landing soon

normandy/classifier/admin/__init__.py
mythmon: Oh, wait. Shouldn't this be done with app configs now?

normandy/classifier/admin/forms.py
```diff
@@ -0,0 +1,42 @@
...
+from normandy.recipes.models import Country, Locale, ReleaseChannel
+
+
+FakeClient = namedtuple('FakeClient', (
+    'locale',
+    'country',
+    'request_time',
+    'release_channel',
+    'user_id'
+))
```
mythmon: That we need this is pretty unsatisfying. I wish we could use the normal Client class for this, but that is complex. Maybe `Client` should be dumb and contain no logic, except in a `Client.from_request` constructor.

Out of scope for this PR though, I think.

normandy/classifier/templates/admin/classifier_preview.html
```diff
@@ -0,0 +1,78 @@
...
+{% block content %}
+  <form method="get">
+    {{ form.non_field_errors }}
+    <div class="client-properties">
+      <div class="field">
+        {{ form.locale.errors }}
+        {{ form.locale.label_tag }}
+        {{ form.locale }}
+      </div>
+      <div class="field">
```
mythmon: This stuff is really repetitive. Can it be in a loop, or auto generated by the form?

normandy/classifier/templates/admin/classifier_preview.html
```diff
@@ -0,0 +1,78 @@
...
+        {{ form.request_time.errors }}
+        {{ form.request_time.label_tag }}
+        {{ form.request_time }}
+      </div>
+    </div>
+    <div class="submit-row">
+      <input class="default" type="submit" value="Match Recipes"></input>
+    </div>
+  </form>
+  {% if bundle %}
```
mythmon: I think this should always include a list of matched recipes (if there is input in the above form), and show "No recipes matched" if there aren't any.

normandy/recipes/models.py
```diff
@@ -123,7 +146,7 @@ def match_times(recipe, client):
 
 @Recipe.register_matcher
 def match_sample_rate(recipe, client):
-    if recipe.sample_rate:
+    if recipe.sample_rate is not None:
```
mythmon: Nice catch

normandy/classifier/admin/__init__.py
Osmose: Sort've. The problem is whether `AppConfig.ready` runs before the main urlconf is constructed, because that's when the admin URLs are generated and all ModelAdmins need to be registered before that point. `autodiscover` simply imports all the `app.admin` modules it can find, and which includes this `__init__` file.

Even if we had an `AppConfig` running `autodiscover` we'd have to do this import anyway. It's the one downside to having `admin` be a package instead of a module, and I think it's fine.

normandy/classifier/admin/forms.py
Osmose: Oh, `from_request`. I like that.

normandy/classifier/templates/admin/classifier_preview.html
Osmose: It can only be in a loop if we generate the list of fields in the view due to the limitations of Django templates. The best I think we can do is write an [inclusion tag](https://docs.djangoproject.com/en/1.9/howto/custom-template-tags/#inclusion-tags) and list them out explicitly, still.

Osmose:
Addressed feedback. Again.

normandy/classifier/templates/admin/classifier_preview.html
mythmon: Bleh. Django templates always feel so limiting. Lets just leave it
the way it is now, I think.

Osmose: Tooooooo laaaaaaaate

mythmon:
wooh. r+

PR #57, opened 2016-03-02
Allow SECURE_PROXY_SSL_HEADER to be set via the environment.

Osmose:
@mythmon r?


normandy/settings.py
```diff
@@ -151,6 +151,7 @@ class Development(Base):
 
 class Production(Base):
     """Settings for the production environment."""
+    SECURE_PROXY_SSL_HEADER = values.Value(None)
```
mythmon: This needs to be a `TupleValue`. [docs](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/x-forwarded-headers.html#x-forwarded-proto).

mythmon:
r+

PR #58, opened 2016-03-02
Fix a binding issue with the self-repair driver.

Osmose:
@mythmon r?

### DIFF
--- a/normandy/selfrepair/static/js/self_repair_driver.js
+++ b/normandy/selfrepair/static/js/self_repair_driver.js
@@ -149,14 +149,30 @@ function runRecipe(recipe) {
 }
 
 
+/**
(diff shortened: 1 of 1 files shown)


normandy/selfrepair/static/js/self_repair_driver.js
```diff
@@ -149,14 +149,30 @@ function runRecipe(recipe) {
...
+
 // Actually fetch and run the recipes.
 fetchRecipes().then((recipes) => {
     let chain = Promise.resolve();
 
     for (let recipe of recipes) {
-        chain.then(() => {
-            return runRecipe(recipe);
-        });
+        chain.then(buildRecipeCallback(recipe));
```
mythmon: Why not just `chain.then(runRecipe.bind(null, recipe))`?

PR #60, opened 2016-03-02
Use Python Slim docker base image

mythmon:
Locally, this makes images that are 40% smaller than before (895.2 MB  to 537.7 MB). This should translate into faster CI and faster deploys.

The difference between `python:3.5.1` and `python:3.5.1-slim` is largely that `python:3.5.1` installs several Debian packages we don't care about, including MySQL bindings. This new Dockerfile installs only the ones we need.

Note: this PR and the first deploy is going to have slower CI since it will miss the cache.

@relud r?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,23 @@
-FROM python:3.5.1
+FROM python:3.5.1-slim
 WORKDIR /app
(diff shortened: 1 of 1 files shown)


relud:
r+

mythmon:
The upload to dockerhub went from 2m54s to 1m35s. :100:

Osmose:
:tada:

PR #59, opened 2016-03-02
Add /__health__ API that exposes Django system checks

mythmon:
This makes progress on [bug 1251466](https://bugzilla.mozilla.org/show_bug.cgi?id=1251466)

@Osmose r?

PR #61, opened 2016-03-03
Add database system check

mythmon:
This makes more progress on bug 1251466.

This builds on top of #59. That should be reviewed and merged first.

@Osmose r? after you are done with #59?

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -21,9 +21,6 @@ test:
   override:
     - >
       docker run --net host -e DJANGO_CONFIGURATION=Test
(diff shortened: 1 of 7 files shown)


Osmose:
> On the one hand, running the checks might find some silly errors with models, so could be good to run in CI. On the other hand, I don't know how to fix this, so I wonder if removing the check from CI is reasonable.

I wouldn't add it until we fix it, but I also wouldn't mind having it run. Maybe file a bug about it and remove it?

Osmose:
r+wc, nice work!

PR #63, opened 2016-03-04
add /__lbheartbeat__ so the elb can see the application is responsive

relud:
not sure if it should return `Response('', status=status.HTTP_200_OK)` or just `Response('')` or something else. Expected behavior is to return 200 with an empty body.

### DIFF
--- a/normandy/health/api/views.py
+++ b/normandy/health/api/views.py
@@ -36,7 +36,7 @@ def version(request):
 @api_view(['GET'])
 def lbheartbeat(request):
     # lets the load balancer know the application is running and available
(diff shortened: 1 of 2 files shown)


mythmon:
200 is the default return code, so it is redundant right now, but this is fine.

PS: Technically [`HTTP 204 No Content`](http://http.cat/204) would be semantically better, but I imagine the load balancers don't like that?

relud:
pretty sure any 2xx would be fine.

relud:
I was wrong, docs say it must be 200 :/

mythmon:
r+

PR #69, opened 2016-03-10
Add system check for action hash consistency

mythmon:
This seemed like a sane check to have, plus I got some practice writing system checks.

@Osmose r?

### DIFF
--- a/normandy/recipes/models.py
+++ b/normandy/recipes/models.py
@@ -225,12 +225,15 @@ class Action(models.Model):
     def get_absolute_url(self):
         return reverse('action-detail', args=[self.name])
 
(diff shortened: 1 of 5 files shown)


normandy/recipes/models.py
```diff
@@ -225,12 +225,15 @@ def __str__(self):
     def get_absolute_url(self):
         return reverse('action-detail', args=[self.name])
 
+    def compute_hash(self):
```
Osmose: It's verbose, but maybe `compute_implementation_hash`?

Osmose:
r+, change that attribute name or leave it as you like. Great work!

PR #70, opened 2016-03-10
Add migration system check

mythmon:
fixes bug 1255583

@Osmose r?

### DIFF
--- a/normandy/health/checks.py
+++ b/normandy/health/checks.py
@@ -1,8 +1,9 @@
-from django.core.checks import Error, register as register_check
+from django.core.checks import Info, Warning, Error, register as register_check
 from django.db import connection
 from django.db.utils import OperationalError
 
 
+WARNING_UNAPPLIED_MIGRATION = 'normandy.health.W001'
 ERROR_CANNOT_CONNECT_DATABASE = 'normandy.health.E001'
 ERROR_UNUSABLE_DATABASE = 'normandy.health.E002'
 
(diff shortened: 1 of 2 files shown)


normandy/health/checks.py
```diff
@@ -21,5 +22,31 @@ def database_connected(app_configs, **kwargs):
     return errors
 
 
+def migrations_applied(app_configs, **kwargs):
+    from django.db.migrations.loader import MigrationLoader
+    errors = []
+
+    # Load migrations from disk/DB
+    loader = MigrationLoader(connection, ignore_no_migrations=True)
```
Osmose: Interestingly, instead of getting the graph of migrations, [Django generates a "migration plan"](https://github.com/django/django/blob/1845bc1d1007751a7f65c66aeddc35f032f6bf41/django/core/management/base.py#L430) to get a list of unapplied migrations.

The migration API is sufficiently undocumented that I have no opinion on which method is better. I'm curious, did you choose your method specifically for some reason?

Osmose:
r+, tested locally and it works like a charm. Nice job!

normandy/health/checks.py
mythmon: I based this code on the [showmigrations management command](https://github.com/django/django/blob/1845bc1d1007751a7f65c66aeddc35f032f6bf41/django/core/management/commands/showmigrations.py). I didn't see the method you linked to.

PR #71, opened 2016-03-14
Make migration check catch db errors

mythmon:
Checks generally need to be resilient to the DB being down so that the heartbeat page works when the DB is down. This makes the migration check do that.

When we figure out how to run the checks in CI, we should make them test this case too.

@Osmose r?


Osmose:
r+

PR #73, opened 2016-03-15
Try to make no database queries in fetch_bundle

mythmon:
This doesn't completely eliminate the database accesses in fetch_bundle, but it reduces them. The rest of the reduction will come after the make the relation between `Recipe`s and `Action`s not use a through-table. At that point, this might just work.

@Osmose r?

### DIFF
--- a/normandy/classifier/tests/test_api.py
+++ b/normandy/classifier/tests/test_api.py
@@ -1,3 +1,6 @@
+from django.db import connection
+from django.test.utils import CaptureQueriesContext
+
 import pytest
 from rest_framework.reverse import reverse
 
@@ -13,7 +16,7 @@ class TestFetchBundleAPI(object):
         assert res.status_code == 200
         assert res.data == {'recipes': []}
 
-    def test_it_serves_recipes(self, client):
+    def test_it_serves_recipes(self, client, django_cache):
         recipe_action = RecipeActionFactory(arguments={'foo': 'bar'})
         res = client.post('/api/v1/fetch_bundle/')
 
@@ -37,7 +40,7 @@ class TestFetchBundleAPI(object):
             ],
         }]
 
-    def test_it_filters_by_locale(self, client):
+    def test_it_filters_by_locale(self, client, django_cache):
         english = LocaleFactory(code='en-US')
         german = LocaleFactory(code='de')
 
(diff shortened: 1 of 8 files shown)


normandy/classifier/tests/test_api.py
```diff
@@ -13,7 +16,7 @@ def test_it_works(self, client):
         assert res.status_code == 200
         assert res.data == {'recipes': []}
 
-    def test_it_serves_recipes(self, client):
+    def test_it_serves_recipes(self, client, django_cache):
```
Osmose: Does this trigger some sort've mocking?

normandy/classifier/tests/test_api.py
```diff
@@ -52,3 +55,22 @@ def test_it_filters_by_locale(self, client):
...
+        recipe_action = RecipeActionFactory(recipe__enabled=True)
+        recipe = recipe_action.recipe
+
+        # Warm up the cache
+        res1 = client.post('/api/v1/fetch_bundle/', {'locale': 'de'})
+        assert res1.status_code == 200
+        assert res1.data['recipes'][0]['name'] == recipe.name
+
+        # Fire!
+        queries = CaptureQueriesContext(connection)
```
Osmose: This is so cool!

normandy/classifier/tests/test_api.py
```diff
@@ -52,3 +55,22 @@ def test_it_filters_by_locale(self, client):
...
+        with queries:
+            res2 = client.post('/api/v1/fetch_bundle/', {'locale': 'de'})
+            assert res2.status_code == 200
+        assert res1.data == res2.data
+        assert (len(queries), list(queries)) == (0, [])
```
Osmose: Why is just checking for length not enough?

normandy/conftest.py
```diff
@@ -11,3 +14,15 @@ def api_client():
     client = APIClient()
     client.force_authenticate(user=user)
     return client
+
+
+@pytest.yield_fixture()
+def django_cache(request):
```
Osmose: This works, but I wonder if we could make a decorator instead like the database one? If not, or if that'd be a pain, or if that'd be a very un-pytesty way to do this, then don't bother and keep this.

Osmose:
r+, merge after you've made changes or decided not to make them. Nice work!

normandy/classifier/tests/test_api.py
mythmon: Oh, this was mostly for debugging. Doing it this way showed me both a counter of how many tests were remaining, and also what the problematic queries were. I think this should just be comparing to the empty list (which gives really useful debugging when the test fails).

normandy/conftest.py
mythmon: I tried to make a marker like `@pytest.market.django_db` (which, btw, just turns around and uses a fixture). I couldn't get it to work though. I can learn more about markers and try to get it working that way.

normandy/classifier/tests/test_api.py
mythmon: It is! It is also a little verbose. I think I might want to add a little helper that makes it more straightforward. For example, not needing to grab the Django DB connection object would be nice.

PR #75, opened 2016-03-16
[bug 1253716] Salt hashes with revision id

rehandalal:
@mythmon r?


normandy/recipes/models.py
```diff
@@ -127,6 +128,12 @@ def __repr__(self):
     def __str__(self):
         return self.name
 
+    def save(self, *args, **kwargs):
+        if self.revision_id:
+            self.revision_id += 1
```
mythmon: This will result in nothing ever having a positive `revision_id`, right? Since the default is 0? That doesn't seem right.

mythmon:
Can you add a test that `revision_id` gets incremented on save, and also that it is included in the hash? These would be useful as regression tests in the future.

rehandalal:
Test added ^

normandy/recipes/models.py
```diff
@@ -129,7 +129,7 @@ def __str__(self):
         return self.name
 
     def save(self, *args, **kwargs):
-        if self.revision_id:
+        if self.revision_id is not None:
```
mythmon: When would the `revision_id` be `None`, anyways? The default is 0, and the database doesn't allow nulls, right?

rehandalal: yeah i guess so. I'll ditch the line.

rehandalal:
updated ^

mythmon:
r+

PR #77, opened 2016-03-17
Use settings.CDN_URL for implementation URLs

mythmon:
Fixes bug 1257007.

@Osmose r?

### DIFF
--- a/normandy/base/checks.py
+++ b/normandy/base/checks.py
@@ -1,0 +1,19 @@
+from django.conf import settings
+from django.core.checks import Warning, register as register_check
+
+
+WARNING_MISCONFIGURED_CDN_URL = 'normandy.base.W001'
+
+
+def setting_cdn_url_has_trailing_slash(app_configs, **kwargs):
+    errors = []
+
+    if settings.CDN_URL is not None and settings.CDN_URL[-1] != '/':
+        msg = 'The setting CDN_URL should end in a slash'
+        errors.append(Warning(msg, id=WARNING_MISCONFIGURED_CDN_URL))
+
+    return errors
+
+
+def register():
+    register_check(setting_cdn_url_has_trailing_slash)
--- a/normandy/base/apps.py
+++ b/normandy/base/apps.py
@@ -1,0 +1,12 @@
+from django.apps import AppConfig
+
+from normandy.base import checks
+
+
+class BaseApp(AppConfig):
+    name = 'normandy.base'
+    label = 'base'
+    verbose_name = 'Normandy'
+
+    def ready(self):
+        checks.register()
--- a/normandy/recipes/api/fields.py
+++ b/normandy/recipes/api/fields.py
@@ -1,5 +1,6 @@
 from contextlib import closing
 
+from django.conf import settings
 from django.core.files.base import ContentFile
 
 from rest_framework.fields import CharField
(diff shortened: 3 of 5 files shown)


normandy/base/api/relations.py
```diff
@@ -0,0 +1,57 @@
...
+        if settings.CDN_URL is None:
+            url = request.build_absolute_uri(url)
+        else:
+            assert url[0] == '/'
+            url = settings.CDN_URL + url[1:]
+
+        return preserve_builtin_query_params(url, request)
+
+
+class CDNHyperlinkedIdentityField(CDNHyperlinkedRelatedField):
```
Osmose: Instead of re-implementing this entire class, could you write your change as a mixin and apply it to classes that inherit from the original related and identity fields.

normandy/base/api/relations.py
```diff
@@ -0,0 +1,57 @@
...
+
+class CDNHyperlinkedRelatedField(HyperlinkedRelatedField):
+
+    def __init__(self, **kwargs):
+        # super is going to override self.reverse, so stash it and restore it
+        reverse = self.reverse
+        super().__init__(**kwargs)
+        self.reverse = reverse
+
+    def reverse(self, viewname, args=None, kwargs=None, request=None, format=None, **extra):
```
Osmose: Overriding reverse is weird; [the code says](https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/relations.py#L254) it's only a property on the class for testing purposes.

I think it's more natural to override `get_url`. You can even avoid code duplication completely by setting the `request` argument to `None` if `CDN_URL` is filled before calling the supermethod; `request` is only ever used for building absolute URLs. If you don't include it, you'll get a non-absolute URL that you can then prepend the CDN url to.

That might be a bit unclear. I think it'd look like:

```python
def get_url():  # pretend there are args here
    if settings.CDN_URL is not None:
        request = None
        url = super().get_url(args_with_modified_request)
        # Append the CDN url and you're done!
    else
        return super().get_url()
```

normandy/base/api/relations.py
```diff
@@ -0,0 +1,57 @@
...
+        if format is not None:
+            kwargs = kwargs or {}
+            kwargs['format'] = format
+
+        url = django_reverse(viewname, args=args, kwargs=kwargs, **extra)
+        if settings.CDN_URL is None:
+            url = request.build_absolute_uri(url)
+        else:
+            assert url[0] == '/'
+            url = settings.CDN_URL + url[1:]
```
Osmose: Should probably use `urljoin` here and avoid the assert.

normandy/base/tests/test_api_relations.py
```diff
@@ -0,0 +1,37 @@
...
+    class DummySerializer(ModelSerializer):
+        class Meta:
+            model = Action
+
+        self_url = CDNHyperlinkedIdentityField(view_name='action-detail', lookup_field='name')
+
+    def test_no_setting(self, settings, rf):
+        settings.CDN_URL = None
+        request = rf.get('/')
+        obj = ActionFactory()
```
Osmose: Call an action an action.

normandy/base/api/relations.py
mythmon: [`urllib.parse.urljoin`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin) will see the leading `/` in `url` and treat it as a absolute path, stripping out any path components in `CDN_URL`.

That is, if `CDN_URL` is `https://example.com/cdn/`, then `urljoin` will strpi out the `cdn/` part.

mythmon:
@Osmose tag, you're it

normandy/base/checks.py
```diff
@@ -0,0 +1,19 @@
...
+
+WARNING_MISCONFIGURED_CDN_URL = 'normandy.base.W001'
+
+
+def setting_cdn_url_has_trailing_slash(app_configs, **kwargs):
+    errors = []
+
+    if settings.CDN_URL is not None and settings.CDN_URL[-1] != '/':
+        msg = 'The setting CDN_URL should end in a slash'
+        errors.append(Warning(msg, id=WARNING_MISCONFIGURED_CDN_URL))
```
Osmose: I think this should be an error instead of a warning.

Osmose:
r+wc, nice work!

PR #78, opened 2016-03-17
Fix bug 1257263: Only one action per recipe.

Osmose:
The changeset is large, but the vast majority of it is test fixes. Things to note:

- I'm tempted to rename `Action` to `Implementation` in the future given what the API looks like now, or to change that key in the API to `action` instead. Feedback on this point is welcome.
- The changes to the order of fields in the admin were made from feedback that @gregglind gave to me in person.
- Turns out the adminsortable library was what added FontAwesome to the Django admin automatically. Without it, we need to include FontAwesome manually. Laaaaaaaame.

@rehandalal r?


mythmon:
Test failure seem related.

normandy/classifier/tests/test_api.py
```diff
@@ -56,11 +51,26 @@ def test_it_filters_by_locale(self, client):
         recipe_names = set(r['name'] for r in res.data['recipes'])
         assert recipe_names == {'german', 'any', 'both'}
 
+    def test_it_filters_by_locale_with_json(self, api_client):
```
mythmon: Why did you add this test? Is there something special about the json output of this api?

Osmose: This is to test [the fix in the FetchBundle view](https://github.com/mozilla/normandy/pull/78/files/0f39707be504fe890e761ed05499cf94b3a84b2e#diff-e8a07d3ca4f0d4fd6d4ea0aadb2180ddR27) where we were pulling from `request.POST` instead of `request.data`. We want to pull from `request.data` to support different encodings, like JSON. I have it reusing the filter test because the other tests only test when nothing is sent, and would pass even if data was sent as JSON because the locale param would default to None.

mythmon: Oh, interesting. So the main point here is that it uses both POST data and query strings. Maybe a comment about that would be interesting.

rehandalal: +1 for comment

Osmose:
Comment added.

rehandalal:
lgtm. r+

PR #79, opened 2016-03-21
[Bug 1253717] Keep history of changes to Actions / Recipes

brittanystoroz:
I added the [django-reversion](http://django-reversion.readthedocs.org/en/latest/) library & its middleware to keep track of changes to recipes/actions as per [bug 1253717](https://bugzilla.mozilla.org/show_bug.cgi?id=1253717). You'll have to run `manage.py migrate` and `manage.py createinitialrevisions` to populate the new reversion tables. 

There were a couple of options listed for creating revisions with django-reversion. I went with the middleware but it did come with this [warning](http://django-reversion.readthedocs.org/en/latest/api.html?highlight=json#revisionmiddleware). Do we care about that? I don't totally understand what it means.

We can also add extra metadata to our revisions if we want. As is, it gives us all the serialized data for each revision of a recipe/action so I think we've got what we need, but let me know if there's anything else we might want to keep track of. Each row has its own `revision_id` field which could maybeee get confusing with the `revision_id` field that was added to recipes in Rehan's recent [pull request for salt hashes](https://github.com/mozilla/normandy/pull/75/files)? Thoughts?

I haven't added any sort of front-end for this information yet - just wanted to make sure this was on the right track first.  I can file a new bug and PR for that assuming the work here is all good.

### DIFF
--- a/normandy/recipes/admin.py
+++ b/normandy/recipes/admin.py
@@ -5,9 +5,11 @@ from normandy.base.admin import site as admin_site
 from normandy.recipes import models
 from normandy.recipes.forms import ActionAdminForm, RecipeAdminForm
 
+from reversion.admin import VersionAdmin
+
 
 @admin.register(models.Recipe, site=admin_site)
-class RecipeAdmin(admin.ModelAdmin):
+class RecipeAdmin(VersionAdmin):
     form = RecipeAdminForm
     list_display = [
         'name',
@@ -54,7 +56,7 @@ class RecipeAdmin(admin.ModelAdmin):
 
 
 @admin.register(models.Action, site=admin_site)
-class ActionAdmin(admin.ModelAdmin):
+class ActionAdmin(VersionAdmin):
     form = ActionAdminForm
     list_display = ['name', 'implementation_hash', 'in_use']
     fieldsets = [
--- a/normandy/recipes/models.py
+++ b/normandy/recipes/models.py
@@ -68,6 +68,7 @@ class Country(models.Model):
         return self.code == client_val or self.name == client_val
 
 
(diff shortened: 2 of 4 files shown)


Osmose:
I think this looks fine. Nice work!

The only thing missing is the admin integration. Eventually we'll write our own frontend to this data that views it how we want, but for now using the provided admin integration will be good for shipping this now. 

> There were a couple of options listed for creating revisions with django-reversion. I went with the middleware but it did come with this warning. Do we care about that? I don't totally understand what it means.

I don't think we care about that. The only risk I can think of is that if we're saving revisions in a separate transaction than saving the models themselves, it's possible for saving the revisions to fail without rolling back saving the models. I think that might actually be fine behavior; an error would still presumably be raised, but an error in revisions wouldn't block us from putting out changes. We'd just have some glitches in our audit trail until we fix the issue, which is fine IMO.

> We can also add extra metadata to our revisions if we want. As is, it gives us all the serialized data for each revision of a recipe/action so I think we've got what we need, but let me know if there's anything else we might want to keep track of. 

I can't think of anything else we want to track right now. Eventually we might be able to leverage this for the approval system, but that might not be a great idea. Who knows!

> Each row has its own revision_id field which could maybeee get confusing with the revision_id field that was added to recipes in Rehan's recent pull request for salt hashes? Thoughts?

Nah that's fine, I think.

brittanystoroz:
@Osmose admin integration added, circleci failing on unrelated dependencies ... =|

Osmose:
Yeah @mythmon is fixing those failures at this very moment.

Osmose:
r+, nice work!

PR #80, opened 2016-03-21
[bug 1257660] Include implementation hash in its URL.

Osmose:
Turns out our JavaScript/Plaintext renderers don't handle 404s properly. Also they were in a file named `renders.py`. Fixed all that.

I had to move to using a separate view in order to accept more than one argument in the URL.

@mythmon r?

### DIFF
--- a/normandy/base/api/renderers.py
+++ b/normandy/base/api/renderers.py
@@ -8,12 +8,12 @@ class TextRenderer(renderers.BaseRenderer):
     def render(self, data, media_type=None, renderer_context=None):
         response = renderer_context.get('response') if renderer_context else None
         if response and response.exception:
-            return self.render_error(data)
+            data = self.render_error(data)
 
         return data.encode(self.charset)
 
     def render_error(self, data):
-        return data['detail'].encode(self.charset)
+        return data['detail']
 
 
 class JavaScriptRenderer(TextRenderer):
@@ -21,4 +21,4 @@ class JavaScriptRenderer(TextRenderer):
     format = 'js'
 
     def render_error(self, data):
-        return '/* {} */'.format(data['detail'])
+        return '/* {} */'.format(super().render_error(data))
--- a/normandy/base/api/renderers.py
+++ b/normandy/base/api/renderers.py
@@ -6,9 +6,19 @@ class TextRenderer(renderers.BaseRenderer):
     format = 'txt'
 
     def render(self, data, media_type=None, renderer_context=None):
+        response = renderer_context.get('response') if renderer_context else None
+        if response and response.exception:
+            return self.render_error(data)
+
         return data.encode(self.charset)
+
+    def render_error(self, data):
+        return data['detail'].encode(self.charset)
 
 
 class JavaScriptRenderer(TextRenderer):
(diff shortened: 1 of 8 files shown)


normandy/recipes/api/views.py
```diff
@@ -37,7 +37,21 @@ def update(self, request, *args, **kwargs):
...
 
-    @detail_route(permission_classes=[], renderer_classes=[JavaScriptRenderer])
-    def implementation(self, request, name=None):
-        obj = self.get_object()
-        return Response(obj.implementation)
+
+class ActionImplementationView(generics.RetrieveAPIView):
+    """
+    Retrieves the implementation code for an action. Raises a 404 if the
+    given hash doesn't match the has we've stored.
```
mythmon: typo: has -> hash

normandy/recipes/api/views.py
```diff
@@ -37,7 +37,21 @@ def update(self, request, *args, **kwargs):
...
+    """
+    queryset = Action.objects.all()
+    lookup_field = 'name'
+
+    permission_classes = ()
+    renderer_classes = (JavaScriptRenderer,)
```
mythmon: Personally, I prefer to use lists here instead of tuples. I think Django 1.8 started recommending that for settings too. But it doesn't really matter.

normandy/recipes/api/serializers.py
```diff
@@ -1,13 +1,12 @@
...
+from normandy.recipes.api.fields import ActionImplementationHyperlinkField
 from normandy.recipes.models import Action, Recipe, RecipeAction
 
 
 class ImplementationSerializer(serializers.Serializer):
     hash = serializers.CharField(source='implementation_hash', read_only=True)
-    url = serializers.HyperlinkedIdentityField(
-        view_name='action-implementation',
-        lookup_field='name')
+    url = ActionImplementationHyperlinkField(view_name='action-implementation')
```
mythmon: How do you feel about hard coding (or defaulting) `view_name='action-implementation'` here, considering it is in the class name?

normandy/base/api/renderers.py
```diff
@@ -6,9 +6,19 @@ class TextRenderer(renderers.BaseRenderer):
...
+    def render_error(self, data):
+        return data['detail'].encode(self.charset)
+
 
 class JavaScriptRenderer(TextRenderer):
     media_type = 'application/javascript'
     format = 'js'
+
+    def render_error(self, data):
+        return '/* {} */'.format(data['detail'])
```
mythmon: How about this?

```python
return '/* {} */'.format(super.render_error(data))
```

normandy/recipes/api/serializers.py
Osmose: I tried adding it as a class attribute like the docs suggested and I got errors about `view_name` being a required kwarg. I'll look into it more.

mythmon: Oh, I was thinking of

    def __init__(self, ..., view_name='action-implementation')

Osmose:
Added fixes and rebased.

mythmon:
r+

PR #83, opened 2016-03-24
New action runner

Osmose:
This is a parallel PR to https://github.com/mozilla/normandy-actions/pull/4, adding support for the new method of loading and executing actions. It simplifies a fair amount of code, which is nice.

I'm avoiding documenting any of these changes yet until the API is settled. I also need to decide whether the documentation should live in the normandy service docs, or the action repo docs. Or somewhere else. I dunno.

Also I had to include two new JS libraries, so it was time to switch to something better for managing our frontend dependencies. I'm fairly sure this should work so long as the docker image ships with NPM, but I'm not 100% on that. Guidance on this would be appreciated.

@mythmon r?


Osmose:
Welp, looks like npm doesn't exist according to the test failure. Will look into it.

normandy/selfrepair/static/js/normandy_driver.js
```diff
@@ -0,0 +1,85 @@
...
+
+Object.assign(LocalStorage.prototype, {
+    _key(key) {
+        return `${this.prefix}-${key}`;
+    },
+
+    getItem(key) {
+        return new Promise(resolve => {
+            resolve(localStorage.getItem(this._key(key)))
+        });
```
mythmon: You can use [`Promise.resolve`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve) as a shorthand for this:

```javascript
getItem(key) {
    return Promise.resolve(localStorage.getItem(this._key(key));
}
```

mythmon: Hmm... but that has slightly different semantics, since the get happens immediately instead of async. Nevermind, this should probably stay the way it is.

mythmon:
I didn't test this, but it looks fine to me. We'll need to get NPM working, but that's it.

brittanystoroz:
I played around with this locally and it worked well for me :+1:

Dockerfile
```diff
@@ -2,7 +2,13 @@ FROM python:3.5.1-slim
...
 RUN groupadd --gid 1001 app && useradd -g app --uid 1001 --shell /usr/sbin/nologin app
 RUN apt-get update && apt-get install -y --no-install-recommends \
-    gcc libpq-dev
+    gcc libpq-dev curl apt-transport-https
+
+# Install node from NodeSource
+RUN curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add - && \
+    echo 'deb https://deb.nodesource.com/node_5.x jessie main' > /etc/apt/sources.list.d/nodesource.list && \
+    echo 'deb-src https://deb.nodesource.com/node_5.x jessie main' >> /etc/apt/sources.list.d/nodesource.list && \
+    apt-get update && apt-get install -y nodejs
```
mythmon: This causes us to run apt-get update and install twice. The curl and echo stuff should be moved above line 4, and the install nodejs bit should be on line 5.

mythmon: Bleh. I missed the part about transport-https. Laaaaame

Dockerfile
```diff
@@ -2,7 +2,13 @@ FROM python:3.5.1-slim
 WORKDIR /app
 RUN groupadd --gid 1001 app && useradd -g app --uid 1001 --shell /usr/sbin/nologin app
 RUN apt-get update && apt-get install -y --no-install-recommends \
-    gcc libpq-dev
+    gcc libpq-dev curl apt-transport-https
+
+# Install node from NodeSource
+RUN curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add - && \
+    echo 'deb https://deb.nodesource.com/node_5.x jessie main' > /etc/apt/sources.list.d/nodesource.list && \
+    echo 'deb-src https://deb.nodesource.com/node_5.x jessie main' >> /etc/apt/sources.list.d/nodesource.list && \
```
mythmon: I think we should use node 4.x instead of 5.x

mythmon:
r+

PR #87, opened 2016-03-24
Change sample rate to a float

mythmon:
Fixes bug 1259556.

@Osmose r?


Osmose:
Test failures seem to be due to the removal of PercentField. Gotta keep it around like the other older fields until we do a migration squash.

Otherwise I'm r+ on this.

PR #89, opened 2016-03-25
Refactor Client, and read user_id from request data, if provided

mythmon:
The required part of this is pretty easy, just another field on `Client` and a little logic in the recipe runner. The more complex change isn't strictly needed, and could be excised if it is hated.


mythmon:
@Osmose r?

PR #90, opened 2016-03-28
Switch to bcrypt password hashing

mythmon:
Fixes bug 1257257.

@Osmose r?


requirements.txt
```diff
@@ -191,3 +191,13 @@ greenlet==0.4.9 \
     --hash=sha256:58b2f3a2e7075c655616bf95e82868db4980f3bb6661db70ad02a51e4ddd2252
 django-npm==0.1.4 \
     --hash=sha256:4ace771ef76bf6a9430278d0f23a23d1a7083ec2b00240c4400d3c5a5cd7cea6
+bcrypt==2.0.0 \
```
rehandalal: I'm guessing #91 doesn't affect these?

mythmon: #91 purposely didn't do anything with these new requirements, but these are only the hashes that the patched version of Hashin would have added. So the two PRs should be compatible.

Osmose:
r+, nice work!

PR #91, opened 2016-03-28
Trim hashes

mythmon:
Using the modified version of Hashin found in peterbe/hashin#14, I eliminated a bunch of hashes we don't need.


rehandalal:
lgtm. r+

PR #92, opened 2016-03-29
Add cache-control headers to implementation API

mythmon:
This will make the CDN happy. And really everyone who wants this service to be fast fast fast. I confirmed that Cloudfront (the CDN we will be using) will pay attention tot his cache header.

@Osmose r?

### DIFF
--- a/normandy/recipes/api/views.py
+++ b/normandy/recipes/api/views.py
@@ -54,4 +54,9 @@ class ActionImplementationView(generics.RetrieveAPIView):
         if impl_hash != action.implementation_hash:
             raise NotFound('Hash does not match current stored action.')
 
-        return Response(action.implementation)
+        max_age = 60 * 60 * 24 * 365  # one year in seconds
+        headers = {
+            'Cache-Control': 'public, max-age={}'.format(max_age),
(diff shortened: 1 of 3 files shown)


normandy/recipes/api/views.py
```diff
@@ -54,4 +54,9 @@ def retrieve(self, request, name, impl_hash):
         if impl_hash != action.implementation_hash:
             raise NotFound('Hash does not match current stored action.')
 
-        return Response(action.implementation)
+        max_age = 60 * 60 * 24 * 365  # one year in seconds
```
Osmose: Setting?

normandy/recipes/api/views.py
```diff
@@ -54,4 +54,9 @@ def retrieve(self, request, name, impl_hash):
...
+        headers = {
```
Osmose: Would the `cache_control` decorator work here instead?

Osmose:
r+wc

normandy/recipes/api/views.py
mythmon: TIL

PR #95, opened 2016-03-31
[Bug 1257267] Add new control admin

brittanystoroz:
This adds a super bare-bones, not-cute-yet control app to replace the default admin. I've also implemented some of the authentication logic but it's really buggy and needs some help. There's still a lot to do on this but feel free to pick at it in the meantime. I'll be back on later tonight to update this with more notes and thoughts and questions.

### DIFF
--- a/normandy/control/static/control/admin/sass/control.scss
+++ b/normandy/control/static/control/admin/sass/control.scss
@@ -1,75 +1,100 @@
+@import 'partials/layout';
+
+
 body {
-  background: #f8f8f8;
+  background: $cream;
 }
 
 #header {
-  background: #fff;
-  border-bottom: 1px solid #ddd;
-  box-shadow: 0px 3px 3px #ddd;
-  color: #555;
-  height: 40px;
+  background: $darkRed;
+  border-bottom: 1px solid $darkOrange;
+  color: #fff;
+  height: 45px;
+  line-height: 25px;
+  padding: 10px 40px;
+
+  h1 {
+    font-size: 16px;
+
+    a:link, a:visited {
+      color: $cream;
+    }
+  }
 }
 
 #user-tools {
+  text-transform: none;
+
   a:link, a:visited {
-    color: #ff647f;
+    border-bottom: 0px;
+    color: $orange;
     font-size: 11px;
-    font-weight: 400;
-    text-transform: none;
+    font-weight: 600;
+    text-transform: uppercase;
   }
 
   a:hover {
-    color: #447e9b;
-    text-decoration: none;
     border-bottom: 0px;
+    color: #fff;
+    text-decoration: none;
   }
 }
 
-#branding h1, #branding h1 a:link, #branding h1 a:visited {
-  color: #555;
-  font-size: 16px;
-}
-
-/* just hiding these for now, they might be a little overkill */
 div.breadcrumbs {
-  background: rgba(0,0,0,0);
+  background: $orange;
+  border-bottom: 1px solid #fff;
+  color: $darkOrange;
   font-size: 13px;
-  padding: 8px 40px 5px 40px;
-}
+  font-weight: 600;
+  line-height: 30px;
+  margin-bottom: 35px;
+  padding: 0px 40px;
 
-div.breadcrumbs a {
-  color: rgb(152, 102, 14);
+  a:link, a:visited {
+    color: $darkOrange;
+    font-weight: 400;
+    opacity: 0.75;
+  }
+
+  a:hover {
+    opacity: 1;
+  }
+
+  .separator {
+    display: inline-block;
+    font-weight: 300;
+    padding: 0px 5px;
+  }
+
 }
 
 h2 {
-  background: #ff647f;
-  border-bottom: 1px solid #e7526c;
-  border-top-left-radius: 1px;
-  border-top-right-radius: 1px;
-  color: #fff;
-  font-size: 16px;
-  font-weight: 400;
-  margin: 1em 0 .5em 0;
-  padding: 8px 20px;
-}
+  border-bottom: 1px solid $orange;
+  color: $darkBrown;
+  font-size: 20px;
+  font-weight: 300;
+  margin: 0px;
+  padding: 8px 0px;
 
-ul {
-  padding: 0px 20px;
-}
-
-ul li {
-  list-style-type: none;
-  padding: 5px 0px;
-}
-
-.box {
-  padding: 0 0 20px 0;
-  max-width: 40%;
-  border-bottom-left-radius: 1px;
-  border-bottom-right-radius: 1px;
-  background-color: #fff;
+  span.bold {
+    font-weight: 600;
+  }
 }
 
 #content {
-  padding: 50px 70px;
+  margin: 40px;
+  min-height: 100%;
+  padding: 0px;
+
+  ul.table-list {
+    margin: 0px;
+    padding: 0px;
+
+    li {
+      background: #fff;
+      list-style-type: none;
+      margin-bottom: 1px;
+      padding: 15px 30px;
+    }
+  }
 }
--- a/normandy/control/static/control/admin/sass/control.scss
+++ b/normandy/control/static/control/admin/sass/control.scss
@@ -10,17 +10,19 @@ body {
   height: 40px;
 }
 
-#user-tools a:link, #user-tools a:visited {
-  color: #ff647f;
-  font-size: 11px;
-  font-weight: 400;
-  text-transform: none;
-}
+#user-tools {
+  a:link, a:visited {
+    color: #ff647f;
+    font-size: 11px;
+    font-weight: 400;
+    text-transform: none;
+  }
 
-#user-tools a:hover {
-  color: #447e9b;
-  text-decoration: none;
-  border-bottom: 0px;
+  a:hover {
+    color: #447e9b;
+    text-decoration: none;
+    border-bottom: 0px;
+  }
 }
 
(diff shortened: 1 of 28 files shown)


brittanystoroz:
Things still to do:

[ ] add less/sass and pipeline to compile assets
[ ] switch to jinja2 templates
[ ] cleanup the base_site.html & base.html control templates (they can maybe be combined, and there is a lot of copypasta from the original base templates leftover)

normandy/control/views.py
```diff
@@ -0,0 +1,18 @@
+from django.views import generic
+from django.contrib.auth.mixins import LoginRequiredMixin
+
+from normandy.recipes.models import Recipe
+
+
+class IndexView(LoginRequiredMixin, generic.ListView):
```
brittanystoroz: How do we feel about using these built-in generic views (like generic.ListView)? Are they going to be too limiting for us? Tough to tell how restrictive they are at first glance.

normandy/settings.py
```diff
@@ -12,6 +12,7 @@ class Core(Configuration):
     INSTALLED_APPS = [
         'normandy.base.apps.BaseApp',
         'normandy.classifier',
+        'normandy.control.apps.ControlApp',
```
brittanystoroz: What's the big difference between how classifier is defined and how the rest of the apps are defined? `normandy.classifier` vs `normandy.control.apps.ControlApp`.  Should this be defined more like classifier, since it's really just going to be some views and templates built off of pre-existing apps?

normandy/settings.py
```diff
@@ -179,6 +180,7 @@ class Base(Core):
     CAN_EDIT_ACTIONS_IN_USE = values.BooleanValue(False)
     ADMIN_ENABLED = values.BooleanValue(True)
     ACTION_IMPLEMENTATION_CACHE_TIME = values.IntegerValue(60 * 60 * 24 * 365)
+    LOGIN_REDIRECT_URL = values.Value('/control/')
```
brittanystoroz: Is there a better place in the file to put this?

normandy/settings.py
mythmon: `normandy.control.apps.ControlApp` is the newer style of installing Apps in Django. It should be preferred. classifier is using the old version because we were lazy and haven't yet needed any features provided by the newer version.

normandy/settings.py
mythmon: I'm not convinced this needs to be a setting, but if it should be a setting, this is where it should go.

normandy/control/views.py
mythmon: My general impression of class-based-views (CBV)  and Django generics are that CBVs are a great idea, and that Django's CBVs often don't do things the way I would like. However, it seems like what you have here exactly matches what the generics are doing, so might as well use them. We can always move away from them in the future, if needed.

brittanystoroz:
The registration/login stuff is tricky the way I have it working now. I'm either breaking the auth for the default admin site or having the control app just direct to the default admin/auth templates. (Meaning they will look like the default admin site, not our new one.) 

I was able to override the default auth templates by adding a `/registration` directory under the control app, and adding the auth urls to my [urlconf](https://github.com/brittanystoroz/normandy/blob/control-admin/normandy/control/urls.py#L8). This allowed me to switch out the base templates they were extending so that they would look like the new control admin. This is where I ran into problems with the default admin login pages inheriting the control templates rather than the originals. (Maybe we don't care.)

I tried to fix this by adding an `extra_content` field to my urlconf [urlconf](https://github.com/brittanystoroz/normandy/blob/control-admin/normandy/control/urls.py#L8) that specified a base template variable that I could use to adjust where the registration templates were inheriting from. (Only implemented on the password change form for now, [like this](https://github.com/brittanystoroz/normandy/blob/control-admin/normandy/control/templates/registration/password_change_form.html#L1).) This seemed to work great for the control templates but then completely broke the default admin templates because that variable wasn't being picked up for some reason, even though I had added it to `normandy/urls.py` as well.

brittanystoroz:
Soo I added pipeline & SASS which seems to be working alright locally, but I had to adjust the `STATICFILES_STORAGE` setting that was previously being used by whitenoise. I had the [whitenoise storage inherit from the pipeline storage](https://github.com/mozilla/normandy/pull/95/commits/65dc089065316a546ec0055b52de608065353827), which seems to work great locally but I'm not totally sure what other implications that might have. Anything I should be concerned about?  Is there any particular environment in `settings.py` where `STATICFILES_STORAGE` should *not* be set to this?

Osmose:
> Anything I should be concerned about? Is there any particular environment in settings.py where STATICFILES_STORAGE should not be set to this?

Nah, I'm pretty sure this is how we do it in other projects. [Pontoon, for example.](https://github.com/mozilla/pontoon/blob/master/pontoon/base/storage.py)

mythmon:
I ran this locally. When I loaded the page in my browser, it did a bunch of npm installs. That's probably not good, and we should at least put those things in `package.json`.

mythmon:
Nevermind, I just didn't `npm install` before running things. I see now that they are in `package.json`.

normandy/control/urls.py
```diff
@@ -0,0 +1,13 @@
+from django.conf.urls import url, include
+
+from normandy.control import views as control_views
+from django.contrib.auth.views import login, logout_then_login
+
+urlpatterns = [
+    url(r'^control/', include([
```
Osmose: We want the control interface to only be available on the admin node behind the VPN; we have a setting `ADMIN_ENABLED` to control this: https://github.com/mozilla/normandy/blob/master/normandy/urls.py#L15

This URLConf should only include URLs if the setting is true to avoid having it available outside the VPN.

normandy/control/static/control/admin/sass/control.scss
```diff
@@ -0,0 +1,100 @@
+@import 'partials/layout';
+
+
```
rehandalal: nit: seems like an extra line

rehandalal:
pipeline stuff looks fine to me. i'll play with it locally and confirm it works as expected.

brittanystoroz:
@Osmose + @rdalal updated

normandy/control/urls.py
```diff
@@ -0,0 +1,17 @@
...
+
+from normandy.control import views as control_views
+from django.contrib.auth.views import login, logout_then_login
+
+urlpatterns = []
+
+if settings.ADMIN_ENABLED:
+  urlpatterns += [
+      url(r'^control/', include([
+          url(r'^$', control_views.IndexView.as_view(), name='index'),
```
Osmose: URL names are global across the whole site. Typically we prefix them with something like `normandy.control.index`. Looking at what we've got thusfar in the project, we haven't really been consistent.

@mythmon Which do you like better as URL names: `normandy.control.index` or just `control.index`? Should we also use [URL namespaces](https://docs.djangoproject.com/es/1.9/topics/http/urls/#url-namespaces-and-included-urlconfs)?

normandy/control/templates/control/edit_recipe.html
```diff
@@ -0,0 +1,14 @@
+{% extends "control/admin/base.html" %}
+
+{% block breadcrumbs %}
+  <div class="breadcrumbs">
+    <a href="/control">Recipes</a> <span class="separator">//</span> {{ recipe.name }}
```
Osmose: Not a big deal since this will change anyway, but you'll want to use the `url` template tag to generate this URL.

normandy/control/views.py
```diff
@@ -0,0 +1,19 @@
...
+  template_name = 'control/index.html'
+  context_object_name = 'all_recipes_list'
+  login_url = '/control/login/'
```
Osmose: You could subclass `LoginRequiredMixin` and specify this attribute to make a `ControlMixin` class for all control views that redirects to the right URL.

normandy/control/views.py
```diff
@@ -0,0 +1,19 @@
...
+
+  def get_queryset(self):
+    return Recipe.objects.order_by('-start_time')[:5]
```
Osmose: Why the `[:5]`?

normandy/control/urls.py
rehandalal: i prefer `control.index`.

normandy/control/views.py
brittanystoroz: Whoops, was just testing/playing around. I will remove that.

normandy/control/urls.py
mythmon: I'd imagine something like a namespace of `normandy` and a route name of `control.index`. I haven't looked a lot into namespaces though.

brittanystoroz: I putted around with namespaces when I was overthinking all the registration/auth stuff, and they seemed really useful. Except my impression was that you would use `control` as the namespace, and the url name would still just be `index`. Then the urls in your template would look like `{% url 'control:index' %}`.  I think the answer given here was pretty useful http://stackoverflow.com/questions/19171570/a-real-example-of-url-namespace  ?

rdalal:
You guys seem to have the wrong rdalal again.

mythmon:
Sorry. @rehandalal, we meant you.

normandy/control/urls.py
brittanystoroz: @Osmose @mythmon @rehandalal yay or nay on URL namespacing? I can revert if we don't like. https://github.com/mozilla/normandy/pull/95/commits/8b23cb17ecc53ca9ec830c864c11ad5950d0333b

brittanystoroz:
I think this should be ready to go. All updates have been made based on feedback

rehandalal:
r+

PR #96, opened 2016-04-04
[bug 1248327] Update Normandy driver for new functionality.

Osmose:
Implements the client, locale, location, and saveHeartbeatFlow functions for the Normandy driver. These APIs are documented and used in [a PR on normandy-actions](https://github.com/mozilla/normandy-actions/pull/6).

@mythmon r?

### DIFF
--- a/normandy/selfrepair/static/js/normandy_driver.js
+++ b/normandy/selfrepair/static/js/normandy_driver.js
@@ -37,12 +37,23 @@ Object.assign(LocalStorage.prototype, {
  * Implementation of the Normandy driver.
  */
 let Normandy = {
+    locale: document.documentElement.dataset.locale || navigator.language,
+
     get testing() {
         return new URL(window.location.href).searchParams.has('testing');
     },
 
-    log(message, testOnly=true) {
-        if (!(testOnly && !this.testing)) {
+    _location: {countryCode: null},
+    location() {
+        return Promise.resolve(this._location);
+    },
+
+    log(message, level='debug') {
+        if (level === 'debug' && !this.testing) {
+            return;
+        } else if (level === 'error') {
+            console.error(message);
+        } else {
             console.log(message);
         }
     },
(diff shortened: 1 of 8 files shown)


Osmose:
Whoops, gotta fix the tests.

@mythmon I rebased on master and tried to install the new bcrypt requirements and got:

```
Collecting cffi==1.5.2 (from -r requirements.txt (line 142))
  Downloading cffi-1.5.2-cp35-cp35m-macosx_10_10_x86_64.whl (151kB)
    100% |████████████████████████████████| 153kB 3.9MB/s
THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
    cffi==1.5.2 from https://pypi.python.org/packages/cp35/c/cffi/cffi-1.5.2-cp35-cp35m-macosx_10_10_x86_64.whl#md5=4e247f933774b307d6ce257fbde6ce87 (from -r requirements.txt (line 142)):
        Expected sha256 9aee43865a42243555394d8090eaf97a0c9e0fc5ddc3344fce5a34b17f4b8c94
        Expected     or 96ca8beab869c2a0164b3ec9716dc85713a8e8b480e8d303520e6869dec27216
        Expected     or da9bde99872e46f7bb5cff40a9b1cc08406765efafb583c704de108b6cb821dd
             Got        810c52902a8a8ebdf37c3097930de1e51f0a70167444faa91f426228bc3f5277
```

Presumably this has something to do with your recent hashin patch. We probably don't have the hash for the OSX 10.10 wheel for cffi. Thoughts? Is this something we should just manually fix now or did we run hashin wrongly?

mythmon:
The hashes for bcrypt and cffi were done by hand, since they were done before the hashin patch was complete. I didn't go back and do the automated version aftewards, sorry.

I just ran `hashin -p 3.5 cffi`, and it added two more hashes, including the one you just failed on. I'll make a PR for that.

normandy/selfrepair/static/js/normandy_driver.js
```diff
@@ -79,7 +129,22 @@ let Normandy = {
                 options.learnMoreMessage,
                 options.learnMoreUrl
             );
+
+            resolve(emitter);
```
mythmon: I'm concerned about the order of execution here. What happens when an even gets emitted before the `.then()` that listens to this receives this emitter? I don't think `EventEmitter`s usually queue messages until a listener is around. I'm not sure how to fix this though, except to not use the Promise, and just return an event emitter directly.

mythmon: I'm imagining something like this:

1. `showHeartbeat` immediately fires an event (that is asynchronously dispatched)
2. `.then()` handler that would recieve the event emitter is registered (but not executed)

    > *Queue is now 1. handle event, 2. execute `.then()`*
4. Event is handled, and event emitter does nothing because there are no listeners

So this is really only an issue if `showHeartbeat` is going to immediately, synchronously, fire an event (which is reasonable). I don't know how it is implemented.
5. `.then()` handler runs, binding listener to the event emitter

Osmose: Returning an event emitter doesn't really help; we still have to call `showHeartbeat` before returning, so it might emit an event early before we return.

If this issue occurs, all that happens is that the timestamp for when the dialog was shown to the user is incorrect. Given that this is unlikely (In all my testing I've never once run into the issue of showHeartbeat emitting an event before the emitter listeners get registered), I think it's acceptable to just not handle this case. It'll be pretty clear from our data whenever this actually becomes an issue.

Osmose:
Added a test and fixed another one. They were all in a new file so I didn't add a second commit, just amended the existing one.

This is otherwise ready, I think.

mythmon:
LGTM, but GitHub says you have a merge conflict you need to fix

PR #97, opened 2016-04-04
Add OSX hashes for cffi

mythmon:
This should fix [the problems in #96](https://github.com/mozilla/normandy/pull/96#issuecomment-205149995). These hashes were added automatically by `hashin==0.4`. The previous hashes were added by my dumb fingers looking at the list of hashes on PyPI.

@Osmose r?


Osmose:
r+, install works for me now

PR #101, opened 2016-04-06
Avoid data loss by always populating the arguments editor.

Osmose:
If the argument schema changes for an existing recipe, validation
will fail and the argument editor will be blank. Instead, we want
to preserve the existing data and always populate the editor.

@mythmon r?

### DIFF
--- a/normandy/recipes/static/css/arguments_editor.css
+++ b/normandy/recipes/static/css/arguments_editor.css
@@ -39,3 +39,14 @@
 .arguments-editor .tabs span {
     margin-left: 5px;
 }
+
+/* The JSON editor modal hangs off the button, which is floated right.
+   It's absolutely positioned, so we set it's right edge to the button's
+   right edge to be able to see it.
+
+   We use !important because these are set as style attributes on the DOM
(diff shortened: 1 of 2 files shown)


normandy/recipes/static/css/arguments_editor.css
```diff
@@ -39,3 +39,11 @@
...
     margin-left: 5px;
 }
+
+/* The JSON editor modal hangs off the button, which is floated right.
+   It's absolutely position, so we set it's right edge to the button's
+   right edge to be able to see it. */
+.arguments-editor .modal {
+    left: unset !important;
+    right: 0 !important;
+}
```
Osmose: These are set as style attributes directly on the DOM element by JSONEditor, and they're set outside of the theme system where we can override it. That's why I'm using `!important` to override them. I'm not very proud of this.

mythmon: This should be a comment next to those `!important`s.

Osmose:
Updated

mythmon:
r+

PR #102, opened 2016-04-07
Add test coverage for geolocation, and fix bug in the same

mythmon:
Bug 1262620

@Osmose r?

### DIFF
--- a/normandy/classifier/geolocation.py
+++ b/normandy/classifier/geolocation.py
@@ -18,7 +18,7 @@ def load_geoip_database():
 
     try:
         geoip_reader = Reader(settings.GEOIP2_DATABASE)
-    except IOError as e:
+    except IOError:
         log.warning('Geolocation is disabled: Cannot load database.')
 
 
@@ -28,8 +28,8 @@ def get_country_code(ip_address):
             return geoip_reader.country(ip_address).country.iso_code
         except AddressNotFoundError:
             pass
-        except GeoIP2Error as e:
-            log.warning(e)
+        except GeoIP2Error as exc:
+            log.warning(exc)
             pass
 
     return None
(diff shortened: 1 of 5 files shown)


normandy/conftest.py
```diff
@@ -22,3 +23,13 @@ def django_cache(request):
...
     for cache_name in settings.CACHES.keys():
         caches[cache_name].clear()
+
+
+@pytest.fixture
+def geolocation():
+    """Fixture to load geolocation data."""
+    geolocation_module.load_geoip_database()
+    if not geolocation_module.has_geoip:
+        pytest.skip()
```
mythmon: I want a way to make skips turn into failures in CI. Sumo has this with nose, and it helps us be sure about our CI. I can't find a thing in pytest to do this.

normandy/conftest.py
```diff
@@ -22,3 +23,13 @@ def django_cache(request):
...
     for cache_name in settings.CACHES.keys():
         caches[cache_name].clear()
+
+
+@pytest.fixture
+def geolocation():
+    """Fixture to load geolocation data."""
+    geolocation_module.load_geoip_database()
+    if geolocation_module.geoip_reader is None:
+        pytest.skip()
```
mythmon: I want a way to make skips turn into failures in CI. Sumo has this with nose, and it helps us be sure about our CI. I can't find a thing in pytest to do this.

normandy/classifier/geolocation.py
```diff
@@ -18,15 +18,18 @@ def load_geoip_database():
 
     try:
         geoip_reader = Reader(settings.GEOIP2_DATABASE)
-    except IOError:
+    except IOError as e:
```
willkg: Nit: If you're not using the exception, you don't need the `as e`. Also, I'd suggest calling it `exc` or something that's not a single character.

mythmon: Oops, this was a part of my debugging (I logged the error). I'll fix that.

mythmon:
Ok. This is ready again.

Osmose:
r+, nice work!

PR #106, opened 2016-04-13
[bug 1262620] Fetch IP address from client correctly.

Osmose:
@rehandalal r?

### DIFF
--- a/normandy/classifier/models.py
+++ b/normandy/classifier/models.py
@@ -42,7 +42,11 @@ class Client(object):
 
     @cached_property
     def country(self):
-        ip_address = self.request.META.get('REMOTE_ADDR')
+        try:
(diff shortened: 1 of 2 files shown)


Osmose:
For context, [Amazon ELBs set `X-Forwarded-For` correctly](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/x-forwarded-headers.html), and [our nginx config passes them through to the app](https://github.com/mozilla-services/puppet-config/blob/master/shield/modules/normandy/templates/http_normandy.conf.erb#L46), so the Django app should have access to the header.

rehandalal:
I don't have access to the puppet-config repo so I couldn't check that link out.

But the code looks good to me. r+

PR #109, opened 2016-04-14
[fix bug 1262617] Replace RevisionMiddleware with decorators.

Osmose:
RevisionMiddleware automatically creates revisions when things
change. However, viewing the revert page in the admin interface
triggers a revert (in order to generate the preview data before
actually reverting) within a transaction, and then rolls back the
transaction. But, as far as I can tell, django-reversion doesn't
handle transaction rollback and saves the revision from the revert
anyway.

Removing the middleware fixes this, but we still have to manually
trigger revision creation outside the admin. This commit adds a
ViewSet mixin that triggers revision creation for DRF views.

@mythmon r?

### DIFF
--- a/normandy/recipes/api/views.py
+++ b/normandy/recipes/api/views.py
@@ -1,10 +1,12 @@
 from django.conf import settings
+from django.db import transaction
 from django.http import Http404
 from django.views.decorators.cache import cache_control
 
 from rest_framework import generics, permissions, viewsets
 from rest_framework.exceptions import NotFound
 from rest_framework.response import Response
+from reversion import revisions as reversion
 
 from normandy.base.api.permissions import AdminEnabled
 from normandy.base.api.renderers import JavaScriptRenderer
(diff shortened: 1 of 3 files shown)


normandy/recipes/api/views.py
```diff
@@ -6,14 +6,15 @@
...
+from normandy.base.api.mixins import RevisionMixin
 from normandy.base.api.permissions import AdminEnabled
 from normandy.base.api.renderers import JavaScriptRenderer
 from normandy.recipes.models import Action
 from normandy.recipes.api.permissions import NotInUse
 from normandy.recipes.api.serializers import ActionSerializer
 
 
-class ActionViewSet(viewsets.ModelViewSet):
+class ActionViewSet(RevisionMixin, viewsets.ModelViewSet):
```
mythmon: This mixin shouldn't be able to affect the update method of this class, since that class never calls `super()` (praise be to Python 3). That being said, I don't know how the tests are passing.

Osmose: It totally does call super.

mythmon: Doh. I misread that. I'm still concerned about the order of things. The update method here doesn't create a revision or become atomic until it calls super(), which in this case is fine. The mixin isn't as fool proof as I'd like, but that's probably fine.

mythmon:
r+

PR #110, opened 2016-04-14
[bug 1262889] Pass the request to the Client in the preview view.

Osmose:
@rehandalal r?

The bug here was that `Bundle.for_client` was attempting to match a recipe with a start and end time to a client object that didn't have a `request`. Thus, when it tried to access `request.received_at` to figure out when the request was received, it found a `None` value. The fix passes the request from the view into the form that builds the client.


rehandalal:
nothing stands out to me, and the tests are happy, so r+

PR #111, opened 2016-04-14
API endpoints for recipes

rehandalal:
r?

### DIFF
--- a/normandy/recipes/api/views.py
+++ b/normandy/recipes/api/views.py
@@ -8,9 +8,9 @@ from rest_framework.response import Response
 
 from normandy.base.api.permissions import AdminEnabled
 from normandy.base.api.renderers import JavaScriptRenderer
-from normandy.recipes.models import Action
+from normandy.recipes.models import Action, Recipe
 from normandy.recipes.api.permissions import NotInUse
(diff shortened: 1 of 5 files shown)


normandy/recipes/api/views.py
```diff
@@ -58,3 +58,26 @@ def retrieve(self, request, name, impl_hash):
...
+    def update(self, request, *args, **kwargs):
+        """
+        Intercept PUT requests and have them create instead of update
+        if the object does not exist.
+        """
+        if request.method == 'PUT':
+            try:
+                self.get_object()
+            except Http404:
+                return self.create(request, *args, **kwargs)
```
mythmon: @Osmose I wonder if we should have a `normandy.base.api.ModelViewSet` that we can use as the base for our APIs that have this version of `update`, since I think we are going to want to do this on everything. (It also helps with the concerns I have on #109).

@rehandalal I don't think we need to make any changes to this PR about this.

mythmon:
The viewset and the tests look good. I think we should also have a test for using the API to delete recipes, since the generic viewset includes code for that.

mythmon:
r+

PR #112, opened 2016-04-18
Merge the classifier app with the recipes app.

Osmose:
There's not much point splitting the two up, so we might as well
admin that they're really one big complicated app rather than two
fairly complicated apps.

@mythmon r?


normandy/recipes/tests/test_api.py
```diff
@@ -249,3 +252,25 @@ def test_unavailable_if_admin_disabled(self, api_client, settings):
         res = api_client.get('/api/v1/recipe/')
         assert res.status_code == 403
         assert res.data['detail'] == AdminEnabled.message
+
+
+def test_version(client, mocker):
```
mythmon: This test (and `test_lbheartbeat_makes_no_db_queries`) were in `normandy/health`, and probably got duplicated.

mythmon:
r+

PR #114, opened 2016-04-22
[bug 1266885] Replace recipe filters with filter expressions.

Osmose:
@mythmon r?


normandy/recipes/migrations/0024_auto_20160422_0541.py
```diff
@@ -0,0 +1,44 @@
...
+            name='sample_rate',
+        ),
+        migrations.RemoveField(
+            model_name='recipe',
+            name='start_time',
+        ),
+        migrations.AddField(
+            model_name='recipe',
+            name='filter_expression',
+            field=models.TextField(default='(1 + 1 == 2)'),
```
mythmon: `'(1 + 1 == 2)'` is a weird default. I think `''` would be better. The addon, at least, interprets an empty `filter_expression` as matching anything (just like an empty matcher did before).

mythmon: If we don't like that "empty is match all" behavior, `true` or `false` would be a better default.

Osmose:
Updated!

normandy/selfrepair/static/js/self_repair_runner.js
```diff
@@ -103,14 +100,46 @@ function runRecipe(recipe) {
...
+/**
+ * Match a recipe against a context using its filter expression.
+ * @param  {Recipe} recipe  Recipe fetched from the server.
+ * @param  {object} context Context returned by filterContext.
+ * @promise Resolves with a list containing the recipe and a boolean
+ *     signifying if the filter passed or failed.
+ */
+function matches(recipe, context) {
+    return jexl.eval(recipe.filter_expression, context)
+    .then(value => [recipe, !!value]);
```
mythmon: If we're going with "empty matches everything", you'll need some handling of that here.

Also, Jexl doesn't do multiline expression, and we have the habit of pressing Enter while writing filters. This should join the `filter_expression` into one line.

mythmon:
Besides the tweak to the jexl handling, this looks good.

Though, we need to refiddle the commits and migrations so that we can deploy this cleanly. There needs to be two commits that can be deployed separately. I think the easiest way to do this is to have one commit that simply adds the new field, but doesn't do anything with it, and then another commit that includes all the code changes and removes the old fields. You could also add the filters apis to that first commit, if you want, as long as it works with both the old and new DB.

normandy/selfrepair/static/js/self_repair_runner.js
Osmose: Ah, I misunderstood what you said as meaning that JEXL considers a blank expression to be true. Whoops!

Osmose:
Updated. I've split the change into two commits as suggested; one adds the `filter_expression` field, while the other one adds the rest of the functionality and removes the older filter fields.

After thinking about it, I decided that I'm not a fan of treating empty as true on the server side, as it makes it difficult to detect when a user has forgotten to add a filter expression. So instead, I've made the migration default all existing recipes to have `true` as their filter expression, and set `blank=False` on the filter expression field. This way, existing recipes get a default value that isn't invalid, but new recipes will throw a validation error if the user forgets to add a filter expression.

I also tweaked some things in the self-repair page:
- The JEXL context had no way to access the user's locale. I included it in `normandy.client`.
- As suggested, it now removes all newlines in a filter expression.

Osmose:
Oh yeah, also rebased this on master.

normandy/selfrepair/static/js/self_repair_runner.js
```diff
@@ -3,6 +3,11 @@ import xhr from './xhr.js'
 import Mozilla from './uitour.js'
 import Normandy from './normandy_driver.js'
 import uuid from 'node-uuid'
+import jexl from 'Jexl'
```
mythmon: The package name here should be lowercase.

mythmon:
Thanks!

PR #116, opened 2016-04-23
[bug 1264386] Add plugin data to the client object.

Osmose:
Adds the `plugin` attribute as described in https://github.com/mozilla/normandy-actions/pull/10.

@mythmon r?

### DIFF
--- a/normandy/selfrepair/static/js/normandy_driver.js
+++ b/normandy/selfrepair/static/js/normandy_driver.js
@@ -68,7 +68,19 @@ let Normandy = {
 
     client() {
         return new Promise(resolve => {
-            let client = {};
+            let client = {
+                plugins: {}
+            };
+
+            // Populate plugin info.
+            for (let plugin of navigator.plugins) {
+                client.plugins[plugin.name] = {
+                    name: plugin.name,
+                    filename: plugin.filename,
+                    description: plugin.description,
+                    version: plugin.version,
+                }
+            }
 
             // Keys are UITour configs, functions are given the data
             // returned by UITour for that config.


normandy/selfrepair/static/js/normandy_driver.js
```diff
@@ -68,7 +68,20 @@ let Normandy = {
...
     client() {
         return new Promise(resolve => {
-            let client = {};
+            let client = {
+                plugins: {}
+            };
+
+            // Populate plugin info.
+            for (let k = 0; k < navigator.plugins.length; k++) {
+                let plugin = navigator.plugins[k];
```
mythmon: Why not `for (let plugin of navigator.plugins) {`?

Osmose: Because navigator.plugins isn't actually an Array, it's a PluginArray. But I didn't actually test if that still works.

I could've used `Array.from`, but I somehow thought that `Array.from` wasn't available in the Firefox versions we care about, but looking at the docs, I was wrong. I'll see about fixing this up.

mythmon: `PluginArray` seems to implement the iterator protocol, so it works fine:

```
>> for (let plugin of navigator.plugins) { console.log(plugin); }
Plugin { application/googletalk: MimeType, description: "Version: 5.41.0.0", filename: "libnpgoogletalk.so", version: "", name: "Google Talk Plugin", length: 1, 1 more… }
Plugin { application/o1d: MimeType, description: "Version: 5.41.0.0", filename: "libnpo1d.so", version: "", name: "Google Talk Plugin Video Renderer", length: 1, 1 more… }
```

I don't know about `Array.from` in Firefox though.

Osmose:
Updated.

mythmon:
This makes me want lodash. `_.pluck` would have worked nicely, I think. But it doesn't make me want lodash enough to actually use lodash.

Anyways, r+.

PR #118, opened 2016-04-25
Use git to get current commit, instead of being clever

mythmon:
It seems something changed about how Circle gets the code from GitHub for the master builds. Instead of trying to increase the cleverness I was using to get the current git commit without using git, this just installs git and uses it. This make a bigger image, which slows down CI and deploys, but it should be more reliable.

@Osmose r?


Osmose:
r+

PR #119, opened 2016-04-27
Make DB health checks more helpful

mythmon:
I added these while working with Docker Compose. They were helpful in debugging.

@rehandalal r?


rehandalal:
lgtm. r+

PR #123, opened 2016-04-28
Remove unused Heroku and Travis bits

mythmon:
These bits are for Heroku and Travis. We are using neither of them anymore.

r?


Osmose:
r+

PR #125, opened 2016-04-28
Add webpack to bundle selfrepair assets

brittanystoroz:
@mythmon r? 

You can run `./node_modules/.bin/webpack --config ./normandy/webpack.config.js` to create the bundle manually. `--watch`. I'm not 100% sure if there's anything particular we have to do to set this up for the production environment? We can a prod-specific webpack config if there are significant changes we need to make.

### DIFF
--- a/normandy/selfrepair/static/js/index.js
+++ b/normandy/selfrepair/static/js/index.js
@@ -1,0 +1,3 @@
+import 'babel-polyfill'
+import './normandy_driver.js'
+import './self_repair_runner.js'
--- a/normandy/selfrepair/static/js/index.js
+++ b/normandy/selfrepair/static/js/index.js
@@ -1,3 +1,0 @@
-import 'babel-polyfill'
-import './normandy_driver.js'
-import './self_repair_runner.js'
--- a/normandy/selfrepair/static/js/self_repair_runner.js
+++ b/normandy/selfrepair/static/js/self_repair_runner.js
@@ -2,6 +2,7 @@ import 'babel-polyfill'
 import xhr from './xhr.js'
 import Mozilla from './uitour.js'
 import Normandy from './normandy_driver.js'
+import uuid from 'node-uuid'
 
 let registeredActions = {};
 window.registerAction = function(name, ActionClass) {
--- a/normandy/selfrepair/static/js/self_repair_runner.js
+++ b/normandy/selfrepair/static/js/self_repair_runner.js
@@ -1,3 +1,4 @@
+import 'babel-polyfill'
 import xhr from './xhr.js'
 import Mozilla from './uitour.js'
 import Normandy from './normandy_driver.js'
--- a/normandy/selfrepair/static/js/self_repair_runner.js
+++ b/normandy/selfrepair/static/js/self_repair_runner.js
@@ -1,3 +1,7 @@
+import xhr from './xhr.js'
+import Mozilla from './uitour.js'
+import Normandy from './normandy_driver.js'
+
 let registeredActions = {};
 window.registerAction = function(name, ActionClass) {
     registeredActions[name] = ActionClass;
--- a/normandy/selfrepair/templates/selfrepair/repair.html
+++ b/normandy/selfrepair/templates/selfrepair/repair.html
@@ -8,6 +8,6 @@
     <link rel="icon" href="data:;base64,=">
   </head>
   <body>
-    {% render_bundle 'main' %}
+    {% render_bundle 'selfrepair' %}
   </body>
 </html>
--- a/normandy/selfrepair/templates/selfrepair/repair.html
+++ b/normandy/selfrepair/templates/selfrepair/repair.html
@@ -1,4 +1,5 @@
 {% load staticfiles %}
+{% load render_bundle from webpack_loader %}
 
 <!DOCTYPE html>
 <html lang="en" data-locale="{{ locale }}" data-bundle-url="{% url 'recipes:api.v1.fetch_bundle' %}">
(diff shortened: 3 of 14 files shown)


mythmon:
This error is popping up during tests: "Error reading webpack-stats.json. Are you sure webpack has generated the file and the path is correct?". I wonder if we have to generate a webpack bundle before running the tests?

brittanystoroz:
Test failures are related to not finding the webpack-stats.json file...which I've .gitignored locally. Is there a place for me to add the webpack command so it'll run as part of the build step...?

brittanystoroz:
@mythmon yea it seems like that is the case, I was reading https://github.com/owais/django-webpack-loader#how-to-use-in-production Just not sure where in our project we add something like that

mythmon:
Will we need `webpack-stat.json` in production? If so, we could add a webpack line to the docker file. I think right before we run `collectstatic`. If we don't need it in production, we can add it as a dependencies command in `circle.yml`.

normandy/webpack.config.js
```diff
@@ -0,0 +1,24 @@
+var path = require("path");
```
mythmon: Nit: We should be consistent about quotes.

normandy/selfrepair/static/js/index.js
```diff
@@ -0,0 +1,3 @@
+import 'babel-polyfill'
+import './normandy_driver.js'
+import './self_repair_runner.js'
```
mythmon: Why import `./normandy_driver.js` here? Wouldn't it get included through `self_repair_runner.js`?

normandy/selfrepair/templates/selfrepair/repair.html
```diff
@@ -7,13 +8,6 @@
...
   <body>
-    <script src="{% static 'js/uitour.js' %}"></script>
-    <script src="{% static 'npm/node-uuid/uuid.js' %}"></script>
-    <script src="{% static 'npm/wolfy87-eventemitter/EventEmitter.js' %}"></script>
-    <script src="{% static 'npm/babel-polyfill/dist/polyfill.js' %}"></script>
-    <!-- Cheat so that we get let and other version-limited JavaScript features. -->
-    <script type="application/javascript;version=1.8" src="{% static 'js/xhr.js' %}"></script>
-    <script type="application/javascript;version=1.8" src="{% static 'js/normandy_driver.js' %}"></script>
-    <script type="application/javascript;version=1.8" src="{% static 'js/self_repair_runner.js' %}"></script>
+    {% render_bundle 'main' %}
```
mythmon: Where does the name `main` come from?

mythmon:
This looks nice and simple. Way better than the devserver idea we talked about way back when. I'm going to try and run it locally now.

mythmon:
When running `npm install`, I get

```
npm WARN babel-loader@6.2.4 requires a peer of babel-core@^6.0.0 but none was installed.
npm WARN babel-loader@6.2.4 requires a peer of webpack@1 || ^2.1.0-beta but none was installed.
```

brittanystoroz:
@mythmon  ahh what version of npm are you using? I might be on an old version, npm3 no longer installs peers for you. Will fix.

AFAIK you shouldn't actually need webpack-stats.json in prod.

normandy/selfrepair/templates/selfrepair/repair.html
brittanystoroz: This is just the default bundle reference name, I believe I can override it in the webpack config with something more specific.

mythmon:
```shell
$ npm --version
3.8.7
```

Also, I think `babel-preset-es2015` is missing.

mythmon:
Good news, after installing the missing bits abpve, this worked great. When I visited https://localhost:8000/en-US/repair/, I got exactly one request for the bundle. :fireworks: :fireworks:

normandy/webpack.config.js
```diff
@@ -0,0 +1,26 @@
+var path = require('path');
```
mythmon: Most projects I've seen put this at the root of the git repo. Is there a reason to put it in the `normandy` subdirectory here?

mythmon:
After reading the django-webpack-loader "How to use in production" section, I think adding a call to webpack to `Dockerfile` is the right way to go, so we get a `webpack-stats.json` in the image.

I made this change to `Dockerfile`, and now running the server in Docker just works (I used Docker Compose and [normandy-compose](https://github.com/mythmon/normandy-compose) to do that). I suspect this will make the tests pass too.

```diff
diff --git a/Dockerfile b/Dockerfile
index 7f6a9da..0cef5c2 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,10 +11,14 @@ RUN curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add -
     apt-get update && apt-get install -y nodejs
 
 COPY ./requirements.txt /app/requirements.txt
-RUN pip install -U 'pip>=8' && pip install --upgrade --no-cache-dir -r requirements.txt
+COPY ./package.json /app/package.json
+RUN pip install -U 'pip>=8' && \
+    pip install --upgrade --no-cache-dir -r requirements.txt && \
+    npm install
 
 COPY . /app
-RUN DJANGO_CONFIGURATION=Build ./manage.py collectstatic --no-input && \
+RUN ./node_modules/.bin/webpack && \
+    DJANGO_CONFIGURATION=Build ./manage.py collectstatic --no-input && \
     mkdir -p media && chown app:app media && \
     mkdir -p __version__ && \
     git rev-parse HEAD > __version__/commit && \
```

brittanystoroz:
🎉

normandy/selfrepair/static/js/self_repair_runner.js
```diff
@@ -1,3 +1,8 @@
+import 'babel-polyfill'
+import xhr from './xhr.js'
+import Mozilla from './uitour.js'
+import Normandy from './normandy_driver.js'
```
mythmon: This file needs uuid imported too. I only get an error for this when I run this in docker-compose, not when I run it normally, which is weird.

mythmon: Oh! When I ran it normally, I used `https://normandy/` (my normal test domain). When I tested it with docker-compose, I used `https://localhost:8000`, which doesn't share local storage. This module only needs `uuid` once, and then it caches that value in localstorage, which is per domain. That explains it.

mythmon:
Woot, nice job! One more tiny issue with this (missing an import), and I think this is good to go.

Well, maybe we want to add that bit about `webpack` and/or `webpack --watch` somewhere? I don't know about that.

I've got to step out for a while, but I'm r+ after the import statement is fixed.

mythmon:
Nice. This all works for me now.

PR #127, opened 2016-04-29
Bring in webpack bundling and make it work.

mythmon:
Note: This is a PR against the backports branch, not master. The backports branch is not going to be merged to master.

This PR does three things in two commits

* Backports @brittanystoroz's PR #125 so we can get the sweet sweet goodness of only making one CDN request on the selfrepair page. (commit 1)
* Removes all of Pipeline, since we aren't going to use it long term, and it gets in the way of the webpack work. (commit 2, part 1)
* Removes the vestigial Control modules, since it isn't in use in the branch, and it was using Pipeline. We'll have a new, webpacked version in the future, courtesty of @brittanystoroz. (commit 2, part 2)

I think this is the simplest way to get this to a point we can deploy it without breaking everything.

@brittanystoroz r?


brittanystoroz:
r+

PR #129, opened 2016-04-29
Fix bug 1268743: Improve documentation for self-repair.

Osmose:
@mythmon r?

The ODG file included is the LibreOffice Draw file used to generate the sequence diagram in the PNG. I couldn't find a text-based format I liked for sequence diagrams so I just drew it manually. It's kinda simple but I think a visual description is useful. We can update or remove it later depending on how useful it ends up being.


mythmon:
This seems fine, though you have some trouble with your requirements files in CI. I'm not convinced that the sequence diagram adds much, since it is a trivial sequence. I find more value in those when it is not strictly back and forth. (If we do keep it, I'm fine with how it is included here. I also don't know a good text based format for this, unless we do ascii art).

Also, we should build the docs in CI. We could even have CircleCI save them as artifacts so we can look at them, if we wanted.

PR #130, opened 2016-05-01
Reactify and webpackify control admin

brittanystoroz:
Reactify basic CRUD for control admin.

Cleaned up and squashed some commits, so hopefully this is a little easier to stomach.  Cut out some extraneous commits and filed bugs for remaining work:

[1271272](https://bugzilla.mozilla.org/show_bug.cgi?id=1271272) - Build route & actions for handling recipe history
[1271274](https://bugzilla.mozilla.org/show_bug.cgi?id=1271274) - Validate recipe edit forms
[1271276](https://bugzilla.mozilla.org/show_bug.cgi?id=1271276) - Handle API request failures in UI
[1271278](https://bugzilla.mozilla.org/show_bug.cgi?id=1271278) - Add eslint to normandy for control admin
[1271279](https://bugzilla.mozilla.org/show_bug.cgi?id=1271279) - Test react components
[1266483](https://bugzilla.mozilla.org/show_bug.cgi?id=1266483) - Generate action configuration forms for arguments_json field

Notable brokenness is on adding a recipe, the action form isn't in there yet so our request is failing with missing fields. Adding the action forms requires a bit of refactoring and a decent amount of additional code, so I want to put those in a follow-up PR.


normandy/control/static/control/js/actions/ControlAppActions.js
```diff
@@ -0,0 +1,98 @@
...
+      });
+    }
+  } else {
+    return dispatch => {
+      dispatch(selectedRecipeReceived(null));
+    }
+  }
+}
+
+export function deleteRecipe(recipeId) {
```
brittanystoroz: Hitting this endpoint fails with `CSRF Failed: CSRF token missing or incorrect.` Not sure how to fix.

brittanystoroz: This delete request here fails with `CSRF Failed: CSRF token missing or incorrect.` Not sure how to fix.

normandy/control/static/control/js/actions/ControlAppActions.js
```diff
@@ -0,0 +1,98 @@
...
+  return dispatch => {
+    dispatch(removeRecipeFromCollection(recipeId));
+
+    return apiFetch(`/api/v1/recipe/${recipeId}/`, {
+        method: 'delete',
+        credentials: 'include',
+      }).then(recipe => {
+        dispatch(recipeDeleted(recipe));
+      });
```
brittanystoroz: This delete request here fails with `CSRF Failed: CSRF token missing or incorrect.` Not sure how to fix.

karma.conf.js
```diff
@@ -0,0 +1,79 @@
+// Karma configuration
+module.exports = function(config) {
+    config.set({
+        // base path that will be used to resolve all patterns (eg. files, exclude)
+        basePath: 'normandy/control/tests/',
+
+        // frameworks to use
+        // available frameworks: https://npmjs.org/browse/keyword/karma-adapter
+        frameworks: ['mocha'],
```
Osmose: I'm fine with mocha, but want to note that normandy-actions is using Jasmine currently (because I had issues with sinon and webpack). It'd be nice to be consistent, but I'm also fine with sticking with mocha if we like it better (and possibly switching normandy-actions back to mocha if we can figure out the issues with sinon).

rehandalal: +1 for consistency

rehandalal: i don't have a preference between mocha/jasmine.

normandy/control/static/control/js/actions/ControlAppActions.js
```diff
@@ -0,0 +1,177 @@
...
+    recipe: recipe
+  };
+}
+
+export function shouldFetchRecipes(state) {
+  if (state.recipeCollection.recipes.length === 0) {
+    return true;
+  } else if (state.recipeCollection.recipes.isFetching) {
+    return false;
+  }
```
mythmon: Shouldn't this always return either true or false instead of sometimes returning undefined? What if `recipes.length` is not 0, but it also isn't fetching?

mythmon:
> @mythmon can you r+ that selfrepair still works as expected with these changes?

This seems to be working fine with self-repair. `/en-US/repair` serves the bundled self-repair code as expected. This doesn't have the problem we had before.

normandy/control/static/control/js/actions/ControlAppActions.js
brittanystoroz: Yes. I think I'm gunna go with changing the default state of `recipes` to be null rather than an empty array, that way you can just do a check if the collection exists or not. We shouldn't have to refetch the entire collection from the server more than once, assuming the reducers continue to keep the state accurately up-to-date.

normandy/control/views.py
```diff
@@ -1,40 +1,7 @@
...
-  template_name = 'control/create_recipe.html'
-  success_url = reverse_lazy('control:index')
-
-
-class DeleteView(ControlMixin, generic.DeleteView):
-  model = Recipe
-  success_url = reverse_lazy('control:index')
-  template_name = 'control/delete_recipe.html'
+def IndexView(request):
+  if not request.user.is_authenticated():
```
Osmose: Python uses 4-space indents.

normandy/control/static/control/js/actions/ControlActions.js
```diff
@@ -0,0 +1,173 @@
...
+}
+
+export function makeApiRequest(requestType, settings) {
+  return (dispatch, getState) => {
+    let apiRequestConfig = apiRequestMap[requestType](settings, getState);
+    if (apiRequestConfig) {
+      dispatch(requestInProgress());
+      return apiFetch(apiRequestConfig.url, {
+        ...apiRequestConfig.settings,
+        ...API_REQUEST_SETTINGS
```
rehandalal: the order of these two should probably be flipped?
```es6
...API_REQUEST_SETTINGS,
...apiRequestConfig.settings
```

brittanystoroz:
@rehandalal updated

normandy/settings.py
```diff
@@ -80,8 +78,12 @@ class Core(Configuration):
         'django.contrib.staticfiles.finders.FileSystemFinder',
         'django.contrib.staticfiles.finders.AppDirectoriesFinder',
         'npm.finders.NpmFinder',
-        'pipeline.finders.PipelineFinder',
     ]
+
+    STATICFILES_DIRS = (
```
rehandalal: this should probably move into the `Base` class with `STATIC_ROOT` and `STATIC_URL`

rehandalal:
we should update the docs to include how to run karma

rehandalal:
did my best to make sure everything looked good. i think for the most part it did. since there is still more work to be done i assume we can fix anything i missed as we spot it. so r+

rehandalal:
ps: awesome job @brittanystoroz!

PR #131, opened 2016-05-02
Implement peer signing backend for recipes

rehandalal:
r?

@mythmon: could you point me in the right direction with fixing the failing test?


mythmon:
I bet you'll need to add `approver` to the select related caching code here: https://github.com/mozilla/normandy/blob/master/normandy/recipes/models.py#L323-L337.

mythmon:
Also, to ease deployment, we need to make migrations a two step process. We'll deploy the first, wait for it all to settle, and then deploy the second. This is similar to Sumo's two step ES process. The easiest way to do that is to have one commit that adds fields and migrations, but does not require them in the code, and a second commit that has everything else.

rehandalal:
Cool, I'll add the migration as a first commit.

normandy/recipes/api/views.py
```diff
@@ -88,14 +89,48 @@ def update(self, request, *args, **kwargs):
...
+
+        if recipe.is_approved:
+            return Response({'approver': 'This recipe is already approved.'},
+                            status=status.HTTP_400_BAD_REQUEST)
+        recipe.approver = request.user
+        recipe.save()
+
+        return Response(status=status.HTTP_204_NO_CONTENT)
+
+    @detail_route(methods=['POST',], permission_classes=[permissions.IsAuthenticated,])
```
mythmon: Trailing commas aren't needed for lists, only for tuples.

normandy/recipes/api/views.py
```diff
@@ -88,14 +89,48 @@ def update(self, request, *args, **kwargs):
...
             try:
-                self.get_object()
+                recipe = self.get_object()
             except Http404:
-                return self.create(request, *args, **kwargs)
+                if request.method == 'PUT':
+                    return self.create(request, *args, **kwargs)
+            else:
+                if recipe.is_approved:
+                    recipe.disable(ignore_revision_id=True)
```
mythmon: I think this should unapprove the revision, or possibly both unapprove it and disable it.

normandy/recipes/api/views.py
```diff
@@ -88,14 +89,48 @@ def update(self, request, *args, **kwargs):
...
 
+    @detail_route(methods=['POST',], permission_classes=[permissions.IsAuthenticated,])
+    def approve(self, request, pk=None):
+        recipe = self.get_object()
+
+        if recipe.is_approved:
+            return Response({'approver': 'This recipe is already approved.'},
+                            status=status.HTTP_400_BAD_REQUEST)
+        recipe.approver = request.user
+        recipe.save()
```
mythmon: We probably want to have a permission for users that are allowed to approve recipes, instead of allowing everyone to approve them.

mythmon:
How does this assure that two people are happy with the recipe? Could someone make a recipe, and then approve it themselves? Would we be able to detect that?

normandy/recipes/api/views.py
rehandalal: this does unapprove and disable it:
https://github.com/mozilla/normandy/pull/131/files#diff-c5447b2cc20ceb64379f180e6f3a5d00R130

normandy/recipes/api/views.py
rehandalal: Do we need a separate permission or should we just check something like `user.is_staff()`?

mythmon: I think we should have a separate permission.

normandy/recipes/api/views.py
mythmon: Ah, I missed that. It isn't totally clear from the method name. I was imagining an approved but disabled recipe (though since any change rightfully unapproves, that recipe wouldn't be very useful).

mythmon:
This still needs fixes for the tests. It has also developed a merge conflict.

mythmon:
I talked to @Osmose about this, and we decided it is important that the model changes are included in the commit that has the migration.

rehandalal:
cool, i can squash it

rehandalal:
rebased/squashed & feedbacked ^

mythmon:
I still think we should have a separate permission for approving beyond simply is authorized. We should at least require edit permission on the object, which you can get with the DjangoPermissions permission class.

rehandalal:
@mythmon doesn't this cover that: https://github.com/mozilla/normandy/pull/131/files#diff-9fead1640dab3242809051a2f95d73ffL86

mythmon:
Uh, yes. I missed that, sorry.

r+

rehandalal:
fixed the conflict

PR #133, opened 2016-05-02
Recipe history API endpoint

rehandalal:
r?

### DIFF
--- a/normandy/recipes/api/serializers.py
+++ b/normandy/recipes/api/serializers.py
@@ -38,14 +38,14 @@ class RecipeSerializer(serializers.ModelSerializer):
         ]
 
 
-class BundleSerializer(serializers.ModelSerializer):
+class BundleSerializer(serializers.Serializer):
     recipes = RecipeSerializer(many=True)
     country = serializers.CharField()
 
 
 class RecipeVersionSerializer(serializers.ModelSerializer):
-    date_created = serializers.SerializerMethodField(read_only=True)
-    recipe = serializers.SerializerMethodField(read_only=True)
+    date_created = serializers.DateTimeField(source='revision.date_created', read_only=True)
+    recipe = RecipeSerializer(source='object_version.object', read_only=True)
 
     class Meta:
         model = Version
@@ -53,10 +53,3 @@ class RecipeVersionSerializer(serializers.ModelSerializer):
             'date_created',
             'recipe',
         ]
-
-    def get_date_created(self, obj):
-        return obj.revision.date_created
-
-    def get_recipe(self, obj):
-        return RecipeSerializer(obj.object_version.object,
-                                context={'request': self.context['request']}).data
--- a/normandy/recipes/api/serializers.py
+++ b/normandy/recipes/api/serializers.py
@@ -1,4 +1,5 @@
 from rest_framework import serializers
+from reversion.models import Version
 
 from normandy.recipes.api.fields import ActionImplementationHyperlinkField
 from normandy.recipes.models import Action, Recipe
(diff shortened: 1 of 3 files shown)


normandy/recipes/api/views.py
```diff
@@ -96,6 +99,21 @@ def update(self, request, *args, **kwargs):
...
+        recipe = self.get_object()
+        content_type = ContentType.objects.get_for_model(recipe)
+        versions = Version.objects.filter(content_type=content_type, object_id=recipe.pk)
+
+        data = []
+        for v in versions:
+            data.append({
+                'date': v.revision.date_created,
+                'recipe': RecipeSerializer(v.object_version.object,
+                                           context={'request': request}).data})
```
mythmon: Building `data` should probably be done with a Serializer.

rehandalal: yeah i guess i could make a VersionSerializer

rehandalal:
updated ^

normandy/recipes/api/views.py
```diff
@@ -96,6 +100,14 @@ def update(self, request, *args, **kwargs):
 
         return super().update(request, *args, **kwargs)
 
+    @detail_route(methods=['GET',])
+    def history(self, request, pk=None):
+        recipe = self.get_object()
+        content_type = ContentType.objects.get_for_model(recipe)
+        versions = Version.objects.filter(content_type=content_type, object_id=recipe.pk)
+        data = [RecipeVersionSerializer(v, context={'request': request}).data for v in versions]
```
mythmon: Better would be `data = RecipeVersionSerializer(versions, context={'request': request}, many=True)`. Generally if you have to access `.data` on more than one serializer in a request, something is a little fishy.

normandy/recipes/api/serializers.py
```diff
@@ -37,6 +38,25 @@ class Meta:
         ]
 
 
-class BundleSerializer(serializers.Serializer):
+class BundleSerializer(serializers.ModelSerializer):
```
mythmon: I don't think you want to do this, since `Bundle` isn't a Django model. Was something not working with this?

normandy/recipes/api/serializers.py
```diff
@@ -37,6 +38,25 @@ class Meta:
...
     recipes = RecipeSerializer(many=True)
     country = serializers.CharField()
+
+
+class RecipeVersionSerializer(serializers.ModelSerializer):
+    date_created = serializers.SerializerMethodField(read_only=True)
+    recipe = serializers.SerializerMethodField(read_only=True)
```
mythmon: I think both of these could be done as

```python
date_created = serializer.Serializer(source='revision.date_created')
recipe = RecipeSerializer(source='object_version.object', read_only=True)
```

mythmon: And then you can leave out the `get_date_created` and `get_recipe` methods below. Serializers are pretty smart about how to thread all the context and data objects around.

normandy/recipes/api/serializers.py
rehandalal: d'oh. nope. i originally made `RecipeVersionSerializer` a `Serializer` and then went back to change that and must have changed this instead by mistake.

normandy/recipes/api/serializers.py
rehandalal: ah! i was trying to figure out how to do that. i was trying `revision__date_created` etc. with no luck. thanks!

rehandalal:
updated ^

mythmon:
This looks great. One last thing, can you add a test for it? Something simple like `test_it_works` can be really useful, and makes it a lot easier to add tests for this feature in the future.

rehandalal:
test added ^

mythmon:
Thanks! r+

PR #135, opened 2016-05-04
Improve recipe API endpoint for create and update

rehandalal:
@mythmon r?

### DIFF
--- a/normandy/recipes/tests/test_api.py
+++ b/normandy/recipes/tests/test_api.py
@@ -38,6 +38,7 @@ class TestActionAPI(object):
         assert res.status_code == 200
         assert res.data == [
             {
+                'id': Whatever(),
                 'name': 'foo',
                 'implementation_url': Whatever.endswith(action_url),
                 'arguments_schema': {'type': 'object'}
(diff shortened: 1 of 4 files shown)


mythmon:
We'll also need to update [self_repair_runner.js](https://github.com/mozilla/normandy/blob/master/normandy/selfrepair/static/js/self_repair_runner.js). Do you want to do that here, or should that happen in another PR?

mythmon:
Trying to set `action_id` to the id of a non-existant action results in an ugly 500 with the error `normandy.recipes.models.DoesNotExist: Action matching query does not exist.`. It should return a 40x with a nice error message.

normandy/recipes/tests/test_api.py
```diff
@@ -228,6 +229,17 @@ def test_it_serves_recipes(self, api_client):
         assert res.status_code == 200
         assert res.data[0]['name'] == recipe.name
 
+    def test_it_can_create_recipes(self, api_client):
+        action = ActionFactory()
+
+        res = api_client.post('/api/v1/recipe/', {'name': 'Test Recipe',
+                                                  'action_id': action.id,
+                                                  'arguments': '{}'})
```
mythmon: I think you want this to be `'arguments': {}`. That makes the arguments empty object (which is valid), whereas `'{}'` makes the arguments a string containing two brackets, which isn't valid arguments for an action. We should probably actually be enforcing that somehow.

mythmon:
This looks good, besides a couple little issues. I think this is definitely a better way to build this API.

rehandalal:
updated ^

mythmon:
The `'{}'` vs `{}` thing is still there. Otherwise this looks good. r+wc

PR #136, opened 2016-05-05
Add .dockerignore file

mythmon:
This prevents these files from being copied into Docker images. This makes building them faster, since there is less to copy, and avoids problems with mismatched versions of Node, Python etc.

r?


Osmose:
r+

PR #141, opened 2016-05-06
Implement approval requests

rehandalal:
r?


rehandalal:
Have not implemented API endpoints yet.

normandy/recipes/api/serializers.py
```diff
@@ -21,10 +24,137 @@ class Meta:
...
+        fields = [
+            'id',
+            'created',
+            'creator',
+            'creator_id',
+        ]
+
+    def validate_creator_id(self, attr):
+        try:
+            User.objects.get(id=attr)
```
Osmose: An `exists` check is probably more appropriate here.

rehandalal: yup! silly me.

normandy/recipes/api/serializers.py
```diff
@@ -21,10 +24,137 @@ class Meta:
         ]
 
 
+class ApprovalSerializer(serializers.ModelSerializer):
+    creator = UserSerializer(read_only=True)
+    creator_id = serializers.IntegerField(source='creator.id', write_only=True)
```
Osmose: I think you can use `PrimaryKeyRelatedField` here, which may allow you to drop the three other methods on this class. I'm not 100% on this but the validation and extra logic for `create` and `update` seem like things DRF should handle for us already.

normandy/recipes/api/serializers.py
```diff
@@ -21,10 +24,137 @@ class Meta:
...
+            User.objects.get(id=attr)
+        except User.DoesNotExist:
+            raise serializers.ValidationError('User does not exist.')
+        return attr
+
+    def create(self, validated_data):
+        creator_id = validated_data.pop('creator')['id']
+        validated_data['creator'] = User.objects.get(id=creator_id)
+
+        if 'approval' in validated_data:
```
Osmose: Wouldn't an approval be created separately after an approval request? This might be unnecessary in that case.

normandy/recipes/models.py
```diff
@@ -85,6 +87,13 @@ def arguments(self):
     def arguments(self, value):
         self.arguments_json = json.dumps(value)
 
+    @property
+    def current_approval_request(self):
+        try:
+            return self.approval_requests.get(active=True)
```
Osmose: I think it might be possible to have more than one approval request, which isn't handled here.

normandy/recipes/api/serializers.py
rehandalal: ah! that would be awesome. i'll look into that.

normandy/recipes/api/serializers.py
rehandalal: yeah i should probably make this a read only field

normandy/recipes/models.py
Osmose: To be more specific:

1. There's nothing stopping someone from making multiple approval requests currently, so it's technically possible.
2. I suspect that multiple users will want to make changes to a recipe at the same time (such as the main heartbeat recipe, which hosts multiple surveys) and I'm not sure if we're handling that anywhere, either blocking it or allowing it. I haven't yet looked at the previous peer approval PR though, so I may be missing something there. I'll take a look.

rehandalal:
updated ^

normandy/recipes/api/serializers.py
```diff
@@ -21,10 +24,62 @@ class Meta:
         ]
 
 
+class ApprovalSerializer(serializers.ModelSerializer):
+    creator = UserSerializer(read_only=True)
+    creator_id = serializers.PrimaryKeyRelatedField(source='creator',queryset=User.objects.all(),
```
Osmose: Yessssssssssssssssssss

rehandalal: oh noes. I just noticed the missing space.

Osmose:
Looks good to me besides the multiple approval issue. I think we should make this a WIP PR until the API lands. Nice work!

normandy/recipes/models.py
```diff
@@ -174,3 +183,61 @@ def country(self):
...
+            approval = Approval(approval_request=self, creator=user)
+            approval.save()
+
+            self.active = False
+            self.save()
+
+            self.recipe.approval = approval
+            self.recipe.save()
+        else:
+            raise self.IsNotActive('Approval request has already been closed.')
```
Osmose: I really like this pattern of raising an exception from the class itself, it's neato!

rehandalal:
updated ^

normandy/recipes/api/views.py
```diff
@@ -155,6 +144,101 @@ def disable(self, request, pk=None):
...
+        except IntegrityError:
+            return Response(
+                {'active': 'You can only have one open approval request for a recipe.'},
+                status=status.HTTP_400_BAD_REQUEST)
+
+    @transaction.atomic()
+    @reversion.create_revision()
+    def update(self, request, *args, **kwargs):
+        """
+        Intercept PUT requests and have them create instead of update
```
Osmose: I feel like we tried making a mixin or something for this and failed, but we really should figure out some way to avoid duplicating the code for this.

rehandalal: Yeah `update` and `create` are already added to the class by a mixin, so we can't override them with a new mixin. We could probably create a new class with these methods overridden and then extend that, but the `update` method varies slightly for each of these view sets and the `create` method is slightly different for `ApprovalRequestViewSet

normandy/recipes/models.py
```diff
@@ -94,7 +108,9 @@ def enable(self):
 
     def disable(self, ignore_revision_id=False, *args, **kwargs):
         self.enabled = False
-        self.approver = None
+        self.approval = None
```
Osmose: I'm not sure if disabling a recipe should automatically remove it's approval. For example, if we want to update the action code, we have to disable all recipes using an action, update it, and then re-enable them. In other cases we might want to temporarily disable a recipe and then re-deploy it.

I guess the real question here is: What is approval approving of? Is it simply reviewing a given configuration for a recipe and verify that it is good? That'd be the Github model; require approval when the code changes, but don't require peer approval to deploy / rollback the code after it's been merged. In that model we'd remove the approval when a recipe changes but not when it is disabled.

This probably needs some feedback from User Advocacy.

Osmose:
It's not in this PR, but I noticed that we're removing approvals when a recipe is changed via the API.

1. Do we want to move this to the model level and remove it when it's saved?
2. There are some recipe changes that have no effect on the recipe as run by users and should not remove approval, such as changing the name of the recipe. There's a library called [django-dirtyfields](https://github.com/romgar/django-dirtyfields) that can help detect what fields are changing in a model before it's saved; perhaps we want a whitelist of acceptable changes?

normandy/recipes/api/views.py
```diff
@@ -155,6 +144,101 @@ def disable(self, request, pk=None):
...
+class ApprovalRequestViewSet(viewsets.ModelViewSet):
+    queryset = ApprovalRequest.objects.all()
+    serializer_class = ApprovalRequestSerializer
+    filter_fields = ('active',)
+    permission_classes = [
+        permissions.DjangoModelPermissionsOrAnonReadOnly,
+        AdminEnabledOrReadOnly,
+    ]
+
+    @transaction.atomic()
```
Osmose: There's gotta be a way to use a mixin or class decorator to auto-decorate viewset methods that modify things.

normandy/recipes/api/views.py
```diff
@@ -155,6 +144,101 @@ def disable(self, request, pk=None):
...
+    @reversion.create_revision()
+    def create(self, request, *args, **kwargs):
+        try:
+            return super().create(request, *args, **kwargs)
+        except IntegrityError:
```
Osmose: Actually I lied before; we should have a custom exception for this, and then catch that in `dispatch` so that we don't have to repeatedly catch this exception everywhere.

normandy/recipes/api/views.py
```diff
@@ -155,6 +144,101 @@ def disable(self, request, pk=None):
...
+                status=status.HTTP_400_BAD_REQUEST)
+
+    @transaction.atomic()
+    @reversion.create_revision()
+    @detail_route(methods=['POST'])
+    def approve(self, request, pk=None):
+        approval_request = self.get_object()
+        try:
+            approval_request.approve(request.user)
+        except ApprovalRequest.IsNotActive:
```
Osmose: Same as the IntegrityError, you should move the catching to `dispatch`.

Osmose: Actually I'm a liar, there's a convenient `handle_exception` method you can override instead of overriding `dispatch`.

normandy/recipes/api/views.py
```diff
@@ -155,6 +144,101 @@ def disable(self, request, pk=None):
...
+        return Response(status=status.HTTP_204_NO_CONTENT)
+
+    @detail_route(methods=['POST'])
+    def reject(self, request, pk=None):
+        approval_request = self.get_object()
+        try:
+            approval_request.reject()
+        except ApprovalRequest.IsNotActive:
+            return Response({'active': 'This approval request has already been closed.'},
+                            status=status.HTTP_400_BAD_REQUEST)
```
Osmose: I almost want to make our own set of subclasses that have these status codes already, like how Django has HttpResponseBadRequest and the like. But I'm not 100% on that. Other opinions are welcome!

normandy/recipes/api/views.py
```diff
@@ -155,6 +144,101 @@ def disable(self, request, pk=None):
...
+        return Response(status=status.HTTP_204_NO_CONTENT)
+
+    @detail_route(methods=['POST'])
+    def comment(self, request, pk=None):
+        if 'text' not in request.data:
```
Osmose: Couldn't you use the `ApprovalRequestComment` serializer to perform validation instead of doing custom validation here?

normandy/recipes/api/views.py
Osmose: Also if we always use them both together we should combine them into a single decorator.

normandy/recipes/api/views.py
Osmose: I think with some of the other proposed changes this would become possible. And if not, what other methods do we have? Could a class decorator work? Could we override `dispatch` and do something there?

rehandalal:
Updated and rebased to deal with merge conflict

normandy/base/decorators.py
```diff
@@ -14,3 +18,10 @@ def wrapped_view(*args, **kwargs):
         return view_func(*args, **kwargs)
     wrapped_view.short_circuit_middlewares = True
     return wraps(view_func)(wrapped_view)
+
+
+def reversion_transaction(view_func):
+    def wrapped_view(*args, **kwargs):
+        with transaction.atomic(), revisions.create_revision():
+            return view_func(*args, **kwargs)
+    return wraps(view_func)(wrapped_view)
```
Osmose: bd

normandy/recipes/api/views.py
```diff
@@ -119,42 +99,97 @@ def history(self, request, pk=None):
...
+
+    @detail_route(methods=['POST'])
+    def reject(self, request, pk=None):
+        approval_request = self.get_object()
+        approval_request.reject()
         return Response(status=status.HTTP_204_NO_CONTENT)
 
+    @detail_route(methods=['POST'])
+    def comment(self, request, pk=None):
+        if 'text' not in request.data:
```
Osmose: Did you try using the `ApprovalRequestComment` serializer for this validation? The docs say that DRF might even handle generating a response automatically if you use the `is_valid(raise_exception=True)` method on the serializer.

rehandalal: Oh, no I forgot to try that. I'll do that right now.

normandy/recipes/models.py
```diff
@@ -68,14 +76,17 @@ class Recipe(models.Model):
     # Fields that determine who this recipe is sent to.
     enabled = models.BooleanField(default=False)
     filter_expression = models.TextField(blank=False)
-    approver = models.ForeignKey(User, null=True, blank=True)
+    approval = models.OneToOneField(Approval, related_name='recipe', null=True, blank=True)
+
+    # A tuple of fields that can be edited without causing the recipe to be disabled
+    EDITABLE_FIELDS_WHITELIST = ('name', 'enabled', 'approval',)
```
mythmon: :100: :ok_hand:

Osmose: Oh, I see! Enabled being in the whitelist means that disabling a recipe does not remove its approval! Clever!

normandy/recipes/models.py
```diff
@@ -174,3 +200,64 @@ def country(self):
...
+        pass
+
+    class ActiveRequestAlreadyExists(Exception):
+        pass
+
+    @property
+    def is_approved(self):
+        return self.approval is not None
+
+    def approve(self, user):
```
Osmose: Probably want to wrap this method in a transaction to ensure all three objects get saved together.

rehandalal:
updated ^

Osmose:
ARPLUSSSS

Thanks for sticking with it, really nice work on this. :D

PR #142, opened 2016-05-11
Add API endpoint for recipe versions

rehandalal:
r?

### DIFF
--- a/normandy/recipes/api/views.py
+++ b/normandy/recipes/api/views.py
@@ -156,6 +156,14 @@ class RecipeViewSet(viewsets.ModelViewSet):
         return Response(status=status.HTTP_204_NO_CONTENT)
 
 
+class RecipeHistoryViewSet(viewsets.ReadOnlyModelViewSet):
+    queryset = Version.objects.all()
+    serializer_class = RecipeVersionSerializer
+    permission_classes = [
+        AdminEnabledOrReadOnly,
+    ]
+
+
 class ClassifyClient(views.APIView):
     authentication_classes = []
     permission_classes = []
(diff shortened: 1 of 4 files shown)


brittanystoroz:
r+ from me, gives me what I need to finish the front-end for history 👍

*gives the more python-y people a chance to look at this before merging*

Osmose:
It's a little annoying that we can't do URLs like `/recipes/4/history/7` easily, yet that would be weird too because the ID 7 might not be a history item for recipe 4. Oh well. This is fine.

normandy/recipes/urls.py
```diff
@@ -7,12 +7,14 @@
...
     ClassifyClient,
     RecipeViewSet,
+    RecipeHistoryViewSet,
 )
 
 # API Router
 router = DefaultRouter()
 router.register(r'action', ActionViewSet)
 router.register(r'recipe', RecipeViewSet)
+router.register(r'recipe_history', RecipeHistoryViewSet)
```
Osmose: I think this should be called `recipe_version` to match the pattern of using the name of the thing that you're querying for.

normandy/recipes/api/views.py
```diff
@@ -156,6 +156,14 @@ def disable(self, request, pk=None):
         return Response(status=status.HTTP_204_NO_CONTENT)
 
 
+class RecipeHistoryViewSet(viewsets.ReadOnlyModelViewSet):
+    queryset = Version.objects.all()
```
Osmose: Is it possible to filter this queryset to only versions for Recipes? Or am I wrong in thinking that this might match versions that aren't for recipes?

Osmose:
Looks good for the most part, nice work! I'll check it out again after an update.

rehandalal:
fix0red ^

Osmose:
This could use a test to ensure the filtering out of non-recipe versions is working. Like create an action revision and check if that ID returns a 404.

Other than that, I'm r+wc on this, nice work!

rehandalal:
test added ^

PR #143, opened 2016-05-11
[Bug 1269536] Add recipe preview page

brittanystoroz:
@Osmose r? 

I removed xhr and replaced it with [isomorphic-fetch](https://github.com/matthew-andrews/isomorphic-fetch), which also required [es6-promise](https://github.com/stefanpenner/es6-promise). I added both of these dependencies into the webpack config as a plugin.

I added some line notes to self_repair.js on some other changes I had to make. Let me know if those make sense, I had a little trouble following all the code so my changes might not have been the best solution to the problem I was seeing.


normandy/selfrepair/static/js/self_repair.js
```diff
@@ -0,0 +1,22 @@
+import Normandy from './normandy_driver.js';
+import {fetchRecipes, filterContext, doesRecipeMatch, runRecipe} from './self_repair_runner.js';
+
+// Actually fetch and run the recipes.
+fetchRecipes().then(recipes => {
+  let context = filterContext().then(context => {
```
brittanystoroz: I changed filterContext() to return a promise, rather than just the context object. The way it was [set up earlier](https://github.com/Osmose/normandy/blob/776acddfff1a33f1c65b213b22da1f6c0195dd41/normandy/selfrepair/static/js/self_repair_runner.js#L121-L127), the context object would be returned with the `normandy` property === a promise that was still pending.

normandy/selfrepair/static/js/self_repair.js
```diff
@@ -0,0 +1,22 @@
...
+    // Update Normandy driver with user's country.
+    Normandy._location.countryCode = context.normandy.country;
```
brittanystoroz: Which meant `context.normandy.country` was always going to be undefined.

normandy/selfrepair/static/js/self_repair.js
```diff
@@ -0,0 +1,22 @@
...
+
+    for (let recipe of recipes) {
+      doesRecipeMatch(recipe, context).then(([recipe, match]) => {
```
brittanystoroz: And the context object here wasn't guaranteed to be ready yet, so none of the recipes would come back as matches.

normandy/selfrepair/static/js/self_repair_runner.js
```diff
@@ -27,18 +26,20 @@ function loadAction(recipe) {
...
-                    if (!registeredActions[action.name]) {
-                        reject(new Error(`Could not find action with name ${action.name}.`));
-                    } else {
-                        resolve(registeredActions[action.name]);
-                    }
-                };
-                document.head.appendChild(script);
+            fetch(`/api/v1/action/${action_name}/`).then(function(response) {
+                return response.json().then(action => {
+                    let script = document.createElement('script');
```
mythmon: I usually do this like

```js
fetch(`/api/v1/action/${action_name}/`)
.then(response => response.json())
.then(action => {
  let script = ...
});
```

Returning `response.json()` directly lets you avoid indenting more.

normandy/selfrepair/static/js/self_repair_runner.js
```diff
@@ -114,12 +119,16 @@ function runRecipe(recipe) {
...
+export function filterContext() {
+    return new Promise(resolve => {
+        classifyClient().then(classifiedClient => {
+            resolve({
+                normandy: classifiedClient,
+                telemetry: {},
+                events: {},
+            })
+        })
+    })
```
mythmon: To me, having a promise inside a `new Promise` block seems like an anti-pattern. What do you think about this?

```js
return classifyClient()
.then(classifiedClient => {
  return {
    normandy: classifiedclient,
    telemetry: {},
    events: {},
  }
});
```

Besides this, I don't think we actually need `telemetry` and `events` on this object: Jexl will just return undefined if actions try to access them. That's even true if someone tries to access something nested, like `telemetry.general.user_id`.

normandy/selfrepair/static/js/self_repair.js
Osmose: I did that because JEXL accepts Promises as context values and automatically waits for them to resolve, but your point below about country is still an issue. Nice catch!

Osmose:
I'm going to try and review this after working on moving the actions to the main repo.

normandy/control/static/control/js/routes.js
```diff
@@ -3,13 +3,16 @@ import { Route } from 'react-router';
...
 import RecipeForm from './components/RecipeForm.jsx';
+import RecipePreview from './components/RecipePreview.jsx';
 import DeleteRecipe from './components/DeleteRecipe.jsx';
 
 export default (
   <Route component={ControlApp}>
     <Route path='control/'
       component={RecipeList}
-      ctaButton={{text: 'Add New', icon: 'plus', link: 'recipe/new/'}}
+      ctaButtons={[
```
Osmose: This is out-of-scope for this PR but I wonder if there's a better place we can specify these buttons than the Route components.

normandy/control/static/control/js/components/RecipePreview.jsx
```diff
@@ -0,0 +1,49 @@
...
+
+      this.setState({recipeExecuted: true});
+    }
+  }
+
+  render() {
+    const {recipe} = this.props;
+    if (recipe) {
+      return (
+        <div className="fluid-7">
```
Osmose: One thing I didn't get to with my branch was figuring out if there's something more useful we can put here than just a preview message. Maybe a fake console? A message when the recipe is finished executing? I dunno.

normandy/selfrepair/static/js/normandy_driver.js
```diff
@@ -123,10 +122,13 @@ let Normandy = {
...
         } else {
-            return xhr('POST', 'https://input.mozilla.org/api/v2/hb/', {
+            return fetch('https://input.mozilla.org/api/v2/hb/', {
+                method: 'POST',
                 data: data,
                 headers: {Accept: 'application/json'},
-            }).then(request => JSON.parse(request.responseText));
+            })
+            .then(response => response.json())
+            .then(heartbeatFlow => heartbeatFlow);
```
Osmose: Does this actually do anything?

webpack.config.js
```diff
@@ -24,6 +24,10 @@ module.exports = {
   plugins: [
     new BundleTracker({ filename: './webpack-stats.json' }),
     new ExtractTextPlugin('[name]-[hash].css'),
+    new webpack.ProvidePlugin({
+      'Promise': 'exports?global.Promise!es6-promise',
```
Osmose: Do we actually need this? Self-repair isn't meant to work outside of Firefox, and it was introduced in, like, Firefox 30-something (@gregglind knows for sure) where native Promises were already a thing . The preview will also eventually require the system add-on to be installed, so trying to make it compatible with other browsers will eventually fail.

Osmose:
> I'm going to try and review this after working on moving the actions to the main repo.

I lied. But my _second pass_ will totally happen after I do that.

webpack.config.js
gregglind: I don't know it for sure, but native promise is WAY OLD.  (29!)

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise

normandy/selfrepair/static/js/normandy_driver.js
brittanystoroz: Doesn't look like it o_O https://github.com/mozilla/normandy-actions/blob/master/actions/show-heartbeat/index.js#L73

normandy/control/static/control/js/routes.js
```diff
@@ -22,7 +22,7 @@ export default (
       component={RecipeForm}
       pageTitle="Edit Recipe"
       ctaButtons={[
-        {text: 'Preview', icon: 'eye', link: 'preview/'},
+        {text: 'Preview', icon: 'eye', link: 'preview/?testing=true'},
```
Osmose: Why add this back in? The idea was that the testing argument should completely replace this.

Osmose: Oh, because your other commits removed the modification to runRecipe that made the testing parameter work. Could you include that back in?

Osmose:
r+wc, nice work!

normandy/control/static/control/js/routes.js
brittanystoroz: Whoops, my b. Updated.

PR #144, opened 2016-05-16
WIP: Move actions into service repo

Osmose:
This is a near-done WIP that moves the [normandy-actions](https://github.com/mozilla/normandy-actions) repo into this one, including the action code, tests, and docs.

Only thing missing at this point is tests for the new management command, and some doc updates for the changes.

CC @brittanystoroz


jvehent:
Can we pause on this PR just for a day or two? I'm wondering if we there's a way to require commit signing on actions and what that means for our deployment flow. Sorry for being vague, it's a brainstorming exercise.

Osmose:
> Can we pause on this PR just for a day or two? I'm wondering if we there's a way to require commit signing on actions and what that means for our deployment flow. Sorry for being vague, it's a brainstorming exercise.

Sure, no problem.

What if we required signing only for the tagged commits that get built? Could we enforce that in our CI?

jvehent:
Yes I thing signed tags are the way to go. I'll create another issue to discuss it further, I don't want to block this PR.

Osmose:
Phew. Okay. I think this is ready.

I spun my wheels on trying to remove action code and schemas from the database, but couldn't come up with a good solution. I'm convinced one exists, but I think landing this is a better thing to do than to stand around trying to find one.

Action code lives in `normandy/recipes/static/actions`. Whenever the code or schema changes, you need to run `manage.py update_actions` to update the code in the database. The webpack config for actions will do this automatically if you pass the `--update-actions` flag to webpack. This can be combined with `--watch` so that you can just save actions and refresh the self-repair page during development.

`update_actions` disables recipes being used by actions that are updated. You can pass `--no-disable` to not do that.

We'll need to update the deploy process to call `update_actions` on deploy so that the database has the latest action code. This also means that deploys can potentially automatically disable recipes. I think that's fine.

I call upon @rehandalal for r?, although feedback from @brittanystoroz would be appreciated as well.

brittanystoroz:
This looks good from a functionality perspective so far, I'll take a closer look at the code later today. Only things I'm noticing after playing with it: 

* the recipe now seems to run twice on the recipe preview page?
* I'm not sure the `--update-actions` flag is working with `npm run start`. I rebased this over my implement-action-forms branch and wasn't seeing updates to action schemas unless I ran `manage.py update_actions`

Osmose:
> I'm not sure the --update-actions flag is working with npm run start. I rebased this over my implement-action-forms branch and wasn't seeing updates to action schemas unless I ran manage.py update_actions

I'll test it locally again, but did you restart webpack after rebasing? It needs to pick up a new plugin to run the command automatically.

Osmose:
Oh, if you add the argument to `npm start` or `npm run start` you need to add a `--` to get npm to pass the argument on to webpack. 

```sh
npm start -- --upgrade-actions
```

Not sure if we want to document this somewhere, the docs tell you to run webpack directly so I didn't include that note.

Osmose:
> the recipe now seems to run twice on the recipe preview page?

This should be fixed now.

brittanystoroz:
We should update the docs to run webpack through the npm script.

brittanystoroz:
Minus that doc change this all looks good to me, works as expected & tests are all passing  r+ 🎉

Osmose:
> We should update the docs to run webpack through the npm script.

Updated. Also changed the `npm start` command into two: `npm run build` and `npm run watch`.

mythmon:
Do we need any changes here for the Dockerfile build process? Or are the only changes during app startup with the DB?

Osmose:
> Do we need any changes here for the Dockerfile build process? Or are the only changes during app startup with the DB?

I don't think so. As long as we run the `update_actions` command on the admin node around the same time we run the `migrate` command this should work fine. I considered making the Dockerfile run `npm run build` instead of running webpack directly, but held off as I don't think it's that important.

rehandalal:
phew! sorry it took me so long. okay, i think this is good. r+

PR #145, opened 2016-05-16
Add content signing for Actions

mythmon:
This allows us to serve a `Content-Signature` header with Actions served from the CDN. The signatures are pulled from an [Autograph](https://github.com/mozilla-services/autograph/) server.

This does not add any automatic signature fetching. Newly updated actions won't have signatures, and signatures will quietly expire. The intent here is that we will add these later, possibly as a part of the Action review process. We'll probably set up `./manage.py update_action_signatures` on a cron job so it runs often.

We should be able to deploy these two commits separately, if we want to go for a two-phase deploy to be nice about migrations.

r?


normandy/recipes/models.py
```diff
@@ -149,12 +170,16 @@ def compute_implementation_hash(self):
         return hashlib.sha1(self.implementation.encode()).hexdigest()
 
     def save(self, *args, **kwargs):
-        # Save first so the upload is available.
-        super().save(*args, **kwargs)
```
mythmon: Actions aren't uploaded anymore, they are included in-line in the API, so this wasn't doing anything except getting in the way.

normandy/recipes/tests/__init__.py
```diff
@@ -21,6 +23,10 @@ class Meta:
     name = FuzzyUnicode()
     implementation = 'console.log("test");'
 
+    @factory.lazy_attribute
+    def implementation_hash(action):
+        return hashlib.sha1(action.implementation.encode()).hexdigest()
```
mythmon: Factories don't yet have a Django model when they are being built (when this is called), so I duplicated this logic from the model. There is a test in `test_factories.py` to make sure this stays in sync.

mythmon:
@jvehent If you wanted to see how Normandy will be using Autograph, here it is.

mythmon:
PS: To test this, I've updated [normandy-compose](https://github.com/mozilla/normandy-compose) to spin up an Autograph server as well, and docker images built from this branch should work with it.

docs/ops/config.rst
```diff
@@ -129,7 +129,32 @@ in other Django projects.
...
+    The URL where an Autograph_ server can be reached. If left blank, action
+    signing will be disabled.
+
+    .. _Autograph: https://github.com/mozilla-services/autograph/
+
+..envvar:: DJANGO_AUTOGRAPH_HAWK_ID
+
+    The pre-arranged ID to use for Hawk authentication with Autograph.
+
+..envvar:: DJANGO_AUTOGRAPH_SERCREY_KEY
```
jvehent: nit: `DJANGO_AUTOGRAPH_SECRET_KEY`

normandy/recipes/utils.py
```diff
@@ -46,3 +53,53 @@ def deterministic_sample(rate, inputs):
...
+            encoded_implementation = base64.b64encode(item.encode('utf8')).decode('utf8')
+            signing_request.append({
+                'template': 'content-signature',
+                'input': encoded_implementation,
+            })
+
+        res = self.session.post(url, json=signing_request)
+        res.raise_for_status()
+        signing_responses = res.json()
+        return [res['content-signature'] for res in signing_responses]
```
jvehent: r+

jvehent:
Looks good to me. I might suggest adding logs when requesting signatures maybe, in mozlog format.

mythmon:
@jvehent do you mean [this mozlog](http://mozbase.readthedocs.io/en/latest/mozlog.html)?

@rehandalal can you review this?

normandy/recipes/migrations/0027_auto_20160506_1944.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
```
rehandalal: depending on whose PR lands first (this one or #141) this migration might cause a conflict.

normandy/recipes/models.py
```diff
@@ -149,12 +170,16 @@ def compute_implementation_hash(self):
...
-
         # Update hash
+        old_hash = self.implementation_hash
         self.implementation_hash = self.compute_implementation_hash()
-        super().save(update_fields=['implementation_hash'])
+
+        # Invalidate signature if content has changed
+        if self.implementation_hash != old_hash:
```
rehandalal: maybe we should wait for #141 to land because I added django-dirtyfields which you could use for this

normandy/recipes/tests/__init__.py
rehandalal: yay for tests!

normandy/recipes/models.py
mythmon: That's probably a good idea

normandy/recipes/migrations/0027_auto_20160506_1944.py
rehandalal: since #141 merged, this will need to be changed.

mythmon:
I talked to @jvehent about this, and we're going to have to do signing differently. I'm going to close this PR and open a new one later. I'll re-use a lot of stuff from this PR though.

PR #146, opened 2016-05-17
[Bug 1271272] Recipe history

brittanystoroz:
Adds a history route & component for viewing a recipe's revisions.

The history page is kind of sparse right now, but I think it might make sense to integrate some of the approval stuff here, which will fill it out a bit more. Thoughts on that tk.

I added some rough notification text to the recipe form page to let users know they are viewing a revision rather than the current state of the recipe. This should eventually be broken out into a separate notification component so we can do more with it (e.g. handle api response error/success messages) but I think that should be saved for another PR. 

@rehandalal  r?

### DIFF
--- a/normandy/control/static/control/admin/sass/control.scss
+++ b/normandy/control/static/control/admin/sass/control.scss
@@ -182,6 +182,28 @@
 }
 
 
+/* Recipe History */
+li {
+  p {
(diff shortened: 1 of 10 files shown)


rehandalal:
this is great! r+

PR #147, opened 2016-05-19
[Bug 1266483] Implement action forms

brittanystoroz:
Stopping here before this gets too unruly. 

This adds in the action forms for adding/editing a recipe. I'd still like to add some component tests for this, since the forms are a bit more complex than our other components, but this is at least ready for review.

This does not include any form validation / API response handling. Implementing these pieces required some significant refactoring, so I made another branch to PR with that after this one lands.

Anybody dying to r+ this? @Osmose @rehandalal @mythmon


normandy/control/static/control/js/components/ActionForm.jsx
```diff
@@ -0,0 +1,58 @@
...
+class ActionForm extends React.Component {
+  constructor(props) {
+    super(props)
+  }
+
+  shouldComponentUpdate(nextProps, nextState) {
+    const currentProps = _.pick(this.props, 'initialValues', 'values', 'fields', 'selectedAction');
+    const incomingProps = _.pick(nextProps, 'initialValues', 'values', 'fields', 'selectedAction');
+
+    return (!_.isEqual(currentProps, incomingProps) || !_.isEqual(this.state, nextState));
```
mythmon: I don't understand how you picked these 4 files. It seems like you only use `fields` and `arugments_schema` in `render()`. Why include `selectedAction`, `initialValues`, and `values`? Why not include `argument_schema`?

Was there a performance problem without this? I can't really imagine calling render to really be that expensive if the DOM doesn't change.

normandy/control/static/control/js/components/RecipeForm.jsx
```diff
@@ -1,27 +1,93 @@
...
+    this.changeAction = this.changeAction.bind(this);
+
+    this.state = {
+      availableActions: [],
+      selectedAction: null,
+    };
+  }
+
+  getAvailableActions(recipeId) {
+    apiFetch(`/api/v1/action/`)
```
mythmon: Nit, maybe. Should we only use backticks for things with variables in it, and use quotes otherwise? Or is it fine to use backticks here, style wise?

normandy/control/static/control/js/components/RecipeForm.jsx
```diff
@@ -1,27 +1,93 @@
...
+              });
+            }
+            return action;
+          })
+        });
+
+        this.setState({
+          availableActions
+        });
+    });
```
mythmon: This is super nitty, so feel free to ignore me. With `.then` on a new line, indenting it always results in these double-detents, which seem weird to me (and harder to read).

I usually do this:

```js
fetchTheThing('foo')
.then(foo => {
   foo.bar();
});
```

instead of this

```
fetchTheThing('foo')
  .then(foo => {
    foo.bar();
  });
```

Either way, I think we should standardize on one or the other. Since you're writing most of this code, you should go with whatever you like better. (Can we encode this in eslint?)

normandy/control/static/control/js/components/ActionForm.jsx
brittanystoroz: Good question. I can take out `initialValues` and `selectedAction`, and replace them with `arguments_schema`. I still need `values` because the input values are set to `values[inputName]`, and we need a re-render in order to see the changes we make to each field. (If you take `values` out, you can type in an input field and it will look as if nothing is happening.)

Without this, re-rendering occurs an INSANE amount of times in redux-forms. It fires after literally anything you could possibly do with your mouse or keyboard (onBlur, for example) because there are so many things changing in the form state at any given time.  It just seemed excessive and unnecessary. Also, any time the recipeForm re-rendered, the actionForm would go along for a ride as well. It was definitely laggy while I had devtools opened, but that could've been devtools-specific. I can test it out and see how it works if I remove these lifecycle methods.

brittanystoroz: Performance seems fine with `shouldComponentUpdate` removed & dev tools closed. Will add back in if anyone notices otherwise.

normandy/control/static/control/js/components/RecipeForm.jsx
Osmose: Backticks only when necessary IMO

normandy/control/static/control/js/components/RecipeForm.jsx
```diff
@@ -1,34 +1,93 @@
...
 import composeRecipeContainer from './RecipeContainer.jsx'
+import ActionForm from './ActionForm.jsx'
 
-
-class RecipeForm extends React.Component {
+export class RecipeForm extends React.Component {
   constructor(props) {
     super(props);
     this.submitForm = this.submitForm.bind(this);
+    this.getAvailableActions = this.getAvailableActions.bind(this);
```
Osmose: /me throws his hands up and curses React and JavaScript

normandy/control/static/control/js/components/ActionForm.jsx
```diff
@@ -0,0 +1,51 @@
...
+      case 'console-log':
+        childForm = (<ConsoleLogForm fields={fields} />);
+        break;
+      default:
+        childForm = childForm;
+    }
+
+    return (
+      <div id="action-configuration">
+        <i className="fa fa-caret-up fa-lg"></i>
```
Osmose: Probably not for this PR but an `<FAIcon>` or `<Icon>` or `<FontAwesome>` component might be neat someday.

normandy/control/static/control/js/utils/formHelpers.js
```diff
@@ -0,0 +1,65 @@
...
+export function generateFieldsFromSchema(jsonSchema) {
+  let props = jsonSchema.properties;
+  return processProperties(orderPropNames(props), props);
+}
+
+export function formatLabel(labelName) {
+  return labelName.replace( /([A-Z])/g, " $1" );
+}
+
+export {
```
Osmose: Aren't these already exported?

normandy/control/static/control/js/components/action_forms/HeartbeatForm.jsx
```diff
@@ -0,0 +1,106 @@
+import React from 'react'
+import { reduxForm } from 'redux-form'
+import { _ } from 'underscore'
+import { formatLabel } from '../../utils/formHelpers.js'
+
+
+class SurveyForm extends React.Component {
+  constructor(props) {
```
Osmose: This seems unnecessary if it's just calling `super`.

normandy/control/static/control/js/components/ActionForm.jsx
```diff
@@ -0,0 +1,51 @@
...
+    let childForm = 'No action form available';
+
+    switch(name) {
+      case 'show-heartbeat':
+        childForm = (<HeartbeatForm fields={fields} />);
+        break;
+      case 'console-log':
+        childForm = (<ConsoleLogForm fields={fields} />);
+        break;
+      default:
```
Osmose: This `default` clause seems unnecessary.

normandy/control/static/control/js/components/RecipeForm.jsx
```diff
@@ -42,6 +101,22 @@ class RecipeForm extends React.Component {
...
           </div>
         </div>
+        <div className="row">
+          <div className="fluid-3">
+            <label>Action</label>
+            <select {...action_name} onChange={this.changeAction}>
+              <option>Select an action</option>
+              {
+                availableActions.map(action => {
+                  return (<option key={action.name} value={action.name}>{action.name}</option>)
```
Osmose: The `return` is unnecessary if you use the short-return form of the double-arrow function. Also destructuring can help:

```js
availableActions.map(({name}) => (
  <option key={name} value={name}>{name}</option>
))
```

normandy/control/static/control/js/components/action_forms/HeartbeatForm.jsx
```diff
@@ -0,0 +1,106 @@
...
+    let containerClass = 'fluid-8';
+    if (selectedSurvey) {
+      headerText = selectedSurvey.title.initialValue || "New survey"
+      containerClass += ' active';
+    }
+
+    return (
+      <div id='survey-form' className={containerClass}>
+        { selectedSurvey &&
+          <a className="return-to-defaults" href="#" onClick={(e) => {
```
Osmose: An inline handler like this is really hard to read vs a method on the component class.

normandy/control/static/control/js/components/action_forms/HeartbeatForm.jsx
```diff
@@ -0,0 +1,106 @@
...
+            e.preventDefault();
+            setSelectedSurvey()
```
Osmose: This is missing a semicolon, but I also find the behavior of "show the default form when nothing or null is passed" is weird. A dedicated method for showing the defaults would be better.

Osmose:
Review is going to be slow because I'm slow, will comment when I'm done.

normandy/control/static/control/js/components/action_forms/ConsoleLogForm.jsx
```diff
@@ -0,0 +1,27 @@
...
+  constructor(props) {
+    super(props)
+  }
+
+  render() {
+    const { fields } = this.props;
+    return (
+      <div className="row">
+        {
+          Object.keys(fields).map(fieldName => {
```
Osmose: This feels really weird. I get why: If we want to depend on the argument schema for defining these forms, we need to do this to handle new fields being added.

Osmose:
@brittanystoroz The heartbeat form as written is pretty brittle to schema changes. Same with the parser for the arguments schema itself; it isn't generic enough to handle _any_ argument schema, just the ones we currently have.

On the other hand, I do suspect individual survey fields are the main fields that would change, and the form handles those changing no problem. So it's not entirely brittle.

What do you think about dropping the argument-schema parsing completely and just hardcoding the fields? I'm guessing based on this PR not doing it that there's a problem with it vs relying on the schema?

normandy/control/static/control/js/components/RecipeForm.jsx
brittanystoroz: that looks nice 😍

normandy/control/static/control/js/components/ActionForm.jsx
brittanystoroz: Agreed. I will file a low-priority bug for this 👍

brittanystoroz:
@Osmose agreed re: brittleness. I figured we would just add flexibility/handle different types of schemas as they came in, but right now it's definitely a hybrid of trying to be flexible but hard-coding individual action forms anyway soooo I'm not opposed to doing a full-blown hard-code.  It will remove of a lot of the helper code. If I had added specific form-config files for each action, I'd say let's try for a generic/flexible approach. I will push a hard-coded implementation today and we can see how that looks.

brittanystoroz:
Updated with redux form fields hardcoded.

normandy/control/static/control/js/components/RecipeForm.jsx
```diff
@@ -1,34 +1,90 @@
...
+        action.fields = reduxFormFields[action.name];
+
+        if (selectedActionName === action.name) {
+          this.setState({
+            selectedAction: action
+          });
+        }
+
+        return action;
+      });
```
mythmon: `Array.map` is generally used for making a new list that has be transformed, not for modifying a list in place. In fact, this is sort of throwing away a lot of things, since it is making an entirely new list and then not savings it. This would probably be better as `availableActions.forEach`, which doesn't build the new list, or as a `for (let action in availableActions) { ... }`.

normandy/control/static/control/js/components/RecipeForm.jsx
```diff
@@ -1,34 +1,90 @@
...
+  }
 
+  render() {
+    const { fields: { name, filter_expression, enabled, action_name }, recipe, recipeId, handleSubmit, viewingRevision } = this.props;
+    const { availableActions, selectedAction } = this.state;
     return (
       <form onSubmit={handleSubmit(this.submitForm)} className="crud-form">
-        { viewingRevision ?
-          <p className="notification info">
+        { viewingRevision &&
```
mythmon: That's a neat trick.

normandy/control/static/control/js/components/RecipeForm.jsx
brittanystoroz: I do this every time, I don't know why. `.map` has some sick hold on me.

normandy/control/static/control/js/components/action_forms/HeartbeatForm.jsx
brittanystoroz: So the reason I like depending on `selectedSurvey` for what to show there is that it will toggle between showing a survey/showing the defaults automatically since it's updating with each state change.

If I add a `showDefaults` method, I'd honestly just want it to unset the `selectedSurvey` like I'm already doing. That on it's own might still be better, since `showDefaults` is more explicit/reads more semantically. Do you imagine that method doing anything different or is that the solution you had in mind?

Osmose: > Do you imagine that method doing anything different or is that the solution you had in mind?

That is exactly what I had in mind. You're totally right on using `selectedSurvey` itself.

brittanystoroz:
This has been updated with all feedback so far. I've added additional component tests for the console-log form and show-heartbeat forms. Testing the remaining functionality on the RecipeForm component is a little tricky right now. See my struggles here: https://twitter.com/brittanystoroz/status/738714184624525312 Looking into solutions but might never find one ¯\_(ツ)_/¯

normandy/control/static/control/js/components/action_forms/ConsoleLogForm.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+
+class ConsoleLogForm extends React.Component {
+  render() {
+    const { fields } = this.props;
+    return (
+      <div className="row">
+        <div key="message" className="fluid-3">
```
Osmose: Does this need a key? I thought you only need that for things in lists.

normandy/control/static/control/js/components/action_forms/HeartbeatForm.jsx
```diff
@@ -0,0 +1,105 @@
...
+      <div className="row">
+        <div className="fluid-4">
+          <div className="row">
+            <label>Survey ID</label>
+            <input type="text" field={fields.surveyId} {...fields.surveyId} />
+          </div>
+          <div className="row array-field">
+            <h4>Surveys</h4>
+
+            <a className="button add-field" onClick={this.addSurvey.bind(this)}><i className="fa fa-plus"></i> Add Survey</a>
```
Osmose: Do we have the babel plugin to use the [function bind operator](https://babeljs.io/docs/plugins/transform-function-bind/)?

```js
onClick={::this.addSurvey}
```

normandy/control/static/control/js/utils/formHelpers.js
```diff
@@ -0,0 +1,23 @@
+/* Redux forms with complex/nested data structures require
+ * a specialized fields array to handle objects and arrays.
+ * Eg: ['surveyId', 'surveys[].title', 'defaults.thankYouMessage']
+ */
+const reduxFormFields = {
```
Osmose: I think it'd be clearer to store these in the form files instead of in a shared helpers file.

package.json
```diff
@@ -19,10 +19,12 @@
     "classnames": "^2.2.5",
     "cssmin": "^0.4.3",
     "font-awesome": "^4.5.0",
+    "fs": "0.0.2",
     "isomorphic-fetch": "^2.2.1",
     "jexl": "^1.1.3",
     "jquery": "^2.2.3",
     "json-editor": "^0.7.23",
+    "json-schema-ref-parser": "^3.1.2",
```
Osmose: We don't need this dependency anymore.

normandy/control/static/control/js/components/ActionForm.jsx
```diff
@@ -0,0 +1,45 @@
...
+      case 'console-log':
+        childForm = (<ConsoleLogForm fields={fields} />);
+        break;
+    }
+
+    return (
+      <div id="action-configuration">
+        <i className="fa fa-caret-up fa-lg"></i>
+        <div className="row">
+          <p className="help fluid-4">{arguments_schema.description || arguments_schema.title }</p>
```
Osmose: If we move this to the child form and make each form include a description itself instead of pulling it from the schema we can remove dependence on the schema entirely, which would be nice.

normandy/control/static/control/js/components/RecipeForm.jsx
```diff
@@ -42,6 +97,22 @@ class RecipeForm extends React.Component {
             <textarea field={filter_expression} {...filter_expression} />
           </div>
         </div>
+        <div className="row">
+          <div className="fluid-3">
```
Osmose: I think this entire `div` could be a standalone stateless component to help make the code more readable:

```jsx
<ActionNameSelect 
  className="fluid-3" 
  field={action_name}
  onChange={this.changeAction}
/>
```

normandy/control/static/control/js/components/action_forms/HeartbeatForm.jsx
```diff
@@ -0,0 +1,105 @@
...
+
+class SurveyForm extends React.Component {
+  render() {
+    const { selectedSurvey, fields, showDefaults } = this.props;
+    const surveyObject = selectedSurvey || fields.defaults;
+    let headerText = 'Default Survey Values';
+    let containerClass = 'fluid-8';
+    if (selectedSurvey) {
+      headerText = selectedSurvey.title.initialValue || 'New survey';
+      containerClass += ' active';
```
Osmose: We have [classnames](https://www.npmjs.com/package/classnames), which makes conditional classes a little more readable. Maybe it'd help here?

normandy/control/static/control/js/components/action_forms/HeartbeatForm.jsx
```diff
@@ -0,0 +1,105 @@
...
+          <div className="row array-field">
+            <h4>Surveys</h4>
+
+            <a className="button add-field" onClick={this.addSurvey.bind(this)}><i className="fa fa-plus"></i> Add Survey</a>
+
+            { fields.surveys.length ?
+              <ul>
+                {
+                  fields.surveys.map((survey, index) =>
+                    <li key={index} className={_.isEqual(survey, selectedSurvey) ? 'active' : ''} onClick={this.setSelectedSurvey.bind(this, survey)}>
```
Osmose: I don't feel as strongly here, but maybe another stateless component would make this more readable as well?

normandy/control/static/control/js/components/action_forms/HeartbeatForm.jsx
```diff
@@ -0,0 +1,105 @@
...
+    return (
+      <div id='survey-form' className={containerClass}>
+        { selectedSurvey &&
+          <span className="return-to-defaults" onClick={showDefaults}>
+            <i className="fa fa-long-arrow-left pre"></i> Return to defaults
+          </span>
+        }
+        <h4>{headerText}</h4>
+        {
+          Object.keys(surveyObject).map(fieldName =>
```
Osmose: I am undecided as to whether I think we should stick with this (where we specify the fields in the field list we give to Redux Form, and then pull from that list dynamically here) or go with fully explicit forms (where we specify the fields in the field list as well as having individual components here for each field). If we were fully explicit this would look something like (assuming we also add a `TextField` component):

```jsx
<TextField className="row" label="message" field={surveyObject.message} />
<TextField className="row" label="thanksMessage" field={surveyObject.thanksMessage} />
<TextField className="row" label="engagementURL" field={surveyObject.engagementURL} />
```

Either would probably be fine, so maybe this should remain unchanged.

Osmose:
I've never written tests for React stuff before so I'm super green, but I wonder if instead of testing if the HTML we output matches what we expect, we instead should be testing that after interacting with the forms the data they submit matches what we expect? As it is these tests are very tightly coupled to the HTML that gets output, and while we can't completely avoid that, we can limit it.

I'm imagining tests like "If I click the new survey button, edit this field, edit a default field, and click the other survey and click the delete button, the form outputs JSON of the form {BLAH}".

Osmose:
I really like where this PR has gone. It was a much easier read and I'm happy with most of the changes. Most of my comments are nits, except the tests question and the possible addition of some more stateless components. Nice work!

normandy/control/static/control/js/utils/formHelpers.js
brittanystoroz: That makes things a little tricky, because we actually need the form fields ready and available in RecipeForm.jsx. They have to be passed into the ActionForm component as soon as it initializes. Dynamically adding and changing redux form fields is not really possible without completely destroying and re-initializing the form.

normandy/control/static/control/js/components/ActionForm.jsx
brittanystoroz: Actually, if I do this, my previous line note might be irrelevant. (re: moving the list of form fields into the specific form files.) Will try and report back.

normandy/control/static/control/js/utils/formHelpers.js
Osmose: If it's a big problem then don't worry about it.

normandy/control/static/control/js/components/action_forms/HeartbeatForm.jsx
brittanystoroz: It's a little bit easier to iterate through the list with `Object.keys` right now because we toggle between showing the default fields and a selected survey's fields. Selected surveys will have a title & weight field associated with them, defaults will not. So we'd need to add some conditional logic in to determine whether or not those should be rendered if we decide to hardcode the fields rather than iterating the keys.

brittanystoroz:
@Osmose Ok -- everything from round 2 of review has been fixed or at least addressed in some way. 

The tests are hard. You're right that they are too tightly coupled to the HTML. If we do a [shallow render](https://facebook.github.io/react/docs/test-utils.html#shallow-rendering) like in the recipeForm test, that's really all we can test - the HTML. I ran into a lot of problems trying to do a full render in order to test the methods on the components (setting up a store, having to import all dependencies for child components, etc), so testing the HTML at least provided *some* kind of validation that interactions were working as expected.

That said, obviously all the tests are broken now anyway. So I will work on those. But feel free to review the other changes in the meantime.

normandy/control/static/control/js/components/action_forms/HeartbeatForm.jsx
Osmose: +1, leave it!

normandy/control/static/control/js/components/action_forms/HeartbeatForm.jsx
```diff
@@ -9,14 +10,17 @@ class SurveyForm extends React.Component {
     const { selectedSurvey, fields, showDefaults } = this.props;
     const surveyObject = selectedSurvey || fields.defaults;
     let headerText = 'Default Survey Values';
-    let containerClass = 'fluid-8';
+    let containerClasses = classNames({
+      'fluid-8': true,
```
Osmose: \<nit\>You can just pass `'fluid-8'` as a positional argument and it'll work. \</nit\>

Osmose:
Looked over the changes, they all look good to me! Nice work on this!

I might even go so far as to say we should just remove the tests and merge, and spend some time together in London trying to figure out how we want to address the testing issues. Once the tests are removed, I'm r+ on this.

How does that sound?

brittanystoroz:
Yes please. Tests removed.

PR #148, opened 2016-05-20
Accept lower case booleans in recipes API filters.

mythmon:
I was having the hardest time getting the addon to filter this, because I was using `?enabled=true`.

@rehandalal r?


rehandalal:
nice! r+

PR #149, opened 2016-05-24
Use timezone.now for model defaults

mythmon:
Without this I get things like 

```python
>>> from normandy.recipes.models import Approval
>>> from django.contrib.auth.models import User
>>> me = User.objects.get()
>>> a = Approval(creator=me)
>>> a.save()
/home/mythmon/src/normandy/normandy/venv/lib/python3.5/site-packages/django/db/models/fields/__init__.py:1453:
RuntimeWarning: DateTimeField Approval.created received a naive datetime (2016-05-24 17:53:06.309489) while time zone support is active.
```

@rehandalal r?


rehandalal:
makes sense. r+

PR #150, opened 2016-05-25
Don't show revision notification if clicking first history record

brittanystoroz:
Quick fix on the recipe history stuff. The top record listed in the RecipeHistory component is always the current recipe, not an outdated revision. If a user clicks on this record, it should bring you to the edit page for the recipe and not display a notification that you are viewing an older version.


rehandalal:
lgtm, r+

PR #154, opened 2016-06-07
Make '/' uri respond with 200

casebenton:
fixes bug 1250575

r?


normandy/base/tests/test_views.py
```diff
@@ -0,0 +1,3 @@
+def test_index(client):
+    url = '/'
+    assert client.get(url).status_code == 200
```
Osmose: Might as well just put `'/'` directly in the method call.

Osmose:
r+, nice work!

PR #156, opened 2016-06-08
Document the concepts behind SHIELD and a full example.

Osmose:
PROSE PROSE PROSE

This change adds better documentation to describe the concepts behind how Normandy works, the mitigations we have in place to prevent compromise, and a specific, detailed example of how to go from nothing to an action being executed by the system add-on. It's intended for multiple audiences, but especailly for security and Firefox QA.

@mythmon r?

@casebenton r? I'd like your feedback as someone who is still very new to Normandy and would benefit from these kind of docs existing. Are we missing anything here that you don't understand? Do you find the docs useful?

@jvehent r? I'm probably missing a few things in the threat model section, so feedback on that in particular (and also on the entire document really) would be super-helpful.


docs/dev/concepts.rst
```diff
@@ -0,0 +1,175 @@
...
+
+Runtime
+-------
+The runtime is an execution environment that can affect a running instance of
+Firefox. Upon activation (typically a few moments after Firefox launches), the
+runtime:
+
+1. Downloads :ref:`recipes <recipes>` from the Normandy service.
+2. Evaluates the recipe filters and filters out recipes that do not match the
+   client the runtime is installed within.
```
jvehent: To prevent fraudulent injections in the filtering logic, we should verify signatures prior to evaluating the filters.

docs/dev/concepts.rst
```diff
@@ -0,0 +1,175 @@
...
+------------
+Since the goal of SHIELD is to allow Mozilla to perform certain privileged
+actions quickly without shipping full updates to Firefox, it is a tempting
+target for compromising Firefox users. Normandy includes several security
+controls to help mitigate this risk.
+
+Action and Recipe Signing
+^^^^^^^^^^^^^^^^^^^^^^^^^
+Actions and recipes that are downloaded by the runtime are signed using the
+autograph_ service. The runtime verifies the signature upon downloading the
```
jvehent: I'd refer to the protocol as well `signed according to the Content-Signature protocol as provided by the autograph service`.
Unfortunately, we don't yet have a spec for Content-Signature for you to link to.

docs/dev/concepts.rst
```diff
@@ -0,0 +1,175 @@
...
+service.
+
+This helps prevent compromise of a single account from compromising the entire
+service, since two accounts need to be compromised to publish a recipe.
+
+Action Sandbox
+^^^^^^^^^^^^^^
+Actions are executed within a JavaScript sandbox by the runtime. The sandbox
+limits the access of the JavaScript to prevent it from modifying Firefox in ways
+that haven't been reviewed and approved beforehand.
```
jvehent: It would be interested to document what that approval process look like. Maybe there should be a `secreview+` from the seceng team when the sandbox changes? I don't know what's appropriate.

jvehent:
r+ with comments. Thanks for writing this! :+1:

mythmon:
This seems accurate and understandable to me. We should probably poke @casebenton, since he probably didn't notice the notification.

Osmose:
Benton gave me feedback in person; he didn't find the code examples terribly useful but otherwise liked the docs.

casebenton:
I like the full example with code and an accompanying description in example.rst. I think that the short code snippets in concepts.rst were more difficult to understand at first read through, but they made more sense after I read through again.

docs/dev/concepts.rst
Osmose: I agree that that should be documented, but not here. I think the wiki is a better place for documentation of processes.

PR #157, opened 2016-06-09
Sort recipe list by enabled & last_updated

brittanystoroz:
I have a feeling we might decide against sorting by enabled AND last updated, but this is what we originally discussed back in the day. (Since recipes are going to be edited more often while they're disabled, it might wind up making sense to nix the enabled sorting and just go with last_updated)

Fixes bug 1266480 r?


mythmon:
There is silly test failure here, and a lint failure (there is a line in the migration that is too long). Besides that, it looks fine to me.

r+ after CI passes.

normandy/recipes/tests/test_serializers.py
```diff
@@ -15,6 +15,7 @@ def test_it_works(self, rf):
         assert serializer.data == {
             'name': recipe.name,
             'id': recipe.id,
+            'last_updated': Whatever(),
```
brittanystoroz: @mythmon is there a better way to fix this test? If I try to set it to something like `timezone.now()` I get a failure where `last_updated` doesn't match up: 

```
{'last_updated': '2016-06-09T19:32:08.115291Z'} != {'last_updated': datetime.datetime(2016, 6, 9, 19, 32, 8, 120002, tzinfo=<UTC>)}
```

normandy/recipes/migrations/0029_recipe_last_updated.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.9 on 2016-06-09 19:23
```
rehandalal: fyi: we've been adding: 
```python
# flake8: noqa
```
to get the linter to ignore migrations

normandy/recipes/tests/test_serializers.py
mythmon: Can you do something like `recipe.last_updated`? I guess that would still have the type error. Maybe convert it to a string with `recipe.last_updated.isoformat()`?

normandy/recipes/tests/test_serializers.py
```diff
@@ -15,6 +15,7 @@ def test_it_works(self, rf):
         assert serializer.data == {
             'name': recipe.name,
             'id': recipe.id,
+            'last_updated': recipe.last_updated.replace(tzinfo=None).isoformat() + 'Z',
```
brittanystoroz: Took a little extra magic and is not pretty but this seems to be the only way to format the date to match O_o

mythmon: Oh, that's pretty gross. I can't find an easy way to make it better though :\

rehandalal: FWIW, i think `Whatever()` is a better option.

brittanystoroz: I like `Whatever()` better as well. Less ugly and maybe it's not something that's super important to test the exact value of, just that it exists.

mythmon: I wish there was a way to assert it was a date in the right format, without getting all ugly. I'm fine with `Whatever()` though.

brittanystoroz:
Reverted to `Whatever()`

PR #158, opened 2016-06-09
Fixed show-heartbeat action's handling of telemetry data

casebenton:
Updated the driver to take `surveyId`, `surveyVersion`, and `testing` as options rather than
bundled in a single object option. Updated index.js to pass these options on directly to the
UITour.showHeartbeat call in the options object. Updated documentation to reflect this.

fixes bug 1277700

### DIFF
--- a/docs/dev/driver.rst
+++ b/docs/dev/driver.rst
@@ -55,7 +55,7 @@ The driver object contains the following attributes:
    :param learnMoreUrl: URL to open when the "Learn More" button is clicked.
    :param surveyId: Extra data to be stored in telemetry.
    :param surveyVersion: Extra data to be stored in telemetry.
-   :param testing: Extra data to be stored in telemetry when Normandy is in testing.
+   :param testing: Extra data to be stored in telemetry when Normandy is in testing mode.
    :returns: A Promise that resolves with an event emitter.
 
    The emitter returned by this function can be subscribed to using ``on``
--- a/docs/dev/driver.rst
+++ b/docs/dev/driver.rst
@@ -53,9 +53,9 @@ The driver object contains the following attributes:
       the user won't be shown anything.
    :param learnMoreMessage: Text to show on the "Learn More" button.
    :param learnMoreUrl: URL to open when the "Learn More" button is clicked.
-   :param extraTelemetryArgs: Object containing extra arguments to send to
-      telemetry associated with this heartbeat call. Defaults to an empty
-      object.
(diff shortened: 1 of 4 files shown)


normandy/recipes/static/actions/show-heartbeat/index.js
```diff
@@ -116,26 +116,28 @@ export default class ShowHeartbeatAction extends Action {
...
 
-        let extraTelemetryArgs = {
-            surveyId: surveyId,
-            surveyVersion: this.recipe.revision_id,
-        };
-        if (this.normandy.testing) {
-            extraTelemetryArgs.testing = 1;
-        }
-
+        // Object declared for showHeartbeat call below, so that optional
```
Osmose: +1 for adding comments, but this particular fact isn't necessary to point out. It's pretty clear why you're storing the data in a variable.

normandy/recipes/tests/actions/show-heartbeat.js
```diff
@@ -186,11 +184,9 @@ describe('ShowHeartbeatAction', function() {
 
         await action.execute();
         expect(this.normandy.showHeartbeat).toHaveBeenCalledWith(jasmine.objectContaining({
-            extraTelemetryArgs: {
-                surveyId: 'my-survey',
-                surveyVersion: 42,
-                testing: 1,
-            },
+            surveyId: 'my-survey',
+            surveyVersion: 42,
```
Osmose: You can leave out `surveyId` and `surveyVersion` here since this test is only about checking if the `testing` argument is present. The others are handled in a different test.

docs/dev/driver.rst
```diff
@@ -53,9 +53,9 @@ The driver object contains the following attributes:
       the user won't be shown anything.
    :param learnMoreMessage: Text to show on the "Learn More" button.
    :param learnMoreUrl: URL to open when the "Learn More" button is clicked.
-   :param extraTelemetryArgs: Object containing extra arguments to send to
-      telemetry associated with this heartbeat call. Defaults to an empty
-      object.
+   :param surveyId: Extra data to be stored in telemetry.
+   :param surveyVersion: Extra data to be stored in telemetry.
+   :param testing: Extra data to be stored in telemetry when Normandy is in testing.
```
Osmose: "Is in testing mode" would be better.

Osmose:
I helped @casebenton test this locally and we found that it doesn't fix the stated problem yet. He's investigating.

Osmose:
r+, nice work!

PR #159, opened 2016-06-12
Add notification bar for handling form submissions

brittanystoroz:
This adds a super bare-bones notification component to let you know if you successfully added/updated a recipe or not. We'll probably want to make it a bit more robust in the future (e.g. handling multiple messages) but I think for our current use case this will do.

I hard-coded some generic success/error messages for the form submission actions, but have another branch with more specific error handling [here](https://github.com/brittanystoroz/normandy/tree/validate-forms). Just wanted to split these PRs up to keep 'em tiny.

Fixes bug 1271276 r?

demo gifs tktk

### DIFF
--- a/normandy/control/static/control/js/actions/ControlActions.js
+++ b/normandy/control/static/control/js/actions/ControlActions.js
@@ -7,6 +7,7 @@ export const RECIPES_RECEIVED = 'RECIPES_RECEIVED';
 export const SINGLE_RECIPE_RECEIVED = 'SINGLE_RECIPE_RECEIVED';
 
 export const SET_SELECTED_RECIPE = 'SET_SELECTED_RECIPE';
+export const SET_NOTIFICATION = 'SET_NOTIFICATION';
 
 export const RECIPE_ADDED = 'RECIPE_ADDED';
 export const RECIPE_UPDATED = 'RECIPE_UPDATED';
@@ -30,7 +31,8 @@ const apiRequestMap = {
         settings: {
           method: 'get'
         },
-        actionOnSuccess: recipesReceived
+        actionOnSuccess: recipesReceived,
+        errorNotification: 'Error fetching recipes.'
       };
     } else {
       return null;
@@ -43,7 +45,8 @@ const apiRequestMap = {
       settings: {
         method: 'get'
       },
-      actionOnSuccess: singleRecipeReceived
+      actionOnSuccess: singleRecipeReceived,
+      errorNotification: 'Error fetching recipe.'
     };
   },
 
@@ -64,7 +67,9 @@ const apiRequestMap = {
         data: recipeInfo,
         method: 'post'
       },
-      actionOnSuccess: recipeAdded
+      actionOnSuccess: recipeAdded,
+      successNotification: 'Recipe added.',
+      errorNotification: 'Error adding recipe.'
     };
   },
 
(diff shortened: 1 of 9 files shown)


brittanystoroz:
Notification bar looks like this: 
validation-demo 

Fixed-positioned to the top of the page, figure we can use it for additional messages in the future. It'll disappear and remove the message automatically after ~10 seconds if you don't click the 'x'.  The field-level validation isn't in this PR but will come in a follow-up one

normandy/control/static/control/js/components/NotificationBar.jsx
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  render() {
+    const { notification, dispatch } = this.props;
+    return (
+      <div id="notification-bar" className={classNames({ 'active': this.state.isVisible })}>
+        { notification &&
+          <p className={classNames('notification', notification.messageType)}>
+            { notification.message }
+            <i className="fa fa-lg fa-times remove-message" onClick={::this.hideNotificationBar}></i>
```
mythmon: ::nice

normandy/control/static/control/js/components/NotificationBar.jsx
```diff
@@ -0,0 +1,61 @@
...
+      isVisible: false
+    }
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.notification) {
+      this.setState({
+        isVisible: true
+      });
+      this.setVisibilityTimer();
```
mythmon: This use of component local state and state changing in reaction to props like this makes me nervous, though it does seem like a nice API. How would you feel about managing the `isVisible` state in the Redux store? I suppose that would make the action creators more complicated. Would it make anything better/worse?

normandy/control/static/control/js/actions/ControlActions.js
```diff
@@ -98,11 +105,14 @@ function requestInProgress() {
   };
 }
 
-function requestComplete(status) {
-  return {
-    type: REQUEST_COMPLETE,
-    status: status
-  };
+function requestComplete(result) {
+  return (dispatch) => {
```
mythmon: Nit: do we care about `(dispatch) =>` vs `dispatch =>`?

mythmon:
This looks good to me. I have some concerns about the visibility state, but I'd be fine with it if doing it in Redux is painful.

I also wonder if we want to be able to support multiple notifications at once, so having an array of notifications that pop off after 10 seconds/being closed instead of a singleton. That might be premature though.

normandy/control/static/control/js/actions/ControlActions.js
brittanystoroz: I don't have a preference, as long as we're consistent. I feel like if we have to use parens when there are two or more params, we might as well use parens for all of them.

brittanystoroz: Also @potch just mentioned Airbnb released some ES6 style guides that can hook into eslint that we might want to consier.

brittanystoroz:
Yea I figured we'd want to change the component to handle an array rather than just a single message at a time, but for our current use case I think this is fine. It won't be a hard change to make but I'd rather wait until we need to do it.

I *might* be able to get rid of the `isVisible` state altogether. Will try and report back.

mythmon:
lgtm.

PR #160, opened 2016-06-14
Fix leftover nits

brittanystoroz:
Some lefover nits that I missed in my last pull request. Using positional arguments for classes that are definite with `classNames`. @casebenton r?


casebenton:
Looks good!
r+

PR #161, opened 2016-06-14
Add reactable for sorting and filtering recipe list

brittanystoroz:
Added a nifty package called [reactable](https://github.com/glittershark/reactable) that lets you sort and filter tables of data. 

All the columns are asc/desc sortable by clicking on the column headers
The search input filters based on the recipe `name` and `action_name` properties
Switch filter does an AND to limit to all/enabled/disabled recipes with the current search text. (We could also do this for `is_approved` if we want)

There is some not-so-pretty code around filtering the list. Reactable currently only supports filtering by a single string (there are PRs in to fix that in the near future). In order to filter by multiple columns with different conditions, I added `searchQuery` to the component state which combines both the search text & current enabled/disabled filter (i.e. "lorem ipsum dolor:enabled" will return only enabled recipes that contain "lorem ipsum dolor" in either `name` or `action_name`.

Fixes bug 1266490 r?

### DIFF
--- a/normandy/control/static/control/admin/sass/partials/_common.scss
+++ b/normandy/control/static/control/admin/sass/partials/_common.scss
@@ -131,6 +131,7 @@ input, select, textarea {
 table {
   border-collapse: collapse;
   font-family: $SourceSansPro;
+  table-layout: fixed;
   width: 100%;
 
   a:link, a:visited {
@@ -147,8 +148,37 @@ thead {
   th {
     outline: 0;
     padding: 5px 0px;
+    position: relative;
     &:hover { cursor: pointer; }
-    i { color: $secondaryColor; }
+
+    &.reactable-th-name { width: 30%; }
+
+    span::after {
+      color: $secondaryColor;
+      content: "\f0dc";
+      display: inline-block;
+      font: normal normal 400 16px/18px FontAwesome;
+      margin: 0 0 0 10px;
+      position: absolute;
+    }
+
+    &.reactable-header-sort-asc {
+      span::after {
+        color: $cta;
+        content: "\f0de";
+        font-size-adjust: 0.85;
+        margin-top: 6px;
+      }
+    }
+
+    &.reactable-header-sort-desc {
+      span::after {
+        color: $cta;
+        content: "\f0dd";
+        font-size-adjust: 0.85;
+        margin-top: -5px;
+      }
+    }
   }
 }
 
--- a/normandy/control/static/control/admin/sass/partials/_common.scss
+++ b/normandy/control/static/control/admin/sass/partials/_common.scss
@@ -63,9 +63,10 @@ input, select, textarea {
   }
 }
 
-.switch-label {
+.switch span {
   color: $darkBrown;
   cursor: pointer;
+  font-size: 12px;
   float: left;
   line-height: 26px;
   padding: 0px;
@@ -74,15 +75,12 @@ input, select, textarea {
   text-transform: none;
   width: 33%;
   z-index: 2;
-}
 
-.switch-label:active { font-weight: 600; }
-.switch-input { display: none; }
-
-.switch-input:checked + .switch-label {
-  font-weight: 600;
-  text-shadow: 0 1px rgba(255, 255, 255, 0.25);
-  transition: 0.15s ease-out;
+  &.active {
+    font-weight: 600;
+    height: 26px;
+    text-shadow: 0 1px rgba(255, 255, 255, 0.25);
+  }
 }
 
 .switch-selection {
@@ -91,9 +89,15 @@ input, select, textarea {
   box-shadow: inset 0 1px rgba(255, 255, 255, 0.5), 0 0 2px rgba(0, 0, 0, 0.2);
   display: block;
   height: 26px;
+  left: 5px;
   position: absolute;
+  top: 5px;
   transition: left 0.15s ease-out;
   width: 33%;
+  z-index: 1;
+
+  &.position-1 { left: 33%; }
+  &.position-2 { left: 65%; }
 }
 
 
@@ -139,6 +143,13 @@ thead {
   font-weight: 600;
   font-size: 13px;
   font-family: $OpenSans;
+
+  th {
+    outline: 0;
+    padding: 5px 0px;
+    &:hover { cursor: pointer; }
+    i { color: $secondaryColor; }
+  }
 }
 
 tr:nth-child(even) {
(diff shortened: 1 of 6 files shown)


brittanystoroz:
screen shot 2016-06-14 at 11 40 17 am

normandy/control/static/control/js/components/RecipeList.jsx
```diff
@@ -1,41 +1,38 @@
...
-class BooleanIcon extends React.Component {
-  render() {
-    switch(this.props.value) {
-      case true: return <i className="fa fa-lg fa-check green">&nbsp;</i>;
-      case false: return <i className="fa fa-lg fa-times red">&nbsp;</i>;
-    }
+const BooleanIcon = (props) => {
+  switch(props.value) {
+    case true: return <i className="fa fa-lg fa-check green">&nbsp;</i>;
+    case false: return <i className="fa fa-lg fa-times red">&nbsp;</i>;
```
mythmon: :+1:

brittanystoroz:
@casebenton this already had a brief once-over but I think you could dig into this a bit and make sure everything is working as expected. 

@Osmose do you think the description I provided above of how it works is ok? (search text input: filters `name` and `action` property values, filtering: toggles based on the `enabled` property) Let me know if you think we want to search or filter by anything else.

normandy/control/static/control/admin/sass/partials/_common.scss
```diff
@@ -134,6 +138,13 @@ thead {
   font-weight: 600;
   font-size: 13px;
   font-family: $OpenSans;
+
+  th {
+    outline: 0;
+    padding: 5px 0px;
+    &:hover { cursor: pointer; }
+    i { color: $secondaryColor; }
+  }
```
casebenton: Small detail, but maybe add something like: 
`&.reactable-header-sort-asc i {color: $green;}
&.reactable-header-sort-desc i {color: $red;}`
To give an indication that sorting is taking place, like this 
screen shot 2016-06-22 at 9 56 45 am

casebenton:
Functionality looks good! I couldn't find anything that didn't behave properly.

normandy/control/static/control/admin/sass/partials/_common.scss
Osmose: Showing one arrow or the other rather than switching color would be nice too.

normandy/control/static/control/js/components/RecipeList.jsx
```diff
@@ -44,27 +44,89 @@ class RecipeList extends React.Component {
...
+
+        <div className="fluid-8">
+          <Table id="recipe-list" sortable={true} hideFilterInput
+            filterable={[
+              { column: 'name', filterFunction: ::this.parseFilter },
+              { column: 'action', filterFunction: ::this.parseFilter }
+            ]}
+            filterBy={this.state.searchQuery}>
+            <Thead>
+              <Th column="name">Name <i className="fa fa-sort post">&nbsp;</i></Th>
```
Osmose: Capitalized tag names as an API is so weird to me.

normandy/control/static/control/js/components/RecipeList.jsx
```diff
@@ -1,41 +1,41 @@
...
-        <td><BooleanIcon value={recipe.is_approved} /></td>
-        <td>{ moment(recipe.last_updated).fromNow() }</td>
-      </tr>
-    )
-  }
+const SwitchFilter = (props) => {
+  const { options, selectedFilter, updateFilter } = props;
+  return (
+    <div className="switch">
+      <div className={classNames('switch-selection', `position-${options.indexOf(selectedFilter)}`)}>&nbsp;</div>
```
Osmose: No need to use the `classNames` function here.

normandy/control/static/control/js/components/RecipeList.jsx
```diff
@@ -1,41 +1,41 @@
...
-        <td>{recipe.name}</td>
-        <td>{recipe.action}</td>
-        <td><BooleanIcon value={recipe.enabled} /></td>
-        <td><BooleanIcon value={recipe.is_approved} /></td>
-        <td>{ moment(recipe.last_updated).fromNow() }</td>
-      </tr>
-    )
-  }
+const SwitchFilter = (props) => {
+  const { options, selectedFilter, updateFilter } = props;
```
Osmose: You can put this destructuring right in the function arguments instead of `props`.

normandy/control/static/control/js/components/RecipeList.jsx
```diff
@@ -44,27 +44,89 @@ class RecipeList extends React.Component {
...
-              <td>Approved</td>
-              <td>Last Updated</td>
-            </tr>
-          </thead>
-          <tbody>
-            {
-              this.props.recipes.map(recipe =>
-                <RecipeDataRow recipe={recipe} dispatch={this.props.dispatch} key={recipe.id} />
+      <div>
+        <div id="secondary-header" className="fluid-8">
```
Osmose: This whole bar feels like it should be another component.

Osmose:
> There is some not-so-pretty code around filtering the list. Reactable currently only supports filtering by a single string (there are PRs in to fix that in the near future). In order to filter by multiple columns with different conditions, I added searchQuery to the component state which combines both the search text & current enabled/disabled filter (i.e. "lorem ipsum dolor:enabled" will return only enabled recipes that contain "lorem ipsum dolor" in either name or action_name.

Could you instead filter based on the enabled state before passing the recipes to reactable? Then reactable would only be doing the text filtering.

Osmose:
Behavior-wise this looks fine, but I think the hoops we're jumping through to get reactable to filter the way we want aren't worth it. I'm hoping that I'm not wrong that we can do enabled filtering before passing the recipes to reactable to make this cleaner.

brittanystoroz:
@Osmose I moved the filtering out of reactable's control & fixed all the nits

@casebenton Added an up/down arrow for the currently sorted column. I stuck with just an 'active' blue color instead of red/green...I think red/green colors should be reserved more for indicating successes/errors.

normandy/control/static/control/js/components/RecipeList.jsx
```diff
@@ -1,70 +1,123 @@
...
+
+    if (filterStatus === 'All') {
+      this.setState({
+        filteredRecipes: null,
+        selectedFilter: filterStatus
+      });
+    } else {
+      let enabledState = (filterStatus === 'Enabled') ? true : false;
+      this.setState({
+        filteredRecipes: recipes.filter(recipe => recipe.enabled === enabledState),
```
Osmose: 👍

Osmose:
r+, looks nice, nice work!

PR #162, opened 2016-06-15
Remove peer approval checks and add enabled field to control interface.

Osmose:
Once this is merged we should create a branch for peer approval that reverts this commit.

@rehandalal r?

### DIFF
--- a/normandy/control/static/control/admin/sass/control.scss
+++ b/normandy/control/static/control/admin/sass/control.scss
@@ -114,7 +114,10 @@
 .crud-form {
   padding: 30px 45px;
 
-  input[type="checkbox"] { opacity: 0; }
+  input[type="checkbox"] {
+    width: auto;
+    margin: 0 5px;
+  }
   input[type="number"] { width: 20%; }
 
   .fluid-1 { width: 6.25%; }
--- a/normandy/control/static/control/admin/sass/control.scss
+++ b/normandy/control/static/control/admin/sass/control.scss
@@ -118,6 +118,7 @@
     width: auto;
     margin: 0 5px;
   }
+
   input[type="number"] { width: 20%; }
 
   .fluid-1 { width: 6.25%; }
--- a/normandy/control/static/control/js/components/form_fields/CheckboxField.jsx
+++ b/normandy/control/static/control/js/components/form_fields/CheckboxField.jsx
@@ -1,7 +1,6 @@
 import React from 'react';
 
-const CheckboxField = (props) => {
-  const {label, field, containerClass} = props;
+export default function CheckboxField({label, field, containerClass}) {
   return (
     <div className="row">
       <div className={containerClass}>
@@ -12,6 +11,4 @@ const CheckboxField = (props) => {
       </div>
     </div>
   );
-};
-
-export default CheckboxField;
+}
--- a/normandy/control/static/control/js/components/form_fields/CheckboxField.jsx
+++ b/normandy/control/static/control/js/components/form_fields/CheckboxField.jsx
@@ -1,0 +1,17 @@
+import React from 'react';
+
+const CheckboxField = (props) => {
+  const {label, field, containerClass} = props;
+  return (
+    <div className="row">
+      <div className={containerClass}>
+        <label>
+          <input type="checkbox" field={field} {...field} />
+          {label}
+        </label>
+      </div>
+    </div>
+  );
+};
+
+export default CheckboxField;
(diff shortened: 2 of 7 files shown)


normandy/control/static/control/admin/sass/control.scss
```diff
@@ -114,7 +114,10 @@
 .crud-form {
   padding: 30px 45px;
 
-  input[type="checkbox"] { opacity: 0; }
+  input[type="checkbox"] {
+    width: auto;
+    margin: 0 5px;
+  }
```
rehandalal: nit: blank line here

normandy/recipes/models.py
```diff
@@ -119,22 +118,10 @@ def __str__(self):
...
-                if field in dirty_fields:
-                    dirty_fields.pop(field)
-            if dirty_fields:
-                self.disable()
-
-            # Increment the revision ID
-            self.revision_id += 1
-
-        super(Recipe, self).save(*args, **kwargs)
+        # Increment the revision ID
```
rehandalal: probably want to make this:
```python
if self.is_dirty(check_relationship=True):
    self.revision_id += 1
```

normandy/recipes/tests/test_api.py
```diff
@@ -115,10 +115,13 @@ def test_it_serves_recipes(self, api_client):
...
-                                                  'action_name': action.name,
-                                                  'arguments': {},
-                                                  'filter_expression': 'whatever'})
+        res = api_client.post('/api/v1/recipe/', {
+            'name': 'Test Recipe',
+            'action_name': action.name,
+            'arguments': {},
+            'filter_expression': 'whatever',
+            'enabled': True
+        })
```
mythmon: :+1:

normandy/recipes/tests/test_api.py
```diff
@@ -208,12 +211,14 @@ def test_it_can_enable_recipes(self, api_client):
...
         assert recipe.enabled
 
+    @pytest.mark.xfail
     def test_cannot_enable_unapproved_recipes(self, api_client):
         recipe = RecipeFactory(approval=None, enabled=False)
 
         res = api_client.post('/api/v1/recipe/%s/enable/' % recipe.id)
         assert res.status_code == 400
 
+    @pytest.mark.xfail(strict=False)
```
mythmon: Why is this `strict=False` and the other is not?

normandy/recipes/tests/test_api.py
```diff
@@ -208,12 +211,14 @@ def test_it_can_enable_recipes(self, api_client):
         recipe = Recipe.objects.all()[0]
         assert recipe.enabled
 
+    @pytest.mark.xfail
```
mythmon: `xfail` takes an optional `reason` kwarg. I think we should use that for all xfails.

normandy/control/static/control/js/components/form_fields/CheckboxField.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+const CheckboxField = (props) => {
```
mythmon: I'm not terribly excited about the `const Foo = (args) => { }` style of function declaration. This is a style nit though, so feel free to ignore this.

normandy/recipes/tests/test_api.py
```diff
@@ -212,12 +215,14 @@ def test_it_can_enable_recipes(self, api_client):
...
         assert recipe.enabled
 
+    @pytest.mark.xfail(reason='disabled approvals temporarily')
     def test_cannot_enable_unapproved_recipes(self, api_client):
         recipe = RecipeFactory(approval=None, enabled=False)
 
         res = api_client.post('/api/v1/recipe/%s/enable/' % recipe.id)
         assert res.status_code == 400
 
+    @pytest.mark.skip(reason='disabled approvals temporarily')
```
mythmon: I don't think I'm happy with the idea of having unconditional skip tests. In sumo we had tests that were skipped in certain environments (like a developer without Elasticsearch). We made skips failures in CI. The only reason we don't do that in Normandy already is that I couldn't figure out how to do it.

I think we should either delete these tests, or leave them unmodified. They exist on the peer review branch. They don't provide any value here.

normandy/control/static/control/js/components/form_fields/CheckboxField.jsx
Osmose: I was trying to be consistent with other parts of the codebase, but I too like a function declaration over this style. I'll change it and we'll see what happens when linting lands.

PR #164, opened 2016-06-15
Switch to babel stage 0

mythmon:
We were already pulling in  `transform-function-bind`, which is literally half of `stage-0`. Lets just be simple about it.

@brittanystoroz r?


brittanystoroz:
r+

PR #165, opened 2016-06-15
Add jexl validation

casebenton:
fixes bug 1275658

### DIFF
--- a/normandy/control/static/control/js/components/RecipeForm.jsx
+++ b/normandy/control/static/control/js/components/RecipeForm.jsx
@@ -3,6 +3,8 @@ import { connect } from 'react-redux'
 import { Link } from 'react-router'
 import { destroy, reduxForm, getValues } from 'redux-form'
 
+import jexl from 'jexl'
+
 import apiFetch from '../utils/apiFetch.js';
 import ControlActions from '../actions/ControlActions.js'
 import composeRecipeContainer from './RecipeContainer.jsx'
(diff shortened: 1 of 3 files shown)


casebenton:
@Osmose 
Others already looked at this. 
r?

brittanystoroz:
@casebenton add this chunk of CSS [here](https://github.com/mozilla/normandy/blob/master/normandy/control/static/control/admin/sass/control.scss#L128) and just check that it looks ok. (Should right-align the error message, make it a little smaller and red.)

```
  .validation-error {
    color: $red;
    float: right;
    font-size: 11px;
    text-transform: lowercase;
  }
```

normandy/control/static/control/js/components/RecipeForm.jsx
```diff
@@ -3,6 +3,8 @@ import { connect } from 'react-redux'
 import { Link } from 'react-router'
 import { destroy, reduxForm, getValues } from 'redux-form'
 
+import jexl from 'jexl'
```
mythmon: Nit: I think the import structuring we talked about means that this shouldn't have a blank line above it. So `react-router`, `redux-form`, and then `jexl`, with no blanks.

normandy/control/static/control/js/components/RecipeForm.jsx
```diff
@@ -30,19 +32,32 @@ export class RecipeForm extends React.Component {
     });
   }
 
+  validateForm(formValues) {
+    let jexlExpression = formValues.filter_expression;
+    return jexl.eval(jexlExpression, {});
```
Osmose: Why return?

casebenton: I belive that's so that lines 45 through 59 can handle the response from the jexl.eval 
@brittanystoroz is that right?

Osmose: Whoops yeah I keep forgetting it's a promise and not a value, ignore me.

Osmose:
r+, nice work!

PR #166, opened 2016-06-15
Add stable sampling to jexl filters

mythmon:
Also refactor babel configuration a bit, which makes reasoning about it a lot easier.

The Python version of the code is at [`normandy/recipes/utils.py`](https://github.com/mozilla/normandy/blob/b99cc7134330aa584d51ecab4fa8615ab0e8d071/normandy/recipes/utils.py#L27). The tests are ported from the Python versions. The changed from `7fff...` to `8000...` (etc) because of rounding errors.

@rehandalal r?

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["react", "es2015", "stage-0"]
+}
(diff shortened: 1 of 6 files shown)


rehandalal:
r+

PR #167, opened 2016-06-16
Remove custom update/create methods from Recipe serializer.

Osmose:
Due to our custom update/create methods on the recipe serializer, we can't edit filter expressions via the control interface, as the custom methods don't alter the field at all. The whole reason we have them is to support using the `name` field on actions as the value on serialized Recipes, but this is provided by DRF already, so we don't need them.

This changes the API; the action_name field is now simply called action. Most of the changeset is renaming things from action_name to action.

@mythmon r?

### DIFF
--- a/normandy/control/static/control/js/components/ActionForm.jsx
+++ b/normandy/control/static/control/js/components/ActionForm.jsx
@@ -34,7 +34,7 @@ export default reduxForm({
         break;
     }
 
-    if (props.recipe && props.recipe.action_name === props.name) {
+    if (props.recipe && props.recipe.action === props.name) {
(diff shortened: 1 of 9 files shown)


mythmon:
This is definitely the right way to do this: we should have done it this way to begin with. Changing the APIs is lame, but fine.

In the future we could get all these benefits without changing the API by saying:

```python
action_name = serializers.SlugRelatedField(
    source='action', slug_field='name', queryset=Action.objects.all())
```

(ie: add a `source` parameter). We don't need to do this, just something to keep in mind if we have to do this on an API we've declared "done".

r+

PR #170, opened 2016-06-21
Remove console clutter

brittanystoroz:
Removing all the excessive logging that was happening in the console. @casebenton r?


casebenton:
looks good! r+

PR #171, opened 2016-06-21
Remove API Fetch in favor of fetch

brittanystoroz:
This removes our `apiFetch` util and replaces it with the standard `fetch` implementation. This refactoring allows us to have more control over the side effects that occur in response to API requests. It'll set us up nicely for implementing the rest of the form validation...(it automatically takes care of surfacing any validation errors returned in the response of the API request)

(It also fixes the bug where the list view doesn't appropriately get repopulated after deleting a recipe.)

@Osmose and/or @rehandalal  r?


normandy/control/static/control/js/actions/ControlActions.js
```diff
@@ -183,26 +148,33 @@ function setNotification(notification) {
...
+    return fetch(apiRequestConfig.url, {
+      ...API_REQUEST_SETTINGS,
+      ...apiRequestConfig.settings
+    })
+    .then(response => {
+      if (response.status >= 400) {
         dispatch(requestComplete({ status: 'error', notification: apiRequestConfig.errorNotification }));
-      });
-    }
+        return response.json().then(err => { throw err; });
```
Osmose: Does the error need to be thrown here? I'd think that the promise chain would surface the error to the code calling the API automatically.

mythmon: This is turning the success-case (`.then()`) into an error (`throw`).

normandy/control/static/control/js/actions/ControlActions.js
```diff
@@ -19,55 +17,41 @@ const BASE_API_URL = '/api/v1/recipe/';
...
       },
-      actionOnSuccess: singleRecipeReceived,
-      errorNotification: 'Error fetching recipe.'
+      errorNotification: 'Error fetching recipes.'
     };
   },
 
-  fetchSingleRevision(recipeInfo) {
+  fetchSingleRecipe(recipeInfo) {
+    let fetchUrl = (recipeInfo.recipeId ? `${BASE_API_URL}${recipeInfo.recipeId}/` : `/api/v1/recipe_version/${recipeInfo.revisionId}/`)
```
Osmose: it's weird to only use `BASE_API_URL` in one spot and a full URL in another. Also `BASE_API_URL` isn't actually the base API url, it's just the URL to the recipe API itself.

Also why is `fetchSingleRecipe` handling fetching two different types of objects?

normandy/control/static/control/js/components/RecipeForm.jsx
```diff
@@ -49,14 +53,18 @@ export class RecipeForm extends React.Component {
...
-      }
-    })
+        submitAction = 'addRecipe';
+        successAction = (response) => {
+          dispatch(recipeAdded(response));
+          dispatch(push(`/control/recipe/${response.id}/`));
+        };
+      };
+
+      return dispatch(makeApiRequest(submitAction, requestBody)).then(response => successAction(response));
```
Osmose: I actually think setting up the variables makes it harder to see what this method is doing vs the old method of calling dispatch in two different spots. It's a bit more typing but less state for the programmer to store in their head when reading this.

normandy/control/tests/actions/controlActions.js
```diff
@@ -70,79 +74,84 @@ describe('controlApp Actions', () => {
     const expectedAction = { type: actionTypes.RECIPES_RECEIVED, recipes: fixtureRecipes }
     spyOn(window, 'fetch').and.returnValue(successPromise(fixtureRecipes));
 
-    return store.dispatch(controlActions.makeApiRequest('fetchAllRecipes')).then(() => {
+    return store.dispatch(actionTypes.makeApiRequest('fetchAllRecipes'))
+    .then((response) => {
+      store.dispatch(actionTypes.recipesReceived(response));
```
Osmose: The test description says that we're testing that this action is dispatched on a successful fetch, but you've changed the API so that this action isn't automatically dispatched. Manually dispatching it in the test and then asserting below that is was dispatched doesn't test anything.

At this point this test is about testing that the `makeApiRequest` action calls `window.fetch` with the proper arguments. The description should be updated and the other assertions removed.

normandy/control/tests/actions/controlActions.js
```diff
@@ -70,79 +74,84 @@ describe('controlApp Actions', () => {
...
+
       expect(window.fetch).toHaveBeenCalled();
```
Osmose: `toHaveBeenCalledWith` implies `toHaveBeenCalled`, no need to call both.

normandy/control/tests/actions/controlActions.js
```diff
@@ -70,79 +74,84 @@ describe('controlApp Actions', () => {
...
-
-  it('doesnt create any actions if fetching recipes is not needed', () => {
-    const store = mockStore({ controlApp: { ...initialState, recipeListNeedsFetch: false } }, []);
-
-    store.dispatch(controlActions.makeApiRequest('fetchAllRecipes'));
-    expect(store.getActions()).toEqual([]);
   });
 
-  it('creates SINGLE_RECIPE_RECEIVED when fetching a single RECIPE is successful', () => {
+  it('creates SINGLE_RECIPE_RECEIVED when fetching by RECIPE id is successful', () => {
```
Osmose: Same goes for this test as the one above.

normandy/control/static/control/js/actions/ControlActions.js
Osmose: Duh, silly me.

Osmose:
I like the direction of this change, getting rid of our custom fetch thing and especially the explicit dispatching post-fetch instead of the implied post-fetch actions is a good direction to move in. 👍

brittanystoroz:
Feedback fixed

normandy/control/static/control/js/components/DeleteRecipe.jsx
```diff
@@ -14,8 +14,12 @@ class DeleteRecipe extends React.Component {
             <div className="form-action-buttons">
               <div className="fluid-2 float-right">
                 <input type="submit" value="Confirm" class="delete" onClick={(e) => {
-                  dispatch(ControlActions.makeApiRequest('deleteRecipe', { recipeId }));
-                  dispatch(push(`/control/`));
+                  e.preventDefault();
```
Osmose: This is sufficiently complex that it should probably be a method on the component rather than embedding it here.

Osmose:
r+, nice work!

PR #172, opened 2016-06-22
Add cache headers to "hot-path" APIs

mythmon:
Fixes bug 1277994.

I changed normandy-compose's Nginx proxy to have caching behavior similar to what our production servers will have. I tested this change with that caching behavior, and it resulted in cache hits that lasted 30 seconds.

r? @Osmose

### DIFF
--- a/normandy/base/api/mixins.py
+++ b/normandy/base/api/mixins.py
@@ -2,7 +2,7 @@ from django.conf import settings
 from django.views.decorators.cache import cache_control
 
 
-class CachingMixin:
+class CachingViewsetMixin(object):
     """Modify a ModelViewSet to add caching to read methods"""
 
     @cache_control(public=True, max_age=settings.API_CACHE_TIME)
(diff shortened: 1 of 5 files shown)


normandy/base/api/mixins.py
```diff
@@ -0,0 +1,14 @@
+from django.conf import settings
+from django.views.decorators.cache import cache_control
+
+
+class CachingMixin:
```
Osmose: I can't remember what we decided RE: subclassing `object` or not. A lot of our test classes do, for example, as do some middlewares and other classes around the codebase I found do.

Osmose: Also maybe call this `CachingViewsetMixin` to make it clear what it's a mixin for?

Osmose:
r+wc, nice work!

normandy/base/api/mixins.py
mythmon: I think we decided not to really make a decision. I'd be happy to find a
lint either way for it.

Osmose: Let's go with keeping `(object)` then. Removes confusion for Python 2 devs used to inheriting from it coming in to the codebase.

PR #173, opened 2016-06-23
Add Karma tests to Circle CI

mythmon:
Testing to see if this works.


mythmon:
Alright, I'm happy with this now. To summarize:

* Added a script that runs Karma in a very particular way, allowing external browsers to connect, and then exiting after the first test run has completed.
* Tweaked the Karma config to work better in Circle
* Changed the tests to use fetch-mock instead of adhoc Jasmine spying. This ended up being unneeded, but I think it is a good change.
* Add sha.js to our requirements. I forgot this when I added stable sorting in #166.

Additionally, I changed our Circle CI config to use Circle's Ubuntu 14.04 base image, instead of the default 12.04. The only change I observed here was that we got Firefox 46 instead of Firefox 38. The main difference is that 46 has `window.fetch`, and 38 does not.

@brittanystoroz r?

mythmon:
NB: This collides a bit with #171, mainly in the tests. We'll probably have to re-work on of the PRs after the other lands.

brittanystoroz:
@mythmon kewl, looks good to me 👍 ⚡ fetchMock

r+ after resolving conflicts

mythmon:
I merged master into this branch to fix the merge conflicts. This way, there is a record of how I resolved them, which is sort of neat. Will merge once CI passes.

PR #174, opened 2016-06-23
Do not save recipe revisions with no changes.

Osmose:
Revisions are only saved and the revision number is incremented only
when a recipe actually changes. Includes an update to django-reversion
2.0.4.

@mythmon r?

### DIFF
--- a/normandy/recipes/api/serializers.py
+++ b/normandy/recipes/api/serializers.py
@@ -110,7 +110,7 @@ class ClientSerializer(serializers.Serializer):
 
 class RecipeVersionSerializer(serializers.ModelSerializer):
     date_created = serializers.DateTimeField(source='revision.date_created', read_only=True)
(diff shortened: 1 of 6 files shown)


normandy/recipes/api/serializers.py
```diff
@@ -110,7 +110,7 @@ class ClientSerializer(serializers.Serializer):
 
 class RecipeVersionSerializer(serializers.ModelSerializer):
     date_created = serializers.DateTimeField(source='revision.date_created', read_only=True)
-    recipe = RecipeSerializer(source='object_version.object', read_only=True)
+    recipe = RecipeSerializer(source='_object_version.object', read_only=True)
```
mythmon: Ooh boy, this is fun.

mythmon:
I reviewed the django-reversion changelog, and I think this is fine. Not terribly happy with using a private attribute, but oh well.

r+

PR #175, opened 2016-06-24
Clean up route hierarchy and add better breadcrumbs.

Osmose:
Introspecting the react-router hierarchy is not easy, so I gave up and turned to a library called react-breadcrumbs to handle it.

@brittanystoroz r?


mythmon:
Stealing this review, to get the queue moving.

LGTM r+

PR #176, opened 2016-06-24
Add cache headers to self-repair page

mythmon:
This is very similar to #172. We might be able to more aggressively cache this page (read: longer cache times), but that would probably make our deploys go funny. Even 30 seconds of caching (the default value for `API_CACHE_TIME`) should drastically help our scalibility.

@Osmose r?


Osmose:
r+

PR #177, opened 2016-06-24
Add basic mid-air collision detection to update APIs

mythmon:
@Osmose r?

### DIFF
--- a/normandy/base/api/mixins.py
+++ b/normandy/base/api/mixins.py
@@ -1,6 +1,9 @@
 from django.conf import settings
 from django.views.decorators.cache import cache_control
 
+from normandy.base.api.exceptions import MidairCollisionError
+from normandy.base.api.serializers import LastUpdatedSerializer
+
 
 class CachingViewsetMixin(object):
     """Modify a ModelViewSet to add caching to read methods"""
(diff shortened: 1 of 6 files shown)


mythmon:
Note: This is a PR against the `midair` branch. Branched development, yay!

normandy/base/api/mixins.py
```diff
@@ -12,3 +15,32 @@ def retrieve(self, request, *args, **kwargs):
...
+
+
+class MidairCollisionViewsetMixin(object):
+    """Modify a ModelViewSet to add midair collision to update methods"""
+
+    def update(self, request, *args, **kwargs):
+        if _midair_collision_enabled:
+            # Using a serializer gets some easy validation
+            serializer = LastUpdatedSerializer()
+            request_ts = serializer.to_internal_value(request.data)['last_updated']
```
Osmose: Instead of `to_internal_value`, could you pass `request.data` as the `data` kwarg to the serializer to deserialize it? http://www.django-rest-framework.org/api-guide/serializers/#deserializing-objects

normandy/base/api/mixins.py
```diff
@@ -12,3 +15,32 @@ def retrieve(self, request, *args, **kwargs):
     @cache_control(public=True, max_age=settings.API_CACHE_TIME)
     def list(self, request, *args, **kwargs):
         return super().list(request, *args, **kwargs)
+
+
+_midair_collision_enabled = True
+
+
+def _enable_midair_collision(enabled=True):
```
Osmose: `_toggle_midair_collision` would be a better name for this.

normandy/recipes/tests/test_api.py
```diff
@@ -275,6 +278,25 @@ def test_detail_view_includes_cache_headers(self, api_client):
         assert 'max-age=' in res['Cache-Control']
         assert 'public' in res['Cache-Control']
 
+    def test_midair_collisions(self, api_client):
+        recipe = RecipeFactory(name='first')
+        original_last_updated = recipe.last_updated
+
+        url = '/api/v1/recipe/{id}/'.format(id=recipe.id)
```
Osmose: Why aren't we reversing these URLs again?

Osmose:
The global feels a bit weird but I can't think of anything better except maybe for a setting, but this isn't really something we want to be configurable, either. Oh well. Nice work!

normandy/base/api/mixins.py
mythmon: Yes, though it ends up being slightly longer:

```
serializer = LastUpdatedSerializer(data=request.data)
request_ts = serializer.validated_data['last_updated']
```

The call to `is_valid` is required, without it, accessing `validated_data` throws an error (I wish it would just run the validation, instead of raising an error). This is probably more readable though, what do you think?

normandy/recipes/tests/test_api.py
mythmon: Originally, we wanted to assert not only that the API worked as expected, but that the URLs were correct, since that is part of our public interface. I'm not entirely convinced we need to use that style everywhere though. Maybe we could have QA tests that validate the URLs, or just have a few tests that have hardcoded URLs, with comments as to why.

normandy/base/api/mixins.py
Osmose: Yeah, I think that's better.

normandy/recipes/tests/test_api.py
Osmose: Maybe. Let's be consistent with the surrounding code for this PR though.

mythmon:
Feedback'd ^

Osmose:
r+, nice work!

PR #178, opened 2016-06-25
Add 404 handler and refactor root component.

Osmose:
Phew. Okay. So, this took way longer than I anticipated.

The initial change was easy; add a NoMatch component, display it when other routes don't match. Sure, whatever. The problem was that I wanted to test my route matching.

I eventually got it working, with the following changes:
- Moved the `babel-runtime` import from `self_repair_runner.js` to `self_repair.js`. The former is imported by the tests, which already have a runtime injected into them. The latter is an entry point that isn't in the tests, and without this change, the tests fail due to multiple runtimes being loaded on the same page.
- Removed the `initialState` argument from `controlStore`. This isn't strictly necessary but it was convenient in the middle of my refactoring and I don't see a whole lot of value in passing the initial state in; we're never going to pass another value besides what we always do.
- Moved the root component into a standalone component and moved creation of the store, history, and root component instance into a helper method. This makes setting up tests a lot easier since we can reuse `createApp` there as well.

@brittanystoroz r? Although I feel like we might be flooding you with review requests. Maybe @mythmon instead?

### DIFF
--- a/normandy/control/static/control/admin/sass/control.scss
+++ b/normandy/control/static/control/admin/sass/control.scss
@@ -115,6 +115,20 @@
 }
 
 
+/* 404 NoMatch Page */
+.no-match {
+  padding: 30px 45px;
+
+  h2 {
+    margin-bottom: 15px;
+  }
+
+  p {
+    margin-bottom: 10px;
+  }
+}
+
+
 /* Add & Edit Forms */
 .crud-form {
   padding: 30px 45px;
--- a/normandy/selfrepair/static/js/self_repair.js
+++ b/normandy/selfrepair/static/js/self_repair.js
@@ -1,3 +1,4 @@
+import 'babel-polyfill'
 import Normandy from './normandy_driver.js';
 import {fetchRecipes, filterContext, doesRecipeMatch, runRecipe} from './self_repair_runner.js';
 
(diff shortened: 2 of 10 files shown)


normandy/control/static/control/admin/sass/control.scss
```diff
@@ -111,6 +111,20 @@
...
+  padding: 30px 45px;
+
+  h2 {
+    margin-bottom: 15px;
+  }
+
+  p {
+    margin-bottom: 10px;
+  }
+}
```
mythmon: Yes, this is CSS.

normandy/selfrepair/static/js/self_repair.js
```diff
@@ -1,3 +1,4 @@
+import 'babel-polyfill'
```
mythmon: Moving this here just moves the problem elsewhere, and while we aren't likely to need to import this from test code, we might. I think there are better approaches, like getting babel to inject this automatically, or including it as a webpack entry point (I'm not sure if/how either of those work at the moment though)

mythmon:
This refactoring seems good. I like the simplicity of `createApp`. I think we should try and make the babel polyfill better before merging this though.

Osmose:
Local testing with the webpack method worked fine for me, so I've updated this PR with it.

mythmon:
Cool. r+, but you have merge conflicts to fix.

PR #179, opened 2016-06-26
Add status indicator for the current revision.

Osmose:
Includes a mild refactoring of the history view. Reuses the preview status indicator.

@brittanystoroz r?

### DIFF
--- a/normandy/control/static/control/js/components/RecipeHistory.jsx
+++ b/normandy/control/static/control/js/components/RecipeHistory.jsx
@@ -14,11 +14,11 @@ class RecipeHistory extends React.Component {
 
   getHistory(recipeId) {
     apiFetch(`/api/v1/recipe/${recipeId}/history/`)
-      .then(response => {
-        this.setState({
-          revisionLog: response
-        })
-      });
+    .then(response => {
+      this.setState({
+        revisionLog: response
+      })
+    });
   }
 
   componentDidMount() {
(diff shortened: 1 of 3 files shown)


normandy/control/static/control/js/components/RecipeHistory.jsx
```diff
@@ -30,32 +30,64 @@ class RecipeHistory extends React.Component {
...
         </ul>
       </div>
     )
   }
 }
 
+class HistoryItem extends React.Component {
+  render() {
+    const { revision, recipe } = this.props;
+    const isCurrent = revision.recipe.revision_id === recipe.revision_id;
```
rehandalal: I am getting:
```
TypeError: revision.recipe is undefined
```
when i go to the history page and it is coming from this line.

mythmon: Adding proptypes to the component will likely help track down the issue, and we should probably start doing that on new components.

Osmose: I added some PropTypes to this but it didn't really track down the issue. But it was good to add. I have some open questions about proptypes I'll be asking later.

Buuuuuuuuut I suspect I know the problem. @rehandalal , have you updated your Python requirements lately? If you haven't, you probably don't have the latest reversion, which changed the property that the version instance is stored in for revisions from `revision.object_version` to `revision._object_version`.

I'm willing to bet that the serializer silently fails when it can't find the source for a property, and thus the recipe field from the `RecipeVersionSerializer` is empty.

Which sucks. It shouldn't be silently failing to serialize that. If that's the case it's a bug.

rehandalal: yep it was failing silently. working now!

normandy/control/static/control/js/components/RecipeHistory.jsx
```diff
@@ -30,32 +30,76 @@ class RecipeHistory extends React.Component {
...
+              />
+            )}
         </ul>
       </div>
     )
   }
 }
 
+class HistoryItem extends React.Component {
+  static propTypes = {
```
rehandalal: 👍🏼

rehandalal:
looks good! r+

PR #183, opened 2016-06-29
Change `routeParams` to `params` when accessing URL data

brittanystoroz:
Refactoring the routes in #175 broke the ability to fetch recipe data from the URL if you're doing a hard refresh. For some reason we lose `this.props.routeParams` with the new way routes are nested. Wrapping the `recipeContainer` component in `withRouter` fixes it, though I'm not convinced it's the best solution. Just wanted to put up a quick fix for now.

### DIFF
--- a/normandy/control/static/control/js/components/RecipeContainer.jsx
+++ b/normandy/control/static/control/js/components/RecipeContainer.jsx
@@ -43,7 +43,7 @@ export default function composeRecipeContainer(Component) {
     }
 
     return {
-      recipeId: state.controlApp.selectedRecipe || parseInt(props.routeParams.id) || null,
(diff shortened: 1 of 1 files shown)


Osmose:
The switch to `params` instead of `routeParams` actually fixes this without `withRouter`. [`routeParams`](https://github.com/reactjs/react-router/blob/master/docs/API.md#routeparams) is a subset of `params` that only has parameters for the component's route. Since we nest routes now, the `id` parameter is lost in `routeParams` but remains in `params`.

Remove the `withRouter` stuff and I think this is the right fix. Nice work!

brittanystoroz:
@Osmose updated

Osmose:
r+, hooray! Thanks for catching this. :D

PR #184, opened 2016-06-30
Pass engagementButtonLabel to showHeartbeat properly.

Osmose:
Also includes some refactoring to make testing the Normandy driver possible / easier.

@mythmon r?

### DIFF
--- a/normandy/selfrepair/static/js/self_repair.js
+++ b/normandy/selfrepair/static/js/self_repair.js
@@ -1,8 +1,7 @@
-import Mozilla from './uitour.js';
 import NormandyDriver from './normandy_driver.js';
 import { fetchRecipes, filterContext, doesRecipeMatch, runRecipe } from './self_repair_runner.js';
 
-const driver = new NormandyDriver(Mozilla.UITour);
+const driver = new NormandyDriver();
 driver.registerCallbacks();
 
 // Actually fetch and run the recipes.
--- a/normandy/selfrepair/static/js/self_repair.js
+++ b/normandy/selfrepair/static/js/self_repair.js
@@ -1,16 +1,20 @@
-import Normandy from './normandy_driver.js';
+import Mozilla from './uitour.js';
+import NormandyDriver from './normandy_driver.js';
 import { fetchRecipes, filterContext, doesRecipeMatch, runRecipe } from './self_repair_runner.js';
+
+const driver = new NormandyDriver(Mozilla.UITour);
+driver.registerCallbacks();
 
 // Actually fetch and run the recipes.
 fetchRecipes().then(recipes => {
-  filterContext().then(context => {
+  filterContext(driver).then(context => {
     // Update Normandy driver with user's country.
-    Normandy._location.countryCode = context.normandy.country;
+    driver._location.countryCode = context.normandy.country;
 
     for (const recipe of recipes) {
       doesRecipeMatch(recipe, context).then(([, match]) => {
         if (match) {
-          runRecipe(recipe).catch(err => {
+          runRecipe(recipe, driver).catch(err => {
             console.error(err);
           });
         }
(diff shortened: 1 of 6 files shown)


normandy/control/static/control/js/components/RecipePreview.jsx
```diff
@@ -1,7 +1,8 @@
-import React from 'react'
-import classNames from 'classnames'
-import composeRecipeContainer from './RecipeContainer.jsx'
-import {runRecipe} from '../../../../../selfrepair/static/js/self_repair_runner.js';
+import React from 'react';
+import classNames from 'classnames';
+import composeRecipeContainer from './RecipeContainer.jsx';
+import { runRecipe } from '../../../../../selfrepair/static/js/self_repair_runner.js';
+import NormandyDriver from '../../../../../selfrepair/static/js/normandy_driver.js';
```
mythmon: 😞 We should find a way to avoid all these dots in the path. They aren't good.

mythmon:
"some refactoring" is a bit of an understatement. Would it be possible to split the refactoring changes into a separate commit?

Osmose:
> "some refactoring" is a bit of an understatement. Would it be possible to split the refactoring changes into a separate commit?

I could. If you view the diff with whitespace changes disabled it's actually much more manageable (I'm a monster and switched the file to 2-spaces during this commit to match the rest of the project). But if that's still too much I totally can.

normandy/control/static/control/js/components/RecipePreview.jsx
Osmose: If I could make all imports relative to the static folders I would. In fact, I think I'll look into that at some point.

brittanystoroz: http://moduscreate.com/es6-es2015-import-no-relative-path-webpack/

mythmon: Nice. That looks like a great idea. We'll just have to make sure none of the places we want to import from collide with any package names.

mythmon:
Even with `?w=1`, it is hard to tease out what was a change just to refactor and what was a change in behavior. Maybe I'm just being lazy though. Plus, you can't comment in `?w=1` mode

normandy/selfrepair/tests/test_normandy_driver.js
```diff
@@ -0,0 +1,38 @@
+import NormandyDriver from '../static/js/normandy_driver.js';
+
+describe('Normandy Driver', () => {
+  describe('showHeartbeat', () => {
+    it('should pass all the required arguments to the UITour helper', async function() {
```
mythmon: Lint: prefer arrow functions. Eslint has some issues with `async () => {}`, but it is valid. If this is problematic for you, leave it and I'll fix it in my linter branch (which can deal with `async () => {}`).

normandy/selfrepair/static/js/self_repair.js
```diff
@@ -1,16 +1,20 @@
-import Normandy from './normandy_driver.js';
+import Mozilla from './uitour.js';
+import NormandyDriver from './normandy_driver.js';
 import {fetchRecipes, filterContext, doesRecipeMatch, runRecipe} from './self_repair_runner.js';
 
+let driver = new NormandyDriver(Mozilla.UITour);
```
mythmon: This is just the default value, right?

mythmon:
LGTM, besides a couple nits and much easier to review in those chunks. Thanks for breaking it up.

r+

normandy/selfrepair/static/js/self_repair.js
Osmose: Yeah but I wanted to be explicit about it. I dunno.

Osmose:
@mythmon Had to rebase this after the linting changes and it was sufficiently complex that another quick glance over this would be appreciated. :D

mythmon:
Nice!

PR #185, opened 2016-06-30
*raises white flag*

brittanystoroz:
Here is a small sample of the thousand tests I wrote and scrapped. I don't know about any of these. The biggest problem I've run into with the integration tests is not being able to spy on things. You'll notice a nice ugly hack I put in to spy on a `dispatch` property for the component. I tried spying on that `dispatch` method from every object and component I could think of and would always be met with 'expected spy to be called with {} but it was never called'. 

There are a lot of places where I can/should just check the state of the store instead of spying on dispatch (the redux-form actions, mainly), but when we're conditionally dispatching our own actions (i.e. in the submit method we want to dispatch either `updateRecipe` or `addRecipe` depending on current props & state.) it would be nice to check, especially when there are promises involved, as state isn't going to be updated immediately. I also wanted to spy on `window.fetch` so I could fake success/failure responses after submitting but again, it's never getting called.

I think in general everything is probably just a little too nesty. Even doing unit tests where I only import the `RecipeForm` class on its own isn't really possible without still doing all of the `Provider` and `store` setup to keep the action form components happy.  Leaving this branch here so someone else can take a look, my brain is tired.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -13,8 +13,8 @@ module.exports = function(config) {
             'node_modules/babel-polyfill/dist/polyfill.js',
             'node_modules/jasmine-promises/dist/jasmine-promises.js',
             'normandy/control/tests/index.js',
-            'normandy/recipes/tests/actions/index.js',
-            'normandy/selfrepair/tests/index.js',
+            // 'normandy/recipes/tests/actions/index.js',
+            // 'normandy/selfrepair/tests/index.js',
         ],
 
         // preprocess matching files before serving them to the browser
--- a/normandy/control/static/control/js/components/RecipeForm.jsx
+++ b/normandy/control/static/control/js/components/RecipeForm.jsx
@@ -23,7 +23,7 @@ export class RecipeForm extends React.Component {
 
   changeAction(event) {
     const { dispatch, fields } = this.props;
-    let selectedActionName = event.currentTarget.value;
+    let selectedActionName = event.target.value;
 
     dispatch(destroy('action'));
     fields.action.onChange(event);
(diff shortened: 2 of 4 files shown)


mythmon:
That seems really tough.

These tests seem to be working at several levels all at once. On the one hand, you're dealing with the very black box idea of monitoring and manipulating the DOM, and on the other hand you're dealing with the white box idea of introspecting the actions and the store. I noticed you called this an integration test, so it does make sense to test the DOM, but I'm not sure if all the detailed testing is worth this trouble.

Do you think it would be worthwhile to change the scope of these tests to only be unit tests, and test each part in isolation? Test that each action creator creates the expected action(s), that reducers modify the state in the expected way, that a component--given the right props--renders the expected code and calls the right handlers?

Then we could use something like Marionette or Selenium to do full, end-to-end, black box integration tests that assert things like "when this url is loaded, and that button is hit, it creates a recipe in the DB" (or something).

This would probably be more work. I know I don't like writing Selenium tests. But I think it will also give us more confidence in the system. Plus, e2e tests can be used to verify other environments, like stage or new deploys.

mythmon:
On a more specific level, I think that doing something in a `beforeAll` function, and then trying to test that it happened with spies in an actual `it('should ...', () => {})` test seems a bit odd. I could imagine Jasmine being designed such that spies actually get reset before tests run. Did you ever try putting all of the code in one test, instead of splitting it between `beforeAll` and `it` functions?

brittanystoroz:
I will redo these with enzyme and see how that helps.

brittanystoroz:
Closing this as it was a bit easier to start over on a new branch after all the time that had passed between working on this. Will open a new non-WIP PR with 2nd attempt.

PR #187, opened 2016-07-01
Multiple notification support

Osmose:
This got bigger than I expected so I decided to land just the multiple notification support separately from adding new notifications to the app.

The animations leverage `react-addons-css-transition-group` now. I couldn't find a good way to mix React and the keyframe-based animations. I also had to use an explicit height for notifications that assumes one line because you can't transition a non-explicit height property.

The tests are a little different-looking than the tests around them, such as being grouped in `describe` blocks a little deeper, and using an async function instead of Promises directly. I think these will be a cleaner style for our tests, but I'm happy to make them match their surrounding code if ya'll don't agree.

@brittanystoroz r?


Osmose:
@mythmon has convinced me of a good way to test the React component, I'll be adding an extra commit shortly to this.

Osmose:
omg enzyme is amazing

Also this is ready.

PR #189, opened 2016-07-06
Add notifications for client-side validation errors & deleting recipes

brittanystoroz:
Fixes bug 1283937 r? @Osmose

### DIFF
--- a/normandy/control/static/control/js/actions/ControlActions.js
+++ b/normandy/control/static/control/js/actions/ControlActions.js
@@ -95,6 +95,8 @@ const apiRequestMap = {
       settings: {
         method: 'DELETE',
       },
+      successNotification: 'Recipe deleted.',
+      errorNotification: 'Error deleting recipe.',
     };
   },
 };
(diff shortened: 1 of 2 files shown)


normandy/control/static/control/js/components/RecipeForm.jsx
```diff
@@ -47,6 +47,10 @@ export class RecipeForm extends React.Component {
 
     return this.validateForm(combinedFormValues)
     .catch(error => {
+      dispatch(showNotification({
```
Osmose: How hard would testing this be? Could we use a simulated click with enzyme to trigger the submit?

Osmose:
Code-wise this looks fine, and testing locally it works fine. I'm curious about that test though. If it's difficult, then we should just merge this and lump it in with the rest of the functionality that will get tests after we deploy.

r+, nice work!

normandy/control/static/control/js/components/RecipeForm.jsx
brittanystoroz: I kindaaaaa wanna wait. Mocking out the forms is a little more involved and we already have tests for the notification actions, reducers and component so all we'd really be testing here is that it's called.

PR #180, opened 2016-06-27
Recipe signing

mythmon:
* Adds signatures to Recipes
* Adds utilities to get signatures for byte strings
* Add management command to get signatures for Recipes
* Tweak test settings
* Change default DRF JSON renderer to output canonical JSON, so it can be hashed/signed

@Osmose r?

PR #190, opened 2016-07-07
Recipe signing

mythmon:
Retargeting #180 at a branch.

@Osmose r?


normandy/recipes/api/serializers.py
```diff
@@ -119,3 +119,26 @@ class Meta:
...
+    public_key = serializers.ReadOnlyField()
+
+    class Meta:
+        model = Signature
+        fields = ['timestamp', 'signature', 'x5u', 'public_key']
+
+
+class SignedRecipeSerializer(serializers.ModelSerializer):
+    signature = SignatureSerializer()
+    recipe = serializers.SerializerMethodField()
```
Osmose: The docs suggest you can just use `RecipeSerializer` as the field and get the same effect: http://www.django-rest-framework.org/api-guide/relations/#nested-relationships

normandy/recipes/models.py
```diff
@@ -121,16 +150,40 @@ def __repr__(self):
     def __str__(self):
         return self.name
 
-    def save(self, *args, **kwargs):
+    def save(self, *args, skip_last_updated=False, **kwargs):
         # Increment the revision ID and save the reversion if we've
```
Osmose: We don't really need this comment anymore with the individual comments below.

normandy/recipes/models.py
```diff
@@ -121,16 +150,40 @@ def __repr__(self):
...
         # changed
         if self.is_dirty(check_relationship=True):
+            dirty_fields = self.get_dirty_fields(check_relationship=True)
+
+            # Any change will invalid the signature
```
Osmose: invalidate

normandy/recipes/models.py
```diff
@@ -121,16 +150,40 @@ def __repr__(self):
...
+            dirty_field_names = list(dirty_fields.keys())
+            if dirty_field_names != ['signature'] and self.signature is not None:
+                self.signature.delete()
```
Osmose: I'm worried about deleting the signature before the changes are saved. Would a transaction help? Or maybe just removing the signature from the recipe?

normandy/recipes/models.py
Osmose: Also, will any of the changes after this line invalidate the signature? Should this be done last?

normandy/recipes/utils.py
```diff
@@ -46,3 +54,68 @@ def deterministic_sample(rate, inputs):
...
+    If Autograph signing is not configured using `settings.AUTOGRAPH`,
+    raises `ImproperlyConfigured`. If the Autograph server can't be reached
+    or returns an HTTP error, an error will be thrown by `requests`.
+    """
+
+    def __init__(self):
+        self.check_config()
+        self._session = None
+
+    @property
```
Osmose: Django's `cached_property` would remove some boilerplate here.

normandy/recipes/utils.py
```diff
@@ -46,3 +54,68 @@ def deterministic_sample(rate, inputs):
...
+                raise ImproperlyConfigured(msg)
+
+    def sign_data(self, content_list):
+        """
+        Fetches Signatures objects from Autograph for each item in `content_list`.
+
+        The items in `content_list` must be bytes objects.
+        """
+        ts = timezone.now()
+        url = '{}sign/data'.format(settings.AUTOGRAPH_URL)
```
Osmose: Maybe use `urljoin` to be durable against missing slashes in the setting?

normandy/recipes/utils.py
```diff
@@ -46,3 +54,68 @@ def deterministic_sample(rate, inputs):
...
+            })
+
+        res = self.session.post(url, json=signing_request)
+        res.raise_for_status()
+        signing_responses = res.json()
+
+        from normandy.recipes.models import Signature  # avoid circular import
+        signatures = []
+        for res in signing_responses:
+            sig = Signature(
```
Osmose: I saw that you originally had this returning the signed data and created signatures elsewhere, and then changed it to return saved signature objects. Having it return dicts or something would lower the coupling with the models, avoid the circular import, etc.

Osmose:
Just to check my understanding: Signatures are meant to be deleted when they're invalidated, right?

Also, this PR doesn't include creating signatures outside of the management command. That's waiting for peer signing, right? And will block this branch from getting merged?

normandy/recipes/tests/test_factories.py
```diff
@@ -0,0 +1,27 @@
+from datetime import datetime
```
Osmose: Testing the factories?!

normandy/recipes/tests/test_serializers.py
```diff
@@ -36,3 +37,55 @@ def test_it_uses_cdn_url(self, rf, settings):
...
+        serializer = SignedRecipeSerializer(instance=recipe, context={'request': rf.get('/')})
+
+        assert serializer.data == {
+            'signature': {
+                'signature': 'fake signature',
+                'timestamp': Whatever(),
+                'x5u': Whatever(),
+                'public_key': Whatever(),
+            },
+            'recipe': {
```
Osmose: Is it worth specifying this explicitly vs. just calling `RecipeSerializer` here?

normandy/recipes/models.py
```diff
@@ -121,16 +150,40 @@ def __repr__(self):
...
+
+            # Check for changes that should disable the recipe
```
Osmose: Hmmmmm... so this got removed when we removed the peer approval changes from master. The `EDITABLE_FIELDS_WHITELIST` property got left in (which it shouldn't have been).

I think we should not include this in this PR.

Osmose:
Missing tests for `skip_last_updated` and for the management command.

Alright, I think that covers my first pass. Nice work thusfar!

normandy/recipes/api/serializers.py
mythmon: I tried that first. The problem is that this serializer is serializing a `Recipe`, and needs to pass itself to the nested serializer. If this line were just `recipe = RecipeSerializer()`, it would try and look for `self.recipe`, which doesn't exist. I also couldn't figure out anything I could supply as the `source` for the serializer to make it point to itself either.

normandy/recipes/tests/test_factories.py
mythmon: I wasn't sure there were working just right. So I tested it.

normandy/recipes/models.py
mythmon: Doing this last won't help, unfortunately. But I think we can bail early to make sure things work out. I'll write some tests to be sure.

normandy/recipes/utils.py
mythmon: We've talked about this before. `urljoin` makes the situation worse, not better.

```python
>>> from urllib.parse import urljoin
>>> urljoin('https://example.com/api/1/', 'sign/data')
'https://example.com/api/1/sign/data' # good...
>>> urljoin('https://example.com/api/1/', '/sign/data')
'https://example.com/sign/data' # bad, only kept hostname
>>> urljoin('https://example.com/api/1', 'sign/data')
'https://example.com/api/sign/data' # bad, stripped off "1"
>>> urljoin('https://example.com/api/1', '/sign/data')
'https://example.com/sign/data' # bad, only kept hostname
```

We should probably write a helper util that does this.

normandy/recipes/utils.py
mythmon: I think the creation of this is complex enough that it *should* be a model. Returning a dict just makes callers do something like `[Signature(**sig_data) for sig_data in autographer.sign_data()]`, which doesn't seem better to me.

The key part for me is that I don't see Autographer being used for anything except generating Signature instances. Maybe it should be a part of the Signature model?

normandy/recipes/models.py
mythmon: I blame git's bad conflict resolution UI.

normandy/recipes/utils.py
Osmose: I swear to god `urljoin` is the worst-named thing

normandy/recipes/utils.py
Osmose: > I think the creation of this is complex enough that it should be a model. Returning a dict just makes callers do something like [Signature(**sig_data) for sig_data in autographer.sign_data()], which doesn't seem better to me.

I think we'll just have to agree to disagree, I think having callers create signatures is fine and I think that separating the model creation from interacting with the Autograph server is a cleaner design. 

> The key part for me is that I don't see Autographer being used for anything except generating Signature instances. Maybe it should be a part of the Signature model?

Maybe as a compromise, have the Autograph class return dicts, and have a class method on the model that generates signature instances and calls the Autograph class?

normandy/recipes/tests/test_models.py
```diff
@@ -122,6 +164,18 @@ def test_update_signatures(self, mocker):
         signatures = list(Recipe.objects.all().values_list('signature__signature', flat=True))
         assert signatures == ['fake signature', 'fake signature']
 
+    def test_update_signatures_increments_revision_id(self, mocker):
```
Osmose: We talked offline about removing this test, let's not keep it.

normandy/recipes/tests/test_models.py
```diff
@@ -103,6 +120,31 @@ def test_canonical_json(self):
         expected = expected.encode()
         assert recipe.canonical_json() == expected
 
+    def test_signature_is_invalidated(self):
+        recipe = RecipeFactory(name='unchanged', signed=True)
+        recipe.name = 'changed'
+        recipe.save()
+        recipe = Recipe.objects.get(id=recipe.id)
```
Osmose: You can use [refresh_from_db](https://docs.djangoproject.com/en/1.9/ref/models/instances/#django.db.models.Model.refresh_from_db) here and below.

normandy/recipes/tests/test_models.py
```diff
@@ -103,6 +120,31 @@ def test_canonical_json(self):
...
+    def test_setting_signature_doesnt_change_canonical_json(self):
+        recipe = RecipeFactory(name='unchanged', signed=False)
+        serialized = recipe.canonical_json()
+        recipe.signature = SignatureFactory()
+        recipe.save()
+        recipe = Recipe.objects.get(id=recipe.id)
+        assert recipe.signature is not None
+        assert recipe.canonical_json() == serialized
+
+    def test_cant_change_serializer_and_other_fields(self):
```
Osmose: I think you mean `test_cant_change_signature_and_other_fields`.

Osmose:
r+wc if the tests pass after you remove that failure

PR #192, opened 2016-07-11
Moved API tests that had been part of mozilla-services to be part of this repo

chartjes:
@mythmon r?


rehandalal:
The contents of `CONTRACT-TESTS.md` should probably be moved to the docs.

And a little nit. I prefer `requirements-test.txt` to `test-requirements.txt`. But I'll let @mythmon decide if he cares about that.

CONTRACT-TESTS.md
```diff
@@ -0,0 +1,20 @@
+# API Contract Tests
+
+It's highly recommended to use [Virtualenv](https://virtualenv.pypa.io/en/latest/)
+in order to have an isolated environment.
+
+From the root directory of the project
+
+1. `virtualenv venv`
+2. `source venv/bin/activate` to turn on the virtualenv
+2. `./venv/bin/pip install -r test-requirements.txt`
```
mythmon: These docs seem to collide with our normal setup docs. For example, we suggest naming the virtualenv `venv` as well, and the two aren't compatible. It probably also needs to be called out that these need to be Python 2.

test-requirements.txt
```diff
@@ -0,0 +1,5 @@
+configparser
+flake8
+jsonschema
+pytest
+requests
```
mythmon: These should at least pin versions, and preferably use hashing to assure we get the same requirements each time.

CONTRACT-TESTS.md
mythmon: Oh wait, can these tests run in Python 3? I assumed, but then I didn't notice any Python2-only libraries in the requirements below. If that's the case, definitely defer to the standard setup docs for setting up an environment.

mythmon:
+1 to everything @rehandalal said.

This has broken the CI build for other tests, since we also use pytest. This PR either needs to make the two pytest setups into one, or modify `circle.yml` to specify which pytest to use.

conftest.py
```diff
@@ -0,0 +1,16 @@
+# Configuration file for running our tests
+import pytest
+
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--env",
+        action="store",
+        required=False,
```
mythmon: This should either be required, or have a default. Having neither produces strange errors like `configparser.NoSectionError: No section: None`.

mythmon:
Besides that `--env` issue, I think this is ready to merge. After that happens, I can get these to run in our automation.

chartjes:
@mythmon r? Added in a default value for --env

mythmon:
WFM

PR #193, opened 2016-07-12
Fix prop values in ControlApp

brittanystoroz:
`this` context is undefined in the `ControlApp` function the way it's been refactored. We have to explicitly pass in the objects we want to use otherwise `this.props` is always undefined.


mythmon:
r+

PR #194, opened 2016-07-12
Document the basics of how to write filter expressions.

Osmose:
Fixes bug 1284598.

@mythmon r?

### DIFF
--- a/docs/user/filter_expressions.rst
+++ b/docs/user/filter_expressions.rst
@@ -81,7 +81,7 @@ additional arguments passed in the expression:
 
 .. code-block:: javascript
 
-   'January 7th, 1980'|date // == a date object
+   '1980-01-07'|date // == a date object
 
 .. _JEXL Readme: https://github.com/TechnologyAdvice/Jexl#jexl---
 
@@ -165,8 +165,8 @@ function is the value being transformed.
 
 .. js:function:: date(dateString)
 
-   Parses a string as a date and returns a Date object. Strings should be in a
-   format recognized by `Date.parse()`_.
+   Parses a string as a date and returns a Date object. Date strings should be
+   in `ISO 8601`_ format.
 
    :param string dateString:
       String to parse as a date.
@@ -175,7 +175,7 @@ function is the value being transformed.
 
       '2011-10-10T14:48:00'|date // == Date object matching the given date
 
-   .. _Date.parse(): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse
+   .. _ISO 8601: https://www.w3.org/TR/NOTE-datetime
 
 Examples
 --------
@@ -201,5 +201,7 @@ This section lists some examples of commonly-used filter expressions.
    // Match users located in the US who have Firefox as their default browser
    normandy.country == 'US' && normandy.isDefaultBrowser
 
-   // Match users with the Flash plugin installed
+   // Match users with the Flash plugin installed. If Flash is missing, the
+   // plugin list returns `undefined`, which is a falsy value in JavaScript and
+   // fails the match. Otherwise, it returns a plugin object, which is truthy.
    normandy.plugins['Shockwave Flash']
--- a/docs/user/filter_expressions.rst
+++ b/docs/user/filter_expressions.rst
@@ -163,6 +163,32 @@ This section describes the transforms available to filter expressions, and what
 they do. They're documented as functions, and the first parameter to each
 function is the value being transformed.
 
+.. js:function:: stableSample(input, rate)
+
+   Randomly returns ``true`` or ``false`` based on the given sample rate. Used
+   to sample over the set of matched users.
+
+   Sampling with this transform is stable over the input, meaning that the same
+   input and sample rate will always result in the same return value. The most
+   common use is to pass in a unique user ID and a recipe ID as the input; this
+   means that each user will consistently run or not run a recipe.
+
+   Without stable sampling, a user might execute a recipe on Monday, and then
+   not execute it on Tuesday. In addition, without stable sampling, a recipe
+   would be seen by a different percentage of users each day, and over time this
+   would add up such that the recipe is seen by more than the percent sampled.
+
+   :param input:
+      A value for the sample to be stable over.
+   :param number rate:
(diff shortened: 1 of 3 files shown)


docs/user/filter_expressions.rst
```diff
@@ -0,0 +1,205 @@
...
+Filter expressions are written using a language called JEXL_. JEXL is an
+open-source expression language that is given a context (in this case,
+information about the user's browser) and evaluates a statement using that
+context. JEXL stands for "JavaScript Expression Language" and uses JavaScript
+syntax for several (but not all) of its features.
+
+.. note:: The rest of this document includes examples of JEXL syntax that has
+   comments inline with the expressions. JEXL does **not** have any support for
+   comments in statements, but we're using them to make understanding our
+   examples easier.
```
mythmon: This sounds like a good patch to submit to JEXL.

docs/user/filter_expressions.rst
```diff
@@ -0,0 +1,205 @@
...
+.. _JEXL: https://github.com/TechnologyAdvice/Jexl
+
+JEXL Basics
+-----------
+The `JEXL Readme`_ describes the syntax of the language in detail; the following
+section covers the basics of writing valid JEXL expressions.
+
+.. note:: Normally, JEXL doesn't allow newlines or other whitespace besides
+   spaces in expressions, but filter expressions in Normandy allow arbitrary
+   whitespace.
```
mythmon: This should probably go away now?

docs/user/filter_expressions.rst
```diff
@@ -0,0 +1,205 @@
...
+
+   normandy.locale == 'en-US' // == true if the client's locale is en-US
+
+Another unique feature of JEXL is transforms, which modify the value given to
+them. Transforms are applied to a value using the ``|`` operator, and may take
+additional arguments passed in the expression:
+
+.. code-block:: javascript
+
+   'January 7th, 1980'|date // == a date object
```
mythmon: This won't work, right now.

```
> new Date('January 7th, 1980')
Invalid Date
```

The values that work right now are, according the JS spec, ISO8601 format, or whatever the browser feels like today. Probably best to stick with ISO8601, like `1980-01-07`.

docs/user/filter_expressions.rst
```diff
@@ -0,0 +1,205 @@
...
+.. js:attribute:: normandy.request_time
+
+   Date object set to the time and date that the user requested recipes from
+   Normandy. Useful for comparing against date ranges that a recipe is valid
+   for.
+
+   .. code-block:: javascript
+
+      // Do not run recipe after January 1st.
+      normandy.request_time < '2011-01-01'|date
```
mythmon: :+1:

docs/user/filter_expressions.rst
```diff
@@ -0,0 +1,205 @@
...
+Transforms
+----------
+This section describes the transforms available to filter expressions, and what
+they do. They're documented as functions, and the first parameter to each
+function is the value being transformed.
+
+.. js:function:: date(dateString)
+
+   Parses a string as a date and returns a Date object. Strings should be in a
+   format recognized by `Date.parse()`_.
```
mythmon: I'd be stronger here, and say that strings should be in ISO8601 format, which isn't technically true, but is probably a better idea.

docs/user/filter_expressions.rst
```diff
@@ -0,0 +1,205 @@
...
+   Normandy. Useful for comparing against date ranges that a recipe is valid
+   for.
+
+   .. code-block:: javascript
+
+      // Do not run recipe after January 1st.
+      normandy.request_time < '2011-01-01'|date
+
+Transforms
+----------
```
mythmon: I think you're missing `|stableSample(rate)` here.

docs/user/filter_expressions.rst
```diff
@@ -0,0 +1,205 @@
...
+   (
+      normandy.request_time > '2011-01-01T00:00:00+00:00'|date
+      && normandy.request_time < '2011-01-07T00:00:00+00:00'|date
+   )
+
+   // Match users located in the US who have Firefox as their default browser
+   normandy.country == 'US' && normandy.isDefaultBrowser
+
+   // Match users with the Flash plugin installed
+   normandy.plugins['Shockwave Flash']
```
mythmon: This might need some explaining for people who aren't familiar with JS's truthiness behaviors.

mythmon:
That's all I've got

docs/user/filter_expressions.rst
Osmose: IMO not until they do a release and/or we're using the version that supports this.

mythmon: :+1:

docs/user/filter_expressions.rst
Osmose: Current master doesn't have `stableSample` available yet so I didn't include it.

mythmon: I think `stableSample` is an important part of the S&I test, and something we should document. I could add it in my PR that makes `stableSample` available, if you'd prefer.

mythmon:
These docs look good. Do we want to add the missing bits (stable sample and user id) to this, so S&I can read them early? Or do we want to add those docs as those features land? If you want to land this as is, go ahead and merge it.

docs/user/filter_expressions.rst
```diff
@@ -163,6 +163,32 @@ This section describes the transforms available to filter expressions, and what
...
+   :param input:
+      A value for the sample to be stable over.
+   :param number rate:
+      A number between ``0`` and ``1`` with the sample rate. For example,
+      ``0.5`` would be a 50% sample rate.
+
+   .. code-block:: javascript
+
+      // True 50% of the time, stable per-user per-recipe.
+      (normandy.userId + normandy.recipe.id)|stableSample(0.5)
```
mythmon: I was expecting `[normandy.userId, normandy.recipe.id]|stableSample(0.5)`. Not that it would really be much different in JS.

mythmon:
Nit about the input, otherwise r+

docs/user/filter_expressions.rst
Osmose: Agreed, that's better! Fixed.

PR #196, opened 2016-07-13
Validate arguments

brittanystoroz:
This is a little hacky right now. I had trouble massaging the error response data into a format that would mean something to redux form. Other things that need taking care of:

- [x] convert weight field value to integer before submitting
- [x] don't submit fields with empty string values / convert their values to undefined...errr are we ok with empty string values? should we only yell about it if the field is required?
- [x] add indicator in surveys list if a survey has fields that are invalid

### DIFF
--- a/normandy/control/static/control/admin/sass/control.scss
+++ b/normandy/control/static/control/admin/sass/control.scss
@@ -276,6 +276,20 @@
           padding: 5px 3em;
         }
 
+        &.invalid {
+          color: $red;
+          font-weight: 700;
+
+          &:before {
+            border: none;
+            color: $red;
+            content: "\f071";
+            font: normal normal 400 11px/1 FontAwesome;
+            left: -1.5em;
+            margin-right: 4px;
+          }
+        }
+
         &:before {
           border-color: transparent $darkBrown;
           border-style: solid;
--- a/normandy/control/static/control/admin/sass/control.scss
+++ b/normandy/control/static/control/admin/sass/control.scss
@@ -283,7 +283,7 @@
           &:before {
             border: none;
             color: $red;
-            content: "\f071";
+            content: "\f071"; // error icon, fa-exclamation-triangle
             font: normal normal 400 11px/1 FontAwesome;
             left: -1.5em;
             margin-right: 4px;
--- a/normandy/control/static/control/js/components/RecipeForm.js
+++ b/normandy/control/static/control/js/components/RecipeForm.js
@@ -2,6 +2,7 @@ import React, { PropTypes as pt } from 'react';
 import { Link } from 'react-router';
 import { push } from 'react-router-redux';
 import { destroy, reduxForm, getValues } from 'redux-form';
+import { _ } from 'underscore';
 
 import { makeApiRequest, recipeUpdated, recipeAdded, showNotification }
   from '../actions/ControlActions.js';
@@ -63,6 +64,7 @@ export class RecipeForm extends React.Component {
     const recipeFormValues = getValues(formState.recipe);
     const actionFormValues = getValues(formState.action);
     const combinedFormValues = { ...recipeFormValues, arguments: actionFormValues };
+
     const requestBody = { recipe: combinedFormValues, recipeId };
 
     return this.validateForm(combinedFormValues)
--- a/normandy/control/static/control/js/components/RecipeForm.js
+++ b/normandy/control/static/control/js/components/RecipeForm.js
@@ -157,9 +157,15 @@ export default composeRecipeContainer(reduxForm({
   const formatErrors = payload => {
     let errors = payload;
 
+    /* If our payload is an object, process each error in the object
+       Otherwise, it is a string and will be returned immediately */
     if (_.isObject(payload)) {
       const invalidFields = Object.keys(payload);
       if (invalidFields.length > 0) {
+        /* If our error keys are integers, it means they correspond
+           to an array field and we want to present errors as an array
+           e.g. { surveys: {0: {title: 'err'}}, {2: {weight: 'err'}} }
+           =>   { surveys: [{title: 'err'}, null, {weight: 'err'}] } */
         errors = isNaN(invalidFields[0]) ? {} : [];
 
         invalidFields.forEach(fieldName => {
--- a/normandy/control/static/control/js/components/RecipeForm.js
+++ b/normandy/control/static/control/js/components/RecipeForm.js
@@ -1,7 +1,7 @@
 import React, { PropTypes as pt } from 'react';
 import { Link } from 'react-router';
 import { push } from 'react-router-redux';
-import { destroy, reduxForm, getValues } from 'redux-form';
+import { destroy, stopSubmit, reduxForm, getValues } from 'redux-form';
 import { _ } from 'underscore';
 
 import { makeApiRequest, recipeUpdated, recipeAdded, showNotification }
@@ -64,7 +64,6 @@ export class RecipeForm extends React.Component {
     const recipeFormValues = getValues(formState.recipe);
     const actionFormValues = getValues(formState.action);
     const combinedFormValues = { ...recipeFormValues, arguments: actionFormValues };
-
     const requestBody = { recipe: combinedFormValues, recipeId };
 
     return this.validateForm(combinedFormValues)
@@ -155,10 +154,37 @@ export default composeRecipeContainer(reduxForm({
     ? props.location.state.selectedRevision
     : null;
 
+  const formatErrors = payload => {
+    let errors = payload;
+
+    if (_.isObject(payload)) {
+      const invalidFields = Object.keys(payload);
+      if (invalidFields.length > 0) {
+        errors = isNaN(invalidFields[0]) ? {} : [];
+
+        invalidFields.forEach(fieldName => {
+          errors[fieldName] = formatErrors(payload[fieldName]);
+        });
+      }
+    }
+
+    return errors;
+  };
+
+  const onSubmitFail = errors => {
+    const { dispatch } = props;
+    const actionFormErrors = errors['arguments']; // eslint-disable-line dot-notation
+
+    if (actionFormErrors) {
+      dispatch(stopSubmit('action', formatErrors(actionFormErrors)));
+    }
+  };
+
   return {
     fields,
     initialValues: selectedRecipeRevision || props.recipe,
     viewingRevision: selectedRecipeRevision || props.location.query.revisionId,
     formState: state.form,
+    onSubmitFail,
   };
 })(RecipeForm));
--- a/normandy/control/static/control/js/components/RecipeForm.js
+++ b/normandy/control/static/control/js/components/RecipeForm.js
@@ -173,7 +173,7 @@ export default composeRecipeContainer(reduxForm({
 
   const onSubmitFail = errors => {
     const { dispatch } = props;
-    const actionFormErrors = errors['arguments']; // eslint-disable-line dot-notation
+    const actionFormErrors = errors.arguments;
 
     if (actionFormErrors) {
       dispatch(stopSubmit('action', formatErrors(actionFormErrors)));
--- a/normandy/control/static/control/js/components/action_forms/HeartbeatForm.js
+++ b/normandy/control/static/control/js/components/action_forms/HeartbeatForm.js
@@ -21,8 +21,6 @@ export const HeartbeatFormFields = [
   'surveys[].weight',
 ];
 
-const formatLabel = labelName => labelName.replace(/([A-Z])/g, ' $1').toLowerCase();
-
 const SurveyListItem = props => {
   const { survey, surveyIndex, isSelected, deleteSurvey, onClick } = props;
   return (
@@ -51,9 +49,11 @@ const SurveyForm = props => {
   const { selectedSurvey, fields, showDefaults } = props;
   const surveyObject = selectedSurvey || fields.defaults;
   let headerText = 'Default Survey Values';
+  let showAdditionalSurveyFields = false;
   let containerClasses = classNames('fluid-8', { active: selectedSurvey });
 
   if (selectedSurvey) {
+    showAdditionalSurveyFields = true;
     headerText = selectedSurvey.title.initialValue || 'New survey';
   }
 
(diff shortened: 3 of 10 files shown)


normandy/recipes/api/serializers.py
```diff
@@ -102,6 +104,25 @@ class Meta:
...
+            errorResponse.append({
+                'message': error.message,
+                'field': json.JSONEncoder().encode(list(error.path))
+            })
+
+        if len(errorResponse):
+            raise serializers.ValidationError(errorResponse)
+
+        return value
+
```
brittanystoroz: Is this where this should go? I tried putting the validation logic all in `/validators.py` but I wasn't able to import the `Action` model there for some reason, which I needed in order to query the schema.

Osmose:
Whenever this is ready this smells like @rehandalal 's expertise to me

normandy/recipes/api/serializers.py
rehandalal: Seems like the right place to me. you probably couldn't import `Action` due to some kind of circular import issue. But I think this should be here anyway.

brittanystoroz: Cool, I will add some tests for this.

normandy/recipes/api/serializers.py
```diff
@@ -102,6 +103,43 @@ class Meta:
             'is_approved',
         ]
 
+    def validate_arguments(self, value):
```
rehandalal: looks like a lot happening here. might make sense to include some tests.

normandy/recipes/tests/test_serializers.py
```diff
@@ -28,6 +31,46 @@ def test_it_works(self, rf):
             'is_approved': recipe.is_approved
         }
 
+    def test_it_validates_arguments(self):
```
brittanystoroz: @rehandalal ummm, are these on the right track?

I want to test the serializer with 3 different sets of data -- 
1. with an invalid action name, as it would not be able to find an arguments_schema to validate against
2. with a valid action, but invalid arguments, so it returns the right error formatting for `arguments`
3. with all valid data so it raises no errors

Is there a concept of isolating these different scenarios in pytest? Like in JavaScript you would wrap them all in a `describe` block then isolate each case into an `it` block.

normandy/recipes/tests/test_serializers.py
```diff
@@ -28,6 +31,46 @@ def test_it_works(self, rf):
...
+        schema = json.loads(
+            open('normandy/recipes/static/actions/console-log/package.json').read()
```
brittanystoroz: @rehandalal Right now I'm just pulling in the `console-log` schema to validate against. This one is a bit too simple for all the complex logic in `validate_arguments`. I can switch it to the `show-heartbeat` schema orrrrr is it bad practice to depend on these files and I should just mock up an entirely fake one?

normandy/recipes/tests/test_serializers.py
```diff
@@ -28,6 +31,46 @@ def test_it_works(self, rf):
...
+            'action': 'console-log', 'arguments': {'message': ''}
+        })
+
+        serializer.is_valid()
+        assert serializer.errors['arguments'] == {'message': 'This field may not be blank.'}
+        assert pytest.raises(serializers.ValidationError)
+
+        # If the action can be found, and the arguments passed
+        # are accurate, serializer should be valid
+        serializer = RecipeSerializer(data={
```
brittanystoroz: Does python care that I'm continuously resetting `serializer`? It doesn't seem to. I know JavaScript would probably not like this.

normandy/recipes/api/serializers.py
brittanystoroz: @rehandalal Added some serializer tests with a bunch of questions for you. I'll also add some api tests to make sure the right responses are returned

normandy/control/static/control/admin/sass/control.scss
```diff
@@ -276,6 +276,20 @@
...
         }
 
+        &.invalid {
+          color: $red;
+          font-weight: 700;
+
+          &:before {
+            border: none;
+            color: $red;
+            content: "\f071";
```
mythmon: It would be great to have a comment that describes what this character is.

normandy/control/static/control/js/components/RecipeForm.js
```diff
@@ -153,10 +154,37 @@ export default composeRecipeContainer(reduxForm({
     ? props.location.state.selectedRevision
     : null;
 
+  const formatErrors = payload => {
+    let errors = payload;
+
+    if (_.isObject(payload)) {
+      const invalidFields = Object.keys(payload);
+      if (invalidFields.length > 0) {
+        errors = isNaN(invalidFields[0]) ? {} : [];
```
mythmon: That's a pretty confusing incantation. What's going on here?

Overall, it might be a good idea to add a comment to this function that describes the kind of data it is expecting.

normandy/control/static/control/js/components/RecipeForm.js
```diff
@@ -153,10 +154,37 @@ export default composeRecipeContainer(reduxForm({
...
+        });
+      }
+    }
+
+    return errors;
+  };
+
+  const onSubmitFail = errors => {
+    const { dispatch } = props;
+    const actionFormErrors = errors['arguments']; // eslint-disable-line dot-notation
```
mythmon: Why disable `dot-notation`? Is there a reason to write this as `errors['arguments']` instead of `errors.arguments`?

normandy/control/static/control/js/components/action_forms/HeartbeatForm.js
```diff
@@ -141,6 +149,7 @@ class HeartbeatForm extends React.Component {
                       survey={survey}
                       surveyIndex={index}
                       isSelected={_.isEqual(survey, selectedSurvey)}
+                      hasErrors={_.some(survey, field => field.invalid)}
```
rehandalal: 👍🏼

normandy/recipes/tests/test_serializers.py
rehandalal: Nope totally fine.

normandy/recipes/api/serializers.py
```diff
@@ -102,6 +103,43 @@ class Meta:
...
+        try:
+            schema = Action.objects.get(name=self.initial_data.get('action')).arguments_schema
+        except:
+            raise serializers.ValidationError('Could not find arguments schema.')
+
+        schemaValidator = Validator(schema)
+        errorResponse = {}
+        errors = sorted(schemaValidator.iter_errors(value), key=lambda e: e.path)
+
+        for error in errors:
```
mythmon: It might be a good idea here to add a comment that shows the expected shape of the data here too.

normandy/recipes/tests/test_serializers.py
rehandalal: Usually I'd say mock up a fake one because: the schema may change, which would possibly change the test coverage. It also allows you to very specifically target aspects of the validation function to ensure it works correctly. But for now this should be fine. Maybe toss a TODO comment in to say we should probably do a more extensive test at some point.

mythmon: It is probably fine to use this schema file directly (or the heartbeat one). Mocking out an entirely fake one has the advantage of being able to get exactly the data you need to test the behavior you want.

mythmon: Also, you should do this as

```python
with open('normandy/recipes/static/actions/console-log/package.json') as f:
    schema = json.load(f)
```

This way the file is automatically closed, and also `json.load` will read directly from files.

normandy/recipes/tests/test_serializers.py
rehandalal: You could refactor it to have another function that takes the data and verifies the assertions, but since it's only 3 sets of data I think that would be overkill. This looks good to me.

normandy/recipes/tests/test_serializers.py
```diff
@@ -28,6 +31,46 @@ def test_it_works(self, rf):
...
+            arguments_schema=schema['normandy']['argumentsSchema']
+        )
+
+        # If the action specified cannot be found, raise validation
+        # error indicating the arguments schema could not be loaded
+        serializer = RecipeSerializer(data={
+            'action': 'action-that-doesnt-exist', 'arguments': {}
+        })
+
+        serializer.is_valid()
```
mythmon: It's probably a good idea to assert the value this returns.

normandy/control/static/control/js/components/RecipeForm.js
rehandalal: out of curiosity, why couldn't you use dot notation here?

normandy/recipes/tests/test_serializers.py
```diff
@@ -28,6 +31,46 @@ def test_it_works(self, rf):
...
+        assert serializer.errors['arguments'] == ['Could not find arguments schema.']
+        assert pytest.raises(serializers.ValidationError)
```
mythmon: I don't think this does what you want it to. Usually this is used as

```python
with pytest.raises(SomeExceptionType):
    something_that_raises_that_type()
```

Just calling `pytest.raises(...)` alone doesn't actually assert anything.

normandy/recipes/tests/test_serializers.py
mythmon: Python doesn't care. How would JavaScript care?

That being said, you might want to break each of these sections into a separate test, if they are independent.

normandy/recipes/validators.py
```diff
@@ -1,8 +1,33 @@
...
+        if requirement not in instance or instance[requirement] == '':
+            error = jsonschema.ValidationError(
+                'This field may not be blank.',
+                path=[requirement]
+            )
+            yield error
+
+
+# Construct validator as extension of Json Schema Draft 4.
+Validator = jsonschema.validators.extend(
```
rehandalal: Maybe rename this to `JSONSchemaValidator` so it's a little clearer what it's validating and it the name won't conflict with any future validators we may write.

normandy/recipes/validators.py
```diff
@@ -1,8 +1,33 @@
 import json
+import jsonschema
 
 from django.core.exceptions import ValidationError
 
 
+# Add path to required validator so we can get property name
+def _required(validator, required, instance, schema):
+    '''Validate 'required' properties.'''
```
mythmon: Nit: Use triple-double quotes (`"""`), not triple-single quotes (`'''`).

normandy/recipes/validators.py
```diff
@@ -1,8 +1,33 @@
...
+    if not validator.is_type(instance, 'object'):
+        return
+
+    for index, requirement in enumerate(required):
+        if requirement not in instance or instance[requirement] == '':
```
mythmon: You could write this as `if instance.get(requirements, '') != '':`.

rehandalal: +1

normandy/control/static/control/js/components/RecipeForm.js
rehandalal: haha. sorry for the double post. i was reviewing this at the same time @mythmon was.

normandy/recipes/tests/test_serializers.py
rehandalal: i'm just guessing Brittany is referring to ESLint complaining about reassigning variables that are arguments of a functions.

normandy/recipes/tests/test_serializers.py
rehandalal: the more i think about it, maybe just mock up a fake one now.

mythmon:
I stole this to review. I have some comments, mostly about the Python. Also, @brittanystoroz promised some API tests. Your turn!

normandy/recipes/tests/test_serializers.py
rehandalal: to expand on this you probably want to do something like this:
```python
with pytest.raises(serializers.ValidationError):

serializer.errors['arguments'] == ['Could not find arguments schema.']
```
```

File: normandy/control/static/control/js/components/RecipeForm.js (hunk 8 above)
Comment (brittanystoroz) [inline normandy/control/static/control/js/components/RecipeForm.js]: There was previously a problem using `arguments` with redux-form because it was a reserved keyword. Now that we're processing that data and sending it as `actionFormErrors` that's not an issue any more. I've changed this to `errors.arguments`

File: normandy/recipes/api/serializers.py (hunk 10 above)
Comment (brittanystoroz) [inline normandy/recipes/api/serializers.py]: Added a comment [here](https://github.com/brittanystoroz/normandy/blob/7558686c5c86587ca361b3b9a3011804923c8613/normandy/recipes/api/serializers.py#L117-L120)  

But could/should I also just link to the `ValidationError` docs in jsonschema? [#jsonSchema.ValidationError](http://python-jsonschema.readthedocs.io/en/v0.8.0/errors/#jsonschema.ValidationError)

File: normandy/recipes/tests/test_serializers.py [hunk 16]
```diff
@@ -28,6 +29,117 @@ def test_it_works(self, rf):
...
+            serializer.is_valid(raise_exception=True)
+
+        assert serializer.errors['arguments'] == ['Could not find arguments schema.']
+
+    # If the action can be found, raise validation error
+    # with the arguments error formatted appropriately
+    def test_validation_with_wrong_arguments(self):
+        ActionFactory(
+            name='show-heartbeat',
+            arguments_schema={
```
rehandalal: maybe pull this out into something like:
```python
ARGUMENTS_SCHEMA = {
    "required": ["surveyId", "surveys"],
    "properties": {
        "surveyId": {"type": "string"},
        "surveys": {
            "type": "array",
            "minItems": 1,
            "items": {
                "properties": {
                    "title": {"type": "string"},
                    "weight": {"type": "integer", "minimum": 1}
                },
                "required": ["title", "weight"]
            },
        },
    },
}
```
at the top of the file or in `__init__.py`.

normandy/recipes/tests/test_serializers.py
```diff
@@ -28,6 +29,117 @@ def test_it_works(self, rf):
...
+            'arguments': {
+                'surveyId': 'lorem-ipsum-dolor',
+                'surveys': [
+                    {'title': 'adipscing', 'weight': 1},
+                    {'title': 'consequetar', 'weight': 1}
+                ]
+            }
+        })
+
+        serializer.is_valid()
```
rehandalal: should probably be:
```python
assert serializer.is_valid()
```

normandy/recipes/tests/test_serializers.py
brittanystoroz: I was just about to ask about that. Will do

brittanystoroz:
Ok -- all review should be taken care of. I added an API test that asserts a `400` response when the arguments are invalid. Should I add a test that asserts a `201` with valid arguments?

mythmon:
The updates look fine to me. @rehandalal what do you think?

rehandalal:
lgtm. r+ 🚢 

nice work @brittanystoroz!

PR #198, opened 2016-07-15
Do not render the control app on the login page.

Osmose:
This fixes a bug where the login page showed the 404 page of the
control app since the login page is separate from the frontend app.

I couldn't think of any good unit tests that we'd want to keep around for this. Someday I think we'll just allow login via the API and have frontend-based login views, but that day is not today.

@rehandalal r?


rehandalal:
lgtm. r+

PR #199, opened 2016-07-15
Add userId and recipe to filter expression context for self-repair.

Osmose:
Fixes bug 1286896

@mythmon r?


mythmon:
It would be nice to see a test that involves `userId`.

Would it be possible to write a test that calls `doesMatchRecipe`? Do we have this test already?

mythmon:
Nice.

PR #201, opened 2016-07-19
Move most client side code outside of Django's grasp

mythmon:
This makes JS import paths much shorter, and moves everything out of the crazy `normandy/app/static/app/js/app/component/foo.js` structure we had.

I'm not sure if this is what we want. I do think it is an improvement. What do y'all think?

r?


Osmose:
Thoughts on whether we should land on a branch instead of master to keep it out of the next release? I'm on the fence myself.

Osmose:
I like this! I'd actually go a step farther and move the actions in here as well. After years of dealing with per-app JavaScript I'm about ready to give up on it.

mythmon:
I was planning that we could land this after we tag the release, but I could retarget this to a branch. I don't think it matters much, since we won't have continued development before this can hit master.

I was a little unclear about moving the actions, but I could go back and look at that again. They are more ingrained in the Django app, so it is a bit harder to detangle. I figured I'd do the easier parts and then get feedback.

Another thing I'm unsure about is the name of the top level directory, and the move from `control/js/*.js` to `control/*.js`. I'd be happy to change those, if anyone has opinions.

Osmose:
I like `client` and have seen that in other projects (specifically node ones but that's fine).

Remove `js` seems fine to me. I'd be interested in someday splitting up frontend files by the components they support rather than their file types, as well.

package.json
```diff
@@ -9,7 +9,8 @@
   "scripts": {
     "watch": "webpack --config ./webpack.config.js --watch",
     "build": "webpack --config ./webpack.config.js",
-    "test": "karma start"
+    "test": "karma start",
+    "lint": "eslint normandy client webpack.config.js karam.conf.js"
```
rehandalal: i like that we're also linting `webpack.config.js`

mythmon: That's actually new in this PR.

rehandalal: yeah, i like it!

rehandalal: one annoying thing about this is when there is a lint error. npm errors out too and spits out all of this:
```
npm ERR! Linux 3.13.0-79-generic
npm ERR! argv "/home/rehandalal/.nvm/versions/node/v4.0.0/bin/node" "/home/rehandalal/.nvm/versions/node/v4.0.0/bin/npm" "run" "lint"
npm ERR! node v4.0.0
npm ERR! npm  v2.14.2
npm ERR! code ELIFECYCLE
npm ERR! normandy@ lint: `eslint normandy client webpack.config.js karam.conf.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the normandy@ lint script 'eslint normandy client webpack.config.js karam.conf.js'.
npm ERR! This is most likely a problem with the normandy package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     eslint normandy client webpack.config.js karam.conf.js
npm ERR! You can get their info via:
npm ERR!     npm owner ls normandy
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /home/rehandalal/Projects/normandy/npm-debug.log
```

and creates the file `npm-debug.log`.

rehandalal: i was able to get around this with:
```json
"lint": "eslint normandy client webpack.config.js karam.conf.js || true"
```
but maybe there's a better way?

mythmon: I think you're right. I made it `npm run lint` so linting was standardized, but the output kind of sucks. I think the JUnit output makes this ok though? Also I'd have to do `node_modules/.bin/eslint client ...` because PATH issues.

Why `|| true`? That makes the command always pass, which I don't think is what we want.

rehandalal: it makes the command always pass so that npm doesn't error out but you still get all the lint output. you just don't get all the `npm ERR!` output and it doesn't spit out an `npm-debug.log` file each time you run.

rehandalal: that being said i'm sure there is a better way than `|| true`. that was my terrible 5 second fix.

mythmon: That makes sense. One reason I'm hesitant to do this is we rely on the exit code for CI. So if we do `|| true`, we can't use this for CI. I did some googling and I don't think there is any way to keep the exit code but not have npm lose its shit.

I think a better approach would be using [therapist](https://github.com/rehandall/therapist) instead of npm scripts to run lints, once that is ready.

rehandalal: +1

brittanystoroz: `karam.conf.js` should be `karma.conf.js` ...?

mythmon: @brittanystoroz yes, right

mythmon:
I rebased this and fixed some merge conflicts. This should be ready to merge now.

brittanystoroz:
r+

PR #207, opened 2016-08-02
Upgrade to therapist 0.3.1

rehandalal:
I found a couple issues that are outlined here:
https://github.com/rehandalal/therapist/blob/master/CHANGES.md

r?


mythmon:
Thanks!

PR #210, opened 2016-08-03
Test that the API does not set cookies

mythmon:
If we set cookies on the anonymous API requests normal clients make, we would totally destroy our caches. So lets not do that.

@rehandalal r?


rehandalal:
agression tests lgtm. r+

PR #211, opened 2016-08-04
Add ProductionReadOnly environment to settings.

Osmose:
This environment will be used on public webheads that shouldn't
make changes to the database. Mainly we want to disable things that
modify cookies (like SessionMiddleware) on these nodes.

@mythmon r?

### DIFF
--- a/normandy/settings.py
+++ b/normandy/settings.py
@@ -31,17 +31,12 @@ class Core(Configuration):
         'django.contrib.staticfiles',
     ]
 
+    # Middleware that ALL environments must have. See the Base class for
+    # details.
     MIDDLEWARE_CLASSES = [
         'normandy.base.middleware.RequestReceivedAtMiddleware',
         'django.middleware.security.SecurityMiddleware',
-        'normandy.base.middleware.ShortCircuitMiddleware',
-
-        'django.contrib.sessions.middleware.SessionMiddleware',
         'django.middleware.common.CommonMiddleware',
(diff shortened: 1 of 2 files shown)


normandy/settings.py
```diff
@@ -136,6 +129,25 @@ class Base(Core):
     DEBUG = values.BooleanValue(False)
     ADMINS = values.SingleNestedListValue([])
 
+    # Middleware that _most_ environments will need. Subclasses can
+    # override this list.
+    EXTRA_MIDDLEWARE_CLASSES = [
+        'django.contrib.sessions.middleware.SessionMiddleware',
+        'django.middleware.common.CommonMiddleware',
```
mythmon: I think we want to keep this one.

normandy/settings.py
```diff
@@ -136,6 +129,25 @@ class Base(Core):
...
+        'django.middleware.csrf.CsrfViewMiddleware',
+        'django.contrib.auth.middleware.AuthenticationMiddleware',
+        'django.contrib.auth.middleware.SessionAuthenticationMiddleware',
+        'django.contrib.messages.middleware.MessageMiddleware',
+        'django.middleware.clickjacking.XFrameOptionsMiddleware',
```
mythmon: I think we want to keep this one too.

mythmon: I think we want to keep this one too.

mythmon:
r+

PR #215, opened 2016-08-10
Reversion improvements

Osmose:
So like I was going to work on real bugs but got distracted thinking about how I couldn't tell which revisions were made during `update_actions` and which ones weren't. Thus, this PR!

1. Updates `update_actions` to set a comment on reversions it creates.
2. A tiny tweak to simplify a view by using a helper I found.
3. Adds comments to the history page. This got a bit out of control in terms of size, but whatever.

   Because the styling of the history page got difficult using flexbox, and because the data is tabular anyway, I converted the page to use a table. To make unit testing some of my changes easier, I also split the history component into a state container and a presentational component. The container should really be managing state in redux instead of on itself, but that tidbit can wait until a later refactoring. I couldn't think of good ways to test the container and it's mostly unchanged from the old code anyway, so I didn't test it, but I did test the presentational stuff.

@rehandalal r?

### DIFF
--- a/normandy/control/static/control/admin/sass/control.scss
+++ b/normandy/control/static/control/admin/sass/control.scss
@@ -330,33 +330,24 @@
 
 /* Recipe History */
 .recipe-history {
-  .history-item {
-    display: flex;
-    flex-direction: row;
+  td {
+    text-align: left;
+  }
 
-    .revision-number,
-    .revision-created {
-      line-height: 20px;
-      margin: 0 20px;
-    }
+  .revision-number {
+    font-weight: 600;
 
-    .revision-number {
-      flex: 0;
-      font-weight: 600;
-    }
+    /* Force table cell to smallest width */
+    white-space: nowrap;
+    width: 1px;
+  }
 
-    .revision-created {
-      flex: 1;
-    }
+  .status-indicator {
+    margin: 0 20px; /* Reset .status-indicator margin */
+  }
 
-    .revision-status {
-      flex: 1;
-      margin: 0 20px; /* Reset .status-indicator margin */
-    }
-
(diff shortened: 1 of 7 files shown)


normandy/control/static/control/admin/sass/control.scss
```diff
@@ -330,33 +330,24 @@
...
+  }
 
-    .revision-number {
-      flex: 0;
-      font-weight: 600;
-    }
+  .revision-number {
+    /* Force table cell to smallest width */
+    width: 1px;
+    white-space: nowrap;
```
rehandalal: nit: wh < wi

normandy/control/static/control/js/components/RecipeHistory.js
```diff
@@ -29,39 +29,52 @@ class RecipeHistory extends React.Component {
...
+    return <HistoryList recipe={recipe} dispatch={dispatch} revisions={revisions} />;
   }
 }
 
-class HistoryItem extends React.Component {
+export function HistoryList({ recipe, revisions, dispatch }) {
+  return (
+    <div className="fluid-8 recipe-history">
+      <h3>Viewing revision log for: <b>{recipe ? recipe.name : ''}</b></h3>
+      <table>
```
rehandalal: tables for tabular data makes sense! +1

rehandalal:
lgtm. r+wc

normandy/control/static/control/admin/sass/control.scss
Osmose: I AM ASHAMED

rehandalal: Shame!

PR #218, opened 2016-08-16
Remove leftover models for server-side filters

mythmon:
We had a bunch of code leftover from when we had server side filtering. We decided to leave it in place until after the control interface was ready so we could be sure we didn't need it.

We don't need it.

🔥🔥🔥
🔥🔥🔥
🔥🔥🔥

@Osmose r?


Osmose:
r+ 🔥

PR #219, opened 2016-08-17
Use mozlog format in production

mythmon:
Fixes bug 1273331

@Osmose r?

### DIFF
--- a/normandy/settings.py
+++ b/normandy/settings.py
@@ -148,6 +148,40 @@ class Base(Core):
         """
         return Core.MIDDLEWARE_CLASSES + self.EXTRA_MIDDLEWARE_CLASSES
 
+    LOGGING_USE_JSON = values.BooleanValue(False)
+
+    def LOGGING(self):
+        return {
+            'version': 1,
+            'disable_existing_loggers': True,
+            'formatters': {
+                'json': {
+                    '()': 'mozilla_cloud_services_logger.formatters.JsonLogFormatter',
+                    'logger_name': 'normandy',
+                },
+                'development': {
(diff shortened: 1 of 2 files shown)


normandy/settings.py
```diff
@@ -148,6 +148,40 @@ def MIDDLEWARE_CLASSES(self):
...
+            'version': 1,
+            'disable_existing_loggers': True,
+            'formatters': {
+                'json': {
+                    '()': 'mozilla_cloud_services_logger.formatters.JsonLogFormatter',
+                    'logger_name': 'normandy',
+                },
+                'development': {
+                    'format': '%(levelname)s %(asctime)s %(name)s %(message)s',
+                }, },
```
Osmose: Put that `}` on another line please.

normandy/settings.py
```diff
@@ -148,6 +148,40 @@ def MIDDLEWARE_CLASSES(self):
...
+            'handlers': {
+                'console': {
+                    'level': 'DEBUG',
+                    'class': 'logging.StreamHandler',
+                    'formatter': 'json' if self.LOGGING_USE_JSON else 'development',
+                },
+            },
+            'root': {
+                'handlers': ['console'],
```
Osmose: Maybe make the setting `LOGGING_HANDLERS` and use it here and below? Would allow for a bit more flexibility.

Osmose:
r+wc, nice work!

normandy/settings.py
mythmon: I don't think this is something we need to be flexible right now. It is easy enough to add later if we need

PR #221, opened 2016-08-18
Delete babel dep

jmm:
Hello, while it doesn't really hurt anything by being there, you should be able to delete the `babel` dep as babel@6 doesn't really do anything. (FYI I haven't run the tests though.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,6 @@
     "wolfy87-eventemitter": "4.3.0"
   },
   "devDependencies": {
-    "babel": "6.5.2",
     "babel-core": "6.11.4",
     "babel-eslint": "6.1.2",
     "babel-loader": "6.2.4",


mythmon:
Ah, I didn't know. Thanks for the patch!

jmm:
No problem! The role of the packages and the changes in v6 weren't always real clear ([see for example](http://stackoverflow.com/q/32544685/1034448)) or documented great (and I've seen the `babel@6` dep in other projects). It was decided to move the CLI from `babel` to `babel-cli` for v6 and we talked about having `babel` be the API and eliminating `babel-core`. That would've been a nice situation to end up with (`babel` = API, `babel-cli` = CLI), but there was concern about the confusion factor. So ultimately `babel` just got deprecated. See also https://babeljs.io/blog/2015/10/31/setting-up-babel-6.

PR #224, opened 2016-08-19
CI: Save docker cache before tests run

mythmon:
CircleCI saves the cache after building dependencies, before running the tests. Without this change, the cache is never updated.

@relud r?

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -9,8 +9,12 @@ dependencies:
   override:
     - docker info
     # use circleci's docker cache workaround
-    - mkdir -p ~/cache/
-    - if [ -e ~/cache/docker/image.tar ]; then echo "Loading image.tar"; docker load -i ~/cache/docker/image.tar || rm ~/cache/docker/image.tar; fi
+    - |
+      mkdir -p ~/cache/
+      if [ -e ~/cache/docker/image.tar ]; then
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -13,10 +13,18 @@ dependencies:
...
     - docker build -t normandy:build .
+    # pull other docker images used below
+    - docker pull giorgos/takis
     # write the sha256 sum to an artifact to make image verification easier
     - docker inspect -f '{{.Config.Image}}' normandy:build | tee $CIRCLE_ARTIFACTS/docker-image-shasum256.txt
-    # Get MaxMind GeoIP database
+    # get MaxMind GeoIP database
     - cd ~/cache/ && ~/normandy/bin/download_geolite2.sh
+    # clean up old cached Docker image and save the new one
+    - >
```
relud: why not use `- |` and leave out the `;`'s ?

relud:
comment is non-blocker, r+

circle.yml
mythmon: Ignorance. I didn't know that was an option. I'll try it out

mythmon:
This broke because `python:3.5.1-slim` has been removed from Dockerhub. #225 fixes that, and I'll rebase this after it has landed.

PR #226, opened 2016-08-19
Remove navigator field from heartbeat action

brittanystoroz:
Removing the optional `navigator` prop from the show heartbeat action as it was causing problems in the addon. Also fixed some indentation while I was in there.


mythmon:
Looks good. r+ when CI is ready

PR #227, opened 2016-08-19
Cache action requests to the server in the self-repair runner.

Osmose:
Splits the loadAction function in two to make testing easier and
the design cleaner.

Fixes bug 1293475

@rehandalal r?


rehandalal:
this lgtm. r+

PR #228, opened 2016-08-19
Minify webpack code in prod

mythmon:
In my testing, this takes the control bundle from 2.29MB to 727KB in production. It should also silence all the prop types in production. They are still there, just quieter.

@Osmose r?

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -46,7 +46,7 @@ test:
         echo Starting Firefox
         firefox localhost:9876
       ) &
-      bin/ci/docker-run.sh -p 9876:9876 node bin/ci/karma-ci.js
+      bin/ci/docker-run.sh -e NODE_ENV=production -p 9876:9876 node bin/ci/karma-ci.js
     # Start the app, and run acceptance tests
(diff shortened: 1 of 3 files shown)


mythmon:
Note: This also should be running the JS tests on minified code in CI.

circle.yml
```diff
@@ -46,7 +46,7 @@ test:
         echo Starting Firefox
         firefox localhost:9876
       ) &
-      bin/ci/docker-run.sh -p 9876:9876 node bin/ci/karma-ci.js
+      bin/ci/docker-run.sh -e NODE_ENV=production -p 9876:9876 node bin/ci/karma-ci.js
```
Osmose: This doesn't have any effect because the karma tests use a different webpack config than the main one (I still see proptype errors in my test runs). Which is a separate bug. We'll want this in the long term though.

Osmose:
I kinda wish our webpack config was a bit more readable than it is, but that's for another PR.

This looks fine, and when run locally makes much smaller and faster-executing JS. r+. Nice work!

PR #229, opened 2016-08-20
Add Redux DevTools

Osmose:
Adds Redux DevTools to the control interface to make debugging easier. I've already used these to track down a bug, they're incredibly useful.

Also had to add the `PRODUCTION` and `DEVELOPMENT` globals to eslint so that it wouldn't be angry. Also also, I noticed that we weren't emitting sourcemaps in development mode, so I tacked on a commit adding that now that we can distinguish between the two.

@mythmon r?


Osmose:
Tests are failing because Karma isn't using the same config as the main Webpack config.

Osmose:
Okay. So the karma tests use the same webpack config that we use for building the site now, which fixes the issues with `DEVELOPMENT` not being defined. It's a little janky because it doesn't use the action config for building and testing the actions, but that was true before this PR anyway so whatever. Getting webpack and karma to play nicely is a bit rough.

Once I got that done, I got test failures from fetch 404ing. Webpack was injecting `isomorphic-fetch` by surrounding the module in a closure, which means that `fetch-mock`'s changes to `window.fetch` weren't visible to the code under test.

We're not running our tests in Node anymore, and the control interface probably doesn't need to support Firefox versions old enough to not support fetch, so I removed `isomorphic-fetch`.

The build keeps failing to download dependencies, not entirely sure why. Will look into this, but I think the code is ready.

Osmose:
Rebase fixed everything!

docs/dev/workflow.rst
```diff
@@ -163,3 +163,12 @@ To add a new action:
 
 To update an existing action, follow steps 4 and 5 above after making your
 changes.
+
+Redux DevTools
+--------------
+The control interface includes the `Redux DevTools`_ in development mode to help
+debug issues. To toggle the DevTools, hit ``Ctrl-H``. You can change the side of
+the screen the tools are docked on using ``Ctrl-Q``, and can resize the tools by
```
mythmon: `Ctrl-Q` is exit for most Firefox users.

mythmon:
Besides having a bad hotkey for docking the tools, this looks good to me.

I think that the devtools code will still be included in our production bundle, which makes it bigger. I think we can't really fix this without webpack 2.0. It also isn't active, so it won't slow down or bloat the app, so it should be fine.

mythmon:
r+

docs/dev/workflow.rst
Osmose: Whoops.

mythmon:
@Osmose any news here? Do you want to merge it with the bad hotkey? Or change it?

Osmose:
I'm willing/planning to change it but honestly I don't think anyone is even going to use that hotkey, so if you're fine with it let's merge it.

PR #232, opened 2016-08-25
control: Check for UITour on preview page, and give help if not available

mythmon:
I put a little implicit state machine in the previewer, it seems to work alright. The `ExecuteStatus` component is a bit verbose. I'd be interested in any ideas to trim it down a bit.

@brittanystoroz r?


brittanystoroz:
This is nice! Small merge conflict that needs to be resolved.

Re: cutting down the `ExecuteStatus` component, it would be nice if the switch case was just handling the inner text of the element. The only thing that's really changing about the element itself is the className on the icon. (class list includes `fa-circle` if the status is `executed`, otherwise it includes `fa-circle-thin`). You could do what you did with the `statusColor` variable and set a default `iconClass` variable to `fa-circle-thin`, then only change it under the `executed` case. Then you could at least move the span & icon elements down [here](https://github.com/mythmon/normandy/blob/c857575d9c98aa7539f3c96120cc4c3f6c8b854a/client/control/components/RecipePreview.js#L139). I don't know if it'll actually save you *lines* of code, but it'll certainly cleanup some of the repeat.

brittanystoroz:
Also, does the `statusEl` need to be wrapped in a `span` if we're already wrapping it in a `div`? I can't quite remember why I set it up that way, probably a styling thing. But it seems a little unnecessary looking at it now.

mythmon:
`statusEl` is wrapped in a `<span>` because we can't do `statusEl = <i className="fa fa-circle-thin pre" /> Initializing;`. Originally it was `[<i className="fa fa-circle=thin pre" />, 'Initializing']`, which works, but React got grumpy about not setting a key on the icon, and it was just a little messier.

You're write, moving the text and icon into prefs instead of repeating the JSX all the time helped. Want to merge this now?

client/control/components/RecipePreview.js
```diff
@@ -82,4 +99,52 @@ class RecipePreview extends React.Component {
...
+  switch (status) {
+    case 'start':
+    case 'uitourFound': {
+      statusColor = 'blue';
+      statusText = 'Initializing';
+      break;
+    }
+    case 'attempting': {
+      statusColor = 'green';
+      statusIcon = 'fa-circle-thin';
```
brittanystoroz: This should be `circle-thin` since you're already adding the `fa-` in line 138

mythmon: Nice catch!

brittanystoroz:
r+

PR #237, opened 2016-08-27
Fix webpack errors during tests

Osmose:
This does two things:

- Removes the config that silences _all_ webpack output during test runs, and adds a config that only shows webpack middleware errors (and a few minor status lines from webpack itself).
- Changes the regexes used in `index.js` files in the tests to look for files that start with `test_`, and renames all the test files to match this. The tests were showing a warning message due to entry points being pulled in by the `index.js` files (I think it was the `index.js` files themselves, but I'm not sure). This fixes that by requiring test files to explicitly start with a `test_` prefix.

r?


brittanystoroz:
I'm getting the following errors when I run the tests:

```bash
ERROR in ./~/enzyme/build/react-compat.js
Module not found: Error: Cannot resolve module 'react/addons' in /Users/mozilla/github/mozzy/normandy/node_modules/enzyme/build
 @ ./~/enzyme/build/react-compat.js 40:16-39 41:46-69

ERROR in ./~/enzyme/build/react-compat.js
Module not found: Error: Cannot resolve module 'react/lib/ReactContext' in /Users/mozilla/github/mozzy/normandy/node_modules/enzyme/build
 @ ./~/enzyme/build/react-compat.js 42:23-56

ERROR in ./~/enzyme/build/react-compat.js
Module not found: Error: Cannot resolve module 'react/lib/ExecutionEnvironment' in /Users/mozilla/github/mozzy/normandy/node_modules/enzyme/build
 @ ./~/enzyme/build/react-compat.js 72:4-45
```

Although they don't seem to prevent the tests from actually running. Did you get these same errors er nah? A little googling sent me to this [issue](https://github.com/airbnb/enzyme/issues/47) and it looks like the solution would be to add the following to the webpack object in our `karma.conf.js`:

```javascript
      externals: {
        'react/addons': true,
        'react/lib/ReactContext': true,
        'react/lib/ExecutionEnvironment': true,
      },
```

Osmose:
Ah, yeah. It didn't occur to me that I should add it to the config in `karma.conf.js` because I was busy with #229, which adds those externals to the webpack config and uses it in place of the custom config for Karma, but that'd be a fine solution for this PR too.

I can add them next time I have a chance to work on this.

Osmose:
Updated!

brittanystoroz:
Just needs a rebase, otherwise r+

PR #239, opened 2016-08-28
Bind RecipeForm.changeAction to the class instance.

Osmose:
I noticed locally that the recipe form wasn't showing the action forms when I changed the action when creating a new recipe, and tracked it down to not binding this function properly. The dev server doesn't have this problem, weirdly. Not sure why I'm only seeing it locally.

I wanted to create a test for this, but the right test IMO is to test that the console-log form is shown when changing the select box via a user interaction, and we don't really have a great setup yet for running those kind of tests. It's also awkward to this via unit tests because the component as written is generic, so a test that relies specifically on console-log seems to not fit the unit tests for the component.

I'm open to feedback on better ways to test this. Otherwise, r?

### DIFF
--- a/client/control/components/RecipeForm.js
+++ b/client/control/components/RecipeForm.js
@@ -33,6 +33,7 @@ export class DisconnectedRecipeForm extends React.Component {
     };
 
     this.submitForm = ::this.submitForm;
+    this.changeAction = ::this.changeAction;
   }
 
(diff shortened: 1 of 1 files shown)


brittanystoroz:
r+

I believe I have some local branch with WIP tests for this kind of interaction. I will push and PR what I had started and continue the conversation there.

PR #240, opened 2016-08-28
Log navigator.oscpu via the client classification endpoint.

Osmose:
Prior to implementing operating system filtering, I'd like to collect some data from users about what the value of `navigator.oscpu` is, since we'll be using that value to filter by OS.

This is marked as Do Not Merge because I'd like a data review from @rjweiss first, which happens via Bugzilla flags AFAICT. I (or anyone, really) can remove that tag once the data review on Bugzilla is approved.

Fixes [bug 1296028](https://bugzilla.mozilla.org/show_bug.cgi?id=1296028).

@mythmon r? Since you're probably most familiar with how logging should be done.

### DIFF
--- a/normandy/settings.py
+++ b/normandy/settings.py
@@ -255,6 +255,7 @@ class Base(Core):
     ACTION_IMPLEMENTATION_CACHE_TIME = values.IntegerValue(60 * 60 * 24 * 365)
     NUM_PROXIES = values.IntegerValue(0)
     API_CACHE_TIME = values.IntegerValue(30)
+    LOG_OSCPU_RATE = values.FloatValue(0.001)
     # Autograph settings
     AUTOGRAPH_URL = values.Value()
     AUTOGRAPH_HAWK_ID = values.Value()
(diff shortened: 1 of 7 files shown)


mythmon:
I like these changes. r+ on this PR, once rweiss approves the data loggin.

normandy/settings.py
```diff
@@ -255,6 +255,7 @@ def LOGGING(self):
     ACTION_IMPLEMENTATION_CACHE_TIME = values.IntegerValue(60 * 60 * 24 * 365)
     NUM_PROXIES = values.IntegerValue(0)
     API_CACHE_TIME = values.IntegerValue(30)
+    LOG_OSCPU_RATE = values.FloatValue(0.001)
```
mythmon: New settings need to go [in the docs](https://github.com/mozilla/normandy/blob/73bd7183f1c4c692308ad866397dc915350cbf42/docs/ops/config.rst).

mythmon:
I added the new setting to the docs. This is ready to merge once CI finishes.

PR #243, opened 2016-08-30
Add test for changing the action field

brittanystoroz:
This adds a test for changing the action field in the recipe forms as brought up in https://github.com/mozilla/normandy/pull/239. This is more of an integration test, I just plopped it into our existing recipe form test file. Maybe we want to split our tests into `unit` and `integration` directories?

### DIFF
--- a/client/control/components/RecipeForm.js
+++ b/client/control/components/RecipeForm.js
@@ -47,7 +47,7 @@ export class DisconnectedRecipeForm extends React.Component {
 
   changeAction(event) {
     const { dispatch, fields } = this.props;
-    const selectedActionName = event.currentTarget.value;
+    const selectedActionName = event.target.value;
 
     dispatch(destroy('action'));
     fields.action.onChange(event);
--- a/client/control/tests/components/test_RecipeForm.js
+++ b/client/control/tests/components/test_RecipeForm.js
@@ -28,4 +28,22 @@ describe('<RecipeForm>', () => {
       <RecipeForm params={{}} location={{ query: '' }} />
     </Provider>);
   });
+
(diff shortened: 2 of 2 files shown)


client/control/tests/components/RecipeForm.js
```diff
@@ -28,4 +28,22 @@ describe('<RecipeForm>', () => {
       <RecipeForm params={{}} location={{ query: '' }} />
     </Provider>);
   });
+
+  describe('changeAction', () => {
+    const store = controlStore();
+    const recipeFormWrapper = mount(<Provider store={store}>
```
Osmose: These will be executed during test discovery instead of during the test itself; you should move them to a `beforeEach` function.

client/control/tests/components/RecipeForm.js
```diff
@@ -28,4 +28,22 @@ describe('<RecipeForm>', () => {
...
+      <RecipeForm params={{}} recipe={{ action: 'show-heartbeat' }} location={{ query: '' }} />
+    </Provider>);
+
+    const disconnectedFormWrapper = recipeFormWrapper.find('DisconnectedRecipeForm').get(0);
+    disconnectedFormWrapper.changeAction({ target: { value: 'console-log' } });
+
+    it('should update the selectedAction in state', () => {
+      expect(disconnectedFormWrapper.state.selectedAction).toEqual({ name: 'console-log' });
```
Osmose: I kinda expected both of these assertions should be part of a single test, rather than split between two tests with a common `beforeEach` block.

brittanystoroz:
Updated

Osmose:
Nice work!

PR #247, opened 2016-08-31
Allow editing actions, regardless of recipe usage

mythmon:
This just makes our lives harder, especially during deploys. Since the actions are part of this repo, I don't think the case we were protecting against is valid anymore.

@Osmose r?


Osmose:
You should also remove the `--no-update` argument to the management command and alter the webpack config to not use that parameter anymore.

Otherwise, I'm r+ on this. Nice work!

PR #259, opened 2016-09-08
Signature verification

mythmon:
I wrote some tests to verify that recipe signatures actually sign the data the recipe is supposed to be. This revealed a problem where recipes auto-signed on save were signing the wrong data. I fixed that bug.

r?

### DIFF
--- a/normandy/recipes/tests/test_models.py
+++ b/normandy/recipes/tests/test_models.py
@@ -1,3 +1,4 @@
+import hashlib
 from datetime import datetime, timedelta
 from unittest.mock import patch
 
@@ -162,11 +163,29 @@ class TestRecipe(object):
         ]
 
         recipe = RecipeFactory(signed=False)
-        assert recipe.signature is None
         recipe.update_signature()
         recipe.save()
         assert recipe.signature is not None
         assert recipe.signature.signature == 'fake signature'
+
+    def test_signatures_update_correctly_on_enable(self, mocker):
+        mock_autograph = mocker.patch('normandy.recipes.models.Autographer')
+
+        def fake_sign(datas):
+            sigs = []
+            for d in datas:
+                sigs.append({'signature': hashlib.sha256(d).hexdigest()})
+            return sigs
+
+        mock_autograph.return_value.sign_data.side_effect = fake_sign
+
+        recipe = RecipeFactory(enabled=False, signed=False)
+        recipe.enabled = True
+        recipe.save()
+        recipe.refresh_from_db()
+
+        assert recipe.signature is not None
+        assert recipe.signature.signature == hashlib.sha256(recipe.canonical_json()).hexdigest()
 
 
 @pytest.mark.django_db
--- a/normandy/recipes/utils.py
+++ b/normandy/recipes/utils.py
@@ -122,27 +122,10 @@ def verify_signature(data, signature, pubkey):
     if isinstance(data, str):
         data = data.encode()
 
-    # add data template
+    # Add data template
     data = b'Content-Signature:\x00' + data
-
-    signature = base64.b64decode(un_urlsafe(signature))
-    pubkey = un_urlsafe(pubkey)
-
+    signature = base64.urlsafe_b64decode(signature)
     verifying_pubkey = ecdsa.VerifyingKey.from_pem(pubkey)
-    verified = verifying_pubkey.verify(
-        signature,
-        data,
-        hashfunc=hashlib.sha384,
-        sigdecode=ecdsa.util.sigdecode_string)
-
-    print(dir(verifying_pubkey))
+    verified = verifying_pubkey.verify(signature, data, hashfunc=hashlib.sha384)
 
     return verified
-
-
-def un_urlsafe(input):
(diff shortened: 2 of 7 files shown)


mythmon:
@chartjes can you review the changes to the contract tests here?

rehandalal:
i couldn't see any issues and circle is happy. r+ from me.

but i'll leave this open so @chartjes can review the contract tests.

package.json
```diff
@@ -26,7 +26,7 @@
     "json-editor": "0.7.28",
     "json-loader": "0.5.4",
     "moment": "2.14.1",
-    "node-sass": "3.9.1",
+    "node-sass": "3.9.3",
```
Osmose: Unrelated?

normandy/recipes/utils.py
```diff
@@ -115,3 +116,33 @@ def sign_data(self, content_list):
...
+
+
+def verify_signature(data, signature, pubkey):
+    if isinstance(data, str):
+        data = data.encode()
+
+    # add data template
+    data = b'Content-Signature:\x00' + data
+
+    signature = base64.b64decode(un_urlsafe(signature))
```
Osmose: Could you just use `base64.urlsafe_b64decode` here?

normandy/recipes/utils.py
```diff
@@ -115,3 +116,33 @@ def sign_data(self, content_list):
...
+        data,
+        hashfunc=hashlib.sha384,
+        sigdecode=ecdsa.util.sigdecode_string)
+
+    print(dir(verifying_pubkey))
+
+    return verified
+
+
+def un_urlsafe(input):
```
Osmose: Could you document this function to explain what it's doing?

normandy/recipes/utils.py
```diff
@@ -115,3 +116,33 @@ def sign_data(self, content_list):
...
+    pubkey = un_urlsafe(pubkey)
+
+    verifying_pubkey = ecdsa.VerifyingKey.from_pem(pubkey)
+    verified = verifying_pubkey.verify(
+        signature,
+        data,
+        hashfunc=hashlib.sha384,
+        sigdecode=ecdsa.util.sigdecode_string)
+
+    print(dir(verifying_pubkey))
```
Osmose: A debug statement?

normandy/recipes/utils.py
```diff
@@ -115,3 +116,33 @@ def sign_data(self, content_list):
...
+
+    return verified
+
+
+def un_urlsafe(input):
+    input = input.replace("_", "/")
+    input = input.replace("-", "+")
+    if len(input) % 4 > 0:
+        input += "=" * (4 - len(input) % 4)
```
Osmose: Wait, we're decoding stuff but you still need to pad it? That seems odd to me.

normandy/recipes/utils.py
rehandalal: i thought this was doing the opposite of `urlsafe_b64decode` but maybe i'm wrong.

normandy/recipes/tests/test_models.py
```diff
@@ -172,12 +173,30 @@ def test_update_signature(self, mocker):
         ]
 
         recipe = RecipeFactory(signed=False)
-        assert recipe.signature is None
```
Osmose: So we no longer expect signatures to be empty at first?

normandy/recipes/utils.py
mythmon: Autograph is weird. It uses this not-quite-base64 thing that uses `_-` instead of `/+`, and doesn't include the padding, which we need for `base64.b64decode`. Maybe `base64.urlsafe_b64decode` will deal with all this for us, I'll give it shot. Otherwise I'll add a docstring to this function.

package.json
mythmon: This was on master, and was needed to get the tests to pass. I can probably get rid of it with some rebasing.

normandy/recipes/utils.py
mythmon: Oops

normandy/recipes/tests/test_models.py
mythmon: Not when we have a working (or mocked) Autograph, since we automatically sign in `save()` now.

normandy/recipes/utils.py
Osmose: They _don't_ use padding? whyyyyyyyyyyyyyyyyyyyyyyy

Osmose:
I'm r+wc for the non-contract-tests bits, I'm good with either removing the `un_urlsafe` function or documenting it. A test or two for it would be nice but not required if it stays.

mythmon:
Turns out `un_urlsafe` isn't needed. We can use `base64.urlsafe_b64decode`, and ecdsa will automatically detect urlsafe b64, it seems. I'll be updating the Autograph code (which this is based on).

chartjes:
Nicely done with the contract test updates -- it hadn't occurred to me to explicitly check that requests were okay, so raise_for_status() is a nice addition.

PR #263, opened 2016-09-19
Test that Vary: Cookie isn't sent on the hot path

mythmon:
Having a `Vary` header with a value that includes "Cookie" will destroy any performance we might hope to have. This PR adds an external test to verify we don't send this kind of header on performance sensitive servers.

@chartjes r? (please use the new Github review feature, since merging PRs is now gated on having a approved review)


mythmon:
Thanks!

I'm using my admin rights to merge this without a qualifying review.

PR #265, opened 2016-09-20
Add driver & runner tests

brittanystoroz:
Handful of extra driver/repair runner tests

### DIFF
--- a/client/selfrepair/tests/test_normandy_driver.js
+++ b/client/selfrepair/tests/test_normandy_driver.js
@@ -43,6 +43,7 @@ describe('Normandy Driver', () => {
     const driver = new NormandyDriver();
 
     afterEach(() => {
+      expect(fetchMock.calls().unmatched).toEqual([]);
       fetchMock.restore();
     });
 
@@ -123,6 +124,7 @@ describe('Normandy Driver', () => {
 
     it('will not log debug level unless in testing mode', () => {
       spyOn(console, 'log');
+      driver.testing = false
       driver.log('lorem ipsum');
       expect(console.log.calls.count()).toEqual(0);
     });
--- a/client/selfrepair/tests/test_normandy_driver.js
+++ b/client/selfrepair/tests/test_normandy_driver.js
@@ -1,4 +1,7 @@
+import fetchMock from 'fetch-mock';
+
 import NormandyDriver from '../normandy_driver.js';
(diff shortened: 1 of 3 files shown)


client/selfrepair/tests/test_normandy_driver.js
```diff
@@ -36,21 +39,106 @@ describe('Normandy Driver', () => {
     });
   });
 
+  describe('saveHeartbeatFlow', () => {
+    const driver = new NormandyDriver();
+
+    afterEach(() => {
+      fetchMock.restore();
```
mythmon: Should we do this in `afterEach` or `beforeEach`? I don't really care either way, but we should probably be consistent. We did it in `beforeEach` in [another test](https://github.com/brittanystoroz/normandy/blob/149c63c7afb47268c723f0fbbf0bffa2699c5814/client/control/tests/actions/test_controlActions.js#L13-L20).

mythmon: We should have something like `expect(fetchMock.calls().unmatched).toEqual([])`, to check that the tests didn't make any errant fetch calls.

client/selfrepair/tests/test_normandy_driver.js
```diff
@@ -36,21 +39,106 @@ describe('Normandy Driver', () => {
...
+      expect(UUID_ISH_REGEX.test(uuid)).toBe(true);
+    });
+  });
+
+  describe('log', () => {
+    const driver = new NormandyDriver();
+
+    it('will not log debug level unless in testing mode', () => {
+      spyOn(console, 'log');
+      driver.log('lorem ipsum');
```
mythmon: Should probably set `driver.testing = false` here to be explicit (also, to match the "should POST flow data to input" test).

client/selfrepair/tests/test_normandy_driver.js
brittanystoroz: Hmmm I think it actually makes more sense in `afterEach`. Moving it to `beforeEach` would look weird because we still want to call `fetchMock.get(url)` there.  (In the other example you linked to, it didn't matter as much because our `fetchMock.get` calls varied for each test and had to be defined separately in each `it` block.)

I'll switch the other test case to also use `restore()` in `afterEach`.

mythmon:
CI has some lint nits to fix (note that warning don't need to be fixed, only errors). Besides that, r+

PR #266, opened 2016-09-20
Test server heartbeat

mythmon:
Recently, @relud added support for running the contract tests during deploys. Failures in the contract tests will mark a deploy as failed, and are generally much nicer than finding the same problems after the deploy.

This PR adds a test that checks the `/__heartbeat__` endpoint of the server, which contains several internal checks.

@chartjes r?


mythmon:
@Osmose can you take a quick look at the settings changes here?

mythmon:
The reason that @chartjes's approval didn't mark this as ok to merge is because he doesn't have write access to the repo, if anyone is curious. A review from @Osmose should get this in a mergeable state.

PR #272, opened 2016-09-21
Optional TestRail Support

chartjes:
@mythmon r?


docs/dev/api-tests.rst
```diff
@@ -8,6 +8,50 @@ To run these tests, use the following command from the root project directory.
...
+
+    py.test --server=<server> -v \
+    --no-ssl-cert-check \
+    --testrail=contract-tests/testrail.cfg contract-tests/
+
+The values for ``<server>`` are the same as before.
+
+`--no-ssl-cert-check` tells pytest to not check if the SSL certificate for the
+TestRail install is valid -- we are using a self-signed certificate and the
+underlying code throws an exception unless we tell it not to check.
```
mythmon: This is a huge red flag to me. I'm not comfortable integrating with a system that uses a self signed certificate.

chartjes: We host the service ourselves, you need an account on TestRail and VPN access just to use it

chartjes: By "we" I mean Mozilla

docs/dev/api-tests.rst
```diff
@@ -8,6 +8,50 @@ To run these tests, use the following command from the root project directory.
...
+If you want to send results of the test run to TestRail, you need the following
+pre-requisites:
+
+1. You have VPN access to the staging environment
+2. You have an account on https://testrail.stage.mozaws.net
+3. Copy `testrail.cfg.dist` to `testrail.cfg`
+4. Modify that TestRail configuration file with your email, TestRail password and TestRail user ID
+
+
+Then, for any test case you want to report results for, make sure the
```
mythmon: Nit: there is an extra blank line here.

docs/dev/api-tests.rst
relud: Is this the testrail instance being run by @milescrabill ? because his should have a valid cert, and if not that can be fixed.

docs/dev/api-tests.rst
```diff
@@ -8,6 +8,50 @@ To run these tests, use the following command from the root project directory.
...
+.. code-block:: python
+
+    from pytest_testrail.plugin import testrail
+
+and then you use a decorator to indicate this test case will be reporting
+results. Here's an example:
+
+.. code-block:: python
+
+    @testrail('C5603')
```
mythmon: Where does `C5603` come from? How do I get an ID for a new test if I write one?

docs/dev/api-tests.rst
relud: also, does this disable all cert checking, because that would be bad...

chartjes:
It's an optional switch -- by default it checks for a valid certificate. Once the TestRail server has a non-self-signed SSL cert you won't need that switch

docs/dev/api-tests.rst
mythmon: Needing an account is not a argument against having good SSL. In fact, it is an argument for why we *need* good SSL. If we can't validate the site, an attacker could intercept the account credentials we send to the site.

Being on the VPN is a stronger argument, but VPNs aren't perfect, and there a many situation where this could still cause issues. Assuming everything else is working securely is not a good security practice, and [building redundant security into every layer](https://en.wikipedia.org/wiki/Defense_in_depth_(computing)) is a much better approach.

Having bad SSL on a site we rely on is poor form in my book. Additionally, disabling cert checking on globally in the tests compromises other sites used in the tests as well. If this is something that is temporary, and a production version of TestRail is coming, then I can let this by. If this is permanent, I don't want to rely on insecure sites for our QA infrastructure.

Is this just a temporary solution until we can get a better instance of TestRail?

milescrabill: @relud The Testrail instance I will own is not yet operational. It's going to be a full replacement of the existing instance and will be in prod (per sec. advice). It will _not_ have a self-signed cert.

chartjes: Perhaps I did a very bad job of explaining this, but the --no-ssl-cert-check is completely optional and we won't have to use it once we have a TestRail instance without a self-signed certificate.

requirements.txt
```diff
@@ -78,9 +78,8 @@ pytz==2015.7 \
     --hash=sha256:fbd26746772c24cb93c8b97cbdad5cb9e46c86bbdb1b9d8a743ee00e2fb1fc5d \
     --hash=sha256:99266ef30a37e43932deec2b7ca73e83c8dbc3b9ff703ec73eca6b1dae6befea \
     --hash=sha256:8b6ce1c993909783bc96e0b4f34ea223bff7a4df2c90bdb9c4e0f1ac928689e3
-requests==2.9.1 \
-    --hash=sha256:113fbba5531a9e34945b7d36b33a084e8ba5d0664b703c81a7c572d91919a5b8 \
-    --hash=sha256:c577815dd00f1394203fc44eb979724b098f88264a9ef898ee45b8e5e9cf587f
+requests==2.11.1 \
```
chartjes: Yes, the pytest-testrail plugin requires that version of Requests

docs/dev/api-tests.rst
chartjes: Fixed.

docs/dev/api-tests.rst
chartjes: The ID is generated by the TestRail admin when you create the test case. I added some text explaining that.

chartjes:
Yes, the --no-ssl-cert-check stuff is 100% temporary

PR #274, opened 2016-09-26
Sigs on new recipes

mythmon:
This fixes an issue where newly created recipes had the wrong signature. The signature was generated based on data that did not include an ID, since Django hadn't generated one yet.

r?

### DIFF
--- a/normandy/conftest.py
+++ b/normandy/conftest.py
@@ -1,3 +1,5 @@
+import hashlib
+
 from django.core.cache import caches
 from django.conf import settings
 
@@ -33,3 +35,19 @@ def geolocation():
         skip_except_in_ci()
     else:
         return geolocation_module
+
+
+@pytest.fixture
+def mocked_autograph(mocker):
+    mocked = mocker.patch('normandy.recipes.models.Autographer')
+
+    def fake_sign(datas):
+        sigs = []
+        for d in datas:
+            sigs.append({
+                'signature': hashlib.sha256(d).hexdigest()
+            })
+        return sigs
+
+    mocked.return_value.sign_data.side_effect = fake_sign
+    return mocked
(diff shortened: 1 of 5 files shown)


normandy/recipes/tests/__init__.py
```diff
@@ -44,7 +44,7 @@ class Meta:
     @factory.post_generation
     def signed(self, create, extracted=False, **kwargs):
```
Osmose: A comment somewhere in here describing why you're using the canonical JSON for the signature would be useful.

normandy/conftest.py
```diff
@@ -33,3 +35,19 @@ def geolocation():
...
+
+@pytest.fixture
+def mocked_autograph(mocker):
+    mocked = mocker.patch('normandy.recipes.models.Autographer')
+
+    def fake_sign(datas):
+        sigs = []
+        for d in datas:
+            sigs.append({
+                'signature': hashlib.sha256(d).hexdigest()
```
Osmose: You're using `sha256` in enough places that I wonder if you should factor it to a common place in case we change how signatures are generated for testing in the future? I'm not entire sure about that though.

normandy/recipes/models.py
```diff
@@ -154,6 +154,11 @@ def save(self, *args, skip_last_updated=False, **kwargs):
 
             if should_sign:
                 try:
+                    if self.id is None:
+                        # Save now, to get an ID for the signature.
+                        super().save(*args, **kwargs)
+                        # Change from insert to update, so the save() call at the end doesn't fail
+                        kwargs['force_insert'] = False
```
Osmose: Why do we need to save the object after creating the signature?

Osmose: The save-sign-save flow needs to be wrapped in a transaction.

PR #276, opened 2016-09-27
Fix Karma failing on webpack

mythmon:
- [x] Make Karma fail if plugins can't be loaded
- [x] Fix webpack in the tests
- [x] Fix the failing test

PR #278, opened 2016-09-28
self-repair: Send Accept: application/json when fetching action data

### LINKED ISSUE #277: 43.0.1 is not sampling correctly
The recipe "Out of Date Test 2" is attempting to sample English users running Firefox version 43.0.1 at a rate of 20%. We have evidence from telemetry that there are about 5.8 million people running 43.0.1. We have only gotten 300 offers for survey, which is significantly lower than 20% of 5.8 million, even accounting for locale.

The recipe's filter expression is:

```
(
   [normandy.userId, normandy.recipe.id]|stableSample(0.2)
   && normandy.locale in ['en-US', 'en-AU', 'en-CA', 'en-GB', 'en-NZ', 'en-ZA']
   && normandy.version == '43.0.1'
)
```

mythmon:
Fixes #277

### DIFF
--- a/client/selfrepair/tests/test_self_repair_runner.js
+++ b/client/selfrepair/tests/test_self_repair_runner.js
@@ -17,12 +17,9 @@ const UUID_ISH_REGEX = /^[a-f0-9-]{36}$/;
 
 
 describe('Self-Repair Runner', () => {
-  beforeEach(() => {
-    fetchMock.restore();
-  });
-
   afterEach(() => {
     expect(fetchMock.calls().unmatched).toEqual([]);
+    fetchMock.restore();
   });
 
   describe('classifyClient', () => {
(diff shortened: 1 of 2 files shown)


mythmon:
Keep  in mind that the test added here is not actually validated in CI because of #276.

client/selfrepair/tests/test_self_repair_runner.js
```diff
@@ -17,10 +17,14 @@ const UUID_ISH_REGEX = /^[a-f0-9-]{36}$/;
 
 
 describe('Self-Repair Runner', () => {
-  afterEach(() => {
+  beforeEach(() => {
```
Osmose: Why before? Wouldn't we want to stop mocking fetch at the end of each test?

mythmon: I mis-remembered [this conversation](https://github.com/mozilla/normandy/pull/265#discussion_r79666402) with @brittanystoroz. I thought we had agreed that `beforeEach` was the way to do it, and so I changed it. But I was wrong, so I'll change it back.

PR #280, opened 2016-09-29
Include current git tag in __version__, if available

mythmon:
r?

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -10,4 +10,3 @@ site-packages
 venv
 webpack-stats.json
 webpack-stats-actions.json
-Dockerfile
--- a/normandy/health/api/views.py
+++ b/normandy/health/api/views.py
@@ -13,31 +13,41 @@ from normandy.base.decorators import short_circuit_middlewares
 
 
 _commit = None
+_tag = None
 
 
-def get_commit():
-    global _commit
+def get_version_info():
+    global _commit, _tag
     if _commit is None:
         path = os.path.join(settings.BASE_DIR, '__version__', 'commit')
         try:
             with open(path) as f:
-                _commit = f.read().strip()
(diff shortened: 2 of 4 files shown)


.dockerignore
```diff
@@ -10,4 +10,3 @@ site-packages
 venv
 webpack-stats.json
 webpack-stats-actions.json
-Dockerfile
```
mythmon: Turns out [you can't actually ignore `Dockerfile`](https://www.ctl.io/developers/blog/post/more-docker-image-cache-tips/), since it is needed for the build. Including it here does nothing.

normandy/health/api/views.py
```diff
@@ -13,31 +13,41 @@
 
 
 _commit = None
+_tag = None
 
 
-def get_commit():
-    global _commit
+def get_version_info():
+    global _commit, _tag
```
Osmose: Man we could sure use a `cached_property` for functions.

Probably on the third time we do this.

mythmon:
Tests ^

normandy/health/tests/test_api.py
```diff
@@ -26,3 +28,53 @@ def test_lbheartbeat_makes_no_db_queries(client):
...
+    def version_info_dir(self, tmpdir, settings):
+        views._tag = None
+        views._commit = None
+        settings.BASE_DIR = tmpdir.strpath
+        return tmpdir.mkdir('__version__')
+
+    @pytest.fixture
+    def commit_file(self, version_info_dir):
+        f = version_info_dir.join('commit')
+        f.write('bc685e4be05a182ae819990509c92affa3d882ab\n')
```
Osmose: I've never been able to come up with a good way to pass parameters to pytest fixtures.

PR #281, opened 2016-09-30
Organize dependencies into direct dependencies and sub dependencies

mythmon:
[Inspired by Kuma](https://github.com/mozilla/kuma/tree/master/requirements).

It's a bit hard to tell, but I didn't add any requirements, or change any versions or hashes. I only moved them around, and removed the following

* `boto==2.39.0` - from when we were going to store things in S3 directly
* `docutils==0.12` - I'm not sure about this one, but everything works with out it
* `wheel==0.26.0` - I think this snuck in with a `pip freeze`
* `django-adminplus==0.5` - Not using this in the admin anymore.

I would have also removed `django-countries==3.4.1`, but it is being used in a migration. We could eliminate this if we squished some migrations, but I don't want to do that in this PR.

I tested this by

* Pip installing everything into an empty virtualenv
* Running the tests
* Running `./manage.py collectstatic`
* Building the docker image
* Building the docs
* Running the contract tests against `./bin/runsslserver.sh`

r?


mythmon:
Well, I guess I was wrong about docutils.

Dockerfile
```diff
@@ -10,10 +10,10 @@ RUN curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add -
     echo 'deb-src https://deb.nodesource.com/node_4.x jessie main' >> /etc/apt/sources.list.d/nodesource.list && \
     apt-get update && apt-get install -y nodejs
 
-COPY ./requirements.txt /app/requirements.txt
+COPY ./requirements /app/requirements
 COPY ./package.json /app/package.json
 RUN pip install -U 'pip>=8' && \
-    pip install --upgrade --no-cache-dir -r requirements.txt && \
+    pip install --upgrade --no-cache-dir -r requirements/default.txt -c requirements/constraints.txt && \
```
rehandalal: neat!

PR #282, opened 2016-10-01
Do not include hashes in client assets in development.

Osmose:
Currently, built assets have hashes in their filenames by default.
When working locally and using the watch command, this has the side
effect of generating a bunch of nearly-identical assets with different
filenames, and over time can take up a lot of space.

In a development environment there's no reason to use unique hashes in
asset filenames, so disabling this behavior will help avoid the buildup
of assets.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -14,7 +14,7 @@ const production = process.env.NODE_ENV === 'production';
 var plugins = [
   new BundleTracker({ filename: './webpack-stats.json' }),
   new webpack.optimize.OccurrenceOrderPlugin(true),
-  new ExtractTextPlugin('[name]-[hash].css'),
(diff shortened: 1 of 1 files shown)


Osmose:
Failed tests, yay!

Also locally I've noticed a bit of slowness in loading since switching to this, but it might be unrelated (I'm also playing around with the new Redux form so I'm not sure whether this change is the one that caused it).

Osmose:
Whoo, failures were transient.

A double check on whether my slowdown issues are related to this PR would be appreciated though.

brittanystoroz:
Just needs rebase r+

brittanystoroz:
Oh just kidding I guess, weird. r+ merging!

PR #283, opened 2016-10-03
Stub shield-study action

brittanystoroz:
Stubs in an action for allowing users to participate in a SHIELD study.


mythmon:
Test failures seem to be a Python package hash change. I'll look into this.

mythmon:
The hash change is gone now, but the contract tests need updated now.

mythmon:
I think that's everything

Issue #279, opened 2016-09-29
Validate localstorage in self-repair runner


mythmon:
The self-repair runner should validate that local storage is reliable, and values stored there actually persist. This is to avoid a problem where storage is not actually resilient for some users (some privacy-focused addons cause this).

One way to do this would be to store a counter in localstorage that increments every time the runner starts. Any actions that rely on reading data from localstorage should check this counter and make sure it is at least 2 before proceeding.

We probably shouldn't count past 2 here, actually. Values over 2 won't really help us. But we do need more information than a simple boolean, since we need to check that the values persists at least two runs (the first run sets the value, and in the second run the value is persisted).

Osmose:
Should we also update actions that use the storage and have them abort execution early if storage isn't available?

mythmon:
We could do this by making any promises for the storage driver fail if storage isn't available. That should make any actions abort early. We should test this to be sure though.

Osmose:
This was resolved by #313.

PR #291, opened 2016-10-12
Add storage check before running shield study action

brittanystoroz:
We don't want the shield study action to run if it's already been previously displayed to a user. I used a similar approach as the heartbeat action, saving a value to localStorage that denotes it's been shown.

I took a tiny stab at validating the localstorage with the strategy @Osmose mentioned [here](https://github.com/mozilla/normandy/issues/279#issuecomment-250794727) (I didnt quite understand the counter option, maybe someone can clarify it for me a little further and it's something I can take care of.) 

If you toggle the `dom.storage.enabled` pref in firefox you'll see the action execution stops if localstorage is unavailable. I tried to write a test for this but I'm having trouble writing a jasmine test that is _expected_ to fail. [My attempt](https://github.com/brittanystoroz/normandy/blob/82266f2030c81ce6cb804a638f516d60a3aa08bd/client/actions/tests/test_shield-study.js#L27-L34) seems to run the action execution and err out (as it should), but before it actually gets a chance to run the assertion.

Thoughts??

### DIFF
--- a/client/actions/shield-study/index.js
+++ b/client/actions/shield-study/index.js
@@ -1,13 +1,34 @@
 import { Action, registerAction } from '../utils';
 
 export default class ShieldStudyAction extends Action {
-  constructor(normandy, recipe) {
-    super(normandy, recipe);
-    this.storage = normandy.createStorage(recipe.id);
+  setHasBeenShown() {
+    this.storage.setItem('studyHasBeenShown', true);
+  }
+
+  async getHasBeenShown() {
+    const hasBeenShown = await this.storage.getItem('studyHasBeenShown');
+    return hasBeenShown;
   }
 
   async execute() {
+    try {
+      this.storage = await this.normandy.createStorage(this.recipe.id);
+    } catch (error) {
+      throw new Error(error);
+    }
+
+    const hasBeenShown = await this.getHasBeenShown();
+    const shouldShowStudyPrompt = (
+      this.normandy.testing
+      || hasBeenShown === null
+    );
+
+    if (!shouldShowStudyPrompt) {
+      return;
+    }
+
     this.normandy.showStudyConsentPage(this.recipe.arguments);
+    this.setHasBeenShown();
   }
 }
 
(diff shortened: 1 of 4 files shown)


client/actions/shield-study/index.js
```diff
@@ -1,13 +1,34 @@
...
+  async getHasBeenShown() {
+    const hasBeenShown = await this.storage.getItem('studyHasBeenShown');
+    return hasBeenShown;
   }
 
   async execute() {
+    try {
+      this.storage = await this.normandy.createStorage(this.recipe.id);
+    } catch (error) {
+      throw new Error(error);
```
mythmon: You don't need to catch and then rethrow the error. You can just let it fail, leaving out all the try/catch stuff.

brittanystoroz:
Ahhh I see, I will rethink and try again

brittanystoroz:
Here is a second attempt at checking the storage durability. Let me know if this looks more appropriate.

I'm still running into a problem with this [test](https://github.com/brittanystoroz/normandy/blob/82266f2030c81ce6cb804a638f516d60a3aa08bd/client/actions/tests/test_shield-study.js#L27-L34) erring out before confirming the assertion that it threw an error.

client/actions/shield-study/index.js
```diff
@@ -6,8 +6,34 @@ export default class ShieldStudyAction extends Action {
...
+  async getHasBeenShown() {
+    const hasBeenShown = await this.storage.getItem('studyHasBeenShown');
+    return hasBeenShown;
+  }
+
   async execute() {
+    const storageIsDurable = await this.storage.isDurable();
+
+    if (!storageIsDurable && !this.normandy.testing) {
+      throw new Error('Storage durability unconfirmed');
```
brittanystoroz: Good call. My only problem with that would be that I'm using `this.getItem` in the `_setDurabilityStatus` method. So if `getItem` returned `null` or rejected, we would never actually be able to increment the durability status value.

Any thoughts on getting around that? I could add a check in `getItem` that wouldn't reject if we were getting the `storageDurability` key?

mythmon:
@Osmose I added a commit here so that all actions can take advantage of this. Can you review it?

mythmon:
Oops, I didn't notice this was targeted at the shield-study branch. Closing so we can move it over to master.

PR #296, opened 2016-10-18
Fix heartbeat event stuff.

Osmose:
There's a potential race condition in Heartbeat's event emitter, where we might not register events before the heartbeat prompt is shown. This fixes that by using a special event emitter for heartbeat that dispatches each event once, and will run callbacks immediately if the event they're hooked to has already run.

Also includes a fix for [bug 1302791](https://bugzilla.mozilla.org/show_bug.cgi?id=1302791) because I discovered it while originally writing this patch.

This is marked WIP because it lacks tests and should probably be split into two commits, but some feedback on the chosen solution would be appreciated.

@mythmon f?

### DIFF
--- a/client/selfrepair/normandy_driver.js
+++ b/client/selfrepair/normandy_driver.js
@@ -1,4 +1,3 @@
-import EventEmitter from 'wolfy87-eventemitter';
 import uuid from 'node-uuid';
 
 import Mozilla from './uitour.js';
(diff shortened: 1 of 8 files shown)


Osmose:
Oh yeah, this also adds the Engaged event so that we're capturing when users click the engagement button.

client/selfrepair/normandy_driver.js
```diff
@@ -32,23 +31,73 @@ class LocalStorage {
...
     // happened.
     this._uitour.observe((eventName, data) => {
       if (eventName.startsWith('Heartbeat:')) {
-        const flowId = data.flowId;
         const croppedEventName = eventName.slice(10); // Chop off "Heartbeat:"
-        if (flowId in this._heartbeatCallbacks) {
-          this._heartbeatCallbacks[flowId](croppedEventName, data);
+        const callback = this._heartbeatCallbacks[data.flowId];
+        if (callback !== undefined) {
+          callback(croppedEventName, data);
```
mythmon: This is just fixing `_heartbeatCallbacks` to be an object instead of an array, because it is used as an object, not an array, right? There aren't any behavioral changes?

client/selfrepair/normandy_driver.js
```diff
@@ -1,4 +1,3 @@
-import EventEmitter from 'wolfy87-eventemitter';
```
mythmon: I think this is the only place we use `wolfy87-eventemitter`, besides tests. We can probably get rid of the dependency.

client/selfrepair/normandy_driver.js
Osmose: Aha, yeah. I forgot that I packed like four bugs into one fix here.

So it was declared an array but used as an object. So we make it an object. Also, below, we were using `heartbeatCallbacks` instead of `_heartbeatCallbacks`, so the UITour events weren't even triggering properly at all.

It's honestly a bit embarrassing.

Osmose:
@brittanystoroz This should be ready now. r?

client/selfrepair/normandy_driver.js
brittanystoroz: 🎉 🎉 🎉

client/selfrepair/normandy_driver.js
```diff
@@ -32,23 +31,73 @@ class LocalStorage {
...
+ */
+export class HeartbeatEmitter {
+  static EVENTS = [
+    'NotificationOffered',
+    'NotificationClosed',
+    'LearnMore',
+    'Voted',
+    'TelemetrySent',
+    'Engaged',
+  ]
```
brittanystoroz: Do we care about the `SurveyExpired` or `WindowClosed` events? https://dxr.mozilla.org/mozilla-central/source/browser/components/uitour/UITour.jsm#1050-1052

PR #298, opened 2016-10-18
Fix display issues for preview error message

mythmon:
This makes the error message a little easier to visually parse, and also removes the erroneous colon when there is no port in the URL.


brittanystoroz:
😍

PR #300, opened 2016-10-20
Single-survey show-heartbeat and bucket sampling

### LINKED ISSUE #297: in telemetry, unable to distinguish survey variants on the same recipe
## Currently:
- Unified Telemetry packets for recipes use the Survey Id and Survey Version fields *as is*

is a problem because:
- this makes success (uptake/usage) by variant impossible to distinguish

## Some Solutions
1.  **FIX UITour** to report a 3rd field.
2. **Change Messaging Recipe** to somehow glob the variation into one of those fields, either replacing version (meh) or globbing onto id (meh).  If I were **forced to pick**, I would say surveyId should be `${surveyId}:${variation}`.
3.  **Implement Bucketed Sampling** with guaranteed non-overlapping samples.  See @Osmose  for details.  In this system, each recipe only has 1 survey, period.
4.  **Make Analysts Hope and Pray** by doing 2 different recipes, and hope the sampling overlap isn't too much.

Osmose:
This fixes #297 by doing two things:

1. It changes the show-heartbeat action to support only a single survey per-recipe, vastly simplifying the form and allowing us to have a different survey ID for each survey.
2. It adds bucket sampling, which allows us to precisely target recipes to groups of users without overlap, replacing the previous function of survey weights.

r?


client/selfrepair/tests/test_utils.js
```diff
@@ -69,3 +69,101 @@ describe('stableSample', () => {
...
+    for (let i = 0; i < 100; i++) {
+      const total = randInt(100, 10000);
+
+      // Create 10 buckets, with the end of the count as the last.
+      const bucketPoints = [];
+      for (let k = 0; k < 9; k++) {
+        bucketPoints.push(randInt(0, total));
+      }
+      bucketPoints.push(total);
+      bucketPoints.sort((a, b) => a - b);
```
mythmon: I think you can generated bucketPoints more simply like this

```javascript
const bucketCount = 10;
const bucketPoints = [0];
for (let j = 0; j < bucketCount; j++) {
  const delta = randInt(total / bucketCount / 2, total / bucketCount);
  const last = bucketPoints[bucketPoints.length - 1]
  bucketPoints.push(last + delta);
}
```

Then below when you are iterating over this, you don't have to special case 0, because it is already there, and you don't have to special case `threeInARow`.

client/actions/tests/test_show-heartbeat.js
```diff
@@ -2,29 +2,23 @@ import { mockNormandy, pluginFactory } from './utils';
...
-      surveys: [surveyFactory()],
-    },
+      message: 'test message',
+      engagementButtonLabel: '',
+      thanksMessage: 'thanks!',
+      postAnswerUrl: 'http://example.com',
+      learnMoreMessage: 'Learn More',
+      learnMoreUrl: 'http://example.com',
+    }, args),
   }, props);
```
Osmose: > Also, at what point do we make a factoryboy-a-like?

Soon? This is working pretty well, honestly.

client/selfrepair/tests/test_utils.js
```diff
@@ -125,18 +127,6 @@ describe('bucketSample', () => {
           const start = k === 0 ? 0 : bucketPoints[k - 1];
           const count = bucketPoints[k] - start;
```
mythmon: I think you should change this to

```js
for let k = 1; k < bucketPoints.length; k++) {
  const start = bucketPoints[k - 1];
  const count = bucketPoints[k] - start;
  // ...
}
```

since there is the explicit 0 at the beginning now.

Issue #289, opened 2016-10-05
Add ability to filter by number of sync accounts


Osmose:
S&I has been asked to run a mobile promotion via Heartbeat before the end of the year, and to better target users likely to benefit from the promotion, they want to be able to target users by the number of sync clients they have (mobile vs desktop), and by the age of the user's profile and how active they are in using Firefox.

Since the system add-on probably won't hit release before the end of the year, our best bet for this is a patch to Firefox to add this to UITour. A Firefox bug with a WIP patch already exists: https://bugzilla.mozilla.org/show_bug.cgi?id=1305878

Osmose:
After talking with S&I and going through review in the bug, we decided to cut down this feature to just filtering by number of sync accounts. That patch landed and got uplifted to Beta in [bug 1310146](https://bugzilla.mozilla.org/show_bug.cgi?id=1310146).

The remaining work is to add the ability to filter by the new attributes to the self-repair shim and the add-on.

Osmose:
Closing this; this is only waiting on the PR for the add-on, and I expect all the PRs for the add-on to be closed before we move it into this repo.

PR #302, opened 2016-10-20
Add sync client counts to filter expression context. (#289)

Osmose:
r?

### DIFF
--- a/client/selfrepair/normandy_driver.js
+++ b/client/selfrepair/normandy_driver.js
@@ -117,6 +117,9 @@ export default class NormandyDriver {
         },
         sync(data) {
           client.syncSetup = data.setup;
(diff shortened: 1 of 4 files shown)


client/selfrepair/normandy_driver.js
```diff
@@ -117,6 +117,9 @@ export default class NormandyDriver {
         },
         sync(data) {
           client.syncSetup = data.setup;
+          client.syncDesktopDevices = data.desktopDevices;
+          client.syncMobileDevices = data.mobileDevices;
+          client.syncTotalDevices = data.totalDevices;
```
mythmon: These should have `|| 0` at the end, in case this runs on a Firefox that doesn't have the UITour patches.

PR #308, opened 2016-10-26
Use Yarn for NPM packages

mythmon:
Yarn gives us package hashing and deterministic builds, making our NPM dependencies as good as our Python packages.

r?


mythmon:
There seems to be some problem with Yarn, so it isn't working in Circle. I filed yarnpkg/yarn#1485. I'll close this for now.

PR #310, opened 2016-10-31
Update RecipeForm to use Redux-Form 6.

Osmose:
It's here! This rewrites the RecipeForm component to use redux-form 6, and adds a bunch of tests for the form.

r?


Osmose:
@andymikulski r?

client/control/components/Fields.js
```diff
@@ -0,0 +1,48 @@
...
+ * Builds the React component that is rendered for ControlField.
+ */
+export function buildControlField({
+  input,
+  meta: { error },
+  label,
+  className = '',
+  InputComponent,
+  children,
+  ...args // eslint-disable-line comma-dangle
```
andymikulski: Why disable this rule? Was this throwing an error or something?

client/control/components/RecipeContainer.js
```diff
@@ -17,6 +17,12 @@ export default function composeRecipeContainer(Component) {
       }
     }
 
+    componentWillReceiveProps(nextProps) {
+      if (nextProps.recipeId) {
```
andymikulski: Does this need to check against `this.props.recipeId`? e.g.:

```
if (nextProps.recipeId !== this.props.recipeId) { ...
```

client/control/components/RecipeForm.js
```diff
@@ -1,197 +1,206 @@
...
-    const actionFormErrors = errors.arguments;
+    // Throw if we found any errors.
+    if (Object.keys(errors).length > 0) {
+      throw errors;
+    }
+  },
+
+  onSubmit(values, dispatch, { recipeId, updateRecipe, addRecipe }) {
+    // Filter out unwanted keys for submission.
+    const recipe = _.pick(values, [
```
andymikulski: How do you feel about changing `import _ from 'underscore'` to `import { pick } from 'underscore'`, and just going with `... = pick(values, ...`?

If we ever switch to Webpack 2, this would drop the bundle size after tree shaking

andymikulski:
@Osmose can you fix the conflicted file when you get a chance? `client/control/tests/components/test_RecipeForm.js`

client/control/components/Fields.js
Osmose: Yeah. The problem is that it's syntactically invalid to have a comma after the spread argument, but the rule demands commas after all arguments when parameters are split up like this. So I had to disable the rule.

PR #313, opened 2016-11-07
Check durability of storage in driver

mythmon:
#291, retargeted at master with more tests.

Also, the driver now automatically JSON serializes/deserializes things, which is super helpful.

### DIFF
--- a/client/selfrepair/normandy_driver.js
+++ b/client/selfrepair/normandy_driver.js
@@ -2,10 +2,12 @@ import uuid from 'node-uuid';
 
 import Mozilla from './uitour.js';
 
+export const STORAGE_DURABILITY_KEY = '_storageDurability';
+
 /**
  * Storage class that uses window.localStorage as it's backing store.
  */
-class LocalStorage {
+export class LocalStorage {
   /**
    * @param    {string} prefix Prefix to append to all incoming keys.
    */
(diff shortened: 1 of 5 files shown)


client/selfrepair/normandy_driver.js
```diff
@@ -2,27 +2,44 @@ import uuid from 'node-uuid';
...
+  }
+
   _makeKey(key) {
     return `${this.prefix}-${key}`;
   }
 
   async getItem(key) {
-    return localStorage.getItem(this._makeKey(key));
+    const storageIsDurable = await this.isDurable();
+    if (!storageIsDurable && !this.testing) {
```
Osmose: `this.testing` is a property of the driver, not of the LocalStorage class.

Did you mean to add this to the driver instead?

client/selfrepair/normandy_driver.js
```diff
@@ -2,27 +2,44 @@ import uuid from 'node-uuid';
...
+      throw new Error('Storage durability unconfirmed');
+    }
+    const val = localStorage.getItem(this._makeKey(key));
+    try {
+      return JSON.parse(val);
```
Osmose: If you're adding this, you might want to update show-heartbeat implementation to deal with this, and the docs for the storage spec to specify that input and output is JSON-ified during storage. This will be important in the future when storing, say, date objects.

client/selfrepair/normandy_driver.js
```diff
@@ -2,27 +2,44 @@ import uuid from 'node-uuid';
...
+    } catch (e) {
+      this.log(e.toString());
```
Osmose: Same as the above, the storage object has no logging.

client/selfrepair/normandy_driver.js
```diff
@@ -86,6 +103,15 @@ export class HeartbeatEmitter {
 export default class NormandyDriver {
   constructor(uitour = Mozilla.UITour) {
     this._uitour = uitour;
+    this.setDurability();
```
Osmose: I think setting and checking this is the responsibility of the runtime, since we don't necessarily guarantee that only one driver is created per-run. If something causes the driver to be created multiple times, after the first time it will think that it is durable when in fact it is not.

I think the driver API we want is more along the lines of just having an `isStorageDurable()` method.

client/selfrepair/normandy_driver.js
mythmon: I changed this to `this.skipDurability`, and made the driver pass it in based on the value of `NormandyDriver::testing`.

mythmon:
Updated

Issue #303, opened 2016-10-21
Show survey ID on recipe listing for show-heartbeat recipes


Osmose:
(This issue is intended for Andy to work on as his first bug, do not take this!)

The survey ID on show-heartbeat recipes is a useful piece of information that S&I wants to be able to see on the index listing page. However, since it only applies to show-heartbeat recipes, we probably should come up with a more generic solution for showing details about a recipe based on the action.

Here's a mockup I came up with for one possibility of how to display this:

<img width="970" alt="screen shot 2016-10-21 at 10 00 00 am" src="https://cloud.githubusercontent.com/assets/193106/19606839/72073688-977a-11e6-8fe1-f91c0585dcae.png">

This isn't exactly how it should look; we should probably remove the bullets on the list. But it's a starting point.

If possible, we should also make the survey ID searchable in the search box, and more generically, any value in the details column.

andymikulski:
merged in #314

PR #314, opened 2016-11-08
#303 Add metaData and searchData to recipes

andymikulski:
- Add `metaData` and `searchData` properties to received recipes.
- `metaData` is a hash of displayed labels/values
- `searchData` is a string of concat'd metadata values, which is used by the `Table` component for searching/sorting/filtering.
- Desired metadata properties are defined in `DisconnectedRecipeList#ActionMetaData` and collected/applied in `DisconnectedRecipeList#applyRecipeMetadata`

### DIFF
--- a/client/control/components/RecipeList.js
+++ b/client/control/components/RecipeList.js
@@ -99,21 +99,21 @@ class DisconnectedRecipeList extends React.Component {
     // check if there are specific properties/labels we want to display
     const requestedMetaProps = DisconnectedRecipeList.ActionMetaData[recipeAction];
 
-    // if we have a metadata definition to fill..
+    // if we have a metaData definition to fill...
     if (requestedMetaProps) {
+      // ...get the data we want to display
       const foundData = requestedMetaProps(newRecipe);
 
-      // add the data to the metaData collection
+      // ...and add it to the existing metaData collection
+      // (the data comes back as an array of objects,
+      // so we can just concat it to our existing array)
       newRecipe.metaData = newRecipe.metaData.concat(foundData);
     }
 
     // update the searchdata string with whatever the values are
     // (this is used for sorting/filtering/searching)
-    newRecipe.metaData.map(data => {
-      newRecipe.searchData += data.value;
-
-      // lint complains if we dont return
-      return data;
+    newRecipe.metaData.forEach(data => {
+      newRecipe.searchData = `${newRecipe.searchData} ${data.value}`;
     });
 
     return newRecipe;
--- a/client/control/components/RecipeList.js
+++ b/client/control/components/RecipeList.js
@@ -64,13 +64,47 @@ SwitchFilter.propTypes = {
 };
 
 class DisconnectedRecipeList extends React.Component {
-  propTypes = {
-    dispatch: pt.func.isRequired,
-    isFetching: pt.bool.isRequired,
-    recipeListNeedsFetch: pt.bool.isRequired,
-    recipe: pt.object.isRequired,
-    recipes: pt.array.isRequired,
+
+  /**
+   * Given a recipe object, determines what type of recipe it is (based on its `action`),
+   * and then compiles a hash of 'displayed metadata props' and their values. This hash
+   * is saved on the recipe as `metaData`.
+   *
+   * Beyond that, a string of metaData values is created, and attached to the
+   * recipe as the `searchData` property. This is used by the `Table` component
+   * in to search/filter over the metaData.
+   *
+   * @param  {Object} Original recipe received/passed into reducer
+   * @return {Object} Original recipe, but with `metaData` and `searchData` properties added
+   */
+  static applyUniqueRecipeID(recipe) {
+    const newRecipe = { ...recipe };
+
+    switch (recipe.action) {
+      case 'show-heartbeat':
+        newRecipe.metaData = {
+          'Survey ID': recipe.arguments.surveyId,
+        };
+        break;
+      case 'console-log':
+        newRecipe.metaData = {
+          Message: recipe.arguments.message,
+        };
+        break;
+      default:
+        newRecipe.metaData = {};
+        break;
+    }
+
+    // build a string of metadata values,
+    // which is used for table filtering/searching later
+    newRecipe.searchData = Object.keys(newRecipe.metaData)
+      .map(metaProp => newRecipe.metaData[metaProp])
+      .join(' ');
+
+    return newRecipe;
   }
+
 
   constructor(props) {
     super(props);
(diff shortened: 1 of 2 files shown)


client/control/components/RecipeList.js
```diff
@@ -64,14 +64,61 @@ SwitchFilter.propTypes = {
...
+   * @return {Object} Original recipe but with `metaData` and `searchData` properties added
+   */
+  static applyRecipeMetadata(recipe) {
+    const { action: recipeAction } = recipe;
+    const newRecipe = { ...recipe };
+
+    // recipes should have empty metaData/searchData props,
+    // regardless if we set the values or not
+    newRecipe.metaData = {};
+    newRecipe.searchData = '';
```
mythmon: I'd combine these two lines with the newRecipe initializer, like

```js
const newRecipe = { ...recipe, metadata: {}, searchData: '' };
```

client/control/components/RecipeList.js
```diff
@@ -64,14 +64,61 @@ SwitchFilter.propTypes = {
...
+
+    // check if there are specific properties/labels we want to grub
```
mythmon: Nit: s/grub/grab/

client/control/components/RecipeList.js
```diff
@@ -64,14 +64,61 @@ SwitchFilter.propTypes = {
...
+  /**
+   * Recipe metadata properties and associated labels to display
+   * @type {Object}
+   */
+  static ActionMetaData = {
+    'show-heartbeat': {
+      // prop name : displayed label
+      surveyId: 'Survey ID',
+    },
+  };
```
mythmon: What do you think about a slightly more procedural way of doing declaring this data, like this:

```js
static recipeMetadata = {
  'show-heartbeat': recipe => [{ label: 'Survey ID', value: recipe.surveyId }],  
};
```

I think this is a little more direct and easier to understand, and also a bit more flexible if we want to pull more complex data out of recipes (such as values in the filter expression).
```

File: client/control/sass/control.scss [hunk 4]
```diff
@@ -382,3 +382,12 @@
       white-space: nowrap;
     }
 }
+
+/* Recipe List */
+#recipe-list {
```
mythmon: We generally don't use ids for selectors where we can avoid it. I think we can add a classname to the recipe component instead.

client/control/components/RecipeList.js
```diff
@@ -64,14 +64,61 @@ SwitchFilter.propTypes = {
...
+    for (const propName in requestedMetaProps) {
+      if (requestedMetaProps.hasOwnProperty(propName)) {
+        // get the displayed label value
+        const label = requestedMetaProps[propName];
+        // get the actual property value
+        const value = newRecipe.arguments[propName];
+        // save the label:property into the metadata
+        newRecipe.metaData[label] = value;
+
+        // add the value to the existing search data
```
mythmon: The series of comments here is maybe a tick verbose, but it is better to err on the side of over-commenting, so that's fine.

client/control/components/RecipeList.js
```diff
@@ -169,6 +233,14 @@ class DisconnectedRecipeList extends React.Component {
   }
 }
 
+DisconnectedRecipeList.propTypes = {
```
andymikulski: This was to satisfy a console error (something about propTypes being defined on an instance instead of on the class). This also conforms to the conventions used on other components.

client/control/sass/control.scss
andymikulski: I agree wrt using id's for selectors - I was just following the conventions already established in `control.scss`. I'll switch to a class selector instead.

client/control/components/RecipeList.js
andymikulski: ```Warning: propTypes was defined as an instance property on DisconnectedRecipeList. Use a static property to define propTypes instead.```

mythmon: Fair. I think `static propTypes = { ... }` would have also satisfied this. I'm not too picky either way. Do what you prefer.

client/control/components/RecipeList.js
```diff
@@ -87,34 +86,36 @@ class DisconnectedRecipeList extends React.Component {
...
-        const label = requestedMetaProps[propName];
-        // get the actual property value
-        const value = newRecipe.arguments[propName];
-        // save the label:property into the metadata
-        newRecipe.metaData[label] = value;
-
-        // add the value to the existing search data
-        newRecipe.searchData = `${newRecipe.searchData} ${value}`;
-      }
+    // if we have a metadata definition to fill..
```
mythmon: Typo here? Is this a period or an ellipsis?

client/control/components/RecipeList.js
```diff
@@ -64,14 +64,62 @@ SwitchFilter.propTypes = {
...
+      newRecipe.metaData = newRecipe.metaData.concat(foundData);
+    }
+
+    // update the searchdata string with whatever the values are
+    // (this is used for sorting/filtering/searching)
+    newRecipe.metaData.map(data => {
+      newRecipe.searchData += data.value;
+
+      // lint complains if we dont return
+      return data;
```
mythmon: The lint error could be avoided by using `.forEach` instead of `.map`. Alternatively, `for (const data of newRecipe.metaData) {` should work fine.

Separately does this need spaces between values? Maybe it doesn't matter, the search code probably doesn't care.

client/control/components/RecipeList.js
andymikulski: Personally, I prefer to use the `static` keyword, so the propTypes are defined right in the component However, the conventions used for other components uses the `[class].propTypes = {...}` business, so if I used `static` here I'd feel the need to go through and update the rest of the code - but I'd prefer to do that in another branch/PR

client/control/components/RecipeList.js
```diff
@@ -64,14 +64,61 @@ SwitchFilter.propTypes = {
...
+      if (requestedMetaProps.hasOwnProperty(propName)) {
+        // get the displayed label value
+        const label = requestedMetaProps[propName];
+        // get the actual property value
+        const value = newRecipe.arguments[propName];
+        // save the label:property into the metadata
+        newRecipe.metaData[label] = value;
+
+        // add the value to the existing search data
+        newRecipe.searchData = `${newRecipe.searchData} ${value}`;
```
andymikulski: If we didn't have spaces between values, we could use the simple string concatenation. However, apparently if you attempt to do something like
```
searchData += `${value} `
// or
searchData += value + ' ';
```
it throws a `prefer-template` error while linting (http://eslint.org/docs/rules/prefer-template).

We need the values separated to prevent filters from returning the wrong items. e.g. if the values for an item are `123` and `345`, searching for `33` would still display the item (since the computed value is `123345`)

_SO_ basically I'm going to just revert to what the code originally was, unless you prefer another method

mythmon:
Thanks for working through all the lint and review issues! This looks good now, and will go out in the next deploy.

PS: You should associate your Mozilla email with your Github account and/or change your git client to commit with your Github email, so your commits are associated with your Github account here.

Issue #309, opened 2016-10-31
Tie SurveyGizmo Answers to Telemetry Data


Osmose:
With regards to Lead Users, the goal is to send out surveys that we think can identify Lead Users, and then analyze the Telemetry profiles of those users to identify patterns from Telemetry data that we think are good indicators of Lead Users (hardware? customizations?).

Legal bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1313453

## Spec
Add a new boolean argument to the `show-heartbeat` action that is displayed in the admin as a checkbox with the label "Include unique user ID in post-answer-url and Telemetry". When checked, the action should append the unique UUID we generate on the client to the `surveyId` that gets logged in telemetry, separated with a double-colon (`::`). It should also alter the query parameters passed to in the `postAnswerUrl` argument to include the unique UUID.

The UUID in question should be the one we include in filter expressions under `normandy.userId`.

Osmose:
@gregglind @mattgrimes The spec above has us modifying the `surveyId` that ends up in telemetry, are ya'll okay with this change? It lines up with what you suggested in the legal bug.

Are we just waiting for the legal bug to be resolved (and your approval) before working on this?

Osmose:
Confirmed via IRC that the spec above is fine. This is now waiting on the linked legal bug to be approved.

gregglind:
I approve of this.  How it will be handled:

- if the analysis code is naive, then different VARIATIONS will look like different SURVEYS.  This is fine.
-  if the analysis code is smart, it will KNOW to split on `::`.  
- analysts will be SMART and not use `::` as part of a surveyId.

- [ ] confirm that uu / query args handle colon fine.

Osmose:
We've gotten the go-ahead for implementing this by legal. We should double-check with them before deploying it, though. Probably not in this week's release.

mythmon:
Fixed by #315

PR #315, opened 2016-11-08
#309 Add UUID to surveyId

andymikulski:
- Add UUID to surveyId (e.g. `my-survey::USERS-UUID`)
- Add `userId` to query param for the `postAnswerUrl` (for SurveyGizmo)

*Note: I had to run `npm run watch -- --update-actions` to update the heartbeat action with new code - you may have to as well!*

### DIFF
--- a/client/actions/show-heartbeat/index.js
+++ b/client/actions/show-heartbeat/index.js
@@ -1,6 +1,6 @@
 import { Action, registerAction } from '../utils';
 
-const VERSION = 52; // Increase when changed.
+const VERSION = 53; // Increase when changed.
 const LAST_SHOWN_DELAY = 1000 * 60 * 60 * 24 * 7; // 7 days
 
 
@@ -128,11 +128,11 @@ export default class ShowHeartbeatAction extends Action {
     // A bit redundant but the action argument names shouldn't necessarily rely
     // on the argument names showHeartbeat takes.
     const heartbeatData = {
-      surveyId,
+      surveyId: `${surveyId}::${flow.id}`,
       message,
       engagementButtonLabel,
       thanksMessage,
-      postAnswerUrl: this.annotatePostAnswerUrl(postAnswerUrl),
+      postAnswerUrl: this.annotatePostAnswerUrl({ url: postAnswerUrl, userId: flow.id }),
       learnMoreMessage,
       learnMoreUrl,
       flowId: flow.id,
@@ -179,7 +179,7 @@ export default class ShowHeartbeatAction extends Action {
     return Number.isNaN(lastShown) ? null : lastShown;
   }
 
-  annotatePostAnswerUrl(url) {
+  annotatePostAnswerUrl({ url, userId }) {
     // Don't bother with empty URLs.
     if (!url) {
       return url;
@@ -193,6 +193,7 @@ export default class ShowHeartbeatAction extends Action {
       isDefaultBrowser: this.client.isDefaultBrowser ? 1 : 0,
       searchEngine: this.client.searchEngine,
       syncSetup: this.client.syncSetup ? 1 : 0,
+      userId,
     };
 
     // Append testing parameter if in testing mode.
--- a/client/actions/show-heartbeat/index.js
+++ b/client/actions/show-heartbeat/index.js
@@ -1,6 +1,6 @@
 import { Action, registerAction } from '../utils';
 
-const VERSION = 53; // Increase when changed.
+const VERSION = 54; // Increase when changed.
 const LAST_SHOWN_DELAY = 1000 * 60 * 60 * 24 * 7; // 7 days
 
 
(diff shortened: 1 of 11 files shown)


client/actions/show-heartbeat/index.js
```diff
@@ -128,11 +128,11 @@ export default class ShowHeartbeatAction extends Action {
     // A bit redundant but the action argument names shouldn't necessarily rely
     // on the argument names showHeartbeat takes.
     const heartbeatData = {
-      surveyId,
+      surveyId: `${surveyId}::${flow.id}`,
```
Osmose: Ah, the UUID we want to append isn't the flow ID. The flow ID is unique per-interaction with the prompt, so the same user interacting with different surveys will have a different flow ID. We want to use the per-user ID that doesn't change so that S&I can target and correlate answers from the same users over time.

We want to use the `userId` made available to filter expressions (see the [filter expression docs](http://normandy.readthedocs.io/en/latest/user/filter_expressions.html#normandy.userId)). The `getUserId` function in `self_repair_runner.js` can fetch it. I suspect we probably want to include it in the client info returned from `driver.client()` so that actions can use it.

andymikulski:
@Osmose r+?

client/actions/tests/test_show-heartbeat.js
```diff
@@ -195,13 +197,23 @@ describe('ShowHeartbeatAction', () => {
       revision_id: 42,
       arguments: {
         surveyId: 'my-survey',
+        includeTelemetryUUID: true,
```
mythmon: I think this new behavior should be put in a new test, instead of modifying an existing test.

client/actions/tests/test_show-heartbeat.js
```diff
@@ -195,13 +197,23 @@ describe('ShowHeartbeatAction', () => {
...
     const action = new ShowHeartbeatAction(normandy, recipe);
 
     await action.execute();
+
+    // userId should be set
+    expect(normandy.userId).not.toBeNull();
+
+    // userId should be uuid4
+    const UUID_ISH_REGEX = /^[a-f0-9-]{36}$/;
+    expect(UUID_ISH_REGEX.test(normandy.userId)).toBe(true);
```
mythmon: Jasmine has a `toMatch` matcher for regex:

```js
expect(normandy.userId).toMatch(/^[a-f0-9-{36}$/);  // user-id is uuid-ish
```

client/selfrepair/normandy_driver.js
```diff
@@ -128,10 +128,27 @@ export default class NormandyDriver {
...
+   */
   uuid() {
     return uuid.v4();
   }
 
+  /**
+   * Get a user id. If one doesn't exist yet, make one up and store it in local storage.
+   * @return {String} A stored or generated UUID
+   */
+  get userId() {
```
mythmon: Adding this to the driver should be reflected in [the driver docs](http://normandy.readthedocs.io/en/latest/dev/driver.html).

client/selfrepair/normandy_driver.js
```diff
@@ -128,10 +128,27 @@ export default class NormandyDriver {
     }
   }
 
+  /**
+   * Generate a fresh UUID
+   * @return {String} Generated UUID
```
mythmon: :+1: for adding docstrings

client/control/components/action_forms/HeartbeatForm.js
```diff
@@ -18,6 +19,11 @@ export default function HeartbeatForm({ fields }) {
         This action shows a single message or survey prompt to the user.
       </p>
       <div className="fluid-4">
+        <FormField
+          type="checkbox"
+          label="Include unique user ID in post-answer-url and Telemetry"
+          field={fields.includeTelemetryUUID}
+        />
```
mythmon: A minor thing, but this could probably look better visually. Do you know if that would be easy to improve? I think especially putting the checkbox and the label on the same line would nicer?

shot_7335

client/selfrepair/normandy_driver.js
andymikulski: DocBlockr for Sublime Text! https://packagecontrol.io/packages/DocBlockr

client/actions/tests/test_show-heartbeat.js
andymikulski: this was following conventions used in other tests - I've gone ahead and made the change throughout them all

andymikulski:
@mythmon r+?

mythmon:
This looks good, r+. Thanks for trying to keep everything consistent, even we haven't done a good job with that in the past. This can merge after CI passes. (Anyone can do that, but I'll try and keep an eye on it).

PS: The review jargon is `r?`, `r-`, and `r+`. The etymology is that Bugzilla has flags, and one of the flags is "review". You can be "review?" (pending), "review-" (rejected) or "review+" (approved). Then these are shortened to just `/r[?+-]/`. There are other flags like "feedback" (f?, f+, f-) for advice that doesn't care approval.

Given that, "r+?" is a bit silly. Though in practice we mostly just use the Github assignee/review approval stuff nowadays.

PR #317, opened 2016-11-09
Fix propType errors

andymikulski:
- Fixes 'wrong propType' error for SwitchFilter
- Fixes slow/crashy startup behavior (on my machine)

**NOTE**: `Warning: Unknown prop 'props' on <td> tag. Remove this prop from the element.` still exists in console - this is due to the `reactable` package and should be resolved soon. Relevant issue: https://github.com/glittershark/reactable/issues/311

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -44,7 +44,7 @@ if (production) {
 module.exports = [
   {
     context: __dirname,
-    devtool: production ? undefined : 'eval-source-map',
+    devtool: production ? undefined : 'eval',
 
     entry: {
       selfrepair: [
@@ -95,7 +95,7 @@ module.exports = [
     },
   },
   {
-    devtool: production ? undefined : 'eval-source-map',
+    devtool: production ? undefined : 'eval',
 
     entry: {
       'console-log': './client/actions/console-log/index',
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -44,7 +44,7 @@ if (production) {
 module.exports = [
   {
     context: __dirname,
-    devtool: production ? undefined : 'eval',
+    devtool: production ? undefined : 'cheap-module-eval-source-map',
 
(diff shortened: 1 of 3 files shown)


webpack.config.js
```diff
@@ -44,7 +44,7 @@ if (production) {
 module.exports = [
   {
     context: __dirname,
-    devtool: production ? undefined : 'eval-source-map',
+    devtool: production ? undefined : 'eval',
```
Osmose: This means we won't be emitting sourcemaps at all, but the desire is to emit sourcemaps, but only in development mode.

Is it possible for us to emit sourcemaps without seriously slowing down execution locally? Actually, it is a bit weird that just having a source map makes execution of the code slow. Any theories on why that's causing this?

mythmon: The [webpack docs](https://webpack.github.io/docs/configuration.html#devtool) are pretty clear that `eval` will be much faster than `eval-source-map`. Maybe we could use something like `cheap-module-eval-source-map`, which is faster than `eval-source-map` (according to the docs), but still provides line-level source maps? We'd be giving up column information, but that's probably ok. We can always add these back in during a debug session.

@andymikulski is there a noticeable slow down using `cheap-module-eval-source-map` compared to `eval`?

Alternatively, do we want to make this configurable? Something like `npm run watch -- --no-sourcemaps`?

andymikulski: >Actually, it is a bit weird that just having a source map makes execution of the code slow. Any theories on why that's causing this?

Vanilla Firefox is apparently not great at handling a 7mb JS file. Tested in Chrome and it loads fine, tested in FirefoxDeveloperEdition and it loads fine, but in normal FF it hangs and sometimes crashes the whole browser.

I'll switch to `cheap-module-eval-source-map` and just use FFDev for testing.

PR #322, opened 2016-11-14
#301 Add cloning, tests, route

### LINKED ISSUE #301: Clone recipe button
Recipes should have a button on the edit page that opens up a new recipe form with fields populated from the existing recipe. This will make it easier after #300 to create a survey, and then clone it to create variants of the survey.

andymikulski:
Fixes #301 (again)

- Add 'clone' button to existing recipes
- Add 'clone' route (e.g. `../recipe/14/clone/`)
- Add clone tests

### DIFF
--- a/client/control/components/Header.js
+++ b/client/control/components/Header.js
@@ -2,11 +2,13 @@ import React, { PropTypes as pt } from 'react';
 import { Link } from 'react-router';
 import Breadcrumbs from 'react-breadcrumbs';
 
+import absolutePath from '../../utils/absolute-path';
+
 export default function Header({ pageType: { ctaButtons }, currentLocation, routes, params }) {
   let ctaBtns;
   if (ctaButtons) {
     ctaBtns = ctaButtons.map(({ text, icon, link }, index) =>
-      <Link className="button" to={currentLocation + link} key={index}>
+      <Link className="button" to={absolutePath(currentLocation, link)} key={index}>
         <i className={`pre fa fa-${icon}`} /> {text}
       </Link>
     );
--- a/client/control/components/RecipeForm.js
+++ b/client/control/components/RecipeForm.js
@@ -56,8 +56,8 @@ export class RecipeForm extends React.Component {
     const displayedRecipe = this.props.recipe || {};
 
     return isCloning &&
-      (<span className="cloning-message">
-        You are cloning
+      (<span className="cloning-message callout">
+        {'You are cloning '}
         <Link to={`/control/recipe/${displayedRecipe.id}/`}>
           {displayedRecipe.name} ({displayedRecipe.action})
         </Link>.
--- a/client/control/components/RecipeForm.js
+++ b/client/control/components/RecipeForm.js
@@ -53,29 +53,15 @@ export class RecipeForm extends React.Component {
 
   renderCloningMessage() {
     const isCloning = this.props.route && this.props.route.isCloning;
-    const { recipeId } = this.props;
     const displayedRecipe = this.props.recipe || {};
 
-    return isCloning ?
+    return isCloning &&
       (<span className="cloning-message">
-        You're cloning {displayedRecipe.name} ({displayedRecipe.action}).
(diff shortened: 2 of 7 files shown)


andymikulski:
@Osmose @mythmon r?

client/control/components/RecipeForm.js
```diff
@@ -57,8 +99,14 @@ export class RecipeForm extends React.Component {
       );
     }
 
+    const isCloning = route && route.isCloning;
+
+    const submitButtonCaption = recipeId && !isCloning ? 'Update Recipe' : 'Add New Recipe';
+
     return (
       <form className="recipe-form" onSubmit={handleSubmit}>
+        { this.renderCloningMessage() }
```
mythmon: This seems like an odd place to put this. It gives a really high visual priority, since it is the first thing a user will see. On the other hand, it makes sense to have it at the top of the page. Maybe the top right of the page would be better?

We could also punt on the design issue, if you'd rather do a more comprehensive redesign along with the peer approval features.

client/control/tests/components/test_RecipeForm.js
```diff
@@ -112,11 +141,32 @@ describe('<RecipeForm>', () => {
       expect(addRecipe).not.toHaveBeenCalled();
     });
 
-    it("should add a new recipe if it doesn't have an ID", async () => {
+    it('should add a new recipe if it doesn\'t have an ID', async () => {
```
andymikulski: Weird, yeah I'm wondering if that was a result of running `lint --fix`

client/control/components/RecipeForm.js
andymikulski: Ideally, the button would go alongside the 'preview' and 'history' buttons, though I was having trouble displaying that button dynamically up there. I wasn't too concerned with the design at this point, but I can take care of it.

client/control/components/RecipeForm.js
```diff
@@ -37,14 +42,51 @@ export class RecipeForm extends React.Component {
...
   };
 
+  renderCloningMessage() {
+    const isCloning = this.props.route && this.props.route.isCloning;
+    const { recipeId } = this.props;
+    const displayedRecipe = this.props.recipe || {};
+
+    return isCloning ?
+      (<span className="cloning-message">
+        You're cloning {displayedRecipe.name} ({displayedRecipe.action}).
```
andymikulski: Not sure - I'm wondering if it's how I'm splitting up the lines. Easiest fix is to just switch to `You are`, that work?

andymikulski:
aight, since last review:

- Moved 'clone' / 'cancel' buttons to the top navbar (next to 'preview', 'history', etc)
- Corrected bug in Header where relative path links would lead to 404
- Added test for the absolutePath util that Header now uses
- Added 'callout' styles for the `You are now cloning...` message, moved it to the right side of the page*

* This 'design' is most definitely temporary, so I didn't spend much time tweaking it

@mythmon this is ready for another review (minus the `route: pt.object` propType deal)

client/control/components/RecipeForm.js
```diff
@@ -37,14 +42,51 @@ export class RecipeForm extends React.Component {
       action: pt.string.isRequired,
       arguments: pt.object.isRequired,
     }),
-  }
+    // route prop passed from router
+    route: pt.object,
```
mythmon: I didn't realize that `route` was coming from the router directly. I assumed that because this wasn't a connected component that it wasn't at the top level, but I guess I was wrong.

client/control/components/Header.js
```diff
@@ -2,11 +2,13 @@ import React, { PropTypes as pt } from 'react';
 import { Link } from 'react-router';
 import Breadcrumbs from 'react-breadcrumbs';
 
+import absolutePath from '../../utils/absolute-path';
```
mythmon: I find some humor in using a relative import path to import something called "absolute-path". (this is obviously the right think, just amusing)

PR #342, opened 2016-11-16
Webpack/Redux Refactor + Cleanup

andymikulski:
This has enough changes in it that I think it can be pulled at this time. If we want to add a selector lib like Reselect, I'd want to split off into a new branch anyway.

In this branch:

- Object.assigns now use the spread operator
- Most imports no longer require absolute pathing (see `webpack.config.js` under `resolve: alias`)
- Broke out some controlApp properties into their own stores (notifications, recipes). Example state tree:
```
{
  controlApp: {
    isFetching: false,
  },
  recipes: {
    list: [],
    selectedRecipe: null,
    recipeListNeedsFetch: true,
  },
  notifications: [],
  // routing, form, ...
}
```
- Moved reducer business into `reducers/index`, which are imported into `configureStore`
- Cleaned up some Redux stuff (there was some unnecessary initial state business)
- Removed the '.js' from imports - webpack automatically handles the extension for us (this might be personal preference)
- Makes necessary adjustments to reducer tests


andymikulski:
@Osmose @mythmon r?

andymikulski:
This is failing on CircleCI due to some eslint issues:

```
  4:1   error  'components' should be listed in the project's dependencies. Run 'npm i -S components' to add it  import/no-extraneous-dependencies
  4:36  error  Unable to resolve path to module 'components/RecipeContainer'         
```

This is solved with an update to the `.eslintrc` file, included in this PR. It seems as if Circle is using an outdated eslint file or something. You can reproduce the error locally by removing the `settings:importer/resolver` section of .eslintrc and running `npm run lint`.

Is there something we need to do to update Circle's eslint checks? @Osmose @mythmon

webpack.config.js
```diff
@@ -93,6 +93,19 @@ module.exports = [
...
+    resolve: {
+      alias: {
+        app: path.resolve(__dirname, './client/control/app.js'),
+        routes: path.resolve(__dirname, './client/control/routes.js'),
+        stores: path.resolve(__dirname, './client/control/stores'),
+        reducers: path.resolve(__dirname, './client/control/reducers'),
+        actions: path.resolve(__dirname, './client/control/actions'),
+        components: path.resolve(__dirname, './client/control/components'),
+        tests: path.resolve(__dirname, './client/control/tests'),
+        selfrepair: path.resolve(__dirname, './client/selfrepair'),
```
mythmon: I quite like the idea of having these sort of absolute imports. I'm concerned about the predictability of these though. Since they are all at varying depths, it is hard to tell what any given import will actually point to.

I think we should go with a simpler, more predictable model, something like

```js
resolve: {
  alias: {
    actions: path.resolve(__dirname, './client/actions'),
    control: path.resolve(__dirname, './client/control'),
    selfrepair: path.resolve(__dirname, './client/selfrepair'),
    tests: path.resolve(__dirname, './client/tests'),
  },
}
```

andymikulski: I see. I had been swapping them out for easier use in the `control` directory, but you'd like it to point to the directories inside `client`.

I'll get to fixing this now.

andymikulski:
@mythmon I don't think I actually know enough about docker to debug this issue. Can you point me in the right direction? I was attempting to see if the .eslintrc file is what it should be in the container, but I'm having a hard time just looking for the files to begin with. I can at least recreate the issue, FWIW!

mythmon:
I was debugging this by running bash in the container (`docker run --user root -it pr342 bash`), and poking around. There aren't many tools available. I used `cat` to look at files most of the time. The `.eslintrc. file has the expected config. I'm not sure what else to look at. Other useful things may be `docker cp`, which can let you copy files in or out. You can also `apt-get install` any packages you need to debug in the container.

client/control/tests/test_routes.js
```diff
@@ -1,8 +1,9 @@
 import { replace } from 'react-router-redux';
 import { shallow } from 'enzyme';
 
-import { createApp } from '../app.js';
-import NoMatch from '../components/NoMatch.js';
+import { createApp } from 'control/app';
+
+import NoMatch from 'control/components/NoMatch.js';
```
mythmon: This is a super tiny nit, but why the extra space here? Or is there a pattern here that I haven't seen yet?

mythmon:
This looks great, but there is a merge conflict on the RecipeForm.

client/control/tests/test_routes.js
andymikulski: Not sure, this was just a style I had picked up from the various React boilerplates I used to get started

mythmon:
Your merge commit only has one parent, and doesn't actually merge anything. I suspect this is actually therapist's fault. I've seen this before but haven't been able to figure out exactly what is going on to file it.

I'd undo that commit (`git reset --hard 06d7349`), re-fetch everything (`git fetch --all`) and then try the merge again, this time stopping Therapist from running (`git merge upstream/master --no-verify`, assuming your mozilla remote is named upstream).

andymikulski:
Well that was weird. I had to use the `--no-verify` flag when committing to get around Therapist having issue finding the git merge files:

`fatal: could not open '.git/MERGE_HEAD' for reading: No such file or directory`

I thought I was going crazy for a second. Thanks @mythmon !

mythmon:
One of our Python dependencies, click, added a wheel release for the version we use, which threw off our hashes. I've fixed the issue, and CI should pass now.

andymikulski:
It works! Quick, @mythmon , merge it!

Issue #348, opened 2016-11-17
Only display one `show-heartbeat` prompt a day


Osmose:
`show-heartbeat` recipes should store and check a date when they display, and if a `show-heartbeat` recipe has executed and displayed a prompt to the user within the past 24 hours, other `show-heartbeat` recipes should not prompt the user.

This should be separate from checking if the prompt has been shown recently, so that if a user gets 2 recipes, that 1 of them will show one day, and then the next day it will see that it has shown recently and not run, allowing the second recipe to show instead.

andymikulski:
Plan for this is to use a least-recently-used map ([link](https://github.com/rsms/js-lru)) that tracks when heartbeats were last run. In theory, we can use this to pull the ID of the heartbeat that should run next, and store the state in localStorage.

Osmose:
> Plan for this is to use a least-recently-used map (link) that tracks when heartbeats were last run. In theory, we can use this to pull the ID of the heartbeat that should run next, and store the state in localStorage.

Huh, why do you need to store the heartbeats locally? The intent was to have show-heartbeat just abort early if they can't run (if another show-heartbeat has run recently). It's not important that we log that heartbeat and show it later; if a heartbeat action fails on Monday due to collision, and isn't being sent out on Tuesday, then the user in question just shouldn't see it. No need to guarantee display.

andymikulski:
Ah, I see. My misunderstanding, thanks for clarifying @Osmose !

PR #352, opened 2016-11-17
Remove Heartbeat 'last show delay'

### LINKED ISSUE #347: Change show-heartbeat timeout to be indefinite
Currently, `show-heartbeat` recipes will store when they last showed themselves, and will check to see if it has been 7 days before showing themselves again. Instead, we want `show-heartbeat` recipes to only ever prompt the user once, and never show themselves again. 

This is marked as a bug because we did not match S&I's expectations and we suspect users are hitting this as we speak.

andymikulski:
Fixes #347 

- Removes 7-day check for heartbeats
- Heartbeat actions now only run once, ever
- Test mode still fires heartbeats errytime
- Updated tests

### DIFF
--- a/client/actions/show-heartbeat/index.js
+++ b/client/actions/show-heartbeat/index.js
@@ -1,8 +1,6 @@
 import { Action, registerAction } from '../utils';
 
-const VERSION = 54; // Increase when changed.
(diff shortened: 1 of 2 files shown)


andymikulski:
This slightly ties into #348 so I'm going to close this PR and solve both issues (#347 and #348) with one branch

PR #355, opened 2016-11-17
Add controls for users to manage API keys

mythmon:
Before now, the only  way for a user to get an API token was for an admin to generate one. This kind of sucks, so I made some APIs and UI for it. I'm pretty confident about the API design, but I'm not sure about how it is exposed in the UI. Suggestions are welcome here.

I tried to fit in with the changes in #342, but I expect I'll have to deal with merge conflicts one that lands.

@Osmose can you review the Python bits here?

@andymikulski Can you review the JS stuff?


normandy/base/templatetags/app_data.py
```diff
@@ -0,0 +1,27 @@
...
+
+register = Library()
+
+
+@register.tag
+def app_data(parser, token):
+    return AppDataNode()
+
+
+class AppDataNode(Node):
```
Osmose: This should probably just be a [simple tag](https://docs.djangoproject.com/es/1.9/howto/custom-template-tags/#django.template.Library.simple_tag).

normandy/base/tests/test_serializers.py
```diff
@@ -0,0 +1,17 @@
+import pytest
+
+from normandy.base.tests import Whatever, TokenFactory
+from normandy.base.api.serializers import TokenSerializer
+
+
+@pytest.mark.django_db()
+class TestTokenSerializer:
```
Osmose: I think we decided to inherit from object in Python 3 code.

normandy/base/api/views.py
```diff
@@ -0,0 +1,25 @@
...
+class TokenView(views.APIView):
+    """
+    Manages API tokens.
+    """
+    permission_classes = [permissions.IsAuthenticated]
+    serializer = TokenSerializer
+
+    def get(self, request):
+        token, created = Token.objects.get_or_create(user=request.user)
+        serializer = self.serializer(token)
```
Osmose: You should use `self.get_serializer(token)` here.

normandy/base/tests/test_api.py
```diff
@@ -0,0 +1,33 @@
...
+        assert res.data == {
+            'key': token.key,
+            'created': Whatever(),
+            'user': api_client.user.username,
+        }
+
+    def test_it_creates_new_tokens(self, api_client):
+        res = api_client.get('/api/v1/token/')
+        assert res.status_code == 201
+        assert res.data['user'] == api_client.user.username
```
Osmose: Might be nice to test that the token key from the response matches the token in the DB.

client/control/actions/authActions.js
```diff
@@ -0,0 +1,30 @@
+import { BASE_API_URL, API_REQUEST_SETTINGS } from '../constants.js';
+import { showNotification } from './ControlActions.js';
+
+export const AUTH_TOKEN_REQUEST_IN_PROGRESS = 'AUTH_TOKEN_REQUEST';
+export const AUTH_TOKEN_REQUEST_COMPLETE = 'AUTH_TOKEN_REQUEST_SUCCESS';
```
andymikulski: I personally don't love var names THAT_LOOK_LIKE_THIS, generally I'll do something like:

```
const authTokenRequest = {
  inProgress: 'AUTH_TOKEN_REQUEST',
  success: 'AUTH_TOKEN_REQUEST_SUCCESS',
};
// ...
..authTokenRequest.inProgress..
```
But this may be more of a personal preference thing

client/control/actions/authActions.js
```diff
@@ -0,0 +1,30 @@
...
+
+
+export function requestAPIToken() {
+  return async dispatch => {
```
andymikulski: I'm pretty sure the `async` is unnecessary here

client/control/actions/authActions.js
```diff
@@ -0,0 +1,30 @@
...
+    dispatch({ type: AUTH_TOKEN_REQUEST_IN_PROGRESS });
+    const response = await fetch(`${BASE_API_URL}token/`, API_REQUEST_SETTINGS);
+
+    if (response.status >= 400) {
+      dispatch({
+        type: AUTH_TOKEN_REQUEST_COMPLETE,
+        status: 'error',
+      });
+      dispatch(showNotification('Error fetching API token'));
+      throw await response.json();
```
andymikulski: This is interesting - I have not seen anyone just throw a JSON object on errors like this, but it makes sense!

client/control/actions/authActions.js
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    const token = await response.json();
```
andymikulski: Super nitpicky - since both paths (status >= 400 and status OK) call `await response.json()`, it's a liiiiittle DRYer to pull the `const token = await response.json();` declaration above the `if` statement, like so:

```
const jsonResponse = await response.json();
if(response.status >= 400){
  //...
  throw jsonResponse;
}

dispatch({
  // ..
  token: jsonResponse,
})
```

client/control/components/ControlApp.js
```diff
@@ -4,6 +4,8 @@ import Notifications from './Notifications.js';
 import DevTools from './DevTools.js';
 
 export default function ControlApp({ children, location, routes, params }) {
+  const appData = JSON.parse(document.querySelector('#app-data').textContent);
```
andymikulski: I don't _love_ how this works - feels a little hacky to print stuff on the page and then grub it inside JS to transfer data - but I can't immediately think of a better solution, so ¯\\\_(ツ)\_/¯

client/control/components/Header.js
```diff
@@ -32,4 +52,56 @@ Header.propTypes = {
...
+
+export default connect(
+  (state, ownProps) => ({
+    ...ownProps,
+    token: state.auth.token,
+  }),
+)(Header);
+
+
+class APITokenExpander extends Component {
```
andymikulski: In general, I almost never nest/declare multiple classes in the same file. Is this a component that may be used later, like in an admin panel or something? We may want to split this code up.

client/control/components/Header.js
```diff
@@ -32,4 +52,56 @@ Header.propTypes = {
...
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = { expanded: false };
+    this.toggleExpand = ::this.toggleExpand;
+  }
+
+  toggleExpand() {
+    this.setState(state => ({ ...state, expanded: !state.expanded }));
```
andymikulski: You don't need the `...state` operator, as React will create the new state object on its own. E.g.

```this.setState(state => ({ ...state, expanded: !state.expanded }));```
is equivalent to
```this.setState({ expanded: !state.expanded });```
(especially in this case, where `expanded` is the only property in state)

File: client/control/components/Header.js [hunk 12]
```diff
@@ -32,4 +52,56 @@ Header.propTypes = {
...
+  }
+
+  render() {
+    const { token } = this.props;
+    const { expanded } = this.state;
+
+    const expandButton = (
```
andymikulski: Fun performance fact: we can use a [babel transformer](https://babeljs.io/docs/plugins/transform-react-constant-elements/) to optimize cases like this. Not needed for this review, but just a fun fact for future reference

client/control/components/data/QueryAPIToken.js
```diff
@@ -0,0 +1,38 @@
...
+    token: pt.object,
+    requestAPIToken: pt.func.isRequired,
+  }
+
+  componentWillMount() {
+    this.request();
+  }
+
+  request() {
+    if (!this.props.tokenFetching && !this.props.token) {
```
andymikulski: Are there any cases of the token/Fetching to change after the app has been running for a bit? More specifically, do we need a `componentWillReceiveProps` function that compares `newProps.token` and `this.props.token` or something?

client/control/constants.js
```diff
@@ -0,0 +1,10 @@
+export const BASE_API_URL = '/api/v1/';
+
+export const API_REQUEST_SETTINGS = {
+  credentials: 'include',
+  headers: {
+    'X-CSRFToken': document.getElementsByTagName('html')[0].dataset.csrf,
```
andymikulski: This looks a little fragile - probably no way to define this value as a constant somewhere, right?

client/control/sass/control.scss
```diff
@@ -260,3 +269,12 @@
...
   }
 }
+
+/* Buttons */
+.button-icon {
+  background: none;
+  border: none;
+  cursor: pointer;
+  padding: 0;
+  transform: translate(0, -1px);
```
andymikulski: Why the 1px transform? Also, you need a `-moz-` prefix 😀

mythmon: The 1px transform is entirely visual. I thought it looked a little bit better raised up.

The `-moz-` prefix hasn't been needed [since Firefox 16](https://developer.mozilla.org/en-US/docs/Web/CSS/transform#Browser_compatibility). I think we can safely leave it off.

Plus, the preferred way for us to deal with prefixes is something like autoprefixer, which would just deal with all this.

client/control/components/Header.js
mythmon: The `...state` thing was a mistake. I had remembered that in order for state changes to be atomic they had to use the function form if they were based on the previous state, but I can't find any reference to that anymore, or reproduce it in a test, so I guess I'm just misremembering.

client/control/components/data/QueryAPIToken.js
mythmon: newProps.token would be the thing being requested, so it won't need to trigger a new request. It is possible that the API token is deleted on the server, and would need re-requested, but there isn't anyway for this component to tell if that is the case.

Having a `componentWillReceiveProps` that just calls `request` is probably a good idea though.

client/control/constants.js
mythmon: This will be different every page request, so it can't be a constant. We might be able to move it into the `app-data` script tag, but this chunk is just copy/paste to move this to a re-usable place, so I didn't bother.

normandy/base/tests/test_serializers.py
mythmon: We should write a lint for this

client/control/sass/control.scss
mythmon: Though... Now that I look again, it looks very different than I remember, so I'm just going to remove this altogether. I think I had something weird in my environment previously.

normandy/base/api/views.py
mythmon: `Exception Value: 'TokenView' object has no attribute 'get_serializer'`

normandy/base/templatetags/app_data.py
Osmose: The `takes_context` decorator on that doc page seems like it should do the trick.

normandy/base/api/views.py
Osmose: Oh, that method is on `GenericAPIView`, not `APIView`. Whoops.

mythmon:
Oops, it looks like this got closed instead of merged when we decided it was good. Should we resurrect this?

@andymikulski, @rehandalal: do you think this would help with #538?

mythmon:
In #526, we are probably going to lose the ability to do authentication ourselves, so we won't be able to use API keys like this. The user stuff is useful, but the API key stuff isn't needed anymore.

mythmon:
I think we got what we needed out of this PR. I'm going to re-close it.

PR #364, opened 2016-11-22
Updated TestRail test case ID's

chartjes:
We recently moved the test suite for Normandy around inside Testrail. I updated the case numbers and added ones for the performance tests.

r? @mythmon


mythmon:
The original test failure was the problem in #362. I rebased this on master.

PR #366, opened 2016-11-28
Fix #365: Stop writing to the Input Heartbeat API in show-heartbeat.

### LINKED ISSUE #365: Stop sending data to input.mozilla.org in show-heartbeat
We'll need to do this eventually anyway, and Input is consistently unable to handle the traffic from SHIELD recipes anymore, so let's just do it now.

The `show-heartbeat` action should no longer send any data to Input's Heartbeat API. We'll rely only on Telemetry for analyzing Heartbeat data. @MattGrimes already approved this over phone.

@rayborn @jtg567 Just an FYI, we're aiming to merge and deploy this change ASAP. Input's database will still be available, but it won't have any new data after we deploy. Let me know if you have any objections.

Osmose:
@mythmon r?


mythmon:
🔥🔥🔥

PR #357, opened 2016-11-21
#347, #348 Heartbeat Action Throttles

andymikulski:
Fixes #347 (Heartbeat should show once and only once)
Fixes #348 (Heartbeats should show once per 24 hours and that's it)

- Checks if heartbeat has ever been shown
- If it has NOT, it checks if a heartbeat has been shown in the last 24 hours
- If one has NOT, then the heartbeat is shown, and storage is updated with the last heartbeat time.

PR #368, opened 2016-11-29
Heartbeat throttling - #347 #348

### LINKED ISSUE #347: Change show-heartbeat timeout to be indefinite

### LINKED ISSUE #348: Only display one `show-heartbeat` prompt a day

andymikulski:
Fixes #347 (Heartbeat should show once and only once)
Fixes #348 (Heartbeats should show once per 24 hours and that's it)

* Checks if heartbeat has ever been shown
* If it has NOT, it checks if a heartbeat has been shown in the last 24 hours
* If one has NOT, then the heartbeat is shown, and storage is updated with the last heartbeat time.

(This PR replaces #357 due to big merge conflicts)

### DIFF
--- a/client/actions/show-heartbeat/index.js
+++ b/client/actions/show-heartbeat/index.js
@@ -24,12 +24,7 @@ export default class ShowHeartbeatAction extends Action {
    */
   async heartbeatShownRecently() {
     const lastShown = await this.heartbeatStorage.getItem('lastShown');
-    let timeSince = Infinity;
-
-    // If no heartbeat has ever been shown, lastShown will be falsey.
-    if (lastShown) {
-      timeSince = new Date() - lastShown;
-    }
+    const timeSince = lastShown ? new Date() - lastShown : Infinity;
 
     // Return a boolean indicating if a heartbeat
     // has shown within the last HEARTBEAT_THROTTLE ms
--- a/client/actions/show-heartbeat/index.js
+++ b/client/actions/show-heartbeat/index.js
@@ -18,6 +18,33 @@ export default class ShowHeartbeatAction extends Action {
     this.heartbeatStorage = normandy.createStorage('normandy-heartbeat');
   }
 
+  /**
+   * Calculates the number of milliseconds since the last heartbeat from any recipe
+   * was shown. If no heartbeat has ever been shown, returns Infinity, or "forever ago".
+   */
+  async timeSinceLastHeartbeat() {
+    const lastShown = await this.heartbeatStorage.getItem('lastShown');
+    let timeSince = Infinity;
+
+    // If no heartbeat has ever been shown, lastShown will be falsey.
+    if (lastShown) {
+      timeSince = new Date() - lastShown;
+    }
+
+    return timeSince;
+  }
+
+  /**
+   * Checks when this survey was last shown,
+   * and returns a boolean indicating if the
+   * user has seen this survey already or not.
+   */
+  async surveyHasShown() {
+    const lastShown = await this.storage.getItem('lastShown');
+    // If no survey has been shown, lastShown will be falsey.
+    return !!lastShown;
+  }
+
   async execute() {
     const {
       surveyId,
@@ -29,27 +56,10 @@ export default class ShowHeartbeatAction extends Action {
       learnMoreUrl,
     } = this.recipe.arguments;
 
-    const lastShown = await this.getLastShownDate();
-
-    // get the last shown for heartbeats in general
-    const lastHeartbeatShown = await this.getLastHeartbeatDate();
-
-    // we can show the heartbeat if there has not been one shown recently,
-    const canShowHeartbeat = lastHeartbeatShown === null
-    // or if the last one shown was more than HEARTBEAT_THROTTLE ms ago
-      || Date.now() - lastHeartbeatShown >= HEARTBEAT_THROTTLE;
-
-    // the survey should display itself if..
-    const shouldShowSurvey = (
-      // ..we're testing, or..
-      this.normandy.testing ||
-        // ..if this has never been shown..
-        (lastShown === null
-        // ..and we can show it, then we should
-        && canShowHeartbeat)
-    );
-
-    if (!shouldShowSurvey) {
+    if (!this.normandy.testing && (
+      await this.timeSinceLastHeartbeat() < HEARTBEAT_THROTTLE ||
+      await this.surveyHasShown()
+    )) {
       return;
     }
 
@@ -112,12 +122,12 @@ export default class ShowHeartbeatAction extends Action {
 
   async getLastShownDate() {
     const lastShown = await this.storage.getItem('lastShown');
-    return Number.isNaN(lastShown) ? 0 : lastShown;
+    return Number.isNaN(lastShown) ? null : lastShown;
   }
 
   async getLastHeartbeatDate() {
     const lastShown = await this.heartbeatStorage.getItem('lastShown');
-    return Number.isNaN(lastShown) ? 0 : lastShown;
+    return Number.isNaN(lastShown) ? null : lastShown;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


andymikulski:
@rehandalal @mythmon r?

client/actions/show-heartbeat/index.js
```diff
@@ -21,11 +30,25 @@ export default class ShowHeartbeatAction extends Action {
...
 
     const lastShown = await this.getLastShownDate();
+
+    // get the last shown for heartbeats in general
+    const lastHeartbeatShown = await this.getLastHeartbeatDate();
+
+    // we can show the heartbeat if there has not been one shown recently,
+    const canShowHeartbeat = lastHeartbeatShown === null
+    // or if the last one shown was more than HEARTBEAT_THROTTLE ms ago
+      || Date.now() - lastHeartbeatShown >= HEARTBEAT_THROTTLE;
```
rehandalal: this is kind of weird to me. I dont think we should have comments breaking up boolean expressions.

I'll let @mythmon weigh in too.

client/actions/show-heartbeat/index.js
```diff
@@ -21,11 +30,25 @@ export default class ShowHeartbeatAction extends Action {
...
     const shouldShowSurvey = (
-      this.normandy.testing
-      || lastShown === null
-      || Date.now() - lastShown > LAST_SHOWN_DELAY
+      // ..we're testing, or..
+      this.normandy.testing ||
+        // ..if this has never been shown..
+        (lastShown === null
+        // ..and we can show it, then we should
+        && canShowHeartbeat)
```
rehandalal: same thing here. the breakup is weird.

client/actions/show-heartbeat/index.js
```diff
@@ -21,11 +30,25 @@ export default class ShowHeartbeatAction extends Action {
...
+    // or if the last one shown was more than HEARTBEAT_THROTTLE ms ago
+      || Date.now() - lastHeartbeatShown >= HEARTBEAT_THROTTLE;
+
+    // the survey should display itself if..
     const shouldShowSurvey = (
-      this.normandy.testing
-      || lastShown === null
-      || Date.now() - lastShown > LAST_SHOWN_DELAY
+      // ..we're testing, or..
+      this.normandy.testing ||
```
rehandalal: I think the `||` should start out lines rather than end them, that's a little clearer.

client/actions/show-heartbeat/index.js
```diff
@@ -21,11 +30,25 @@ export default class ShowHeartbeatAction extends Action {
...
+        // ..if this has never been shown..
+        (lastShown === null
```
rehandalal: would this ever be `null` since you changed it below to `0`?

client/actions/show-heartbeat/index.js
```diff
@@ -73,16 +96,28 @@ export default class ShowHeartbeatAction extends Action {
...
   }
 
+  setLastHeartbeatDate() {
+    this.heartbeatStorage.setItem('lastShown', Date.now());
+  }
+
   async getLastShownDate() {
     const lastShown = await this.storage.getItem('lastShown');
-    return Number.isNaN(lastShown) ? null : lastShown;
+    return Number.isNaN(lastShown) ? 0 : lastShown;
```
rehandalal: nit: I think this should stay `null`.

client/actions/show-heartbeat/index.js
```diff
@@ -21,11 +30,25 @@ export default class ShowHeartbeatAction extends Action {
     } = this.recipe.arguments;
 
     const lastShown = await this.getLastShownDate();
+
+    // get the last shown for heartbeats in general
+    const lastHeartbeatShown = await this.getLastHeartbeatDate();
+
+    // we can show the heartbeat if there has not been one shown recently,
+    const canShowHeartbeat = lastHeartbeatShown === null
```
rehandalal: Would this ever be `null`?

client/actions/show-heartbeat/index.js
```diff
@@ -73,16 +96,28 @@ export default class ShowHeartbeatAction extends Action {
...
+
   async getLastShownDate() {
     const lastShown = await this.storage.getItem('lastShown');
-    return Number.isNaN(lastShown) ? null : lastShown;
+    return Number.isNaN(lastShown) ? 0 : lastShown;
+  }
+
+  async getLastHeartbeatDate() {
+    const lastShown = await this.heartbeatStorage.getItem('lastShown');
+    return Number.isNaN(lastShown) ? 0 : lastShown;
```
rehandalal: I think this should also return `null` if `isNaN`.

client/actions/show-heartbeat/index.js
mythmon: I think it is a little thing, and it doesn't matter too much. I'd put the comments all together above the entire expression though. Plus, it seems a little odd to pull out `canShowHeartbeat` into its own expression, instead of just having it be a subexpression of `shouldShowSurvey`.

I think that the complexity of this logic is a sign that it could be better abstracted.

Maybe it makes sense to move some more logic into the helper functions. Here I'm using `Infinity` as a trick to represent never, though maybe it would make sense to even move the comparisons into the helpers?

```js
if (!testing && (
  await this.timeSinceLastHeartbeat() < HEARTBEAT_THROTTLE ||
  await this.timeSinceLastSurvey() < LAST_SHOWN_DELAY
)) {
  return;
}

...

/* Calculates the number of milliseconds since the last heartbeat from any recipe
 * was shown. If no heartbeat has ever been shown, returns Infinity, or "forever ago".
 */
async timeSinceLastHeartbeat() {
  const lastShown = await this.heartbeatStorage.getItem('lastShown');
  // If no heartbeat has ever been shown, lastShown will be falsey.
  if (lastShown) {
    return new Date() - lastShown;
  } else {
    return Infinity;
  }
}

/* Calculates the number of milliseconds since the last survey from this recipe
 * was shown. If no survey has ever been shown, returns Infinity, or "forever ago".
 */
async timeSinceLastSurvey() {
  const lastShown = await this.storage.getItem('lastShown');
  // If no survey has been shown, lastShown will be falsey.
  if (lastShown) {
    return new Date() - lastShown;
  } else {
    return Infinity;
  }
}
```

andymikulski:
@mythmon @rehandalal Thanks for the feedback! I've committed the corrections, as well as abstracted the comparison logic into their own helper functions (a la @mythmon 's example).

𝓡?

client/actions/show-heartbeat/index.js
```diff
@@ -1,12 +1,50 @@
...
+   * was shown. Returns a boolean indicating if a heartbeat has been shown recently.
+   */
+  async heartbeatShownRecently() {
+    const lastShown = await this.heartbeatStorage.getItem('lastShown');
+    let timeSince = Infinity;
+
+    // If no heartbeat has ever been shown, lastShown will be falsey.
+    if (lastShown) {
+      timeSince = new Date() - lastShown;
+    }
```
rehandalal: you could simplify this to:
```js
const timeSince = lastShown ? new Date() - lastShown : Infinity;
```

PR #376, opened 2016-12-01
Fix #375: Make LocalStorage options actually optional.

### LINKED ISSUE #375: v26 tests failing
https://bugzilla.mozilla.org/show_bug.cgi?id=1321409

I've hunted down the origin of the `ERROR RUNNING RECIPE: TYPEERROR: N IS UNDEFINED` message, but there still exists a deeper issue.

The initial fix is to correct how the param destructuring/default assignment works in `LocalStorage#constructor` - 

```
  constructor(prefix, opts) {
    // build an options object, even if
    // no options have been passed in
    const options = { ...opts };

    this.prefix = prefix;
    this.skipDurability = options.skipDurability || false;
  }
```

The specific error is that `_ref` is undefined, which is from babel's way of handling destructuring.
Basically, `someFn({ woo })` is converted down into `someFn(_ref){ const woo = _ref.woo; ....` - which means if no object is passed into the function, an error is thrown (since `_ref` doesnt exist to extract `woo` from)

BUT with that fixed, I'm now running into a deeper issue which seems to dive into non-JS code:

screen shot 2016-12-01 at 8 40 30 am

The source map points to the `.catch`'s console.log, and the _only_ place where `normandy.saveHeartbeatFlow` exists is stubbed out in some tests. Not really sure where to go from here - @mythmon any thoughts?

Osmose:
@andymikulski r? You'll probably want to test this manually as well since you were running into issues with your manual testing before.

### DIFF
--- a/client/selfrepair/normandy_driver.js
+++ b/client/selfrepair/normandy_driver.js
@@ -11,7 +11,7 @@ export class LocalStorage {
   /**
    * @param    {string} prefix Prefix to append to all incoming keys.
    */
-  constructor(prefix, { skipDurability = false }) {
+  constructor(prefix, { skipDurability = false } = {}) {
     this.prefix = prefix;
     this.skipDurability = skipDurability;
   }
--- a/client/actions/tests/utils.js
+++ b/client/actions/tests/utils.js
@@ -63,9 +63,6 @@ export function mockNormandy() {
     },
     // this needs to be a valid UUID4 to pass regex tests later
     userId: uuid(),
-    saveHeartbeatFlow() {
-      return Promise.resolve();
-    },
   };
 
   const toSpy = [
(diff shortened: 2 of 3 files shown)


Osmose:
Oh yeah, I also included a test for when `skipDurability` is false, since it is possible even if unused currently.

client/selfrepair/normandy_driver.js
```diff
@@ -11,7 +11,7 @@ export class LocalStorage {
   /**
    * @param    {string} prefix Prefix to append to all incoming keys.
    */
-  constructor(prefix, { skipDurability = false }) {
+  constructor(prefix, { skipDurability = false } = {}) {
```
andymikulski: TIL

Issue #363, opened 2016-11-22
Simplified Filters - Front End Tasks


andymikulski:
Tracking front end tasks for #316 here

- [x] Columns dropdown menu
- [x] Customize show/hiding columns
- [x] Column layout should be stored locally
- [ ] Combo box (text search)
* - [x] ...filters group dropdown
* - [ ] ...adds a text filter upon hitting enter
- [x] Combo box (dropdown selection for filters)
- [x] Active filters section...
* - [x] ...lists filter state
* - [x] ...clicking removes filter
* - [x] ...actually filters list
* - [x] ...has reset filters button
- [x] Last filter view is saved locally (refreshing does not wipe filters)
- [ ] Recipe page Filters section (Searchable multiselect boxes, etc)
- [ ] Connect stores to API
- [ ] Tests

Nice to haves:
- [ ] Formatted/displayed text to match raw API values (e.g. `de` displays as `German (de)`)
- [ ] Keyboard accessibility

andymikulski:
Attaching wireframes here for easier reference

![shield simple filters final-1](https://cloud.githubusercontent.com/assets/2767162/20534787/bdc6b056-b09f-11e6-96dd-9a8a69b866e3.jpg)

![shield simple filters final-2](https://cloud.githubusercontent.com/assets/2767162/20534792/c082dc48-b09f-11e6-9409-f7d94a5272c6.jpg)

![shield simple filters final-3](https://cloud.githubusercontent.com/assets/2767162/20534794/c21277bc-b09f-11e6-97a0-dc69e525156b.jpg)

![shield simple filters final-4](https://cloud.githubusercontent.com/assets/2767162/20534798/c3d1f460-b09f-11e6-8051-27290a3235c8.jpg)

![shield simple filters final-5](https://cloud.githubusercontent.com/assets/2767162/20534799/c5419f80-b09f-11e6-8f04-6494ddd8d641.jpg)

![shield simple filters final-6](https://cloud.githubusercontent.com/assets/2767162/20534802/c6abccba-b09f-11e6-9e34-c9fa6bac3645.jpg)

andymikulski:
hey @Osmose @mythmon regarding the [search/filter combo box](https://cloud.githubusercontent.com/assets/2767162/20534798/c3d1f460-b09f-11e6-8051-27290a3235c8.jpg), if the user types into the text input, does that filter the dropdown menu? Or is it strictly for filtering rows via text search?

Osmose:
[...]

Issue #316, opened 2016-11-08
Add simplified filter controls for recipes


Osmose:
**Blockers**

* [x] #395 - Active filters
* [ ] #396 - Recipe controls

---

We previously stored filter data in the database, and had a built-out form for specifying who recipes should be delivered to. We then switched to using JEXL filter expressions for all filtering, which has a lot of benefits around adding new filters and composing them together. But the UX is still not so good.

We'd like to add some simple filtering back in that gets compiled down to JEXL that is appended to the filter expression. Things we'd like to filter by include:

- Firefox locale
- Geolocated country
- Release channel
- Start and End Time

Along with targetting recipes by these, we'd like to be able to filter by most of these in the recipe listing, and show them in columns in the recipe listing.

Osmose:
I've created a mockup of the UX changes for adding the simple filters: [SHIELD Simple Filters.pdf](https://github.com/mozilla/normandy/files/593123/SHIELD.Simple.Filters.pdf).

Osmose:
@mozilla/strategy-and-insights I'd like to get some feedback on the mockup in the comment above before we start working on it. It's depicts the changes to the recipe listing and recipe form to make common filters easier to enter. The filter expression box will still be available for more complex filters, and can be used in conjunction with simple filters.

What do ya'll think? Any questions?

TylerDowner:
I think it looks great. No specific comments.

raymak:
It looks good. One basic question/comment I have is how can you know that a certain recipe has additional filter expression when you are looking at it in the list?

Osmose:
> It looks good. One basic question/comment I have is how can you know that a certain recipe has additional filter expression when you are looking at it in the list?

Maybe a checkmark/X column flagged on whether the additional filter is empty or not?

raymak:
> It looks good. One basic question/comment I have is how can you know that a certain recipe has additional filter expression when you are looking at it in the list?
> 
> Maybe a checkmark/X column flagged on whether the additional filter is empty or not?

That would be good. Some indication that there's more and this is not all. Then when you click on it it takes you somewhere (preview?) where you can see the additional filters too.

jtg567:
What determines the content of the **Metadata** column? I don't see input controls.

Will it always be empty when action is console-log, and contain **Survey ID:** when action is show-heartbeat?

MattGrimes:
Other than addressing Kamyar's concern, I think it looks great. Very useful.

Osmose:
[...]

PR #380, opened 2016-12-02
#316 Simplified Filters (front end)

andymikulski:
Progress on #363 (and #316 indirectly)

Adds `Columns` dropdown menu
- Columns are shown/hidden based on user input
- Column setups are saved locally (refreshing displays the same column setup you were just looking at)

Adds filtering combobox
- Text searches over group and options
- Support for filters with multiple options (e.g. locales, releases)

Adds "active filters" section
- Filter setups are saved locally (refreshing displays the same filter setup you were just looking at)
- 'Reset filters' button


mythmon:
That is a lot of commits. I don't think this level of granularity is useful when reviewing code, and it is actively harmful when we are doing deploys and reading the git log. Can you squash these down to fewer commits? I'd imagine one would be ideal, but if there are a few big chunks of work two or three commits might be OK (I haven't looked at the changes yet).

Additionally, before publishing a PR you should prefer rebasing onto master, instead of merging master into your branch. It is useful to use merges to update a branch *after* review has started, since it avoids rewriting published history. Before the changes are published though, there isn't any benefit to avoiding rewriting history.

andymikulski:
closing in favor of smaller PRs

PR #382, opened 2016-12-07
Add stylelint

andymikulski:
- Adds SASS linter to ~~Webpack~~ npm scripts, therapist.
- Uses [stylelint-config-standard](https://github.com/stylelint/stylelint-config-standard), with some additional rule changes (see `.stylelintrc`)
- Uses [stylelint-order](https://github.com/hudochenkov/stylelint-order) to force property order (see `.stylelintrc`) - basic order is `[@import/include/extend, custom properties, $variables, styles, nested rules, remaining @'s (media queries etc)]`
- Uses stylelint-order to enforce alphabetical property order
- If stylelint fails when building for production, an error is thrown and the build is halted.

Currently, the linter only runs when starting up webpack - it seems watching doesn't work for stylelint. This should be fine, though not so ideal.

@rehandalal @mythmon ℝ?

Edit: there are a ton of [stylelint rules](http://stylelint.io/user-guide/rules/) - if there's anything that we should include, just comment it here and I'll add it


.stylelintrc
```diff
@@ -0,0 +1,28 @@
...
+      },
+      "custom-properties",
+      "dollar-variables",
+      "declarations",
+      "rules",
+      "at-rules",
+    ],
+    "order/declaration-block-properties-alphabetical-order": true,
+  }
+}
```
rehandalal: nit: newline

webpack.config.js
```diff
@@ -34,10 +35,17 @@ if (production) {
         warnings: false,
       },
     }),
+    new webpack.NoErrorsPlugin(),
```
rehandalal: I'm curious about why we needed this?

andymikulski: Whoops!

mythmon:
I'm really stoked to get CSS lints. I hadn't seen stylelint before, and it looks like exactly what I'd like to see out of a lint tool (flexible config, plug-ins, auto-fixing).

As an implementation detail, I don't understand why to put this in webpack and only run it for prod builds. Our other lint passes (eslint and flake8) are separate tools that run as a part of therapist. Why did you choose a different route for this check?

andymikulski:
@mythmon Currently, our sass is compiled through Webpack and not through a CLI command or anything, so I was looking for a loader or plugin that we could drop in easily. However, in order to add sass linting to the `npm run lint` script, I'll have to add the stylelint command line tool anyway. I was also hoping that by adding the plugin to webpack, it'd get picked up in `watch` mode - but, that's not the case.

Which reminds me - @rehandalal, currently therapist straight up calls `eslint`, correct? Could that change so it runs `npm run lint` instead? That way we can bundle all the lint commands in our `package.json` and therapist could catch them all.

rehandalal:
it definitely could but i'd rather it didn't for a number of reasons

andymikulski:
Oh, actually, turns out that the sass linting DOES get picked up with `--watch`:

screen shot 2016-12-08 at 8 33 49 am

andymikulski:
@mythmon Ready for review.

I've switched the stylelint stuff from using webpack to using the CLI. The stylelint docs (and related issues) all indicate to install with the `-g` flag, probably so npm can handle adding the command to your PATH and all that.

To remedy the `stylelint: command not found` issue, I used a `bin` key in `package.json`, which locally sets the `stylelint` command and lets us use the CLI in things like `npm run lint-css`. Let me know if this works for you.

package.json
```diff
@@ -6,11 +6,16 @@
     "type": "git",
     "url": "git://github.com/mozilla/normandy.git"
   },
+  "bin": {
+    "stylelint": "./node_modules/stylelint/dist/cli.js"
+  },
```
andymikulski: I'm confused - are you suggesting we don't need this bin line at all and `stylelint` should already work without it? Or that we should swap out the `stylelint` in `lint-css` with `node_modules/.bin/stylelint`?

package.json
```diff
@@ -6,11 +6,16 @@
...
   "scripts": {
     "watch": "webpack --config ./webpack.config.js --watch",
     "build": "webpack --config ./webpack.config.js",
     "test": "NODE_ENV=production karma start",
-    "lint": "eslint normandy client webpack.config.js karma.conf.js"
+    "lint": "npm run lint-js; npm run lint-css;",
+    "lint-js": "eslint normandy client webpack.config.js karma.conf.js --fix || true",
+    "lint-css": "stylelint '**/*.s?(a|c)ss' --config .stylelintrc || true"
```
rehandalal: I think instead of putting this here we should just handle all of this in therapist. You'd need to update the therapist.yml file to:
```yaml
actions:
  flake8:
    run: flake8 {files}
    include: "*.py"
    exclude: "docs/"
  eslint:
    run: ./node_modules/.bin/eslint {files} --fix
    include: 
      - "*.js"
      - "*.jsx"
  stylelint:
    run: ./node_modules/bin/stylelint {files}
    include:
      - "*.css"
      - "*.scss"
```

rehandalal: There are several advantages to doing this in therapist:
- It integrates into CI
- It allows you to lint only the files you are changing
- It allows you to lint all the files included in the project (but not all the files in `node_modules` etc.)
- It keeps our linting consistent

andymikulski:
Added to therapist re: @rehandalal 's suggestion. The npm scripts `lint`/`lint-js`/`lint-css` are still there to use while dev'ing, as well.

.therapist.yml
```diff
@@ -6,3 +6,8 @@ actions:
   eslint:
     run: ./node_modules/.bin/eslint {files}
     include: "*.js"
+  stylelint:
+    run: ./node_modules/.bin/stylelint {files}
+    include:
+      - "client/**/*.css"
+      - "client/**/*.scss"
```
rehandalal: you should just have:
```yaml
- "*.css"
- "*.scss"
```
Therapist will take care of the rest.

.therapist.yml
```diff
@@ -6,3 +6,8 @@ actions:
   eslint:
     run: ./node_modules/.bin/eslint {files}
     include: "*.js"
```
rehandalal: We should update this to:
```yaml
- "*.js"
- "*.jsx"
```

package.json
```diff
@@ -6,11 +6,16 @@
     "type": "git",
     "url": "git://github.com/mozilla/normandy.git"
   },
+  "bin": {
```
rehandalal: I don't think you should need this at all? I might be wrong, but your lint stuff should still work without it.

andymikulski: Figured it out - I was convinced I needed this because I couldn't run `stylelint` from the command line without it. npm runs it just fine though, so I don't even need it.

PR #383, opened 2016-12-08
Update reactable to 0.14.1

andymikulski:
Continuation of #317 - removes `Warning: Unknown prop 'props' on <td> tag. Remove this prop from the element` warning in console.


mythmon:
I re-pushed this branch to try and get CircleCI to pick it up. That doesn't seem to have helped.

PR #386, opened 2016-12-15
Update reactable to 0.14.1

andymikulski:
Be sure to run `npm update` after pulling this update locally!


andymikulski:
@rehandalal It looks like there was an issue on the `therapist` step - specifically installing it. Did you change something on Circle wrt therapist?

PR #387, opened 2016-12-15
Sass Linting

andymikulski:
Continuation of #382


andymikulski:
I'm pretty sure this got the stamp approval in #382 but if @rehandalal or @mythmon want to double check it, please feel free!

package.json
```diff
@@ -83,6 +85,9 @@
     "redux-mock-store": "1.1.4",
     "sass-loader": "4.0.1",
     "style-loader": "0.13.1",
+    "stylelint": "^7.6.0",
+    "stylelint-config-standard": "^15.0.0",
+    "stylelint-order": "^0.2.2",
```
rehandalal: we've had issues with not specifying exact versions. i would pin these.

package.json
```diff
@@ -79,15 +79,15 @@
     "postcss-loader": "0.11.1",
     "react-addons-test-utils": "15.2.1",
     "react-tools": "0.13.3",
-    "redux-devtools": "^3.3.1",
-    "redux-devtools-dock-monitor": "^1.1.1",
-    "redux-devtools-log-monitor": "^1.0.11",
+    "redux-devtools": "3.3.1",
+    "redux-devtools-dock-monitor": "1.1.1",
+    "redux-devtools-log-monitor": "1.0.11",
```
rehandalal: oh, good catch on these other ones!

rehandalal:
thanks for sticking with this! so stoked to have this in! nice job!!

PR #384, opened 2016-12-09
Customizable columns

andymikulski:
- Adds column menu
- Saves/loads columns locally
- Adds ColumnMenu, CheckboxList, DropdownMenu
- Refactors RecipeFilters into its own component
- Adds Column redux store, along with actions and selectors

TODO: Tests

PR #388, opened 2016-12-15
Customizable Columns

andymikulski:
Continuation of #384

Adds menu to UI to allow user to customize what columns are shown/hidden, and saves the configuration locally.

- Adds Column Actions, Column Reducer, and Column Selectors (+ tests for all)
- Adds CheckboxList, ColumnMenu, DropdownMenu, RecipeFilters components
- Refactors RecipeList component
- Uses localForage to save/load column state locally
- Compares loaded columns against those hardcoded in the app - if there's a mismatch, the column config is thrown out and the default is shown. This will prevent users from seeing broken columns if we end up changing the column setup later.
- Sets up ground work for more advanced filtering in upcoming PRs

@rehandalal @mythmon R?


client/control/reducers/ColumnReducer.js
```diff
@@ -0,0 +1,91 @@
...
+  saveLocalColumns as saveState,
+} from 'control/actions/ColumnActions';
+
+import {
+  compare,
+} from 'client/utils/hash';
+
+import cloneArrayValues from 'client/utils/clone-array-values';
+
+// this could be sexier
```
mythmon: Lets avoid describing code as sexy.

client/control/sass/partials/_common.scss
```diff
@@ -22,27 +24,29 @@ a:hover {
...
-/* Form Elements */
-label, .label {
-  color: rgba($darkBrown, 0.6);
-  display: block;
-  font: normal normal 400 12px/16px $OpenSans;
-  padding: 5px 0px;
-  text-transform: uppercase;
-}
+// .recipe-form {
+    /* Form Elements */
```
mythmon: nit: I think the indentation got messed up here?

client/control/sass/partials/_common.scss
```diff
@@ -22,27 +24,29 @@ a:hover {
...
-  &:focus {
-    border-color: $darkestCream;
-    outline: 0;
+    &:focus {
+      border-color: $darkestCream;
+      outline: 0;
+    }
   }
-}
+// }
```
mythmon: Oops?

client/control/selectors/ColumnSelector.js
```diff
@@ -0,0 +1,28 @@
...
+ * Given a set of columns, returns an array of
+ * disabled only (hidden) columns
+ *
+ * @param  {Array<Object>} columns Initial set of columns to look through
+ * @return {Array<Object>}         Columns that are disabled/hidden from the user
+ */
+export const getInactiveColumns = columns =>
+  [].concat(columns)
+    .map(col => (!col.enabled ? col : null))
+    .filter(col => col);
```
mythmon: I think we should probably phrase this as `export function getInactiveColumns(columns)`. Unless there is a compelling reason to use the `const ... = () =>` pattern?

client/control/tests/actions/test_columnActions.js
```diff
@@ -0,0 +1,88 @@
...
+              expect(expectedColumns[0].testProperty).toEqual('should exist');
+              // rest of the columns should still match
+              expect(store.getActions()).toContain({
+                type: LOAD_SAVED_COLUMNS,
+                columns: expectedColumns,
+              });
+              // done testing
+              done();
+            })
+        );
```
mythmon: I think this would be clearer as an async function, so we can use await. (We can use those, right?).

client/control/tests/reducers/test_columnReducer.js
```diff
@@ -0,0 +1,129 @@
...
+  initialState,
+} from 'control/tests/fixtures';
+
+/**
+ * Utility function to clone the initialState column array and its values
+ * (This prevents nested objects from being cloned by reference)
+ *
+ * @return {Array} New array with cloned initialState values
+ */
+const cloneInitialArray = () => JSON.parse(JSON.stringify(initialState.columns));
```
mythmon: Ditto on the fat-arrow function here.

client/control/tests/reducers/test_columnReducer.js
```diff
@@ -0,0 +1,129 @@
...
+ */
+const getNondefaultColumn = () => {
+  let foundIndex;
+  initialState.columns
+    .find((col, index) => {
+      foundIndex = index;
+      return !col.enabled;
+    });
+
+  if (!foundIndex) {
```
mythmon: Should this compare specifically to undefined, since `0` is a valid value? Also, should this be an exception instead of just printing a warning? Test code probably shouldn't be tolerant of weird cases.

client/control/tests/reducers/test_columnReducer.js
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * Column reducer tests
+ */
+describe('Column reducer', () => {
+  it('should return initial state by default', () => {
+    expect(appReducer(undefined, {})).toEqual(initialState);
+  });
+
+  it('should have default columns to display', () => {
```
mythmon: I'm not sure I see the value of this test. Is there a reason we need to assert what the default columns are?

client/control/tests/reducers/test_columnReducer.js
```diff
@@ -0,0 +1,129 @@
...
+    // and their shorthand values
+    const foundDefaults = cloneInitialArray()
+      .filter(col => col.enabled)
+      .map(col => col.value);
+
+    // arrays of default enableds should match
+    expect(foundDefaults).toEqual(defaults);
+  });
+
+  it('should handle UPDATE_COLUMN (enable)', () => {
```
mythmon: I think this and the next test should be nested in a describe, like  you did for `LOAD_SAVED_COLUMNS`.

client/control/tests/reducers/test_columnReducer.js
```diff
@@ -0,0 +1,129 @@
...
+      isActive: true,
+    })).toEqual({
+      ...initialState,
+      columns: expectedColumns,
+    });
+  });
+
+  it('should handle UPDATE_COLUMN (disable)', () => {
+    const expectedColumns = cloneInitialArray();
+    expectedColumns[0].enabled = false;
```
mythmon: This assumes the first column will always be a default column. Is that a safe assumption?

client/control/tests/reducers/test_columnReducer.js
mythmon: Also, should this just import and use `client/utils/clone-array-values.js`?

client/utils/hash.js
```diff
@@ -0,0 +1,31 @@
+const memory = {};
+
+export default function hash(obj) {
```
mythmon: 1. Is this used anywhere? I don't see an import for it.
2. We have `sha.js` as a dependency already, so we should use that for hashing instead of writing our own.

client/utils/hash.js
```diff
@@ -0,0 +1,31 @@
...
+    }
+  }
+  return foundHash;
+}
+
+export function compare(base, fork) {
+  const baseHash = hash(base);
+  const forkHash = hash(fork);
+
+  return baseHash === forkHash;
```
mythmon: Is there any reason this couldn't just be

```js
return JSON.stringify(base) === JSON.stringify(fork);
```

I'd expect that to act the same as the hash version, and also be faster.

mythmon:
@rehandalal I think it would be useful for you to go over this too In particular, here are some webcompat things I'm not too familiar with, and I didn't really try to review the sass stuff. (Plus if you see anything else elsewhere, that's cool too)

client/control/components/DropdownMenu.js
```diff
@@ -0,0 +1,142 @@
...
+   * based on the `shouldBind` param.
+   *
+   * @param  {boolean} shouldBind Should the event be attached?
+   * @return {void}
+   */
+  updateWindowBinding(shouldBind) {
+    if (shouldBind) {
+      document.body.addEventListener('click', this.onMenuBlur, true);
+    } else {
+      document.body.removeEventListener('click', this.onMenuBlur);
```
andymikulski: The third argument (`useCapture`) dictates how the handler affects event bubbling, in order to unbind the event we need to call `removeEventListener`.

client/control/components/DropdownMenu.js
```diff
@@ -0,0 +1,142 @@
...
+  toggleVisibility(force) {
+    // if we get a parameter, use that instead of just straight up toggling
+    const newVisibleState = typeof force === 'boolean' ? force : !this.state.isVisible;
+
+    this.setState({
+      isVisible: newVisibleState,
+    });
+
+    // add or remove the event based on visibility
+    this.updateWindowBinding(newVisibleState);
```
andymikulski: I think you're missing some code - `updateWindowBinding` should have this in it:

```updateWindowBinding(shouldBind) {
    if (shouldBind) {
      document.body.addEventListener('click', this.onMenuBlur, true);
    } else {
      document.body.removeEventListener('click', this.onMenuBlur);
    }
```

client/control/components/RecipeFilters.js
```diff
@@ -0,0 +1,83 @@
...
+
+  /**
+   * Loads user's last column display setup,
+   * as well as their last active list filters.
+   *
+   * @return {void}
+   */
+  componentWillMount() {
+    // load the last column setup user was viewing
+    this.props.dispatch(loadLocalColumns());
```
andymikulski: I had seen all the actions go through `this.props.dispatch`, I have no problem switching it over though

client/control/components/DropdownMenu.js
mythmon: Oh, oops. I misread this and thought it was calling `addEventListener` both times, and just not passing the 3rd argument sometimes.

client/control/tests/actions/test_columnActions.js
andymikulski: I was having issues getting async/await to work with the action/localForage, I can dig into it more, though.

mythmon: Barring async/await, a less nested Promise chain would  be easier to follow. If that doesn't work, it makes me wonder if we actually want to be using localForage, as nice as it is.

client/control/tests/reducers/test_columnReducer.js
andymikulski: My thinking was that it would prevent a case where the columns are broken or have no defaults, at which point the table wouldn't render at all. I'll remove the test, though.

client/control/tests/actions/test_columnActions.js
```diff
@@ -36,26 +36,24 @@ describe('Column Actions', () => {
...
 
       const testState = { test: true };
 
-      saveLocalColumns(testState, () => {
-        localForage.getItem(localStorageID, (err, found) => {
-          expect(err).toBeNull();
+      await saveLocalColumns(testState);
+      await localForage.getItem(localStorageID)
+        .then(found => {
           expect(found).toEqual(testState);
```
mythmon: Might as well go all in on await, and not use `.then`.

```js
const found = await localForage.getItem(localStorageID);
expect(found).toEqual(testState);
```

client/control/components/CheckboxList.js
```diff
@@ -3,26 +3,61 @@ import React, { PropTypes as pt } from 'react';
 /**
  * Simple component which lists a bunch of checkboxes,
  * and emits input changes.
+ *
+ * @prop {Function} onInputChange
+ *   Fires when user has changed a checkbox, with
+ *   params (Item index, Checkbox 'checked' status)
```
mythmon: This is value now not index, right?

client/control/components/CheckboxList.js
```diff
@@ -3,26 +3,61 @@ import React, { PropTypes as pt } from 'react';
...
+    this.state = {};
+
+    // Cache of generated event handlers
+    this.handlerCache = {};
+  }
+
   /**
-   * Handler factory. Given an index,
+   * @param  {[type]} newProps [description]
+   * @return {[type]}          [description]
```
mythmon: Looks like auto generated code that needs updated? Or I'd be fine with just removing it. I don't think `componentWillReceiveProps` needs a docblock.

client/control/components/ColumnMenu.js
```diff
@@ -41,16 +41,14 @@ class ColumnMenu extends React.Component {
    * simply updates the component's column state,
    * and notifies the parent of what's selected
    *
-   * Note: it seems a little fragile to use the index for this,
-   * may be useful to switch to something more identifying.
    *
-   * @param  {Number}  columnIndex Integer, index of relevant column
-   * @param  {Boolean} isActive    Is the column now active?
+   * @param  {String}  columnValue  Value of the column ('name', 'endTime', etc)
```
mythmon: Using "value" for this seems a bit strange. It makes sense in the context of check boxes, because HTML forms are odd, but this is a pretty different context. I think this will end up being confusing in the future. Can we use another name? For the human readable one, maybe something like "display", "label", or "title", and for the machine readable one "slug" or "id"?

(If you're not familiar with the term slug, it is sort of a Django-ism. Usually it is just a url-ified version of a title, like "Hello, World!" becoming "hello-world")

client/control/components/CheckboxList.js
andymikulski: Yeah, in retrospect we don't need the `componentWillReceiveProps` at all - the cache is keyed on column slug, so if the options update, the cache is still relevant.

client/control/components/ColumnMenu.js
andymikulski: Yeah, the `value` name had come from the fact that was what the checkbox values were, but `slug` makes more sense! Should have switched it over earlier.

client/control/components/ColumnMenu.js
```diff
@@ -29,6 +29,27 @@ class ColumnMenu extends React.Component {
     </span>
   );
 
+  /**
+   * Given an object, adds necessary properties to
+   * be displayed in a CheckboxList.
+   *
+   * Basically just adds a `value` property from
+   * the col's `slug`. We need this as a function due to some
+   * weird linter issue WRT anon functions returning objects.
```
andymikulski: I actually figured it out shortly after while working on filters - I was trying to return an object from an anon function, like so:
`whatever={ ()=>{ my: 'object' } }`

but the linter would complain unless the returned object is wrapped in parens:
`whatever={ ()=>({ my: 'object' }) }`

mythmon: Yay the linter actually saved us here.

I looked in to this some more, and it turns out that `() => { my: 'object' }` is actually a function that executes the statement `my: 'object'`, which is a string (which does nothing), and a label, which is this one weird trick we inherited from C. Not a function that returns an object.

Issue #395, opened 2016-12-19
Simple filters: Active Filters


andymikulski:
<img width="703" alt="screen shot 2016-12-19 at 4 02 46 pm" src="https://cloud.githubusercontent.com/assets/2767162/21332177/037b296e-c604-11e6-8daa-8328d3fe2647.png">

Mockups can be found here - https://github.com/mozilla/normandy/issues/316#issuecomment-260774698

- [x] Load available filters (locales, names, etc)...
- [x] ...from API
- [x] Display filters in a dropdown list
- [x] Filter dropdown can be filtered by text input
- [x] Clicking a dropdown item (filter) enables that filter
- [x] Typing text into the combobox and hitting 'enter' enables a text filter
- [x] Clicking on an activated filter disables that filter
- [x] Reset button clears all filters
- [x] Upon selecting filter, new recipe list is fetched from API
- [x] Loading message is displayed while loading new recipe list from API
- [x] Cached recipe list is displayed until a new update comes in
- [x] Tests

Bonus objectives:
- [ ] URL updates with filter config (so filter setups can be shared easily)
- [ ] "Displaying _X_ of _Y_ items" text

Reading material:
- [Gist: format for returned filters](https://gist.github.com/andymikulski/e7f281477917bad0183de58127817de5)

andymikulski:
@rehandalal Any thoughts on how the API should consume the various filter params? I'm currently stubbing out code that hits this endpoint: `/api/v1/recipe/?status=disabled&channel=release,aurora&locale=en-uk` - how does that look to you?

rehandalal:
that looks good to me.

i agree about only allowing one text filter at a time.

PR #401, opened 2016-12-22
#395 Active Filters

andymikulski:
Related to #395 

- Adds filter reducer, actions, selectors
- Adds 'active filter' section (ActiveFilters component)
- Adds RecipeCombobox, RecipeCount, ActiveFilters, GroupMenu components
- Adds filter API calls (_not_ the actual API responses, though)
- Adds recipe list caching (user adds filter -> if that filter was loaded already, that list is shown -> api call is still made -> list is updated with fresh server results)
- Adds filter tests
- Adds a handful of missing util tests

NOTE: I'd like @mythmon or @rehandalal to check out the tests, specifically the API tests (the `LOAD_FILTERS` stuff in `test_filterActions`) - I had copied other API tests as an example, but I'm not totally convinced it's as thorough as it could/should be. After that, I'll request a QA review.


andymikulski:
I'd like to refactor some of `client/control/actions/ControlActions` before becomes a bit monolothic, but that may turn into another PR. Either way, at this point, this PR is ready for review.

andymikulski:
To test the filters locally without an endpoint, use this gist to replace the `loadFilters` action - https://gist.github.com/andymikulski/8536d1218343446262cdade1790ebc59

mythmon:
This is a pretty large patch (necessarily so, I think). I didn't look through all of it, but from what I've read, I'm concerned about landing this on master without the APIs to match. I think that it will be a better idea to create a feature branch, and land this PR on it, as well as the APIs for filter options and recipe filtering.

The main concern here is that this feature is not ready for production (since it relies on APIs that are not written), and the goal is that master is always ready for production (at least in theory). As naive an idea as that is, I think it is still a good goal to strive for.

Can we recreate this PR to merge into a feature branch instead of master? Or do you think that's a crazy idea?

andymikulski:
Yep, that makes sense, I'll go ahead and close this PR etc.

andymikulski:
New PR onto the `316-simplified-filters` feature branch is located at #405

PR #405, opened 2017-01-04
#316 Active Filters Component

### LINKED ISSUE #395: Simple filters: Active Filters

andymikulski:
Fixes #395 

-    Adds filter reducer, actions, selectors
-    Adds 'active filter' section (ActiveFilters component)
-    Adds RecipeCombobox, RecipeCount, ActiveFilters, GroupMenu components
-    Adds filter API calls (not the actual API responses, though)
-    Adds recipe list caching (user adds filter -> if that filter was loaded already, that list is shown -> api call is still made -> list is updated with fresh server results)
-    Adds filter tests
-    Adds a handful of missing util tests


client/control/actions/FilterActions.js
```diff
@@ -0,0 +1,113 @@
...
+ * Given an option and its parent group, update the
+ * filter state based on the `isEnabled` prop
+ *
+ * @param  {Object}  group     Group the filter belongs to
+ * @param  {Object}  option    Option that was affected
+ * @param  {Boolean} isEnabled Is the option selected?
+ */
+function selectFilter({ group, option, isEnabled }) {
+  return dispatch => {
+    dispatch({
```
mythmon: Does this need to be a thunked action creator? It looks like this could just return the action directly, instead of the return dispatch => { ... } dance.

client/control/components/ActiveFilters.js
```diff
@@ -0,0 +1,88 @@
...
+        this.props.onFilterSelect(group, option);
+    }
+
+    // return the handling function
+    return this.handlerCache[cacheKey];
+  }
+
+  /**
+   * Render
+   */
```
mythmon: Can we take this comment out, or flesh it out?

client/control/components/ActiveFilters.js
```diff
@@ -0,0 +1,88 @@
...
+  render() {
+    const {
+      className,
+      selectedFilters,
+      onResetFilters,
+    } = this.props;
+
+    // optional className prop
+    const displayedClass = `active-filters ${className || ''}`;
```
mythmon: I think you can do this a little cleaner with [classnames](https://www.npmjs.com/package/classnames), which we already have:

```js
import cx from 'classnames';

let className = undefined;
cx('active-filters', className);  # 'active-filters'
className = 'some-class';
cx('active-filters', className);  # 'active-filters some-class';
```

client/control/components/RecipeCombobox.js
```diff
@@ -0,0 +1,181 @@
...
+
+  /**
+   * Utility function to remove a set of properties
+   * from a given object.
+   *
+   * @param  {Object}         object Object to remove props form
+   * @param  {Array<string>}  list   List of properties to remove
+   * @return {Object}         New object without selected properties
+   */
+  static removeProperties(object, list) {
```
mythmon: This should probably be in a util module, since it is generally useful and not specific to this class.

client/control/components/RecipeCombobox.js
```diff
@@ -0,0 +1,181 @@
...
+    const newObject = { ...object };
+
+    list.forEach(property => {
+      delete newObject[property];
```
mythmon: I usually prefer for-of loops for this sort of thing, but either way is fine

```js
for (const property of list) {
  delete newObject[property];
}
```

client/control/components/GroupMenu.js
```diff
@@ -0,0 +1,97 @@
...
+   * Creates a click handler based on the
+   * group/option passed into it, basically
+   * an event handler factory.
+   *
+   * @param  {Object} group  Group which clicked option belongs to
+   * @param  {Object} option Option which was clicked
+   * @return {Function}      Wrapped handler
+   */
+  handleItemClick(group, option) {
+    return () => this.props.onItemSelect(group, option);
```
mythmon: Since this isn't a event handler, but an event handler factory, I think we should name it differently. Maybe something like `makeClickItemHandler`? Alternatively, it might be simpler to cram it into the markup below, like `onClick={() => onItemSelect('text', searchText)}` (assuming you've pulled `onItemSelect` out of the props already).

client/control/tests/utils/test_controlUtils.js
```diff
@@ -16,4 +19,106 @@ describe('controlApp Utils', () => {
...
+    describe('utils/closest', () => {
+      it('should attempt to find a parent element based on selector', () => {
+        const testContainer = document.createElement('div');
+        testContainer.innerHTML = [
+          '<span class="test">',
+          '<div class="nest">',
+          '<button class="target" />',
+          '</div>',
+          '</span>',
+        ].join('');
```
mythmon: You can use backtick strings for multiline strings. It doesn't fix indentation, but for this, I think that would be fine.

```js
testContainer.innerHTML = `
  <span class="test">
    <div class="nest">
      <button class="target" />
    </div>
</span>
`;
```

I've also had success creating elements for test like this using React/JSX directly.

client/control/tests/actions/test_filterActions.js
```diff
@@ -0,0 +1,190 @@
...
+
+    it('creates an LOAD_FILTERS action', () => {
+      const expectedAction = {
+        type: LOAD_FILTERS,
+        filters: [],
+      };
+
+      store.dispatch(loadFilters())
+      .then(() => {
+        expect(store.getActions()).toContain({ ...expectedAction });
```
mythmon: We should prefer async/await for simple cases like this.

Also, why clone the `expectedAction`? Would `expect(store.getActions()).toContain(expectedAction)` be different?

client/utils/deep-compare.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * Given two objects, deep compares them using JSON.stringify
+ *
+ * @param  {Object}  obj1 Object to compare
+ * @param  {Object}  obj2 Other object to compare
+ * @return {Boolean}      Are both objects completely equal?
+ */
+export default function deepCompare(obj1, obj2) {
```
mythmon: Would [`_.deepEqual`](http://underscorejs.org/#isEqual) work well for this? I expect it would be more efficient, and we already have underscore as a dependency anyways.

client/control/reducers/FiltersReducer.js
```diff
@@ -0,0 +1,197 @@
...
+        newState.push(newGroup);
+      }
+
+      initialState = cloneArrayValues(newState);
+
+      break;
+    }
+
+    case SET_ALL_FILTERS: {
+      newState = cloneArrayValues(action.filters || initialState);
```
mythmon: Is it necessary to clone things from the actions? I've always just used them directly without trouble.

client/control/actions/FilterActions.js
```diff
@@ -0,0 +1,113 @@
...
+ * Dispatches a SET_ALL_FILTERS event.
+ * If called without a state param,
+ * the filters reducer will reset the state
+ * to its initialState.
+ *
+ * The difference between SET_ALL_FILTERS and LOAD_FILTERS
+ * is that LOAD_ parses data from the backend before
+ * inserting into the store. SET_ALL assumes that
+ * the filters are already parsed, and the _app_ is
+ * manipulating the existing store.
```
mythmon: I think this is going to be confusing in the long run. I think a better approach would be to only have `SET_ALL_FILTERS`, and make whatever dispatches `LOAD_FILTERS` do the parsing and expansion there, instead of doing it in the reducer. If I'm understanding the flow correctly, this would make dealing with server formatting and actually making the request close together, instead of being spread across the reducers and action creators.

client/control/reducers/FiltersReducer.js
```diff
@@ -0,0 +1,197 @@
...
+          value: group,
+          label: formatLabel(group),
+          multiple: filters[group].length > 2,
+          options: cloneArrayValues(filters[group]).map(formatFilterOption),
+        };
+
+        newState.push(newGroup);
+      }
+
+      initialState = cloneArrayValues(newState);
```
mythmon: I don't think this is a good idea. This makes it so there is state outside of `state`, and means that actions are no longer pure. This breaks a lot of assumptions Redux makes, including the redux dev tools. It also breaks a lot of expectations developers may have. It is surprising, which isn't really a good idea.

client/control/components/GroupMenu.js
andymikulski: >Alternatively, it might be simpler to cram it into the markup below...

I prefer to break out any anonymous functions that show up in templates (when possible). I think it helps keep the mental model in order when skimming code, and benefits refactoring. For instance, I just added handler caching without having to add any complexity to the template markup.

Anonymous functions are fired on every render too, which is a (very) slight perf hit, iirc it's better practice to not use anon functions in renders - but I could be wrong!

client/control/tests/utils/test_controlUtils.js
andymikulski: Huh, weird. I wonder why I didn't just use backticks in the first place. I didn't consider using JSX - I'll give that a try.

client/control/reducers/FiltersReducer.js
andymikulski: Makes sense 👍.  I've switched the filter reducer to be an object with separate `active`/`list` properties to track what's been modified by the user and should be considered the 'initial state' should the user decide to reset filters.

client/control/reducers/FiltersReducer.js
andymikulski: While running tests, I found that some loops were having side effects on objects in memory. The arrays I was creating were only holding object references, not fresh copies of the object, ex:

```
const thing1 = [{ a: 1 }, { b: 2 }, { c: 3 }];
const thing2 = [].concat(thing1);
thing2[0].a = 5;
console.log(thing1[0].a); // 5
```

Once I copied the array(s) by value instead of by reference, the tests ran successfully. I'm not sure if this is an inherent design issue with my reducer/actions, or if it's just one of those React gotchas (that people usually solve with Immutable.js).

client/control/components/GroupMenu.js
mythmon: Oh, you're right. I forgot about the perf thing. It also makes us re-render more than we should.

I still think the naming issue stands though.

client/control/reducers/FiltersReducer.js
mythmon: Sounds good then. I'm glad the tests found the issues.

Do you think Immutable.js would be useful here? I've tried it both with and without, and I haven't really decided which I prefer. If you'd prefer switching to immutable, I'd support that.

andymikulski:
@mythmon I've addressed the earlier review items, if you could review again that'd be great! Thank you!

mythmon:
Thanks for fixing those. I'll look over those changes. This change is big enough that @rehandalal and I decided he should also do a full review pass. So I'll tag off to him.

andymikulski:
Hey @rehandalal , are you reviewing this PR as-is or should I rebase it first?

rehandalal:
Oops! Accidentally did some silliness and closed this. Reopening...

rehandalal:
#434 has finally landed, so you should probably rebase this on master.

client/utils/clone-array.js
```diff
@@ -0,0 +1,37 @@
+/**
```
andymikulski: The `clone` [doc snippet](http://underscorejs.org/#clone) says it only makes a shallow copy, however we need a deep clone. Underscore itself doesn't offer a `deepClone` function, and [digging into alternatives](https://github.com/jashkenas/underscore/issues/162) leads into using a ~3 year old, test-less [npm package](https://github.com/mateusmaso/underscore.deepclone), or rolling our own (either using a couple of Underscore fn's or using the `JSON.parse(JSON.stringify(...))` business).

IMO I'd prefer to maintain our own util file instead of farming it out to another NPM package (seeing as Underscore doesn't offer what we need). And, fwiw, there are at least tests written already for our utils. What do you think, @rehandalal ?

recipe-server/client/control/actions/FilterActions.js
```diff
@@ -0,0 +1,129 @@
...
+        }
+        const newFilters = [];
+        // format each recipe
+        for (const group in filters) {
+          if (!filters.hasOwnProperty(group)) {
+            break;
+          }
+          const newGroup = {
+            value: group,
+            label: capFirstLetter(group),
```
andymikulski: We actually need it here. The `filters` API endpoint returns the keys for each group in lowercase (e.g. `channels`, `status`), but the group headers[0] need the title casing.

The API could return title cased keys, which would remove the need for this `capFirstLetter` bit. Your call, @rehandalal 


[0]
group headers

rehandalal: ah! sorry I missed that it was the group name. okay, it lives! maybe just rename it to something like `capitalizeFirst` and it should be good to go.

andymikulski: Gotcha. I'm rebasing the branch right now to fix the merge conflicts, I'll be sure to add that change in as well. Thanks!

andymikulski:
After seeing the actual country/locale data in the filter list, I added a `view more` button to declutter the list a little:

truncate

Right now I have it capped at 5 items per filter, but all are shown if the user clicks `view more` or if they are entering a text search (at which point, all matching options are returned).

Along the same vein, text filtering the options now returns only matching options - previously, it would return the whole group of options the text was found in. (E.g. searching `uzbek` would return the whole countries list, instead of just `uzbekistan`).

@mythmon or @rehandalal - https://github.com/mozilla/normandy/pull/405/commits/d4c06e458a15259699f5807b69a488421e9a10ad could stand to be reviewed before merging this PR

recipe-server/client/control/components/GroupMenu.js
```diff
@@ -76,32 +104,62 @@ export default class GroupMenu extends React.Component {
...
-                )
-              }
-            </div>
-          )
+          data.map(group => {
+            const {
+              value,
+              options,
+              label,
+            } = group;
```
mythmon: I think this version is a little more readable:

```js
data.map(({ value, options, label }) => {
   ...
})
```
```

andymikulski: I actually wanted to go with this too, but the event handlers use the `group` object later in that scope, e.g. `...onClick={this.makeClickItemHandler(group, option)}...`.

I can set up the handlers to expect the explicit values from `group`/`option` if you'd prefer. I like to just pass the whole object in so the calling scope doesn't need to know how the handler works.

andymikulski:
While writing the [manual test list](https://gist.github.com/andymikulski/0a61c0742f87c1c3d26804481cc93620), I realized a lot of the functionality could/should have automated tests. So... now we have [40 more tests](https://gist.github.com/andymikulski/121193a38c445148a3faec62f18e4798)! I wrote a bunch of tests for the simple/non-`connect`ed components that build out the active filters, as well as some missing tests for the filter selectors.

@rehandalal / @mythmon , if y'all could review https://github.com/mozilla/normandy/pull/405/commits/e3243c52e3fb3107112fe10aae72f9b24118d884 that would be awesome. Most of the tests are straight-forward.

@chartjes If you want to check out the manual instructions [here](https://gist.github.com/andymikulski/0a61c0742f87c1c3d26804481cc93620), you can find the handful of things I didn't write tests for.

recipe-server/client/control/components/GroupMenu.js
mythmon: Ah, I missed that, sorry. Passing the whole objects makes sense. Lets keep it the way it is.

recipe-server/client/control/actions/FilterActions.js
andymikulski: I'm going to switch from this util file to using [underscore.string](https://epeli.github.io/underscore.string/) - this provides (tested) functionality that we can use in other instances as well

rehandalal:
good job! and nice move on the `underscore.string` addition. thanks for sticking with it through this!!

andymikulski:
Hey @chartjes , can you take a look at this? I have a short [manual test list](https://gist.github.com/andymikulski/0a61c0742f87c1c3d26804481cc93620) you can use to run through the updated functionality (though most have automated tests, now)

PR #409, opened 2017-01-07
Mock Recipe Server and Flake8 Config Fix

Osmose:
Playing around with CircleCI build stuff, don't mind me!


Osmose:
Okay! This should be ready for review.

The first commit moves the flake8 config, since adding more Python files broke our linting setup. The second commit adds the scripts needed for generating static files that mock out the API. We'll be using this to stand up a mock service that desktop QA can test the system add-on against.

I've filed [bug 1330415](https://bugzilla.mozilla.org/show_bug.cgi?id=1330415) for standing up the S3 bucket and domain for the mock server. I dunno how CircleCI handles a deploy step failing, which it will until we get the credentials and bucket name in as environment variables on Circle. That's the only thing I can think of that might block merging this.

r?

mock-recipe-server/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+
+services:
+  testgen:
+    extends:
+      file: normandy-base.yml
+      service: normandy-base
+    links:
+      - database
+      - autograph
+      - proxy
```
mythmon: I'm surprised you can reference link services from other files like this. It seems to work though. 👍

mock-recipe-server/bin/upload.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -eu
+
+BUILD_DIR=$1
+S3_BUCKET=$2
+
+aws s3 rm --recursive "s3://$S3_BUCKET/"
+aws s3 cp --recursive "$BUILD_DIR" "s3://$S3_BUCKET/" --grants read=uri=http://acs.amazonaws.com/groups/global/AllUsers
```
mythmon: I'm not too familiar with the aws tools. @milescrabill, does this look good to upload a directory of files to S3, replacing what is already there?

milescrabill: It might make more sense to use `aws s3 sync --delete "$BUILD_DIR" "s3://$S3_BUCKET/" --grants read=uri=http://acs.amazonaws.com/groups/global/AllUsers`, that way you don't re-copy existing files and only delete those which have been deleted.

milescrabill: I think the way you have it is totally fine though. Just make sure `$S3_BUCKET` is set to avoid errant `rm`'s.

rehandalal:
Since you're doing a flake8 fix you should probably also upgrade therapist to v1.2.0.

Osmose:
Marking as `do not merge` until [bug 1330415](https://bugzilla.mozilla.org/show_bug.cgi?id=1330415) is resolved.

Osmose:
NEVERMIND

PR #410, opened 2017-01-09
New revisions

### LINKED ISSUE #356: Implement a new revision system
We will be ditching `django-reversion` and replacing it with our own revisions. These revisions:

- [ ] Form a linear history, implemented as a linked list
- [ ] Have a comment describing the revision
- [ ] Can be linked to an approval request
- [ ] Are considered the current revision if they are the latest revision which is approved
- [ ] Are considered drafts if they are not approved
- [ ] Are considered "in review" if there is a pending approval request

rehandalal:
Fixes #356 

r?


normandy/recipes/tests/test_models.py
```diff
@@ -37,18 +34,7 @@ def test_recipes_used_by_empty(self):
 
 @pytest.mark.django_db
 class TestRecipe(object):
-    def test_revision_id_increments(self):
```
mythmon: I think we should keep this test, or have one like it.

normandy/recipes/models.py
```diff
@@ -59,130 +54,163 @@ def update_signatures(self):
...
+    def arguments(self, value):
+        self.arguments_json = json.dumps(value)
+
+    @property
+    def restored_recipe(self):
+        recipe = self.recipe
+        recipe.latest_revision = self
+        return recipe
+
+    def save(self, *args, skip_updated=False, **kwargs):
```
mythmon: Is `skip_updated` ever used? I don't see it anywhere.

mythmon:
Separately, I ran the migrations on my local copy of Normandy, and a couple recipes got all their history deleted. I don't really know what to make of that. I'm going to try and reproduce it a little more carefully, and I'm going to read through the migration a bit harder to see if anything comes up.

rehandalal:
There are already tests for the `save` method but I added tests for `update`.

normandy/recipes/tests/test_models.py
```diff
@@ -190,9 +152,38 @@ def fake_sign(datas):
...
+    def test_recipe_force_update(self):
+        recipe = RecipeFactory(name='my name')
+        revision_id = recipe.revision_id
+        recipe.update(name='my name', force=True)
+        assert revision_id != recipe.revision_id
+
 
 @pytest.mark.django_db
 class TestRecipeQueryset(object):
+    def test_revision_id_changes(self):
```
mythmon: Nit: this should be in `TestRecipe`, right?

normandy/recipes/migrations/0035_revision_data.py
```diff
@@ -0,0 +1,68 @@
...
+                    recipe=r, parent=revision, created=v.revision.date_created,
+                    comment=v.revision.comment, user=v.revision.user, name=data.get('name'),
+                    action=action, arguments_json=data.get('arguments_json', ''),
+                    filter_expression=data.get('filter_expression', ''))
+
+                revision.id = hash_revision_data(revision)
+
+                revision.save()
+
+            r.latest_revision = revision
```
mythmon: Would it be ok to assert that revision is not `None` here?

mythmon: Ooh, wait. What happens when a recipe is newly created? It wouldn't have any reversion `Version`s then, right? Shouldn't `r.latest_revision` be based on the current fields in r, not one of the reversions `Version`s?

rehandalal: Yep! I'll do that.

rehandalal: Ah! Yeah, that's probably what the issue is. I thought it creates a revision for the a newly created recipe but I might be wrong. I'll confirm.

rehandalal: It should be creating a reversion when the recipe is created. So that's not it. I'll add the `assert`.

rehandalal: Updated the migration and I think it should work better now

rehandalal:
@mythmon I was wrong about these migrations being easy to make reversible

mythmon:
@rehandalal Do you want to merge without them then? Or do you still want to make them reversible, and it will just take longer?

mythmon:
👏 👏 Thanks! That looks great, and I tested it back and forth and it works great.

(While testing this, I found a different bug, but it was entirely unrelated)

PR #418, opened 2017-01-10
Fix recipe-client-addon README.

Osmose:
Actually what I'm testing is if the linting works on new PRs. #409 has a bunch of linting errors in CI for files it didn't touch, that seem to be because the flake8 config is no longer at the root of the repo, which begs the question: Why did it pass on the original monorepo PR?

And it seemed smart to fix something while I was testing.


Osmose:
Well it looks like my guess was wrong. Oh well. I'll look harder at the failures tomorrow, then.

PR #419, opened 2017-01-11
Use Docker 1.10 during CircleCI builds.

Osmose:
Just like #418, this PR is trying to see if it can break the linting. In this case, the only suspicious-looking thing was the version of docker used, which may have affected how volumes are mounted.


Osmose:
Well, that wasn't it.

PR #423, opened 2017-01-13
Simple filter API changes

rehandalal:
~~Blocked by #410.~~

I still need to:
- [x] Update the `filter_expression` so that it uses the new fields. 

r?


rehandalal:
@andymikulski you should probably check this out. Only commits from 20b24a4 onwards are relevant to this PR.

FYI: You will likely need to make some changes to your simple filtering stuff.

andymikulski:
Looks like the stuff I have in place should be pretty close to what you're returning here. Looking forward to getting it all put together!

normandy/recipes/models.py
```diff
@@ -102,6 +126,22 @@ def revision_id(self):
...
+    @property
+    def channels(self):
+        return self.latest_revision.channels if self.latest_revision else None
+
+    @property
+    def countries(self):
+        return self.latest_revision.countries if self.latest_revision else None
+
+    @property
+    def locales(self):
```
mythmon: So at what point do we make a helper for this?

```python
@latest_revision_property
def countries(self, latest_revision):
    return latest_revision.countries
```

and make it return `None` without calling the function if `latest_revision` doesn't exist?

normandy/recipes/models.py
```diff
@@ -122,18 +162,52 @@ def update_signature(self):
...
+
             data = revision_data
         else:
             is_clean = False
 
         if not is_clean or force:
-            self.latest_revision = RecipeRevision.objects.create(
-                recipe=self, parent=self.latest_revision, **data)
+            self.latest_revision = RecipeRevision.objects.create(recipe=self, parent=revision,
+                                                                 **data)
```
mythmon: Nit: I think the previous version of indentation was prettier.

normandy/recipes/models.py
```diff
@@ -174,18 +248,28 @@ class RecipeRevision(models.Model):
...
-            'filter_expression': self.filter_expression,
+            'extra_filter_expression': self.extra_filter_expression,
+            'channels': list(self.channels.all()),
+            'countries': list(self.countries.all()),
+            'locales': list(self.locales.all()),
         }
 
     @property
+    def filter_expression(self):
+        return self.extra_filter_expression
```
mythmon: Do you plan to make this concatenate everything together to add the locale/language/country filtering?

normandy/recipes/models.py
rehandalal: probably a good idea

normandy/recipes/models.py
rehandalal: agreed.

normandy/recipes/models.py
rehandalal: yep, just committed that

rehandalal:
rebased ^

chartjes:
@rehandalal Am I correct in assuming that all existing tests pass because I see a bunch of code and no new tests...

rehandalal:
@chartjes yes all the existing tests pass and test for the changes

docs/qa/setup.rst
```diff
@@ -48,6 +48,8 @@ Starting the server, and initial setup
 
      $ docker-compose run normandy ./manage.py migrate
      $ docker-compose run normandy ./manage.py update_actions
+     $ docker-compose run normandy ./manage.py update_product_details
+     $ docker-compose run normandy ./manage.py initial_data
      $ docker-compose run normandy ./manage.py createsuperuser
```
mythmon: @Osmose EDI should handle all these manage.py commands.

normandy/recipes/models.py
```diff
@@ -122,18 +163,52 @@ def update_signature(self):
 
     @transaction.atomic
     def update(self, force=False, **data):
```
Osmose: I wasn't involved in the revision PR so I'm a little behind, but I don't quite get what this method is trying to do. It seems like it's avoiding making another revision if a revision with the same state already exists, but does that mean that we can't have two revisions with the same state, but different parents? Wouldn't this make it impossible to represent a recipe going back to a previous state?

Osmose: Also, for finding matching recipe revisions, it seems to make more sense to include the relations in the revision ID hash, and just create a new revision and look for revisions with that hash.

normandy/recipes/models.py
```diff
@@ -174,18 +249,55 @@ class RecipeRevision(models.Model):
...
-            'filter_expression': self.filter_expression,
+            'extra_filter_expression': self.extra_filter_expression,
+            'channels': list(self.channels.all()),
+            'countries': list(self.countries.all()),
+            'locales': list(self.locales.all()),
         }
 
     @property
+    def filter_expression(self):
+        exp = ''
```
Osmose: It'd be easier to make this a list that's joined with `'&&'` at the end, that way you don't have to check if the expression exists for each subsequent clause.

normandy/recipes/tests/__init__.py
```diff
@@ -53,6 +71,33 @@ def signed(self, create, extracted=False, **kwargs):
         else:
             return None
 
+    @factory.post_generation
```
Osmose: Someday we'll factor this pattern out to a helper function. But that might not be today.

normandy/recipes/tests/test_models.py
```diff
@@ -51,27 +54,39 @@ def test_canonical_json(self):
...
+            countries=[CountryFactory(code='CA')],
             enabled=False,
-            filter_expression='2 + 2 == 4',
+            extra_filter_expression='2 + 2 == 4',
+            locales=[LocaleFactory(code='en-US')],
             name='canonical',
         )
         # Yes, this is really ugly, but we really do need to compare an exact
         # byte sequence, since this is used for hashing and signing
+        filter_expression = (
```
Osmose: I don't think this is an appropriate place to test if the filter expression building code, it should have it's own unit tests instead.

normandy/recipes/storage.py
```diff
@@ -0,0 +1,29 @@
...
+from product_details.storage import PDFileStorage
+
+
+class ProductDetailsRelationalStorage(PDFileStorage):
+    """
+    Extends the in-database storage for product_details to provide a
+    database table of locales for other models to have foreign keys to.
+    """
+    @transaction.atomic
+    def update(self, name, content, last_modified):
```
Osmose: Did this have any tests before? If not, it seems pretty testable in terms of the API.

normandy/recipes/models.py
```diff
@@ -174,18 +249,55 @@ class RecipeRevision(models.Model):
...
+        if self.channels.count():
+            channels = ', '.join(["'{}'".format(c.slug) for c in self.channels.all()])
+
+            if len(exp):
+                exp += ' && '
+
+            exp += 'normandy.channel in [{}]'.format(channels)
+
+        if self.extra_filter_expression:
+            prefix = ' && (' if len(exp) else ''
```
Osmose: I think you can just assume the statement will be wrapped in parenthesis even if it's the only thing in the expression, rather than trying to avoid them.

normandy/recipes/api/views.py
```diff
@@ -70,6 +78,28 @@ class RecipeViewSet(CachingViewsetMixin, UpdateOrCreateModelViewSet):
         AdminEnabledOrReadOnly,
     ]
 
+    def get_queryset(self):
```
Osmose: I didn't see any tests for the filtering we've added here, we should have some.

normandy/base/utils.py
```diff
@@ -46,3 +47,10 @@ def get_client_ip(request):
             return ips[-settings.NUM_PROXIES]
         except IndexError:
             return None
+
+
+def filter_m2m(qs, field, values):
```
rehandalal: gah! good spot. i'll add some relevant testing.

normandy/recipes/decorators.py
```diff
@@ -0,0 +1,21 @@
+class LatestRevisionProperty(object):
```
rehandalal: yup!

normandy/recipes/api/serializers.py
```diff
@@ -52,6 +71,21 @@ def update(self, instance, validated_data):
                 'action': Action.objects.get(name=validated_data['action'])
             })
 
+        if 'channels' in validated_data:
```
rehandalal: We need the custom `update` method to call `Recipe.update` which is responsible for creating the revision.

normandy/recipes/api/views.py
rehandalal: I'll add tests. We could potentially use generic filtering but I'm not sure if it would handle the complex many to many filtering we're doing correctly.

normandy/recipes/models.py
rehandalal: This doesn't check all revisions, it only checks the latest_revision to make sure that it has changed. But yes, checking by hash is probably way less complicated and more performant. I'll make the change.

normandy/recipes/storage.py
rehandalal: I'll confirm, and test either way.

normandy/recipes/tests/test_models.py
rehandalal: /me nods

normandy/recipes/models.py
rehandalal: yep, the parens on the extra_filter is the only reason I did it this way.

normandy/recipes/models.py
rehandalal: I lied. Dealing with the hash is more difficult because of the m2m relationships

normandy/recipes/models.py
```diff
@@ -268,34 +271,24 @@ def data(self):
...
+            parts.append('normandy.channel in [{}]'.format(channels))
 
         if self.extra_filter_expression:
-            prefix = ' && (' if len(exp) else ''
-            suffix = ')' if len(exp) else ''
-            exp += '{}{}{}'.format(prefix, self.extra_filter_expression, suffix)
+            parts.append(self.extra_filter_expression)
 
-        return exp
+        return ' && '.join(parts)
```
Osmose: I think you still want to surround the parts with parentheses just to be safe, just do it for all of them always.

normandy/base/utils.py
```diff
@@ -50,7 +50,14 @@ def get_client_ip(request):
...
-    qs.annotate(_count=Count(field)).filter(_count=len(values))
+    """
+    Filters a queryset by an exact list of many to many relations.
+    """
+    values = list(values)
+
+    qs = qs.annotate(_count=Count(field)).filter(_count=len(values))
+
+    if len(values):
+        qs = qs.filter(**{'{}__in'.format(field): values})
```
Osmose: Does `in` do an exact comparison? I would think that this would match all revisions that have any one of the given values in it. Which, if I'm understanding the docstring, is not what we want.

Osmose: Does `in` do an exact comparison? I would think that this would match all revisions that have any one of the given values in it. Which, if I'm understanding the docstring, is not what we want.

normandy/recipes/models.py
Osmose: > This doesn't check all revisions, it only checks the latest_revision to make sure that it has changed.

Maybe I'm missing something? The code fetches revisions that match the latest revision's ID, and then filters out revisions that don't have the same m2m relationships, then filters out revisions that have the same data. Then it says that it's clean if there exist any revisions that match all of that.

None of those filters check if the revisions you're searching for are actually the last revision though. You could do all that filtering and match a revision from months ago that had the same content, and it'd be considered clean even if the latest revision is actually different. If you already have a reference to the latest revision, why do you need to go through all that querying?

normandy/base/utils.py
rehandalal: it's the combination of line 58 (which checks the count) and this line 61 which make sure it's an exact match.

normandy/recipes/models.py
rehandalal: sure.

FYI, as discussed i'll have to write an additional data migration to rehash the revisions

normandy/recipes/models.py
rehandalal: the revision id is unique so you wouldn't match an older revision with the same content. the `latest_revision` is always the last revision. if it is not, something has gone terribly wrong.

rehandalal:
@Osmose updated, and i didn't have to write a data migration for the filter expression change.

PR #425, opened 2017-01-14
recipe-client-addon: Add bucketSampling to filter expressions

### LINKED ISSUE #413: Add bucket sampling to the add-on
#300 added bucket sampling to the self-repair runner; we need to add support for it to the add-on as well. https://github.com/mozilla/normandy-addon/pull/38 is a PR that added it to the add-on, but it needs to be reworked to support the post-moz-central-review add-on.

This is a blocker for the system add-on release. This was originally https://github.com/mozilla/normandy-addon/issues/45 in the add-on repo.

mythmon:
Based on mozilla/normandy-addon#38.

Fixes #413.

TODO:

* [x] Actually run these tests in mozilla-central (I'm waiting on a build right now)
  * [x] Figure out values that make the tests pass

### DIFF
--- a/recipe-client-addon/lib/Sampling.jsm
+++ b/recipe-client-addon/lib/Sampling.jsm
@@ -5,77 +5,121 @@
 "use strict";
 
 const {utils: Cu} = Components;
-Cu.import("resource://shield-recipe-client/lib/LogManager.jsm");
+Cu.import("resource://gre/modules/Task.jsm");
 Cu.importGlobalProperties(["crypto", "TextEncoder"]);
 
 this.EXPORTED_SYMBOLS = ["Sampling"];
 
-const log = LogManager.getLogger("sampling");
-
-/**
- * Map from the range [0, 1] to [0, max(sha256)].
- * @param  {number} frac A float from 0.0 to 1.0.
- * @return {string} A 48 bit number represented in hex, padded to 12 characters.
- */
-function fractionToKey(frac) {
-  const hashBits = 48;
-  const hashLength = hashBits / 4;
-
-  if (frac < 0 || frac > 1) {
-    throw new Error(`frac must be between 0 and 1 inclusive (got ${frac})`);
-  }
-
-  const mult = Math.pow(2, hashBits) - 1;
-  const inDecimal = Math.floor(frac * mult);
-  let hexDigits = inDecimal.toString(16);
-  if (hexDigits.length < hashLength) {
-    // Left pad with zeroes
-    // If N zeroes are needed, generate an array of nulls N+1 elements long,
-    // and inserts zeroes between each null.
-    hexDigits = Array(hashLength - hexDigits.length + 1).join("0") + hexDigits;
-  }
-
-  // Saturate at 2**48 - 1
-  if (hexDigits.length > hashLength) {
-    hexDigits = Array(hashLength + 1).join("f");
-  }
-
-  return hexDigits;
-}
-
-function bufferToHex(buffer) {
-  const hexCodes = [];
-  const view = new DataView(buffer);
-  for (let i = 0; i < view.byteLength; i += 4) {
-    // Using getUint32 reduces the number of iterations needed (we process 4 bytes each time)
-    const value = view.getUint32(i);
-    // toString(16) will give the hex representation of the number without padding
-    hexCodes.push(value.toString(16).padStart(8, "0"));
-  }
-
-  // Join all the hex strings into one
-  return hexCodes.join("");
-}
-
 this.Sampling = {
-  stableSample(input, rate) {
-    const hasher = crypto.subtle;
+  /**
+   * Map from the range [0, 1] to [0, 2^48].
+   * @param  {number} frac A float from 0.0 to 1.0.
+   * @return {string} A 48 bit number represented in hex, padded to 12 characters.
+   */
+  fractionToKey(frac) {
+    const hashBits = 48;
+    const hashLength = hashBits / 4;  // each hexadecimal digit represents 4 bits
 
-    return hasher.digest("SHA-256", new TextEncoder("utf-8").encode(JSON.stringify(input)))
-      .then(hash => {
-        // truncate hash to 12 characters (2^48)
-        const inputHash = bufferToHex(hash).slice(0, 12);
-        const samplePoint = fractionToKey(rate);
+    if (frac < 0 || frac > 1) {
+      throw new Error(`frac must be between 0 and 1 inclusive (got ${frac})`);
+    }
 
-        if (samplePoint.length !== 12 || inputHash.length !== 12) {
-          throw new Error("Unexpected hash length");
-        }
+    const mult = Math.pow(2, hashBits) - 1;
+    const inDecimal = Math.floor(frac * mult);
+    let hexDigits = inDecimal.toString(16);
+    if (hexDigits.length < hashLength) {
+      // Left pad with zeroes
+      // If N zeroes are needed, generate an array of nulls N+1 elements long,
+      // and inserts zeroes between each null.
+      hexDigits = Array(hashLength - hexDigits.length + 1).join("0") + hexDigits;
+    }
 
-        return inputHash < samplePoint;
+    // Saturate at 2**48 - 1
+    if (hexDigits.length > hashLength) {
+      hexDigits = Array(hashLength + 1).join("f");
+    }
 
-      })
-      .catch(error => {
-        log.error(`Error: ${error}`);
-      });
+    return hexDigits;
   },
+
+  bufferToHex(buffer) {
+    const hexCodes = [];
+    const view = new DataView(buffer);
+    for (let i = 0; i < view.byteLength; i += 4) {
+      // Using getUint32 reduces the number of iterations needed (we process 4 bytes each time)
+      const value = view.getUint32(i);
+      // toString(16) will give the hex representation of the number without padding
+      hexCodes.push(value.toString(16).padStart(8, "0"));
+    }
+
+    // Join all the hex strings into one
+    return hexCodes.join("");
+  },
+
+  rangeBucketSample(inputHash, minBucket, maxBucket, bucketCount) {
+    const minHash = Sampling.fractionToKey(minBucket / bucketCount);
+    const maxHash = Sampling.fractionToKey(maxBucket / bucketCount);
+    return (minHash <= inputHash) && (inputHash < maxHash);
+  },
+
+  /**
+   * @promise A hash of `data`, truncated to the 12 most significant characters.
+   */
+  truncatedHash: Task.async(function* (data) {
+    const hasher = crypto.subtle;
+    const input = new TextEncoder("utf-8").encode(JSON.stringify(data));
+    const hash = yield hasher.digest("SHA-256", input);
+    // truncate hash to 12 characters (2^48), because the full hash is larger
+    // than JS can meaningfully represent as a number.
+    return Sampling.bufferToHex(hash).slice(0, 12);
+  }),
+
+  /**
+   * Sample by splitting the input into two buckets, one with a size (rate) and
+   * another with a size (1.0 - rate), and then check if the input's hash falls
+   * into the first bucket.
+   *
+   * @param    {object}  input Input to hash to determine the sample.
+   * @param    {Number}  rate  Number between 0.0 and 1.0 to sample at. A value of
+   *                           0.25 returns true 25% of the time.
+   * @promises {boolean} True if the input is in the sample.
+   */
+  stableSample: Task.async(function* (input, rate) {
+    const inputHash = yield Sampling.truncatedHash(input);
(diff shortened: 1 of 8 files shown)


recipe-client-addon/test/browser_env_expressions.js
```diff
@@ -53,4 +53,12 @@ add_task(function* () {
...
   val = yield EnvExpressions.eval('["test"]|stableSample(0)');
   is(val, false, "Stable sample returns false for 0% sample");
+
+  // Test bucket sample returns true for a known sample
+  val = yield EnvExpressions.eval('["test1"]|bucketSample(0, 5, 10)');
+  is(val, true, "Bucket sample returns true for a known sample");
+
+  // Test bucket sample returns false for a known sample
+  val = yield EnvExpressions.eval('["test2"]|bucketSample(0, 5, 10)');
+  is(val, false, "Bucket sample returns true for a known sample");
```
Osmose: I think you need to update the message to say false instead of true.

recipe-client-addon/lib/Sampling.jsm
```diff
@@ -57,25 +55,72 @@ function bufferToHex(buffer) {
...
+   * The range to check is defined by a start point and length, and can wrap
+   * around the input space. For example, if there are 100 buckets, and we ask to
+   * check 50 buckets starting from bucket 70, then buckets 70-100 and 0-19 will
+   * be checked.
+   *
+   * @param {object}  input Input to hash to determine the matching bucket.
+   * @param {integer} start Index of the bucket to start checking.
+   * @param {integer} count Number of buckets to check.
+   * @param {integer} total Total number of buckets to group inputs into.
+   * @promises {boolean}       True if the given input is within the range of
```
Osmose: Alignment's a bit off here.

recipe-client-addon/lib/Sampling.jsm
```diff
@@ -57,25 +55,72 @@ function bufferToHex(buffer) {
...
-  },
+  /**
+   * Sample by splitting the input into two buckets, one with a size (rate) and
+   * another with a size (1.0 - rate), and then check if the input's hash falls
+   * into the first bucket.
+   *
+   * @param {object} input Input to hash to determine the sample.
+   * @param {Number} rate Number between 0.0 and 1.0 to sample at. A value of
+   *                      0.25 returns true 25% of the time.
+   * @promises {boolean}    True if the input is in the sample.
```
Osmose: Same here RE: alignment. Actually, I dunno what the standard is for this: a hanging indent and per-line alignment, or all the descriptions lined up together?

recipe-client-addon/lib/Sampling.jsm
```diff
@@ -57,25 +55,72 @@ function bufferToHex(buffer) {
...
+}
+
+/**
+ * Return a promise for a truncated hash of `data`.
+ */
+const truncatedHash = Task.async(function* (data) {
+  const hasher = crypto.subtle;
+  const input = new TextEncoder("utf-8").encode(JSON.stringify(data));
+  const hash = yield hasher.digest("SHA-256", input);
+    // truncate hash to 12 characters (2^48), because the full hash is larger
```
Osmose: These comments are too indented.

recipe-client-addon/lib/Sampling.jsm
```diff
@@ -57,25 +55,72 @@ function bufferToHex(buffer) {
...
 }
 
+function rangeBucketSample(inputHash, minBucket, maxBucket, bucketCount) {
+  const minHash = fractionToKey(minBucket / bucketCount);
+  const maxHash = fractionToKey(maxBucket / bucketCount);
+  return (minHash <= inputHash) && (inputHash < maxHash);
+}
+
+/**
+ * Return a promise for a truncated hash of `data`.
```
Osmose: The "Return a promise" bit reads weirdly to me, compared to other methods here that return promises and use `@promises` instead. I'm more interested that it hashes data, and what "truncated" means (as in truncated to 12 characters). I think you should reword this.

recipe-client-addon/lib/Sampling.jsm
```diff
@@ -57,25 +55,72 @@ function bufferToHex(buffer) {
...
-  stableSample(input, rate) {
-    const hasher = crypto.subtle;
-
-    return hasher.digest("SHA-256", new TextEncoder("utf-8").encode(JSON.stringify(input)))
-      .then(hash => {
-        // truncate hash to 12 characters (2^48)
-        const inputHash = bufferToHex(hash).slice(0, 12);
-        const samplePoint = fractionToKey(rate);
-
-        if (samplePoint.length !== 12 || inputHash.length !== 12) {
```
Osmose: Did you want to keep the `samplePoint` and `inputHash` checking?

recipe-client-addon/lib/Sampling.jsm
```diff
@@ -57,25 +55,72 @@ function bufferToHex(buffer) {
...
-          throw new Error("Unexpected hash length");
-        }
-
-        return inputHash < samplePoint;
-
-      })
-      .catch(error => {
```
Osmose: I think logging when we hit an error in evaluating a filter expression is good, but I don't really care whether it's in the utility methods that we expose, or done once by the JEXL-evaluation code itself. Will we still do that somewhere once this is removed?

recipe-client-addon/test/browser_env_expressions.js
```diff
@@ -53,4 +53,12 @@ add_task(function* () {
   // Test stable sample returns true for matching samples
   val = yield EnvExpressions.eval('["test"]|stableSample(0)');
   is(val, false, "Stable sample returns false for 0% sample");
+
+  // Test bucket sample returns true for a known sample
```
Osmose: I like these tests!

I think we should add some tests for stableSample like these that aren't at 100% or 0%, so that we'll know in the future if we break the way input is hashed and could cause currently-running recipes to be sent to different users than they were before.

recipe-client-addon/lib/Sampling.jsm
mythmon: I don't think so. I think this is more consistent now. I'll add a test for this though.

mythmon:
@gijsk Of particular note here is that I moved all the helper functions into the exported object, and then added xpcshell tests for everything in `Sampling.jsm`. The actual details of bucket sampling are pretty straight forward.

recipe-client-addon/lib/Sampling.jsm
```diff
@@ -5,77 +5,121 @@
...
+   */
+  fractionToKey(frac) {
+    const hashBits = 48;
+    const hashLength = hashBits / 4;  // each hexadecimal digit represents 4 bits
+
+    if (frac < 0 || frac > 1) {
+      throw new Error(`frac must be between 0 and 1 inclusive (got ${frac})`);
+    }
+
+    const mult = Math.pow(2, hashBits) - 1;
```
gijsk: Premature optimization suggestion alert, but: we could probably cache this globally.

recipe-client-addon/lib/Sampling.jsm
```diff
@@ -5,77 +5,121 @@
...
+    const inDecimal = Math.floor(frac * mult);
+    let hexDigits = inDecimal.toString(16);
+    if (hexDigits.length < hashLength) {
+      // Left pad with zeroes
+      // If N zeroes are needed, generate an array of nulls N+1 elements long,
+      // and inserts zeroes between each null.
+      hexDigits = Array(hashLength - hexDigits.length + 1).join("0") + hexDigits;
```
gijsk: Err, would:

```js
hexDigits = hexDigits.padStart(hashLength, "0");
```

work just as well?

recipe-client-addon/lib/Sampling.jsm
```diff
@@ -5,77 +5,121 @@
...
+    }
+
+    // Saturate at 2**48 - 1
+    if (hexDigits.length > hashLength) {
+      hexDigits = Array(hashLength + 1).join("f");
```
gijsk: In what circumstance do we hit this? 2**48 - 1 in hex is just 12 * f, and we're using a fraction which we know is <= 1, so the length can never be greater, right? What am I missing?

recipe-client-addon/lib/Sampling.jsm
```diff
@@ -5,77 +5,121 @@
...
+
+    // Join all the hex strings into one
+    return hexCodes.join("");
+  },
 
-      })
-      .catch(error => {
-        log.error(`Error: ${error}`);
-      });
+  rangeBucketSample(inputHash, minBucket, maxBucket, bucketCount) {
```
gijsk: Nitpicking: can we name this method something more meaningful, like "isHashInBuckets" or whatever? I wondered if this was a stats term I didn't know about, but all google gives me is ads for people who have a range of physical (mostly yellow) buckets! :-)

gijsk: Also, please can we have a docstring for this method. :-)

recipe-client-addon/lib/Sampling.jsm
```diff
@@ -5,77 +5,121 @@
...
+    return inputHash < samplePoint;
+  }),
+
+  /**
+   * Sample by splitting the input space into a series of buckets, and checking
+   * if the given input is in a range of buckets.
+   *
+   * The range to check is defined by a start point and length, and can wrap
+   * around the input space. For example, if there are 100 buckets, and we ask to
+   * check 50 buckets starting from bucket 70, then buckets 70-100 and 0-19 will
```
gijsk: uber-nitpicking: if there are 100 buckets... numbered from 1-100? Because then you're saying both 70-100 and 0-19 implying there's a bucket 100 and a bucket 0, in which case the math doesn't work out (you get 31 + 20 = 51 "checked" buckets).

recipe-client-addon/lib/Sampling.jsm
```diff
@@ -5,77 +5,121 @@
...
+    const inputHash = yield Sampling.truncatedHash(input);
+    const wrappedStart = start % total;
+    const end = wrappedStart + count;
+
+    // If the range we're testing wraps, we have to check two ranges: from start
+    // to max, and from min to end.
+    if (end > total) {
+      return (
+        Sampling.rangeBucketSample(inputHash, 0, end % total, total)
+        || Sampling.rangeBucketSample(inputHash, wrappedStart, total, total)
```
gijsk: does this pass eslint on m-c? I would have done:

```js
return a ||
       b;
```

but I don't know that we enforce that style nor do I really mind either way. I've just never seen this before. :-)

recipe-client-addon/lib/Sampling.jsm
```diff
@@ -5,77 +5,121 @@
...
+      );
+    }
+
+    return Sampling.rangeBucketSample(inputHash, wrappedStart, wrappedStart + count, total);
```
gijsk: nit: use `end` rather than `wrappedStart + count`

recipe-client-addon/test/test_Sampling.js
```diff
@@ -0,0 +1,64 @@
+"use strict";
+// Cu is defined in xpc_head.js
+
+Cu.import("resource://gre/modules/Task.jsm", this);
+Cu.import("resource://shield-recipe-client/lib/Sampling.jsm", this);
+
+add_task(function* testStableSample () {
+  // Absolute samples
+  equal(yield Sampling.stableSample("test", 1), true, "stableSample returns true for 100% sample");
```
gijsk: Huh, I didn't realize this was a thing, I thought you needed the `Assert.` prefix. Thanks!

recipe-client-addon/test/test_Sampling.js
```diff
@@ -0,0 +1,64 @@
...
+  // Absolute samples
+  equal(yield Sampling.bucketSample("test", 0, 10, 10), true, "bucketSample returns true for 100% sample");
+  equal(yield Sampling.bucketSample("test", 0, 0, 10), false, "bucketSample returns false for 0% sample");
+
+  // Known samples. The numbers are nonces to make the tests pass
+  equal(yield Sampling.bucketSample("test-0", 0, 5, 10), true, "bucketSample returns true for known matching sample");
+  equal(yield Sampling.bucketSample("test-1", 0, 5, 10), false, "bucketSample returns false for known non-matching sample");
+});
+
+add_test(function testFractionToKey() {
```
gijsk: Hm, can we use add_task for all the tests instead of the mixture of add_task and add_test?

mythmon: I expected the linter to take offense to a generator function without a `yield` statement, but I guess that isn't the case.

mythmon:
Test failures are because of #430.

mythmon:
#430 has merged. I rebased this on master to fix the build.

mythmon:
@chartjes Should we flag you for review on add-on PRs as well? Since we have the monorepo now, we have both code bases in this repo.

PR #426, opened 2017-01-14
Centralize logging config and actually log to the console.

Osmose:
Turns out loggers made with Log.jsm need "appenders" attached to them to actually log things anywhere. Appenders are inherited from parent loggers (determined via dot separate, e.g. foo is a parent logger of foo.bar), so this patch configures the parent logger and creates loggers for the rest of the add-on as children of that parent, which logs to the console. This finally lets us actually see error messages when fetching recipes goes wrong!

I would've liked to write a test for the preference watching, but I suspect that, during test runs, the add-on itself thinks it's being run for the first time and doesn't set anything up. Someday we should figure that out but that's for another bug.

r?

### DIFF
--- a/recipe-client-addon/bootstrap.js
+++ b/recipe-client-addon/bootstrap.js
@@ -6,6 +6,7 @@
 const {utils: Cu} = Components;
 Cu.import("resource://gre/modules/Services.jsm");
 Cu.import("resource://gre/modules/Preferences.jsm");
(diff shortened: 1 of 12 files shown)


mythmon:
Oops. I forgot to wait for @chartjes to review this. I got excited about unblocking add-on work.

@chartjes We can revert this, if you'd like.

chartjes:
All good

--
Chris

PR #427, opened 2017-01-17
Backport changes from mozilla-central

mythmon:
This backports a patch @gijsk made to mozilla-central in bug 1330954, and adds a scripts to make backporting the tiniest bit easier.

### DIFF
--- a/recipe-client-addon/lib/Heartbeat.jsm
+++ b/recipe-client-addon/lib/Heartbeat.jsm
@@ -100,6 +100,7 @@ this.Heartbeat = class {
 
     // so event handlers are consistent
     this.handleWindowClosed = this.handleWindowClosed.bind(this);
+    this.close = this.close.bind(this);
 
     if (this.options.engagementButtonLabel) {
       this.buttons = [{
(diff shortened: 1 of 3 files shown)


recipe-client-addon/lib/Heartbeat.jsm
```diff
@@ -315,12 +313,16 @@ this.Heartbeat = class {
       this.chromeWindow.gBrowser.selectedTab = this.chromeWindow.gBrowser.addTab(this.options.postAnswerUrl.toString());
     }
 
-    if (this.surveyEndTimer) {
-      clearTimeout(this.surveyEndTimer);
-      this.surveyEndTimer = null;
-    }
+    this.endTimerIfPresent("surveyEndTimer");
+
+    this.engagementCloseTimer = setTimeout(() => this.close(), NOTIFICATION_TIME);
```
Osmose: @gijsk Why use the callback form of this if `this.close` is already bound?

gijsk: The binding thing happened in a patch revision (cf. last point of https://bugzilla.mozilla.org/show_bug.cgi?id=1330954#c2 ) and I forgot to update this - good find! We could indeed just pass this.close directly.

gijsk:
Also, sorry about doing this on bugzilla + m-c -- in hindsight, I should have used github. :-(

recipe-client-addon/lib/Heartbeat.jsm
Osmose: Cool, we'll merge this and then fix the nit in another PR to help track the backbackport. :P

PR #428, opened 2017-01-17
Add Storage.clearStorage to add-on for resetting storage.

Osmose:
This is mostly useful right now for use in the Browser Console when manually testing things.

@gijsk I was unsure if I should flag you for review on this as well. Should we be flagging you or another peer here on Github when we make changes to the add-on? If you r+ something on Github, is @mythmon clear to commit to mozilla-central (assuming a try push passes, or whatever other steps are involved with filing a bug and merging that I'm not aware of)?

### DIFF
--- a/recipe-client-addon/lib/Storage.jsm
+++ b/recipe-client-addon/lib/Storage.jsm
@@ -131,4 +131,18 @@ this.Storage = {
       cloneFunctions: true,
     });
   },
+
+  /**
+   * Clear ALL storage data and save to the disk.
+   */
+  clearAllStorage() {
+    return loadStorage()
+      .then(store => {
+        store.data = {};
+        store.saveSoon();
+      })
+      .catch(err => {
+        log.error(err);
+      });
+  },
 };
--- a/recipe-client-addon/test/browser_Storage.js
+++ b/recipe-client-addon/test/browser_Storage.js
@@ -34,4 +34,14 @@ add_task(function* () {
   const complex = {a: 1, b: [2, 3], c: {d: 4}};
   yield store1.setItem("complex", complex);
   Assert.deepEqual(yield store1.getItem("complex"), complex);
+
+  // Check that clearing the storage removes data from multiple
+  // prefixes.
+  yield store1.setItem("removeTest", 1);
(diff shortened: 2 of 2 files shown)


gijsk:
r=me in the sense that the code does what it says.

One question: AIUI this would allow one recipe to clear storage that affects another recipe. I'm... not sure that's always going to be a good idea - other recipes might not expect their stuff to get cleared in the middle of what they were doing... Why do we need it? Or am I misunderstanding whether we use the prefixes to separate different recipes?

> Should we be flagging you or another peer here on Github when we make changes to the add-on? If you r+ something on Github, is @mythmon clear to commit to mozilla-central (assuming a try push passes, or whatever other steps are involved with filing a bug and merging that I'm not aware of)

I think so. Doesn't really matter if the r+ was on github or here. It might be quicker if I look at things as they (individually) get developed rather than doing some kind of per-version batch merge where there'll be a bunch of unrelated changes that I have to separate again in order to make sense of what the m-c merge is doing. :-)

If it's any help, for readability (reader mode), we effectively review and merge on github, and occasionally I or other people update the m-c repo copy manually (usually landing with r=me because there are usually several commits from github in the single m-c merge). So that would be a similar system, AIUI.

gijsk:
Oh, and the counterexample would be pocket, where I've tried before to do reviews of merges, which gets frustrating if there are review comments that need to be addressed, because then you have to go back to github, fix them there, then redo the merge, etc. Might as well just take care of all the reviewing immediately.

Osmose:
> One question: AIUI this would allow one recipe to clear storage that affects another recipe. 

It shouldn't. The only place storage is exposed to the sandbox is via the driver's `createStorage` function, which calls `Storage.makeStorage`. The `Storage` object itself is never made available to the sandbox, so recipes can't call it.

Osmose:
> Why do we need it?

I want this so that testers can call the function from the BrowserConsole when running manual tests, mainly.

Osmose:
@gijsk Added you as a reviewer, wanted to confirm the misunderstanding about whether the sandbox can use this function before merging.

Issue #336, opened 2016-11-14
Option to make messaging studies show once or re-prompt 


Osmose:
#347 changes the `show-heartbeat` action to only prompt the user once, and never again, per-recipe. To accommodate the long-term Release Heartbeat survey, in which we do want to occasionally re-prompt users (after a long time we'll exhaust the user population, so we want to eventually make people eligible to see the prompt again), we should add a new argument to `show-heartbeat` recipes for configuring when a recipe should re-prompt users.

The new argument should be a choice between two options, using radio buttons:

- Do not show this prompt to specific users more than once
- Allow re-prompting users who have already seen this prompt after `X` days since they last saw it

`X` should be a number input that defaults to 30, is disabled if the option isn't selected, and is required if the option is selected.

When run, `show-heartbeat` should store the datetime that it was seen. If the first option is selected, the action should not prompt the user if any datetime is logged If the second option is selected, the action should compare the request time returned by the server with the stored datetime, and if the difference is larger than the specified delay, then it should re-prompt and store the request time as the last-seen date.

MattGrimes:
We would normally only want to reprompt a user when the message is something important enough that we want them to make an explicit choice. We should stop prompting when the user clicks on a button, rates with stars, or answers a question.

Osmose:
@MattGrimes I think we'll also want the option for re-prompting after X days regardless of interaction for the ongoing Heartbeat survey. I'm expecting it to be set to something like 90-180 days. Does that make sense to keep as a third option alongside the user-interaction-specific option?

MattGrimes:
Absolutely if it's something you are willing to build. Hack would be to create a new recipe ID and just use the same bucket.

Osmose:
So then the new spec is 3 options:

- Do not show this prompt to specific users more than once
- Show this prompt until the user interacts with it in any way, and then never again.
- Allow re-prompting users who have already seen this prompt after `X` days since they last saw it

mythmon:
Does it make sense to have separate actions for this? That way we could have unique UI for the different cases, and it may simplify the code in the actions as well.

andymikulski:
After thinking about it for a few days - there's definitely arguments for either approach.

IMO, I think the UX would be better if the `show-heartbeat` action just had these as configurations versus having a bunch of additional actions like `show-heartbeat (once)`, `show-heartbeat (nag)`, etc pollute the actions list. Though, to be fair, multiple actions would be easily filterable on the recipe list page.

[...]

PR #429, opened 2017-01-17
#336 Configurable Reprompting

andymikulski:
- Adds `repeatOption` and `repeatEvery` properties to `show-heartbeat` actions.
  - `repeatOption` select input consists of three options: `once`, `nag`, or `xdays`
    - `once` will display a prompt once and only once, ever
    - `nag` will display until the user interacts with the prompt in a meaningful way
    - `xdays` will show after waiting at least _x_ days since the last time it appeared
  - If `xdays` is selected, an additional form field for `repeatEvery` appears, which dictates how often the prompt is due to show.
- Adds tests for new show-heartbeat configurations

### DIFF
--- a/recipe-server/client/actions/show-heartbeat/index.js
+++ b/recipe-server/client/actions/show-heartbeat/index.js
@@ -19,8 +19,13 @@ export default class ShowHeartbeatAction extends Action {
   }
 
   /**
-   * Calculates the number of milliseconds since the last heartbeat from any recipe
-   * was shown. Returns a boolean indicating if a heartbeat has been shown recently.
+   * Returns a boolean indicating if a heartbeat has been shown recently.
+   *
+   * Checks the saved `lastShown` value against the current time
+   * and returns if the time is under HEARTBEAT_THROTTLE milliseconds.
+   *
+   * @async
+   * @return {Boolean}  Has any heartbeat been shown recently?
    */
   async heartbeatShownRecently() {
     const lastShown = await this.heartbeatStorage.getItem('lastShown');
@@ -34,7 +39,10 @@ export default class ShowHeartbeatAction extends Action {
   /**
    * Checks when this survey was last shown,
    * and returns a boolean indicating if the
-   * user has seen this survey already or not.
+   * user has ever seen this survey or not.
+   *
+   * @async
+   * @return {Boolean}  Has the survey ever been shown?
    */
   async surveyHasShown() {
     const lastShown = await this.storage.getItem('lastShown');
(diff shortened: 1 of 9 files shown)


recipe-server/client/actions/show-heartbeat/index.js
```diff
@@ -16,35 +19,218 @@ export default class ShowHeartbeatAction extends Action {
...
+
+  /**
+   * Checks the repeat argument for this recipe,
+   * then determines if the recipe can be qualified as 'ran'.
+   * This ultimately decides if the prompt is shown at all
+   * to the end user.
+   *
+   * @return {boolean}        Has the heartbeat been shown?
+   */
+  async heartbeatHasRan() {
```
andymikulski: I'm confused. This function checks the recipe to see if it's fulfilled the requirements to be considered 'shown' (that `once` has been shown once, that `nag` has been interacted with, etc). It doesn't determine if anything _should_ run - it just checks if the heartbeat has met its specific criteria for being 'ran'.

The function `shouldNotExecute` (Line 214) determines if anything should run (or specifically, shouldn't). It checks normandy's testing state, the global heartbeat timer, and then checks if this specific heartbeat has ran (via `heartbeatHasRan`).

Perhaps the naming convention needs to be changed for some of these functions?

recipe-server/client/actions/tests/test_show-heartbeat.js
```diff
@@ -82,112 +56,300 @@ describe('ShowHeartbeatAction', () => {
...
+        arguments: {
+          repeatOption: 'xdays',
+          repeatEvery,
+        },
+      };
+
+      it('should not show if less than `repeatEvery` days has elapsed', async () => {
+        const action = new ShowHeartbeatAction(normandy, recipeFactory({ ...xdaysConfig }));
+
+        let idx = 0;
```
andymikulski: Old habit - using `for (let i=0; ...` creates a new variable on each loop, defining it beforehand will only use the one instance. I'll change it.

recipe-server/client/control/components/RecipeForm.js
```diff
@@ -54,15 +55,26 @@ export class RecipeForm extends React.Component {
...
+    const displayedRecipe = this.props.recipe;
+
+    if (!isCloning || !displayedRecipe) {
+      return null;
+    }
+
+    this.cloneCache = this.cloneCache || {};
+    const cache = this.cloneCache;
+
+    if (!cache[displayedRecipe.id]) {
```
andymikulski: I had noticed that I had set up the cloning message to be rendered via 
```{ this.renderCloningMessage() }```, meaning the function fires on each render. Adding the caching at least saved some cycles, but it's not a big deal if it's removed.

recipe-server/client/actions/show-heartbeat/index.js
mythmon: Sorry, I was a bit confused here. It isn't clear to me what "run" means in this context. I was expecting that "run" meant either (1) the action has ever been executed by a runner for this recipe, or if (2) a heartbeat prompt has ever been shown for this recipe.

I think that you are describing run as (3) this recipe has fulfilled all of it's criteria for never being executed again.

The comment "Has the heartbeat ever been shown" makes me think (2). Overall I find this flow pretty hard to follow, since it is split up among a lot of different functions, and there are a lot of ill defined concepts (like "run").

All that is to say: I think changing the comment in the `@return` section will be enough to ward off enough future confusion here.

mythmon:
Looks good. Sorry about the confusion with naming.

recipe-server/client/actions/show-heartbeat/index.js
andymikulski: I rewrote a bunch of the language and ended up renaming the function all together - https://gist.github.com/andymikulski/f2531bdab2bec531ffc0e188f1b62fc2 - Let me know if it looks good and I'll push the changes before merging the branch.

mythmon: +1, that reads a lot clearer to me. Thanks for re-working it!

PR #430, opened 2017-01-18
Pin docker-compose used in mock-server CI to 1.9.0.

Osmose:
r?


Osmose:
So the first commit failed because the pip we have on the CircleCI VM is too old to know about constraints files. So I added a commit to update pip, and I used a requirements file with hashes to verify the pip we're getting. In theory, the Ubuntu image we're using has pip 8.1.1 installed on it, and yet CI seems to be complaining about the hash syntax in the requirements file. 

I tried `sudo pip` for installing, and it seemed to complain about the newlines in the requirements file. Weird. `pip` alone instead said that `--hash` was not an option, but helpfully said it was using pip 7.1.2.

I'm not entirely sure the best way to proceed. Should I try installing pip from the ubuntu software repos? Should I use [pipstrap](https://github.com/erikrose/pipstrap)? Should I try something else?

mythmon:
If we can install a version of pip that supports hashing from Ubuntu's repos, that's ideal. The pipstrap README says 

> When your OS packages pip 8 or you otherwise get a copy of pip 8 you trust onto your servers, you can dispense with Pipstrap.

I haven't checked if Ubuntu has such a package. If not, then pipstrap is the next best option.

mythmon:
@chartjes I'm going to merge this without your review because it is making all of our CI fail. It is still probably worth looking over this, since it touches our CI infra.

PR #432, opened 2017-01-18
#424 Fix ghost recipe bug

### LINKED ISSUE #424: control: Ghost recipes made when updating just-created recipe
When editing a recipe that was just created, there seem to be multiple versions of the recipe added to the Redux store. The extra versions don't end up on the server.

Steps to reproduce:

1. In the control interface, create a new recipe
2. Without leaving the page, make a change to the recipe.
3. Click the "Recipes" bread crumb to go back to the listing. Notice that there are multiple copies of the recipe that was just created, equal to the number of times the recipe was changed.
4. Refresh the page. The recipes go away.

andymikulski:
Fixes #424 

- Fixes issue where adding a new recipe and then updating it would leave extra copies in the recipe list
- Adds `merge-arrays` util, which allows merging/de-duping two or more arrays
- Adds tests for `merge-arrays` util

I wasn't sure what sort of tests to add to prevent this from happening again. Technically, I think the existing tests should have caught the issue to begin with. @mythmon any suggestions on that? (I'll wait to request a QA review until I know what the test plan is)


mythmon:
Test failure is unrelated, and fixed by #430.

Looking through the tests, I see that we test that a recipe with an id updates instead of creating a new one, but we don't actually test the flow of (1) creating a recipe, and then (2) updating it using the same form. We only test what we assume the data would like like in step 2 of that.

How I'd test this would depend on where the bug lies. If the bug is in the component, I'd mount the `RecipeForm` component and simulating submitting the form and then updating the form and saving again. If the bug lies in the reducer, the test I'd do is to fire off a set of actions that reveal the bug (that is, probably, create, and then an update action based on what is in the store). I'd expect the bug is probably closer to the components, but I haven't looked very hard.

Finally, I think this solves the problem well in that it should prevent duplicate ids in the list, which is the underlying issue. However, I wonder if this problem could be addressed better with some design changes instead. In the version of the code in this PR, every time we update the list of recipes we linearly search through all of them and update the one with the correct id, or insert a new value. To me that sounds an awful lot like an object, where the keys are IDs and the values are the recipes.

Do you think it makes sense to change the object stored in the state from a list of recipes to an object that related ids to recipe objects? (or a `Map`? or an `Immutable.Map`?) I used this pattern [in a side project](https://github.com/mythmon/tenbyten/blob/master/src/reducers/players.js), and I think it worked out pretty well.

Like I said, I think what you have here solves the problem well. If you think the array based method is better, then we should stick with it. At this point, you've probably worked with this code more than I have. I think it's your decision.

mythmon:
#430 has merged. Rebasing this on master will fix the build.

andymikulski:
@mythmon I've added 'process' tests (e.g. creating + updating then checking list) as the bug was in the reducer. I've also converted the recipe list into a hash as you suggested, and added a couple selectors for the component or two that needed them.

This should be ready for another review.

recipe-server/client/control/tests/reducers/test_controlAppReducer.js
```diff
@@ -207,18 +211,53 @@ describe('controlApp reducer', () => {
...
+            name: 'Dolor set amet',
+            enabled: true,
+          },
+        },
+      },
+    });
+  });
+
+  describe('Recipe Creation and Updating', () => {
+    it('only has one recipe in the `list` after creating and updating a new recipe', async () => {
```
mythmon: This test looks good, and is probably a valuable thing to have, but I don't think it actually tests the ghost recipe problem. I copied this test and added it to the master branch without any of the other changes. After modifying the call to expect to be the below the test passed. (I changed it to expect the old, array based format, instead of the objects).

```
expect(store.recipes.list).toEqual([
  {
    id: 3,
    name: 'Updated recipe name',
    enabled: true,
  },
]);
```

Since this test didn't fail before this patch, I think we'll have to go with the more complex UI test if we want to verify this.

recipe-server/client/utils/merge-arrays.js
```diff
@@ -0,0 +1,45 @@
+import {
```
mythmon: We can get rid of this file now, right?

recipe-server/client/control/tests/reducers/test_controlAppReducer.js
andymikulski: Aight - took some playing but I figured out the issue. The bug didn't lie with updating the recipe, it was creating the recipe and then loading the recipe list. Creating a recipe stored it in memory, then loading the recipe list from the API would add just append to the list in memory, not checking against dupes.

I had figured it out using a UI test, but in the end a reducer test was all that was really needed, so that's the test that went in. I also confirmed the test failed appropriately on `master`, as well.

This should be all set!

mythmon:
The new test looks great, and fails on master, so I think we're good here. Thanks!

Issue #391, opened 2016-12-19
Create a set of test API responses


Osmose:
We talked with QA during the Hawaii workweek, and in order to facilitate their testing of the system add-on, @mythmon and I came up with a set of interesting test cases to cover:

- Filter Expression targeting
  - Documented normandy attributes (channel, geolocation, version, etc.)
  - Sampling 100% and 0%
  - Telemetry packet attributes
- Basic `console-log` execution
- Basic `heartbeat` execution
  - 5 star rating
  - Engagement button
- Extreme cases
  - Giant filter expression
  - `console-log` with a very large message
  - `show-heartbeat`
    - Very large arguments
    - With an ad blocker installed that resets localstorage on reboot
- Error cases
  - Service unavailable
  - Malformed JSON
  - Invalid signature

We should come up with the response JSON we'd expect from the service in each of these test cases, which will help us set up a test server with fake responses that QA can use in their manual testing.

# Blocked by

* [x] #448 Re-enable mock-recipe-server in CI

PR #436, opened 2017-01-20
Event Emitter and Plugin problems

Osmose:
This fixes two problems I found while working on #391:

- The add-on attempts to clone addon objects into the sandbox, which causes problems. Also, the addon objects don't have filename info like `navigator.plugins` does. My solution was to just use an object derived from the addon object, and remove the filename property, since we're not really using it anyway.

  I ran into a lot of issues trying to write an add-on test for this, mainly trying to mock the list of installed plugins. I opted to leave a test out for now, but future attempts to improve the add-on test coverage should address this.
- The Heartbeat event emitter involves some hairy cloning and unwrapping, mainly due to it being in the sandbox and taking the driver as an argument, but being constructed by code outside the sandbox. To simplify this, I moved it out of the sandbox and removed the driver dependency, since it was only using it to log anyway. 

  From what I read on the original review for the add-on, having the emitter outside the sandbox was mainly a problem due to it having references to other things, but this version is self-contained. I might be wrong on that though. 

  This doesn't really change any behavior, so existing test coverage should be fine here.

r? @mythmon for everything, @gijsk for the add-on bits.


recipe-client-addon/lib/EventEmitter.jsm
```diff
@@ -1,17 +1,16 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
-// This file is meant to run inside action sandboxes
-
 "use strict";
 
+const {utils: Cu} = Components;
+Cu.import("resource://shield-recipe-client/lib/LogManager.jsm");
```
gijsk: My brain is broken and so I don't remember this thing. It doesn't seem to be in this diff. I can't find it in mozilla-central. It also doesn't seem to be in the list of modules to unload in bootstrap.js - I'm assuming that's just an oversight and this got added elsewhere? Or something?

recipe-client-addon/lib/EventEmitter.jsm
```diff
@@ -20,15 +19,13 @@ this.EventEmitter = function(driver) {
       Promise.resolve()
         .then(() => {
           if (!(eventName in listeners)) {
-            driver.log(`EventEmitter: Event fired with no listeners: ${eventName}`);
+            log.info(`EventEmitter: Event fired with no listeners: ${eventName}`);
             return;
           }
-          // freeze event to prevent handlers from modifying it
-          const frozenEvent = Object.freeze(event);
```
gijsk: So I just checked in Firefox's browser console using something like this:

```js
let x = {foo: 5};
x = Object.freeze(x);
y = Cu.cloneInto(y, content) // this only works with e10s disabled
Object.isFrozen(y) // returns false
y.foo = 10; // changes y.foo
```

So I don't think this works as-is. :-(

If the event object is cloned into the sandbox, I wouldn't expect there to be issues with calling `Object.freeze` on it. What exactly happens if you try and freeze the object?

recipe-client-addon/lib/Heartbeat.jsm
```diff
@@ -92,7 +91,7 @@ this.Heartbeat = class {
     }
 
     this.chromeWindow = chromeWindow;
-    this.eventEmitter = eventEmitter;
+    this.eventEmitter = new EventEmitter();
```
gijsk: Maybe you want to clone this into the sandbox directly?

recipe-client-addon/lib/EventEmitter.jsm
Osmose: It's from https://github.com/mozilla/normandy/commit/388cbecf2c8534fea0c9cd8786449c3891ceee08, which I believe got merged before we decided to do reviews on Github and bulk-commit changes to mozilla-central, but we didn't make the bug for it in time before that decision was made. So it'll have to get your review, at which point you'd probably note something like how I missed adding it to the list of modules to unload. :P

If you want, we can just do comments directly on that commit and I'll make a PR with feedback fixes. How does that sound?

recipe-client-addon/lib/EventEmitter.jsm
Osmose: I blame @jimblandy for telling me this was a thing, and myself for not properly testing it.

The problem with freezing as it was before is that the cloned object passed in to `emit` is a proxy object (because it has been cloned into another compartment, so we get an xrayed wrapper? I think) and throws when you try to freeze it. 

So we need to freeze inside of the sandbox, but IIRC there were concerns with the EventEmitter class having a reference to the sandbox and also being cloned into it due to memory leaks and/or potential security issues. So I was hoping to avoid the EventEmitter class itself having a reference to the sandbox.

I'll see if I can find a way around this, and will report back.

recipe-client-addon/lib/EventEmitter.jsm
gijsk: WFM!

recipe-client-addon/lib/EventEmitter.jsm
gijsk: I probably haven't thought this through enough, but I thought that if heartbeat, which constructs the event emitter, clones it into the sandbox (and we throw away the original non-cloned reference, storing only the cloned version as `this.eventEmitter`) the instance that we call `emit` on has access to the object?

Osmose: It does! And that _almost_ works.

The problem with cloning the object into the sandbox is that we can't call `emit` on the cloned emitter without calling `waiveXrays` on it, and if we do that, then the resulting emitter in the privileged compartment can have callbacks attached to it, which the sandbox could then trigger by calling `emit` on the cloned emitter, which seems like a bad idea for potential future engineers who think adding callbacks to the emitter is an easy fix for some problem.

So, I tried re-introducing the sandbox parameter to the emitter, and creating a new emitter specifically for the sandbox that doesn't have `emit`. I'm not entirely sure how to reason about potential leaks now that we've reintroduced the sandbox parameter to the event emitter, so I wanted to check with you. 

@gijsk How does this latest commit look?

recipe-client-addon/lib/Utils.jsm
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Copy a simple object by converting it to JSON and back.
+ * @param  {Object} value Value to clone
+ * @return {Object}       A deep copy of the value, or the original
+ *                        value if it's not an object.
+ */
+this.jsonCopy = function(value) {
+  if (value !== null && typeof value === "object") {
+    value = JSON.parse(JSON.stringify(value));
```
gijsk: Nit: can just use: `Cu.cloneInto(value, {});` instead of going through JSON. :-)

recipe-client-addon/lib/Utils.jsm
```diff
@@ -0,0 +1,14 @@
+this.EXPORTED_SYMBOLS = ["jsonCopy"];
```
gijsk: Now you get to add this to the bootstrap.js list of modules to unload. :-)

recipe-client-addon/lib/EventEmitter.jsm
```diff
@@ -1,34 +1,42 @@
...
-    throw new Error("driver must be provided");
-  }
+const log = LogManager.getLogger("event-emitter");
 
+this.EventEmitter = function(sandboxManager) {
   const listeners = {};
 
   return {
+    createSandboxedEmitter() {
+      return sandboxManager.cloneInto({
```
gijsk: I would expect you to need to pass `{cloneFunctions: true}` for the `on`, `off` and `once` calls to work from the sandbox. Do you not? Do we have tests that verify that calling those from inside the sandbox works?

gijsk: (I checked only the `EventEmitter` browser test, and that doesn't seem to make calls from inside the sandbox...)

recipe-client-addon/lib/EventEmitter.jsm
```diff
@@ -1,34 +1,42 @@
...
           }
-          // freeze event to prevent handlers from modifying it
-          const frozenEvent = Object.freeze(event);
           // Clone callbacks array to avoid problems with mutation while iterating
           const callbacks = Array.from(listeners[eventName]);
           for (const cb of callbacks) {
-            cb(frozenEvent);
+            // Copy event to prevent modifcations leaking between callbacks.
+            const copiedEvent = jsonCopy(event);
+            cb(sandboxManager.cloneInto(copiedEvent));
```
gijsk: Oh, so actually, the `cloneInto` here will guarantee every callback gets a fresh copy anyway.

recipe-client-addon/lib/EventEmitter.jsm
gijsk: I think in principle this looks OK from a leakiness-perspective... pushing to try to doublecheck wouldn't hurt. Just push to linux64 for only debug builds and only mochitest-bc and mochitest-e10s-bc tests and you should have an answer in relatively short order.

Separately though, you could still call `emit()` by eval'ing it inside the sandbox. It's a bit of a faff though, so if this works and offers the functionality we need, great.

Though, thinking about this more... you now have 2 emitters, and you call `emit()` on the one that lives outside the sandbox. But that one won't have the handlers of the one inside the sandbox, right? I though the idea was that code inside a recipe could listen for events from the heartbeat infra? Am I wrong about that, or wrong about how we forward the handling of events, or something else? It looks like the heartbeat tests are using the heartbeat's `eventEmitter` property, not the sandboxed one, so they will keep working, but if I'm right then 'real' recipes might be broken.

recipe-client-addon/lib/Utils.jsm
Osmose: @mossop I TOLD YOU

Mossop: Treat it as a learning opportunity, I learned something too!

mythmon:
I'm holding off on my review until @gijsk's comments are addressed.

recipe-client-addon/lib/EventEmitter.jsm
Osmose: > But that one won't have the handlers of the one inside the sandbox, right?

It will, because the handlers aren't a property of the emitter, but instead are in the scope of the constructor function, and the docs on [`exportFunction`](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Language_Bindings/Components.utils.exportFunction) say that the cloned function has access to its surrounding closure.

Osmose:
I removed the JSON cloning since it was unnecessary, and added a few convenience methods to SandboxManager that were useful in my testing (and also used them in RecipeRunner since they're useful there as well). I also added a test for the emitter executing handlers in the sandbox to ensure that that works properly.

I _think_ this is ready for review again.

gijsk:
Thanks for the added tests!

Issue #420, opened 2017-01-11
Document use of mock server for QA


Osmose:
Splitting this out from #392 since I want to wait until we have the mock server deployed to test out the steps for using it before writing the docs for them.

> We'll want to document how to use the script to update the test service, and document for QA how to use the test service (e.g. how to change Firefox to trust the dev autograph's signatures, or how to clear out the flags to track if a heartbeat recipe has executed so that they can be run again).

Blocked by #392 and [bug 1330415](https://bugzilla.mozilla.org/show_bug.cgi?id=1330415).

Osmose:
This is done and up at https://normandy-mock.dev.mozaws.net/.

PR #437, opened 2017-01-20
mock-recipe-server Test Cases

Osmose:
Phew, finally! Okay. So. This is a bit later than I expected, but I'm 200% convinced it's crazy-useful.

- Adds an index page for the mock API server that lists out instructions on how to run test cases, the test cases and their descriptions, and code that you can run in the Browser Console to run the test case. This is most of #420, with the minor exception of some links in our docs/wiki pointing to the actual live mock server.

  
- Renames `Fixture` to `TestCase`. I'm mildly annoyed that this might be confusing compared to other testing libraries with test case classes, but it's really a better word for what these are.
- Adds all the test cases mentioned in #391. I had to move the serialization logic to the `TestCase` class to override it for two of the tests, and I'm not 100% happy with how I did that move, but I think it does, in fact, belong there.

This one's open for anyone who wants to review it. r?

# ~~Blocked by~~

* [x] #448 Re-enable mock-recipe-server in CI

### DIFF
--- a/mock-recipe-server/api_index.html
+++ b/mock-recipe-server/api_index.html
@@ -1,0 +1,166 @@
+<!DOCTYPE html>
+<html lang="en-us">
+  <head>
+    <title>SHIELD Test Service</title>
+    <style>
+      * {
+        box-sizing: border-box;
+      }
+
+      body {
+        background: #F6F4EF;
+        color: #585755;
+        font-family: sans-serif;
+        font-size: 15px;
+        line-height: 1.4em;
+        margin: 0 auto;
+        width: 720px;
+        word-wrap: break-word;
+      }
+
+      a {
+        font-weight: 600;
+        text-decoration: none;
+      }
+
+      li {
+        margin-bottom: .5em;
+      }
+
+      pre, code {
+        background: #EEE;
+        border: 1px solid #CCC;
+      }
+
+      pre {
+        margin: 0 0 1.5em;
+        padding: 5px;
+      }
+
+      code {
+        box-decoration-break: clone;
+        padding: 2px 4px;
+        position: relative;
+        top: -1px;
+      }
+
+      .testcase {
+        background: #FFF;
+        margin: 0 0 20px;
+      }
+
+      .testcase h3 {
+        background: #585755;
+        color: #FFF;
+        margin: 0;
+        padding: 15px
+      }
+
+      .testcase-content {
+        padding: 15px;
+      }
+
+      .testcase-content p {
+        margin: 0 0 1.5em;
+      }
+
+      textarea {
+        border: 1px solid #ece7db;
+        border-radius: 3px;
+        display: block;
+        height: 6em;
+        overflow: auto;
+        padding: 5px;
+        white-space: pre;
+        width: 100%;
+      }
+    </style>
+  </head>
+  <body>
+    <h1>SHIELD Test Service</h1>
+    <p>
+      This service hosts the data needed for running manual tests of the
+      <a href="https://wiki.mozilla.org/Firefox/SHIELD">SHIELD</a> recipe
+      client. The scripts that generate this page and the data are
+      <a href="https://github.com/mozilla/normandy/tree/master/mock-recipe-server">
+      hosted on Github</a>.
+    </p>
+    <h2>Setting up for Testing</h2>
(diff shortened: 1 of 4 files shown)


Osmose:
Oh yes, another note: This relies on #428 indirectly, in that the Browser Console code uses it. But it doesn't directly depend on the commits.

mythmon:
We disabled the mock-recipe-server in #447 due to integration problems. This PR is now blocked on #448 to re-enable mock-recipe-server.

mock-recipe-server/api_index.html
```diff
@@ -0,0 +1,166 @@
...
+      let testcases = Array.from(document.querySelectorAll('.testcase'));
+      for (let testcase of testcases) {
+        let url = new URL(`/${testcase.dataset.name}/api/v1`, window.location);
+        testcase.querySelector('.url').textContent = url.href;
+        testcase.querySelector('.browser-code').value = (
+          `Cu.import('resource://shield-recipe-client/lib/Storage.jsm');\n` +
+          `Storage.clearAllStorage();\n` +
+          `Services.prefs.setCharPref("extensions.shield-recipe-client.api_url", "${url.href}");\n` +
+          `Cu.import('resource://shield-recipe-client/lib/RecipeRunner.jsm');\n` +
+          `RecipeRunner.start();`
```
mythmon: On the one hand, you can do multiline strings directly with backticks. On the other hand, the indentation is super annoying, so this is probably the right thing to do.

mock-recipe-server/generate.py
```diff
@@ -22,11 +21,9 @@
 
 
 # Now that Django is set up we can import Django things.
-from normandy.base.utils import canonical_json_dumps  # noqa
-from normandy.recipes.api.serializers import ClientSerializer  # noqa
-from normandy.recipes.models import Action  # noqa
+from django.template import Context, Template  # noqa
 
-from fixtures import get_fixtures  # noqa
+from testcases import get_testcases  # noqa
```
mythmon: What's the `#noqa` for here?

mock-recipe-server/testcases.py
```diff
@@ -0,0 +1,375 @@
...
+    @property
+    def description(self):
+        return '<p class="description">{}</p>'.format(self.__doc__)
+
+    def load(self):
+        """
+        Clear out all existing recipes and load this test's data in
+        its place.
+        """
+        Recipe.objects.all().delete()
```
mythmon: It would be really cool if this was a context manager that used database transactions to automatically roll back the changes.

Alternatively, do we need to delete Actions and other objects besides the Recipes?

mock-recipe-server/testcases.py
```diff
@@ -0,0 +1,375 @@
...
+            'true',
+            '1 + 7 == 8',
+            '5 > 4',
+            '((4 * 8) - 12) > -5',
+            '4 in [1, 2, 4]',
+            '{foo: "bar"}.foo == "bar"',
+            '"bar" in "foobarbaz"',
+            'normandy.userId|stableSample(1)',
+        ]
+        truthy_expressions = [random.choice(truthy_choices) for _ in range(1000)]
```
mythmon: Does this actually work? That's impressive.

mock-recipe-server/generate.py
Osmose: The linter complains about imports that aren't at the top of the file, but we can't import anything that relies on Django until after setting up the path stuff. Flake8 sucks are configuring exceptions, so the only good solution I could find was `noqa`, but it's annoyingly broad.

mock-recipe-server/testcases.py
Osmose: Recipes are the only thing I think we need to delete, but a context manager is a great idea.

mock-recipe-server/testcases.py
Osmose: Yep, it just takes a few seconds to evaluate.

mock-recipe-server/testcases.py
Osmose: AFAICT we can't use a transaction here because the code loading data in the database and the code fetching from the database are in two separate containers; I get 500 errors from the service that I believe are because it can't find any recipes in the database.

PR #438, opened 2017-01-20
Approval requests

### LINKED ISSUE #417: Implement approval requests backend
Approval requests will be one-to-one with revisions.

Approval should have a nullable boolean field for `approved`:
- None: No approval or rejection
- True: Approved
- False: Rejected

Closing the approval request should delete the approval request.

This issue only covers creating the approval request object and updating the revisions so that we can identify them as latest draft and latest approved revisions.

rehandalal:
Fixes #417 

f?


recipe-server/normandy/recipes/api/serializers.py
```diff
@@ -170,6 +189,9 @@ class Meta:
             'date_created',
             'recipe',
             'comment',
+            'is_approved',
+            'is_rejected',
+            'is_pending_approval',
```
mythmon: Instead of 3 booleans, what about something like `approval`, that has multiple states, `approved`, `rejected`, or `pending`?

recipe-server/normandy/recipes/api/views.py
```diff
@@ -129,6 +131,26 @@ def disable(self, request, pk=None):
...
+            return Response({'error': 'Recipe has no revisions to be approved.'},
+                            status=status.HTTP_400_BAD_REQUEST)
+        elif recipe.latest_revision.is_rejected:
+            return Response({'error': 'The latest revision has already been rejected. Please make '
+                                      'changes before requesting approval.'},
+                            status=status.HTTP_400_BAD_REQUEST)
+
+        revision = recipe.latest_revision
+        revision.approval_request = ApprovalRequest(revision=revision, user=request.user)
+        revision.save()
```
mythmon: This seems racey. Should we include the current revision id along with the request approval?

recipe-server/normandy/recipes/api/views.py
```diff
@@ -138,6 +160,32 @@ class RecipeRevisionViewSet(viewsets.ReadOnlyModelViewSet):
...
 
+class ApprovalRequestViewSet(viewsets.ReadOnlyModelViewSet):
+    queryset = ApprovalRequest.objects.all()
+    serializer_class = ApprovalRequestSerializer
+    permission_classes = [
+        AdminEnabledOrReadOnly,
+    ]
+
+    @detail_route(methods=['POST'])
+    def approve(self, request, pk=None):
```
mythmon: Should this raise an error if the approval is already approved or rejected? Same with the reject method below.

recipe-server/normandy/recipes/api/views.py
```diff
@@ -138,6 +160,32 @@ class RecipeRevisionViewSet(viewsets.ReadOnlyModelViewSet):
...
+    def reject(self, request, pk=None):
+        approval_request = self.get_object()
+        approval_request.reject(approver=request.user)
+        return Response(ApprovalRequestSerializer(approval_request).data)
+
+    @detail_route(methods=['POST'])
+    def close(self, request, pk=None):
+        approval_request = self.get_object()
+        approval_request.delete()
+        return Response(status=status.HTTP_204_NO_CONTENT)
```
mythmon: I thought that we talked about closing an approval also removing the changes in the approval? Or was that rejecting? Or did we drop that?

recipe-server/normandy/recipes/api/serializers.py
rehandalal: makes sense

recipe-server/normandy/recipes/api/views.py
rehandalal: yeah probably

recipe-server/normandy/recipes/api/views.py
rehandalal: I kind of remember something like that but I don't remember exactly what we decided.

rehandalal:
@mythmon this is ready for review

recipe-server/normandy/recipes/models.py
```diff
@@ -319,6 +346,39 @@ def save(self, *args, **kwargs):
...
+class ApprovalRequest(models.Model):
+    revision = models.OneToOneField(RecipeRevision, related_name='approval_request')
+    created = models.DateTimeField(default=timezone.now)
+    creator = models.ForeignKey(User, on_delete=models.SET_NULL, related_name='approval_requests',
+                                null=True)
+    approved = models.NullBooleanField(null=True)
+    approver = models.ForeignKey(User, on_delete=models.SET_NULL, related_name='approved_requests',
+                                 null=True)
+
+    class AlreadyApproved(Exception):
```
mythmon: I think this name is a little misleading, because it could also be raised by a recipe that has been rejected (not approved). Maybe this should be renamed? `NotPending` might be good. Or maybe have both `AlreadyApproved` and `AlreadyRejected`?

recipe-server/normandy/recipes/tests/test_api.py
```diff
@@ -415,6 +416,71 @@ def test_it_serves_revisions(self, api_client):
...
+        assert res.status_code == 200
+        assert res.data == []
+
+    def test_approve(self, api_client):
+        r = RecipeFactory()
+        a = ApprovalRequestFactory(revision=r.latest_revision)
+        res = api_client.post('/api/v1/approval_request/{}/approve/'.format(a.id))
+        assert res.status_code == 200
+
+        r = Recipe.objects.get(pk=r.pk)
```
mythmon: You can use `r.refresh_from_db()` instead.

recipe-server/normandy/recipes/tests/test_api.py
```diff
@@ -415,6 +416,71 @@ def test_it_serves_revisions(self, api_client):
...
+        assert r.is_approved
+
+    def test_approve_already_approved(self, api_client):
+        r = RecipeFactory()
+        a = ApprovalRequestFactory(revision=r.latest_revision)
+        a.approve(UserFactory())
+
+        res = api_client.post('/api/v1/approval_request/{}/approve/'.format(a.id))
+        assert res.status_code == 400
```
mythmon: It is probably a good idea to assert the text of the error that was returned here.

recipe-server/normandy/recipes/tests/test_models.py
```diff
@@ -296,6 +298,103 @@ def test_revision_id_changes(self):
...
+
+        recipe.update(name='second')
+        assert recipe.name == 'second'
+
+        approval = ApprovalRequestFactory(revision=recipe.latest_revision)
+        approval.approve(UserFactory())
+        assert recipe.name == 'second'
+
+        recipe.update(name='third')
+        assert recipe.name == 'second'
```
mythmon: Ewwww. I guess this is just how it works, but that's really weird.

recipe-server/normandy/recipes/api/views.py
```diff
@@ -137,6 +139,53 @@ class RecipeRevisionViewSet(viewsets.ReadOnlyModelViewSet):
         AdminEnabledOrReadOnly,
     ]
 
+    @detail_route(methods=['POST'])
```
mythmon: This viewset doesn't include `permissions.DjangoModelPermissionsOrAnonReadOnly,` in the permissions, so any user can call this detail route, including anonymous users.

mythmon:
Test failures are related. I also have some questions about how various clients interact with this system that I'm working on. Maybe some docs about the approval workflow would be good?

mythmon:
To figure out if everything worked the way I expected, I wrote a test. Things generally work how I'd expect, with clients who don't care about revisions only getting approved things, so that's cool.

I ran into one major issue, and that is that a user can approve their own approval request. I think we don't want to allow that. Besides that, I think we should change a couple HTTP 200 responses into HTTP 201. Those are minor though.

I'll push that test to this branch.

rehandalal:
I've filed a new issue for writing docs, I'd like to land this so @andymikulski can start working off of it.

recipe-server/normandy/recipes/tests/test_models.py
chartjes: How does that test even pass?!?

rehandalal: it's a bit weird but basically when you `update` it creates a new "draft" so the name doesn't change till you approve the draft.

chartjes: Okay...can you please add a comment block in this test explaining that and then I will approve my review

rehandalal: done!

chartjes:
Again, I want to see a comment block explaining why "assert recipe.name == ‘second’” is being done despite changing the name. I would expect name to be ‘third’.

rehandalal:
@chartjes updated ^

PR #439, opened 2017-01-20
Make docker-compose build the image

rehandalal:
Playing with Circle, don't merge!

PR #434, opened 2017-01-19
New revisions and simple filters

rehandalal:
This is just the code in the [peer-signing](https://github.com/mozilla/normandy/tree/peer-signing) feature branch rebased against master.

r?

PR #440, opened 2017-01-20
Reuse server image  in docker-compose in CI

mythmon:
Hopefully this fixes some of the problems in #439.


mythmon:
@Osmose I think compose  is using the right images now, but something is going wrong with the recipe server ci. can you look at it?

Osmose:
Hmmm. I _was_ getting the error locally, but the last commit fixes it for me. Yet CI is still having issues.

We might have to make a wait-for-postgres script or something.

Osmose:
Permission denied! We're past the postgres hump, with the setup I've been most happy with over all the variations of things I've tried. This is back to erroring out at writing the product_details JSON to the filesystem.

rehandalal:
I've tried switching from `PDFileStorage` to `PDDatabaseStorage` but can't get that to work.

I keep getting:
```
django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.
```

Osmose:
Weird, what's the traceback on that? Are we importing from django too early somewhere?

mythmon:
As for as the permission denied problems, running as the root user instead of the default app user usually fixes that. I'm not sure how to do that with docker-compose though.

rehandalal:
The product details issue should be fixed. The latest version of product details has a bug that was causing the issues with `PDDatabaseStorage` so I rolled back the version. I tried to fix the subsequent issues with mock-recipe-server but I think someone else should probably look at that.

mythmon:
Closing in favor of #434, since we dropped mock-recipe-client in #447.

PR #441, opened 2017-01-20
recipe-client-addon: Add userId to  filter expressions and driver

mythmon:
While working on this, I realized I should also re-arrange the tests, which makes things a little nicer.


recipe-client-addon/lib/EnvExpressions.jsm
```diff
@@ -33,33 +38,53 @@ XPCOMUtils.defineLazyGetter(this, "jexl", () => {
...
+  },
 
-this.EnvExpressions = {
   eval(expr, extraContext = {}) {
-    const context = Object.assign({telemetry: getLatestTelemetry()}, extraContext);
+    // First clone the extra context
+    const context = Object.assign({}, extraContext);
+    // jexl handles promises, so it is fine to include them in this data.
+    context.telemetry = EnvExpressions.getLatestTelemetry();
+    context.normandy = context.normandy || {};
```
Osmose: Should we add a test for this fallback if `normandy` isn't provided? Maybe.

recipe-client-addon/lib/EnvExpressions.jsm
```diff
@@ -33,33 +38,53 @@ XPCOMUtils.defineLazyGetter(this, "jexl", () => {
...
+    const telemetry = {};
+    for (const key in mostRecentPings) {
+      const ping = mostRecentPings[key];
+      telemetry[ping.type] = yield TelemetryArchive.promiseArchivedPingById(ping.id);
+    }
+    return telemetry;
+  }),
+
+  getUserId() {
+    let id = prefs.getCharPref("user_id");
```
Osmose: Do we want a test asserting that this pulls from the pref we expect it to?

recipe-client-addon/lib/EnvExpressions.jsm
```diff
@@ -33,33 +38,53 @@ XPCOMUtils.defineLazyGetter(this, "jexl", () => {
...
+    console.log("pref user_id", id);
+    if (id === "") {
```
Osmose: Should we respect a user-set empty value in case privacy add-ons want to get rid of it? That may be a question for outside of this PR.

recipe-client-addon/lib/EnvExpressions.jsm
```diff
@@ -5,13 +5,18 @@
 "use strict";
 
 const {utils: Cu} = Components;
```
gijsk: This line will want a `Cc` and `Ci` definition as implied by Osmose's error.

recipe-client-addon/lib/EnvExpressions.jsm
```diff
@@ -5,13 +5,18 @@
...
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
 Cu.import("resource://gre/modules/TelemetryArchive.jsm");
 Cu.import("resource://gre/modules/Task.jsm");
 Cu.import("resource://shield-recipe-client/lib/Sampling.jsm");
 
+const {generateUUID} = Cc["@mozilla.org/uuid-generator;1"].getService(Ci.nsIUUIDGenerator);
+
 this.EXPORTED_SYMBOLS = ["EnvExpressions"];
 
+const prefs = Services.prefs.getBranch("extensions.shield-recipe-client.");
```
gijsk: I vaguely recall reading the add-on was switching to Preferences.jsm everywhere. Was there a reason to do something else here?

recipe-client-addon/lib/EnvExpressions.jsm
gijsk: Nit: Firefox style is: `if (!id) {`.

recipe-client-addon/test/browser/browser_EnvExpressions.js
```diff
@@ -5,7 +5,9 @@ Cu.import("resource://gre/modules/TelemetryController.jsm", this);
 Cu.import("resource://gre/modules/Task.jsm", this);
 
 Cu.import("resource://shield-recipe-client/lib/EnvExpressions.jsm", this);
-Cu.import("resource://gre/modules/Log.jsm", this);
+Cu.import("resource://gre/modules/Console.jsm", this);
```
gijsk: You won't need this after removing the debugging, I think.

recipe-client-addon/lib/EnvExpressions.jsm
Osmose: I talked about it with Mythmon and we decided that the fail-fast behavior of `Services.prefs` is desirable in most places. We use Preferences.jsm in a few places where we _need_ defaults, like anything running before the spot where we set default pref values.

gijsk: OK! As long as this was a deliberate decision, that WFM.

recipe-client-addon/lib/EnvExpressions.jsm
```diff
@@ -5,13 +5,18 @@
 "use strict";
 
 const {utils: Cu} = Components;
+Cu.import("resource://gre/modules/Services.jsm");
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
 Cu.import("resource://gre/modules/TelemetryArchive.jsm");
 Cu.import("resource://gre/modules/Task.jsm");
 Cu.import("resource://shield-recipe-client/lib/Sampling.jsm");
 
+const {generateUUID} = Cc["@mozilla.org/uuid-generator;1"].getService(Ci.nsIUUIDGenerator);
```
mythmon: I don't know of a style guide. If there isn't a Mozilla one, we should probably write one for our team, along with a style guide for other JS imports.

gijsk: The only one I'm aware of is https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style#JavaScript_practices , and the unwritten rule to sort imports alphabetically. That and anything eslint flags up. :-)

recipe-client-addon/lib/EnvExpressions.jsm
mythmon: > Should we respect a user-set empty value in case privacy add-ons want to get rid of it? That may be a question for outside of this PR.

I doubt there are many privacy sensitive people that will be ok with Normandy as a whole, but not this particular part.

recipe-client-addon/lib/EnvExpressions.jsm
mythmon: None of the current EnvExpression tests provide context.normandy, and so are only using the fallback. I'll add a test for the non-fallback case.

mythmon:
This is ready for review again.

recipe-client-addon/lib/EnvExpressions.jsm
```diff
@@ -33,33 +38,50 @@ XPCOMUtils.defineLazyGetter(this, "jexl", () => {
...
+      const ping = mostRecentPings[key];
+      telemetry[ping.type] = yield TelemetryArchive.promiseArchivedPingById(ping.id);
+    }
+    return telemetry;
+  }),
+
+  getUserId() {
+    let id = prefs.getCharPref("user_id");
+    if (id === "") {
+      // generateUUID adds leading and trailing "{" and "}". strip them off.
```
Osmose: 👍

recipe-client-addon/test/browser/browser_NormandyDriver.js
```diff
@@ -0,0 +1,22 @@
...
+const {utils: Cu} = Components;
+Cu.import("resource://shield-recipe-client/test/browser/Utils.jsm", this);
+Cu.import("resource://gre/modules/Console.jsm", this);
+
+Utils.Assert = Assert;
+
+add_task(Utils.withDriver(function* uuids(driver) {
+  // Test that it is a UUID
+  const uuid1 = driver.uuid();
+  is(Utils.UUID_REGEX.test(uuid1), true, "valid uuid format");
```
Osmose: Why `is` instead of `ok`?

Osmose:
Nice work, looks good! I tested locally with the following mock server testcase:

```python
class FilterUserId(TestCase):
    """
    Matches if the user's unique ID is "testvalue". The unique ID is
    stored in the <code>extensions.shield-recipe-client.user_id</code>
    preference.
    """
    def load_data(self):
        console_log(
            'FilterUserId executed',
            filter_expression='normandy.userId=="testvalue"'
        )
```

And it worked fine. :D

recipe-client-addon/test/browser/browser_EnvExpressions.js
```diff
@@ -65,4 +66,23 @@ add_task(function* () {
   is(val, true, "Bucket sample returns true for a known sample");
   val = yield EnvExpressions.eval('["test-4"]|bucketSample(0, 5, 10)');
   is(val, false, "Bucket sample returns false for a known sample");
+
+  // Test that userId is available
+  val = yield EnvExpressions.eval("normandy.userId");
+  ok(Utils.UUID_REGEX.test(val), "userId available");
+
+  // test that it pulls from the right preference
+  Preferences.set("extensions.shield-recipe-client.user_id", "fake id");
```
gijsk: Should be able to use:

```js
yield SpecialPowers.pushPrefEnv({set: [["extensions.shield-recipe-client.user_id", "fake id"]]});
```

which will auto-reset to the previous value at the end of the test.

gijsk: (ie then you can omit the additional `Preferences.set` call at the end, and we guarantee we restore a previous value, and we guarantee the pref will be reset even if the test times out / fails / does whatever.)

mythmon:
Rebased on master to fix merge conflict. This is ready to merge once CI passes again.

PR #443, opened 2017-01-23
Bug 1325409: Remove unnecessary cleanup.

Osmose:
Cleanup is already run when the notification is removed, so calling it
in close() causes it to be called twice. Also, when the notification
is removed, it sends a Heartbeat event, which fails if cleanup has
already happened.


mythmon:
I *think* this is right, but it is tricky. I'd be happier if there were tests that failed without this patch, or if there were tests that assured things got removed correctly instead of being leaked. I'll push it to try, and if that is happy, I'll merge this.

PR #446, opened 2017-01-24
Upgrade to Python 3.6

mythmon:
This went easier than I expected, and is working for me locally. Lets see if it breaks in CI.

### DIFF
--- a/docs/dev/install.rst
+++ b/docs/dev/install.rst
@@ -8,7 +8,7 @@ Prerequisites
 This guide assumes you have already installed and set up the following:
 
 1. Git_
-2. `Python 3.5`_, `pip 8`_ or higher, and virtualenv_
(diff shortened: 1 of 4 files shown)


mythmon:
I'm going to save this until after we tag the release for v29. v29 is already going to be big enough, I don't want to add this in as well.

mythmon:
@andymikulski, @chartjes, @rehandalal: I'd like to wait to merge this until both of you can confirm that you have Python 3.6 available. That will mean installing Python 3.6 and re-making your virtualenv. Once you have done that (and don't run into any other problems with this branch) can you approve this PR?

We're not going to merge this until after the next release anyways, so no rush.

mythmon:
@rehandalal and @Osmose both tested Python 3.6 on OSX. I'm confident that this will work for @andymikulski as well.

PR #449, opened 2017-01-24
CI: Pipe all output through cat, so progress bars aren't shown

mythmon:
This should avoid having CI dependency output larger than Circle will allow us to have. Lets check the logs to make sure that's true.


mythmon:
We have other verbosity issues now, but the issue this was supposed to fix was fixed. ☑️

PR #454, opened 2017-01-26
Fix #444: Alias global object to window in recipe-client sandboxes.

### LINKED ISSUE #444: setTimeout not defined in the recipe-client
When running a `show-heartbeat` recipe with the system add-on, I get the following error:

```
ReferenceError: setTimeout is not defined               RecipeRunner.jsm:176:13216
```

It appears to be happening inside the sandbox.

Osmose:
This fixes an issue where our babel-ified action code was trying to call `setTimeout` (because babel compiles to a browser-like environment) and failing because we were defining `window` in the sandbox as a standalone object, instead of aliasing it to the global object.

r?

### DIFF
--- a/recipe-client-addon/lib/RecipeRunner.jsm
+++ b/recipe-client-addon/lib/RecipeRunner.jsm
@@ -151,24 +151,32 @@ this.RecipeRunner = {
           let a = new Action(sandboxedDriver, sandboxedRecipe);
           a.execute()
             .then(actionFinished)
-            .catch(err => sandboxedDriver.log(err, 'error'));
+            .catch(actionFailed);
         };
 
-        window.registerAction = registerAction;
-        window.setTimeout = sandboxedDriver.setTimeout;
-        window.clearTimeout = sandboxedDriver.clearTimeout;
+        this.window = this;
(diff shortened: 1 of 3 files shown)


recipe-client-addon/lib/RecipeRunner.jsm
```diff
@@ -151,24 +151,32 @@ this.RecipeRunner = {
...
-        resolve(result);
+        resolve(clonedResult);
       };
       sandbox.actionFailed = err => {
+        // Error objects can't be cloned, so we just copy the message
+        // (which doesn't need to be cloned) to be somewhat useful.
+        const message = err.message;
         sandboxManager.removeHold("recipeExecution");
-        reject(err);
+        reject(new Error(message));
```
gijsk: Can we `Cu.reportError` the original error so we don't lose the stack?

gijsk: To be clear, I don't need to review this again with or without that added (if there's a reason not to). After spending what's coming up to half my life (uh oh) with bugzilla review+/- flags, I'm struggling with how to use this github approve vs. "request changes" vs. "just leave random comments" thing...

PR #458, opened 2017-01-30
recipe-client-addon: Fix missing import of LogManager in bootstrap::shutdown

mythmon:
Without this, we throw an error on shutdown, preventing the cleanups from happening.

~~This wasn't caught by the lints because the rule that understands `Cu.import` doesn't seem to understand scopes.~~ Correction: `Cu.import` is global, so it doesn't matter if we do it in function-local scopes.

In some cases, we didn't get to the `Cu.import` line in `startup`, so we never imported `LogManager` to begin with.

CC @kmaglione

### DIFF
--- a/recipe-client-addon/bootstrap.js
+++ b/recipe-client-addon/bootstrap.js
@@ -7,6 +7,14 @@ const {utils: Cu} = Components;
 Cu.import("resource://gre/modules/Services.jsm");
 Cu.import("resource://gre/modules/Preferences.jsm");
 Cu.import("resource://gre/modules/Log.jsm");
+Cu.import("resource://gre/modules/XPCOMUtils.jsm");
+
+XPCOMUtils.defineLazyModuleGetter(this, "LogManager",
+  "resource://shield-recipe-client/lib/LogManager.jsm");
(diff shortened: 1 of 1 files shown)


gijsk:
What error? We don't actually use the LogManager in that block.

mythmon:
LogManager is used on line 66, to unregister the preference listener.

mythmon:
The original error, by the way, was

> 1485721284611 addons.xpi WARN Exception running bootstrap method shutdown on shield-recipe-client@mozilla.org: ReferenceError: LogManager is not defined (resource://gre/modules/addons/XPIProvider.jsm -> file:///home/kris/code/mozilla-central/obj-artifact/dist/bin/browser/features/shield-recipe-client@mozilla.org/bootstrap.js:65:3)

gijsk:
I'm noticing this in test runs myself now... did this not merge to m-c yet?

mythmon:
No, it hasn't. I'll be filing the bug for that today.

mythmon:
No, it hasn't. The merge that includes it is in [bug 1336501](https://bugzilla.mozilla.org/show_bug.cgi?id=1336501).

PR #463, opened 2017-01-31
Fix issue with updating arguments

rehandalal:
r?

### DIFF
--- a/recipe-server/normandy/recipes/tests/test_models.py
+++ b/recipe-server/normandy/recipes/tests/test_models.py
@@ -283,6 +283,12 @@ class TestRecipe(object):
         recipe.update(locales=[])
         assert recipe.locales.count() == 0
 
+    def test_recipe_update_arguments(self):
+        recipe = RecipeFactory(arguments_json='')
+        recipe.update(arguments={'something': 'value'})
+        assert recipe.arguments == {'something': 'value'}
+        assert recipe.arguments_json == '{"something": "value"}'
+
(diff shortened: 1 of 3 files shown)


recipe-server/normandy/recipes/tests/test_models.py
```diff
@@ -283,6 +283,12 @@ def test_recipe_update_locales(self):
         recipe.update(locales=[])
         assert recipe.locales.count() == 0
 
+    def test_recipe_update_arguments(self):
+        recipe = RecipeFactory(arguments_json='')
+        recipe.update(arguments={'something': 'value'})
+        assert recipe.arguments == {'something': 'value'}
+        assert recipe.arguments_json == '{"something": "value"}'
```
chartjes: This second assertion seems redundant to me unless the code that turns the arguments into JSON is complicated...

recipe-server/normandy/recipes/tests/test_models.py
```diff
@@ -283,6 +283,12 @@ def test_recipe_update_locales(self):
         recipe.update(locales=[])
         assert recipe.locales.count() == 0
 
+    def test_recipe_update_arguments(self):
```
Osmose: Can you also add an API test that reproduces the error POSTing to the view? That'd be closer to the situation we're actually seeing the error in.

rehandalal:
@Osmose I added the test you requested. I also verified that it was a failing test without the changes.

PR #465, opened 2017-02-01
recipe-server: Add release info for Sentry

mythmon:
I'd like to put this out with v30, I think it will help track things on Sentry. Plus, we can use "Resolved in next version" on Sentry, which will be nice.


mythmon:
This doesn't affect any of our testing, and we don't have tests for other version related things, so I'm going to cancel @chartjes's review on this.

PR #470, opened 2017-02-02
Fix #469: Use extra_filter_expression as name for filter expressions.

### LINKED ISSUE #469: Cannot create new recipes using control interface (extra_filter_expressions: this field is required)
When creating or cloning a recipe in the control interface, the POST to create the recipe return a HTTP 400 and the error `{"extra_filter_expression": ["This field is required."]}`.

Osmose:
Updates the frontend to match the field name for the backend.

I ran into this error yesterday and wrote a fix, but this morning I tried replicating the issue on master and couldn't anymore, and assumed I had some stale JS builds or something. But then #469 happened, so I'm not sure anymore.

I think it's important that we reliably replicate this locally before testing and merging this fix.


mythmon:
I'm stealing this review to keep things moving today.

PR #472, opened 2017-02-02
add logging configs for parquet schemas to recipe-server docker container

relud:


recipe-server/parquet_schemas/normandy.admin.docker.app/request.summary.cfg
```diff
@@ -0,0 +1,37 @@
...
+    optional int64 Severity;
+    optional int64 Pid;
+    optional binary fields_lang (UTF8);
+    optional binary fields_method (UTF8);
+    optional binary fields_path (UTF8);
+    optional binary fields_rid (UTF8);
+    optional binary fields_uid (UTF8);
+    optional binary fields_user_agent_browser (UTF8);
+    optional binary fields_user_agent_os (UTF8);
+    optional binary fields_user_agent_version (UTF8);
```
mythmon: The logs I see just have an `agent` field, unparsed. Can we include that as well as the parsed versions? (Or just that).

Also, where is the code that extracts these values from the mozlog format? (I know we talked about it living elsewhere, but  I'd like to see it).

mythmon:
I'm not sure why, but this CircleCI ran the job for this PR under @relud's repo, and it failed. I can't really do anything about that. Relud, can you check if you have web hooks on your fork of normandy, and if so, remove them and re-push to this branch? (force pushing an "amended" commit will work well)

recipe-server/parquet_schemas/normandy.admin.docker.app/request.summary.cfg
relud: you won't always get the raw agent, because that's PII, but for agents that fail to parse I can keep the agent.

mozlog json is converted to heka/hindsight messages here:
https://github.com/mozilla-services/cloudops-deployment/blob/parquet_rework/libs/puppet/modules/logging/files/hindsight/decoders/mozlog.lua

and then fields are removed/transformed here:
https://github.com/mozilla-services/cloudops-deployment/blob/parquet_rework/libs/puppet/modules/logging/files/hindsight/encoders/common.lua

the default config options are the ones in place unless configured in these files.

relud:
I don't have any webhooks configured, but I can copy the (non-secret) parts of the config from upstream

PR #473, opened 2017-02-02
recipe-server: Update raven

mythmon:
Sentry keeps complaining we use an old version of this library.


mythmon:
Changelog: https://github.com/getsentry/raven-python/blob/master/CHANGES

mythmon:
Docs for integrating with Django: https://docs.sentry.io/clients/python/integrations/django/

It seems I need to tweak our logging config too, due to changes in the library.

mythmon:
Updated ^

PR #479, opened 2017-02-06
Functional tests via Selenium

Osmose:
We've had a few errors during deploys recently with things like the ability to save recipes properly, which reminded me that I never got around to writing marionette/selenium tests. I wanted to learn a bit more about writing selenium tests this weekend anyway, so I worked on this a bit. Whereas the last few times I tried this I was trying to get a version of Firefox that could run recipes, this PR focuses only on testing the functionality of the control interface.

This PR adds another project similar to the `mock-recipe-server` project that builds and runs a docker-compose setup running the recipe server, a selenium server with Firefox, and a container running pytest that executes functional tests.

Future followup work would be:

- Write more tests, this only has a very basic one.
- Refactor the test(s) to use [page objects](http://selenium-python.readthedocs.io/page-objects.html) and other common, useful patterns (former WebQA engineers are well-versed in this).
- Customize which Firefox selenium is using (currently it's hard-coded to 51.0.1 in their docker image) and to install the system add-on and do some in-chrome tests, which AFAIK is a thing Selenium can do in Firefox via Marionette.
- Do the above, or wait until Firefox 52 hits release and the selenium docker images, and run the tests through the nginx proxy over SSL. We need Firefox 52 for this as [a bug prevented versions before that from disabling certificate verification](https://bugzilla.mozilla.org/show_bug.cgi?id=1103196).

One last note, I named the project `functional_tests`, with an underscore, on purpose. The mixture of underscores and dashes throughout the project is a bit confusing, and I think the common pattern is to use underscores for directory paths and filenames and dashes for things like docker image names, slugs, etc. I can change the name if this is a major problem.


functional_tests/bin/wait-for-it.sh
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env bash
```
Osmose: The normandy container can't use `wait-for-it.sh` on itself, as far as I can tell. So I copied it to the `functionaltestrunner` container. I'd like to not have multiple copies lying around but don't have a good alternative.

recipe-server/bin/create_test_admin.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env python
```
Osmose: Django has no way to create a superuser in a script _with a password_. `createsuperuser` can only set a password interactively.

The common alternative is to pipe the Python code to `manage.py shell`, but I had lots of trouble getting this to work in a container run via `docker-compose`. I'd love to do this better.

Osmose:
Marking as WIP while I finagle CircleCI issues.

Osmose:
Yeah that build is "passing" but is actually failing all over the place.

functional_tests/bin/wait-for-it.sh
mythmon: Is `wait-for-it.sh` much different than our use of [`takis`](https://github.com/mozilla/normandy/blob/645bcc4f4cfe9bbb472704162c8b543624c90803/recipe-server/bin/ci/test#L11)? Would using that instead be easier?

Osmose:
Okay, this is ready for review. I rebased on #481 and confirmed that an intentional test failure is surfaced, so the tests appear to be working on CircleCI.

Osmose:
> Firefox has built-in support for webdriver and listening to requests on a specific port. Back in the first half of last year I wrote some tests for Loop (Hello) that tested the video chat sessions, so maybe you might want to look at the use of Marionette for this stuff. I still have the testing code around if you want to look at it.

As I understand it, the driver that the Selenium server this PR uses controls Firefox via Marionette. It's testing on 51.0.1, and [Marionette seems to be the only driver that works on Firefox versions past 47](http://www.theautomatedtester.co.uk/blog/2016/selenium-webdriver-and-firefox-47.html).

IIRC I talked with @AutomatedTester a while ago and he had some example of switching to a chrome context from Selenium, which will let us test the add-on with this stuff too.

chartjes:
Sounds like you're on top of the things I am concerned about. +1

PR #481, opened 2017-02-06
ci: Fail while piping commands to cat

### LINKED ISSUE #480: Master builds are failing but being marked as having passed
See https://circleci.com/gh/mozilla/normandy/1470 for an example. It says the build passed, but the output of the test command is actually failing in every stage.

mythmon:
When we piped all of the CI commands into `cat`, we masked the return codes of the sub-commands. This should fix that, by telling bash to fail a pipeline if any of the commands in it fail, instead of only the last one.

Fixes #480.

### DIFF
--- a/ci/bin/runner.sh
+++ b/ci/bin/runner.sh
@@ -20,6 +20,8 @@ count=0
 step=$1
 shift 1
 
+num_errors=0
+
 for project in "${PROJECTS[@]}"; do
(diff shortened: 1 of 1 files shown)


mythmon:
@Osmose @milescrabill actually, can you re-review this? After some more discussion, I changed it so that it doesn't exit immediately, but runs all the tests and then reports the number of failures.

mythmon:
CI failures are expected, as this unmasked several issues. We'll fix those separately.

PR #485, opened 2017-02-06
Use extra_filter_expression for mock server testcase filters.

Osmose:
Just a tiny fix. It's blocking QA.


Osmose:
Merging this because it's blocking desktop QA, which needs to start ASAP. Also the tests failing is a known thing right now, but the output of the compile step matches correct output.

PR #486, opened 2017-02-07
Testing Task Cluster CI

mythmon:
Nothing to see here, just testing if this works for us at all.


mythmon:
Looks like, at least for now, we can't realistically use Taskcluster to build and publish Docker images, or use docker-compose.

PR #498, opened 2017-02-09
Add comment field to ApprovalRequest

### LINKED ISSUE #488: Backend for comment on approval requests
Create a new `comment` field on approval requests, which will be filled in on approval or rejection, and update API accordingly.

rehandalal:
Fixes #488 

r?


recipe-server/normandy/recipes/tests/test_api.py
```diff
@@ -441,36 +441,54 @@ def test_it_works(self, api_client):
     def test_approve(self, api_client):
         r = RecipeFactory()
         a = ApprovalRequestFactory(revision=r.latest_revision)
-        res = api_client.post('/api/v1/approval_request/{}/approve/'.format(a.id))
+        res = api_client.post('/api/v1/approval_request/{}/approve/'.format(a.id),
+                              {'comment': 'r+'})
         assert res.status_code == 200
 
         r.refresh_from_db()
         assert r.is_approved
```
Osmose: Asserting that the comment got passed through correctly would be nice to do here and in the reject test too.

Osmose: I tried just adding this in myself since it was the only nit I had, but it fails and I'm not quite sure why:

```python
assert r.latest_revision.comment == 'r+'
```

mythmon: I suspect this is because the comment isn't on the revision object, but on the approval. Maybe something like this?

```python
assert r.latest_revision.approval.comment == 'r+'
```

Osmose: Whoops.

Don't mind me.


I'll just be over here.

PR #499, opened 2017-02-09
fix the parquet type of fields_user_agent_version

relud:
issue revealed by new testing scripts in cloudops-deployment

### DIFF
--- a/recipe-server/parquet_schemas/normandy.admin.docker.app/request.summary.cfg
+++ b/recipe-server/parquet_schemas/normandy.admin.docker.app/request.summary.cfg
@@ -26,7 +26,7 @@ message Log {
     optional binary fields_uid (UTF8);
     optional binary fields_user_agent_browser (UTF8);
     optional binary fields_user_agent_os (UTF8);
-    optional binary fields_user_agent_version (UTF8);
+    optional int64 fields_user_agent_version;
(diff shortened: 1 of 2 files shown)


recipe-server/parquet_schemas/normandy.admin.docker.app/request.summary.cfg
```diff
@@ -26,7 +26,7 @@ message Log {
     optional binary fields_uid (UTF8);
     optional binary fields_user_agent_browser (UTF8);
     optional binary fields_user_agent_os (UTF8);
-    optional binary fields_user_agent_version (UTF8);
+    optional int64 fields_user_agent_version;
```
mythmon: It seems odd that this would be an integer. I'd expect at least a floating point number. The original string made even more sense to me, but I see that [the extracter](https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/lpeg/modules/lpeg/common_log_format.lua#L342) doesn't agree with me.

relud: the reason it's an integer is because it's only the browser major version

mythmon: Fair.

PR #502, opened 2017-02-10
Change from `update` to `revise`

rehandalal:
r?


mythmon:
For context: @rehandalal and I discussed how `update` was confusing for a method that didn't directly change the object. For example, this test existed and was correct.

```python
foo = RecipeFactory(name='first')
foo.update(name='second')
assert foo.name = 'first'
```

We decided `revise` was a better verb for this operation, since it creates new revisions.

PR #510, opened 2017-02-13
Make classify-client never use the CDN (via self-describing API)

### LINKED ISSUE #307: Add configuration for domain in classify endpoint
required to serve site through akamai cdn, so that classify hits normandy directly via a url other than self-repair.mozilla.org in prod (such as normandy.services.mozilla.com)

mythmon:
@gijsk only c5eba27 touches the add-on. The rest of the changes might be interesting for context.

* Recipe server
  * Upgrade DRF to 3.5.3.
  * Adds customer router and api root index.
  * Adds settings `APP_SERVER_URL` for non-CDN version of the app
  * Supports `always_dynamic` setting on views to always use dynamic servers, never CDN.
* Self-repair
  * Use dynamic server for classify client url.
* Recipe client add-on
  * Use self-describing URL for recipe, action and classify_client endpoints
* Mock recipe server
  * Save API index

Fixes #307


mythmon:
CC @relud

recipe-client-addon/lib/NormandyApi.jsm
```diff
@@ -70,13 +91,22 @@ this.NormandyApi = {
       const verifier = Cc["@mozilla.org/security/contentsignatureverifier;1"]
         .createInstance(Ci.nsIContentSignatureVerifier);
 
-      if (!verifier.verifyContentSignature(serialized, builtSignature, certChain, "normandy.content-signature.mozilla.org")) {
+      const valid =  verifier.verifyContentSignature(
```
gijsk: Nit: single space before 'verifier'

recipe-client-addon/lib/NormandyApi.jsm
```diff
@@ -49,8 +49,29 @@ this.NormandyApi = {
...
+    if (endpoint[0] === "/") {
+      endpoint = endpoint.slice(1);
+    }
+    return `${base}/${endpoint}`;
+  },
+
   fetchRecipes: Task.async(function* (filters = {}) {
-    const recipeResponse = yield this.get("recipe/signed/", filters);
+    const signedRecipeUrls = (yield this.apiIndex())["recipes-signed"];
+    const recipeResponse = yield this.get(signedRecipeUrls, filters, {enabled: true});
```
gijsk: I'm a bit confused why 'signedRecipeUrls' implies it contains multiple URLs. `get` only allows you to fetch 1 at a time. 

Also note that your test server returns something for `recipe-signed` but not `recipes-signed`, from a quick look, so can you clarify the desired naming here, and what exactly this is supposed to do?

recipe-client-addon/lib/NormandyApi.jsm
```diff
@@ -85,16 +115,20 @@ this.NormandyApi = {
...
+    clientData.request_time = new Date(clientData.request_time);
+    return clientData;
+  }),
 
-  fetchAction(name) {
-    return this.get(`action/${name}/`).then(req => req.json());
-  },
+  fetchAction: Task.async(function* (name) {
+    let actionApiUrl = (yield this.apiIndex())["action-list"];
+    if (actionApiUrl.slice(-1) !== "/") {
```
gijsk: Here and elsewhere, you can use `str.endsWith()` instead of `.slice`, which is a bit easier to read.

gijsk: (and ditto for `startsWith` when passing 1, of course)

recipe-client-addon/test/browser/browser_NormandyApi.js
```diff
@@ -3,19 +3,13 @@
...
-  });
+  const url = "http://mochi.test:8888/browser/browser/extensions/shield-recipe-client/test";
+  yield SpecialPowers.pushPrefEnv({set: [["extensions.shield-recipe-client.api_url", url]]});
 
   // Test that NormandyApi can fetch from the test server.
-  const response = yield NormandyApi.get("browser/test_server.sjs");
+  const response = yield NormandyApi.get(`${url}/browser/test_server.sjs`);
   const data = yield response.json();
-  Assert.deepEqual(data, {test: "data"}, "NormandyApi returned incorrect server data.");
+  Assert.deepEqual(data, {test: "data"}, "NormandyApi returned correct server data.");
```
gijsk: Do we do more in-depth testing of the API stuff elsewhere? Feels like that would increase confidence in the code changes here. :-)

chartjes: Right now we have some contract API tests written in Python and I think they cover API responses quite well. :)

gijsk: Yes, but not what the add-on client does with them. Those python tests aren't "end to end" (integration) tests that pointing the add-on to a test server with a test recipe runs the test recipe successfully, which is what I was kinda hoping we did have somewhere.

chartjes: That's true -- we don't have functional tests (meaning browser + add-on) at this point

recipe-server/normandy/base/api/routers.py
```diff
@@ -0,0 +1,34 @@
...
+    def __init__(self, *args, view=APIRootView, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.view = view
+        self.registered_view_urls = []
+
+    def get_urls(self):
+        """
+        Generate the list of URL patterns, including a default root view for the API.
+        """
+        # Get the superclass of DefaultRouter, skipping over DefaultRouter
```
Osmose: `get_urls` is where the format suffix magic of `DefaultRouter` happens, so if you're skipping over it, you might as well just subclass `SimpleRouter` directly.

recipe-server/normandy/base/api/routers.py
```diff
@@ -0,0 +1,34 @@
...
+
+from normandy.base.api.views import APIRootView
+
+
+class RouterWithDetachedViews(DefaultRouter):
+    """Router that allows for detached routes."""
+
+    def __init__(self, *args, view=APIRootView, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.view = view
```
Osmose: You should call this `self.api_root_view` or something to be specific.

recipe-server/normandy/base/api/routers.py
```diff
@@ -0,0 +1,34 @@
+from django.conf.urls import url
+
+from rest_framework.routers import DefaultRouter
+
+from normandy.base.api.views import APIRootView
+
+
+class RouterWithDetachedViews(DefaultRouter):
+    """Router that allows for detached routes."""
```
Osmose: It's really unclear what "detached" means. I'd call this `MixedViewRouter` or something.

You might also consider splitting the API index stuff and the ability to register non-viewset views into two separate classes.

docs/ops/config.rst
```diff
@@ -179,6 +179,22 @@ in other Django projects.
 
     .. _mozlog: https://github.com/mozilla-services/Dockerflow/blob/master/docs/mozlog.md
 
+.. envvar:: DJANGO_CDN_URL
+
+  :default: ``None``
+
+  The URL of a CDN that is backed by Normandy, if one is in use. This is used to
+  enforce that immutable content is routed through the CDN. Must end with a
+  slash (``/``).
```
relud: Was this not documented before, or was I setting a configuration that wasn't in use before?

recipe-server/normandy/base/api/views.py
```diff
@@ -0,0 +1,43 @@
...
+        for api_url in self.api_urls:
+            if api_url.name is None:
+                continue
+            url_name = api_url.name
+            if namespace:
+                url_name = namespace + ':' + url_name
+
+            try:
+                view_class = getattr(api_url.callback, 'view_class', None)
+                if getattr(view_class, 'always_dynamic', False) and settings.APP_SERVER_URL:
```
Osmose: A flag on a class seems awkward because it only affects this listing, and not anywhere we reverse a URL to the class. I dunno if it's possible, but a better API IMO would be adding a parameter to `router.register` and `router.register_view` called `allow_cdn` or something to control this.

recipe-server/normandy/base/api/views.py
```diff
@@ -0,0 +1,43 @@
...
+    _ignore_model_permissions = True
+    exclude_from_schema = True
+    api_urls = []
+
+    def get(self, request, *args, **kwargs):
+        ret = {}
+
+        namespace = getattr(request.resolver_match, 'namespace', None)
+        for api_url in self.api_urls:
+            if api_url.name is None:
```
Osmose: Should we allow un-named URLs at all?

recipe-server/normandy/recipes/api/views.py
```diff
@@ -62,7 +62,7 @@ def retrieve(self, request, name, impl_hash):
 
 
 class RecipeFilters(django_filters.FilterSet):
-    enabled = CaseInsensitiveBooleanFilter(name='enabled', lookup_type='eq')
+    enabled = CaseInsensitiveBooleanFilter(name='enabled', lookup_expr='eq')
```
Osmose: Just curious, why update django-filters and change this stuff in this PR? I'm fine with it though.

recipe-server/normandy/base/tests/test_api.py
```diff
@@ -0,0 +1,82 @@
...
+import pytest
+
+from django.conf.urls import url
+from django.views.generic import View
+
+from normandy.base.api.views import APIRootView
+
+
+@pytest.mark.django_db
+class TestApiRoot(object):
```
Osmose: I think having unit tests of the router class would be better than having a test manually asserting the structure of our URLs.

recipe-client-addon/lib/NormandyApi.jsm
```diff
@@ -49,8 +49,29 @@ this.NormandyApi = {
...
   },
 
+  apiIndex() {
+    if (!indexPromise) {
+      indexPromise = this.get(prefs.getCharPref("api_url")).then(res => res.json);
+    }
+    return indexPromise;
+  },
+
+  makeApiUrl(endpoint) {
```
Osmose: To be clear, nothing to do here. I'm just sad at JS.

recipe-client-addon/lib/NormandyApi.jsm
```diff
@@ -85,16 +115,20 @@ this.NormandyApi = {
...
+    const response = yield this.get(classifyClientUrl);
+    const clientData = yield response.json();
+    clientData.request_time = new Date(clientData.request_time);
+    return clientData;
+  }),
 
-  fetchAction(name) {
-    return this.get(`action/${name}/`).then(req => req.json());
-  },
+  fetchAction: Task.async(function* (name) {
```
Osmose: Refactoring these to use `yield` is rad. :D

recipe-server/normandy/base/checks.py
```diff
@@ -2,18 +2,41 @@
...
 
-WARNING_MISCONFIGURED_CDN_URL = 'normandy.base.W001'
+ERROR_MISCONFIGURED_CDN_URL_SLASH = 'normandy.base.E001'
+ERROR_MISCONFIGURED_CDN_URL_HTTPS = 'normandy.base.E002'
+ERROR_MISCONFIGURED_APP_SERVER_URL_SLASH = 'normandy.base.E003'
+ERROR_MISCONFIGURED_APP_SERVER_URL_HTTPS = 'normandy.base.E004'
 
 
-def setting_cdn_url_has_trailing_slash(app_configs, **kwargs):
+def setting_cdn_url(app_configs, **kwargs):
```
Osmose: I like system checks, and I like how we use them.

recipe-client-addon/lib/NormandyApi.jsm
mythmon: The API endpoint `/recipe/` is the listing page for all objects of type `recipe`. `recipe/$id/` would  be an individual recipe. `recipe/signed` is a specialization of the listing view that also lists many recipes, and includes signatures.

The singular form of this would include an id or a slug specifying an individual object.

That being said, these names aren't right. It should be `signedRecipesUrl` and `recipesResponse`, most likely. I'll fix that.

recipe-client-addon/test/browser/browser_NormandyApi.js
mythmon: They're aren't any better tests that cover the add-on's interactions with the API right now. I'm not sure how we could do that without risking those tests getting out of date with the real API.

There are unit tests on the server that assert the API has the shape I expected when writing this code. I'm not sure how best to test this in the add-on. Do you have any suggestions?

docs/ops/config.rst
mythmon: It was not documented, but it did work before.

recipe-server/normandy/recipes/api/views.py
mythmon: The version of the router that I based my changes on came from the new version of DRF. That version of DRF required django-filters 1.0.

recipe-client-addon/lib/NormandyApi.jsm
mythmon: I wonder why I keep forgetting these exist.

recipe-client-addon/test/browser/browser_NormandyApi.js
gijsk: You can use an sjs file ( see https://developer.mozilla.org/en-US/docs/Mozilla/httpd.js/HTTP_server_for_unit_tests ) to generate dummy responses for use in a test. It's true that they could get out of date with the server, but presumably when you're updating the server you update the client, and if you update the client without updating the mock server, you'd break things.

It's true that you would only ever guarantee that the test works with the mock server you'd write and so you wouldn't catch subtle differences between the mock server and the 'real' one, but you'd be able to have a bit more confidence in:
1) client changes that should continue to work with the same server
2) server changes that update the mock server in the 'same way' - you get infra to guarantee things work with your mock server, and so now in review you check the mock server is a reasonable representation of the real server, which is easier to do as a human being than "does my client code interact with this orthogonal server code in the right way".

I hope that reasoning makes some sense, though I also think it's something about which reasonable people can disagree. :-)

recipe-client-addon/lib/NormandyApi.jsm
gijsk: OK. For my second paragraph, I mostly meant that this line: https://github.com/mythmon/normandy/blob/b3da82e51da205d0deea69eaa7024170eaba2e51/recipe-server/normandy/base/tests/test_api.py#L18 seems to use `recipe-signed` and this line: https://github.com/mythmon/normandy/blob/b3da82e51da205d0deea69eaa7024170eaba2e51/recipe-client-addon/lib/NormandyApi.jsm#L73 seems to read out `recipeS-signed` (note extra s) for the key in a JSON structure. Now, I haven't checked the rest of the code in detail, so maybe I'm missing something and/or the `recipeS-signed` key (with lowercase s...) already exists pre-patch (... elsewhere?). But it looked odd, is all. :-)

recipe-server/normandy/base/tests/test_api.py
mythmon: This test is more about actually having the expected api. Since part of our public api is (now) having certain names here, I think it is important to have this test. I'll also add more unit-like tests for the router.

recipe-client-addon/lib/NormandyApi.jsm
```diff
@@ -85,16 +115,20 @@ this.NormandyApi = {
...
+    let actionApiUrl = (yield this.apiIndex())["action-list"];
+    if (actionApiUrl.slice(-1) !== "/") {
+      actionApiUrl += "/";
+    }
+    const res = yield this.get(this.makeApiUrl(actionApiUrl + name));
```
mythmon: This sounds like a good idea for v2 of the API.

mythmon:
@Osmose, @gijsk: I've fixed the review issues, and added some tests for NormandyAPI.

recipe-client-addon/test/unit/test_NormandyApi.js
```diff
@@ -0,0 +1,90 @@
...
+    const server = makeMockApiServer();
+    const serverUrl = `http://localhost:${server.identity.primaryPort}`;
+    Services.prefs.setCharPref("extensions.shield-recipe-client.api_url", `${serverUrl}/api/v1`);
+    Services.prefs.setCharPref(
+      "security.content.signature.root_hash",
+      // Hash of the key that signs the normandy dev certificates
+      "4C:35:B1:C3:E3:12:D9:55:E7:78:ED:D0:A7:E7:8A:38:83:04:EF:01:BF:FA:03:29:B2:46:9F:3C:C5:EC:36:04"
+    );
+    yield task(serverUrl);
+    yield new Promise(resolve => server.stop(resolve));
```
gijsk: Should we clear the prefs in a cleanup function?

recipe-client-addon/test/unit/xpcshell.ini
```diff
@@ -1,4 +1,7 @@
 [DEFAULT]
 head = xpc_head.js
+support-files =
+  mock_api/**
 
 [test_Sampling.js]
+[test_NormandyApi.js]
```
gijsk: Nit: alphabetic ordering please.

recipe-client-addon/test/unit/test_NormandyApi.js
```diff
@@ -0,0 +1,90 @@
...
+    if (!index.exists()) {
+      response.setStatusLine("1.1", 404, "Not Found");
+      response.write(`Cannot find path ${index.path}`);
+      response.finish();
+      return;
+    }
+
+    try {
+      const contentsArray = yield OS.File.read(index.path);
+      const contents = new TextDecoder().decode(contentsArray);
```
gijsk: Nit: you can use the options object for `OS.File` to make this less torturous:
```js
const contents = yield OS.File.read(index.path, { encoding: "utf-8" });
```
returns a string directly.

recipe-client-addon/test/unit/test_NormandyApi.js
```diff
@@ -0,0 +1,90 @@
...
+
+add_task(withServer(function* test_fetchRecipes() {
+  const recipes = yield NormandyApi.fetchRecipes();
+  equal(recipes.length, 1);
+  equal(recipes[0].name, "system-addon-test");
+}));
+
+add_task(withServer(function* test_classifyClient() {
+  const classification = yield NormandyApi.classifyClient();
+  Assert.deepEqual(classification, {
```
gijsk: I wonder if both here and for some of the other tests, it might make sense to just read the file independently from the test and assert that the two JSON objects deepEqual each other. We should probably do that as a followup, though.

Osmose: I've always been on the fence about stuff like that, because the argument can be made that if you do that you're not testing that the code transforms the file into the data you expect, you're instead testing that you could write two versions of the transformation that have the same output.

recipe-server/normandy/recipes/api/views.py
```diff
@@ -148,6 +148,7 @@ class ClassifyClient(views.APIView):
     authentication_classes = []
     permission_classes = []
     serializer_class = ClientSerializer
+    always_dynamic = True
```
Osmose: Don't forget to remove this.

mythmon:
Last round of feedback, and I rebased on master to deal with merge conflicts.

Issue #286, opened 2016-10-04
[tracker] Security checklist


jvehent:
## Risk Management
- [x] The service must have performed a Rapid Risk Assessment and have a Risk Record bug (**SVC-RRA**).
## Infrastructure rules
- [x] Use [Intermediate TLS](https://wiki.mozilla.org/Security/Server_Side_TLS#Intermediate_compatibility) (**INFRA-TLS**)
- [x] Set HSTS to 31536000 (1 year) (**INFRA-HSTS**)
- [x] Set HPKP to 5184000 (60 days) (**INFRA-HPKP**)
  - `Public-Key-Pins: max-age=5184000; pin-sha256="WoiWRyIOVNa9ihaBciRSC7XHjliYS9VwUGOIud4PB18="; pin-sha256="r/mIkG3eEpVdm+u/ko/cwxzOMo1bk4TyHIlByibiA5E="; pin-sha256="YLh1dUR9y6Kja30RrAn7JKnbQG/uEtLMkBgFF2Fuihg="; pin-sha256="sRHdihwgkaib1P1gxX8HFszlD+7/gTfNvuAybgLPNis=";`
    - Start with max-age set to 5 minutes and increase gradually
    - Pin to the EV and DV roots of Digicert
  - [x] If the service is not hosted under `services.mozilla.com`, it must be manually added to [Firefox's preloaded pins](https://dxr.mozilla.org/mozilla-central/source/security/manager/tools/PreloadedHPKPins.json#184).
- If service has an admin panels, it must:
  - [x] only be available behind Mozilla VPN (which provides MFA) (**INFRA-ADMINVPN**)
  - [x] require LDAP authentication (**INFRA-ADMINLDAP**)
  - [x] enforce a two-man rule on sensitive changes (**INFRA-2MANRULE**)
## Coding rules

The following rules apply to all web applications: api and websites.
- [x] Sign all commits (**APP-COMMITSIG**)
  - Developers should [configure git to sign all commits](http://micropipes.com/blog//2016/08/31/signing-your-commits-on-github-with-a-gpg-key/) and upload their PGP fingerprint to https://login.mozilla.com
- [x] Detailed logging in mozlog format (**APP-MOZLOG**)
  - Business logic must be logged with app specific codes (errno)
  - Access control failures must be logged at WARN level
- [x] All SQL queries must be parameterized, not concatenated (**APP-SQL**)
- [x] User data must be escaped for the right context prior to reflecting it (**APP-ESCAPE**)
[...]

Issue #508, opened 2017-02-13
Unable to resolve path to module 'underscore.string/titleize'  import/no-unresolved


mythmon:
In [CI build 1529](https://circleci.com/gh/mozilla/normandy/1529?utm_campaign=build-failed&utm_medium=email&utm_source=notification) on master, this error came up:

```
/app/code/recipe-server/client/control/actions/FilterActions.js
  12:22  error  Unable to resolve path to module 'underscore.string/titleize'  import/no-unresolved
```

The build in the PR also had the error, but the error was not reported because of #481.

I don't think we have the underscore library installed that can support this sort of piece-meal usage. I also didn't find `titleize` in the version of underscore installed in recipe-server.

mythmon:
This builds fine in Webpack, which makes me think this is more of a linting issue than a packaging issue. But I don't actually understand how it could work either way.

mythmon:
Since Andy is doing onboarding this week, I'll take this.

PR #511, opened 2017-02-13
#286: Use subresource integrity for static assets.

Osmose:
Couldn't figure out a way to not steal most of the implementation for `render_bundle` from webpack_loader, sadly.

Bonus: The functional test suite helps test that SRI checking actually functions properly, as the JS will fail if it's not. Yay!

r?


recipe-server/normandy/base/tests/test_templatetags.py
```diff
@@ -0,0 +1,52 @@
...
+        get_files.return_value = [
+            {'path': js_file.strpath, 'name': 'file.js', 'url': 'file.js'},
+            {'path': css_file.strpath, 'name': 'file.css', 'url': 'file.css'},
+        ]
+        return get_files
+
+    @pytest.fixture
+    def mock_sri_hash(self, mocker):
+        sri_hash = mocker.patch('normandy.base.templatetags.normandy_webpack_loader.sri_hash')
+        sri_hash.return_value = 'mock-hash'
```
mythmon: I'd rather see this unmocked, or at least use a return value that depends on the inputs so that the tests can assert that the right hash goes with the right file. Maybe something as simple as

```python
def mock_sri_hash(contents):
    return 'hash({})'.format(contents)
```

recipe-server/normandy/base/templatetags/normandy_webpack_loader.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    Modified version of webpack_loader's render_bundle that adds hashes
+    to tags for subresource integrity.
+    """
+    tags = []
+    chunks = get_files(bundle_name, extension=extension, config=config)
+    for chunk in chunks:
+        with open(chunk['path'], 'rb') as f:
+            chunk_hash = sri_hash(f.read())
+        if chunk['name'].endswith(('.js', '.js.gz')):
```
mythmon: I don't think we actually use `.js.gz` files, do we? That sounds like something we shouldn't do. I'd be fine if the extra bit wasn't here.

mythmon:
CI breakage is because of the issue in #508. Rebasing on master will pick up the fix.

recipe-server/normandy/base/templatetags/normandy_webpack_loader.py
Osmose: Yeah, it's leftover from `webpack_loader`. We should drop it.

recipe-server/normandy/base/tests/test_templatetags.py
chartjes: Definitely agree with @mythmon here.

Osmose:
Updated.

Osmose:
@chartjes Can you take a look at this PR soon?

PR #514, opened 2017-02-14
recipe-client-addon: test and build xpi in TC

### LINKED ISSUE #509: recipe-client-addon: Build standalone XPI
There should be a way to build an XPI file for the add-on that works independent of mozilla-central.

mythmon:
Lets see if this works.

fixes #509


mythmon:
This is ready for review now. There was a lot of commits getting CI to work, but I've smooshed them down to just the one above. The taskcluster build takes forever to run, because it has to build Firefox. It won't block our normal PRs (though it will run on them).

To summarize what's going on here:

* Because `.taskcluster.yml` exists, [Taskcluster Github](https://github.com/taskcluster/taskcluster-github) is activated
* It reads `.taskcluster.yml` to decide what tasks to run
* We choose to run a single task, called a *decision task*
* That decision task creates two more tasks:
  * `recipe-client-addon:test`, which builds Firefox and tests the add-on with it
  * `recipe-client-addon:make-xpi` which builds an XPI and saves it as an artifact so we can get it
* The `make-xpi` task *depends on* the `test` task, so it only starts running after the `test` task completes succesfully
* Both tasks depend on the decision task, so they don't start running until the entire *graph* is created sucesfully

.taskcluster.yml
```diff
@@ -0,0 +1,52 @@
...
+      name: normandy:decision
+      description: Normandy decision task
+      owner: "{{ event.head.user.email }}"
+      source: "{{ event.head.repo.url }}"
+    provisionerId: "{{ taskcluster.docker.provisionerId }}"
+    workerType: "{{ taskcluster.docker.workerType }}"
+    scopes:
+      - queue:create-task:aws-provisioner-v1/github-worker
+    payload:
+      env:
```
owlishDeveloper: Everything looks good. I was just wondering if it would make sense to set this explicitly to true or false, just for clarity.

mythmon: Ah, oops. This is actually because there used to be some subkeys, like this:

```yaml
payload:
  env:
    GECKO_DEV_REPO_URL: https://github.com/mozilla/gecko-dev
```

I removed the environment variables without removing the env key. I'll remove it.

mythmon:
@Osmose It looks like nsp broke in CI (not security problems, but bad requests?). Can you look at this?

recipe-client-addon/bin/make-xpi.sh
```diff
@@ -0,0 +1,28 @@
...
+  rm -rf "$TMP_DIR"
+}
+trap cleanup EXIT
+
+while read -r LINE || [[ -n "${LINE}" ]]; do
+  mkdir -p "$(dirname "${DEST}/${LINE}")"
+  cp -r "${BASE_DIR}/${LINE}" "$(dirname "${DEST}/${LINE}")"
+done < "${BASE_DIR}/build-includes.txt"
+
+rm "${DEST}/install.rdf.in"
```
mythmon: This is a part of the Firefox build process. We don't need it for the XPI build process.

recipe-client-addon/bin/tc/test.sh
```diff
@@ -0,0 +1,23 @@
...
+set -eu
+
+# mach wants this
+export SHELL=$(which bash)
+
+apt-get install -y npm curl python2.7 xvfb
+
+# Creates gecko-dev-master
+echo 'Downloading gecko-dev...'
+curl -sL https://github.com/mozilla/gecko-dev/archive/master.tar.gz | tar xz
```
mythmon: I actually started by using taskcluster-vcs, and was steered away from it by garndt. Apparently it isn't well maintained, and is probably going away in the future.

There may be some improvements possible here, but I think there will be much better payoff in improving our total build times.

ci/taskcluster/bin/decision.py
```diff
@@ -0,0 +1,93 @@
...
+            dependencies = [idMaker(name) for name in task.get('dependencies', [])]
+            dependencies.append(decisionTaskId)
+
+            env = task.get('env', {})
+            for key, val in os.environ.items():
+                if key.startswith('GITHUB_'):
+                    env.setdefault(key, val)
+
+            task_id = idMaker(task['name'])
+            res = session.put(f'http://taskcluster/queue/v1/task/{task_id}', data=dumpJson({
```
mythmon: I started by using the python taskcluster library, and after working through some issues related to using the local taskcluster proxy in taskcluster jobs, garndt suggested that I didn't really need to use the library. (specifically, changing the base server is undocumented in the python library).

Overall, the library didn't give me much over just using requests directly, except a layer of obfuscation. I'm still using some useful things from it, like the JSON formatter and the ID generator. Overall I think this is a good balance.

ci/taskcluster/bin/decision.py
```diff
@@ -0,0 +1,93 @@
...
+                'metadata': {
+                    'name': task['name'],
+                    'description': task['description'],
+                    'owner': "mcooper@mozilla.com",
```
mythmon: This should actually pull the email of the user that triggered the build, like in `.taskcluster.yml`. I'll fix that.

ci/taskcluster/bin/decision.py
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env python
+import os
+
+import requests
+from taskcluster import fromNow
+from taskcluster.utils import stableSlugId, dumpJson
+
```
mythmon: I'm not sure why, but flake8 doesn't care about this space. I even ran flake8 manually with `--isolated` (which makes it ignore config files). The only problems it found are line length problems.

-

mythmon:
@Osmose I figured out why the built XPI wasn't working. In Firefox we use `jar.mn` to specify how files are made available as `resource://` files. For the XPI we need to use `chrome.manifest` for that. I've added that file to the add-on, and that made it work for me in Firefox Unbranded 52.

PR #515, opened 2017-02-14
#286: Add CSP via django-csp.

Osmose:
Because CSP prevents the use of eval, we cannot use the eval-based
sourcemap support in Webpack. The other alternative, inline-source-map,
seems to trigger errors in the Django development server due to URLs
that are too long, so the only remaining alternative is to use
external source maps, which seem to work fine.

Adding @jvehent for review to double-check the CSP directive I'm using (it's `'self' https://optional.cdn.net` for `default-src` across the entire site).

### DIFF
--- a/recipe-server/normandy/settings.py
+++ b/recipe-server/normandy/settings.py
@@ -117,18 +117,16 @@ class Core(Configuration):
         if self.CDN_URL:
             srcs.append(self.CDN_URL)
         return srcs
+
     CSP_OBJECT_SRC = "'none'"  # not using <object>, <embed>, and <applet> elements
     CSP_WORKER_SRC = "'none'"  # not using JS Worker, SharedWorker, or ServiceWorkers
     CSP_FRAME_SRC = "'none'"  # not using frames or iframes
+    CSP_BLOCK_ALL_MIXED_CONTENT = True
 
     # these don't fallback to default-src
     CSP_BASE_URI = "'none'"  # not using <base>
     CSP_FRAME_ANCESTORS = "'none'"  # this page isn't iframed elsewhere
     CSP_FORM_ACTION = "'self'"  # we only submit forms to ourselves
-
-    # when https is enabled
-    CSP_UPGRADE_INSECURE_REQUESTS = True
-    CSP_BLOCK_ALL_MIXED_CONTENT = True
 
     # Action names and the path they are located at.
     ACTIONS = {
--- a/recipe-server/normandy/settings.py
+++ b/recipe-server/normandy/settings.py
@@ -38,6 +38,7 @@ class Core(Configuration):
         'django.middleware.security.SecurityMiddleware',
         'django.middleware.common.CommonMiddleware',
         'django.middleware.clickjacking.XFrameOptionsMiddleware',
+        'csp.middleware.CSPMiddleware',
     ]
 
     ROOT_URLCONF = 'normandy.urls'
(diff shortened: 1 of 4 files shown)


jvehent:
302 to @g-k who's familiar with django-csp

chartjes:
I too request a 302 to @g-k

Osmose:
Still waiting on review from a Normandy dev.

recipe-server/normandy/settings.py
```diff
@@ -110,6 +111,25 @@ class Core(Configuration):
         }
     }
 
+    # Content Security Policy
+    def CSP_DEFAULT_SRC(self):
+        srcs = ["'self'"]
+        if self.CDN_URL:
+            srcs.append(self.CDN_URL)
+        return srcs
+    CSP_OBJECT_SRC = "'none'"  # not using <object>, <embed>, and <applet> elements
```
mythmon: nit: Shouldn't there be a blank line between the method and `CSP_OBJECT_SRC`? I mean, the linter didn't care so it is fine, but it looks weird.

recipe-server/normandy/settings.py
```diff
@@ -110,6 +111,25 @@ class Core(Configuration):
...
+    CSP_FRAME_SRC = "'none'"  # not using frames or iframes
+
+    # these don't fallback to default-src
+    CSP_BASE_URI = "'none'"  # not using <base>
+    CSP_FRAME_ANCESTORS = "'none'"  # this page isn't iframed elsewhere
+    CSP_FORM_ACTION = "'self'"  # we only submit forms to ourselves
+
+    # when https is enabled
+    CSP_UPGRADE_INSECURE_REQUESTS = True
+    CSP_BLOCK_ALL_MIXED_CONTENT = True
```
mythmon: I bet you need to be a little more careful with this: not all of our configs use https. In particular, you'll need to change `CSP_UPGRADE_INSECURE_REQUESTS` in `ProductionInsecure`, at least. I think the compose set up uses `ProductionInsecure`.

Osmose: Actually, reading the documentation a little closer, we probably just want `CSP_BLOCK_ALL_MIXED_CONTENT`. It's not recommended to have both of these enabled, as the former treats http pages as https for the purposes of CSP, and the latter blocks http resources on https pages. Having both enabled makes the mixed content check do nothing.

I think having mixed content on for all of our environments works fine; we shouldn't have environments that are both http and https, they should be one or the other.

Osmose: I tested this locally with compose and it seems to work fine.

mythmon:
I just noticed that CSP_REPORT_URI doesn't have docs. Can we add those?

recipe-server/normandy/settings.py
Osmose: @g-k Am I right about using only `CSP_BLOCK_ALL_MIXED_CONTENT` above?

g-k: Yeah, you're right. You can drop upgrade insecure (it was added to help legacy sites upgrade to https).

> Moreover, it means that upgrade-insecure-requests’s effect takes place before the block-all-mixed-content directive would have a chance to block the request. If the former is set, the latter is effectively a no-op.

https://www.w3.org/TR/upgrade-insecure-requests/#mix
https://www.w3.org/TR/upgrade-insecure-requests/#recommendations

Osmose:
> I just noticed that CSP_REPORT_URI doesn't have docs. Can we add those?

Done!

Issue #500, opened 2017-02-09
Keep Python and JS requirements up to date / off insecure versions


Osmose:
#286 recommends we use an automated tool of some kind to monitor our dependencies and either keep them up to date, or at least alert us when they're out of date or have a security vulnerability.

After looking at their suggested tools, I think our best bet is:

- Monitor Python dependencies with https://requires.io/. Have it send emails to a mailing list (normandy-notifications@mozilla.com?).
  - Monthly: Out-of-date notifications
  - Immediately: Insecure notifications
- Run [nsp](https://github.com/nodesecurity/nsp) regularly on our requirements.

Beyond alternatives to the above plan, the main open question is _how_ to run `nsp`. Running it in CI won't ensure we get notified when a vulnerability is found until we make a PR or commit, but we don't have any regular jobs that just run and monitor the code.

@mozilla/normandy Thoughts?

mythmon:
I'm ok with both of these tools.

In the absence of an automated, regular way to run nsp, I think running in CI is a good first step while we explore other options. We make commits to this repo pretty often: in the last year our longest gap was 6 days (around Christmas). In the past month out longest gap has been 2 days.

chartjes:
> On Feb 9, 2017, at 6:59 PM, Michael Kelly <notifications@github.com <mailto:notifications@github.com>> wrote:
> 
> #286 <https://github.com/mozilla/normandy/issues/286> recommends we use an automated tool of some kind to monitor our dependencies and either keep them up to date, or at least alert us when they're out of date or have a security vulnerability.
> 
> After looking at their suggested tools, I think our best bet is:
> 
> Monitor Python dependencies with https://requires.io/ <https://requires.io/>. Have it send emails to a mailing list (normandy-notifications@mozilla.com <mailto:normandy-notifications@mozilla.com>?).
> Monthly: Out-of-date notifications
> Immediately: Insecure notifications
> Run nsp <https://github.com/nodesecurity/nsp> regularly on our requirements.
> 
> 

Some QA projects are already using requires.io <http://requires.io/>, so I recommend that on the Python side.

—
Chris

Osmose:
Requires.io is configured for the team/project, and NSP landed in prod, although it's having some issues: #520 

I think we can call this closed.

PR #517, opened 2017-02-15
#500: Add NSP to lint command in CI.

Osmose:
Currently includes a commit that should cause nsp to fail CI, will remove once that's confirmed working.

### DIFF
--- a/lints/package.json
+++ b/lints/package.json
@@ -20,6 +20,7 @@
     "moment": "2.14.1",
     "node-sass": "3.9.3",
     "node-uuid": "1.4.7",
+    "nsp": "^2.6.2",
     "react": "15.3.1",
     "react-addons-css-transition-group": "15.3.1",
     "react-breadcrumbs": "1.3.16",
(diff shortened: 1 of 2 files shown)


lints/package.json
```diff
@@ -31,6 +32,7 @@
     "redux": "3.6.0",
     "redux-form": "6.1.0",
     "redux-thunk": "2.1.0",
+    "serialize-to-js": "^0.5.0",
```
mythmon: What's this? Is it a peer dependency that we have to install explicitly?

mythmon:
Looks like nsp succesfully caught the issue in 25de33d.

lints/package.json
Osmose: It's a library with a known vulnerability that I'm using to test the CircleCI job.

mythmon: Oh, derp. I didn't look at what the actually changes in the second commit were.

Osmose:
Righto, this is ready for review!

PR #518, opened 2017-02-15
Document approval request workflow and API

### LINKED ISSUE #495: Document approval request workflow and API

rehandalal:
Fixes #495 

f?


recipe-server/normandy/settings.py
```diff
@@ -277,6 +282,9 @@ class Development(Base):
 
     API_CACHE_TIME = values.IntegerValue(0)
 
+    SWAGGER_SETTINGS = Base.SWAGGER_SETTINGS
```
mythmon: Did you mean `Core.SWAGGER_SETTINGS`? Also, I don't think you need this anyways, since it is a subclass.

rehandalal: Oops yeah. I needed it to be able to make the next line happen.

rehandalal: Although maybe there's a cleaner way of making that happen?

rehandalal:
r?

PR #523, opened 2017-02-16
Add logging, standardize logging codes, and add tests for logging.

Osmose:
r?


mythmon:
The flake8 updates are also in #514. I guess we can race to see which gets merged first, and rebase the other.

PR #525, opened 2017-02-16
Downgrade eventlet to 0.17.4 to fix Python 3.6 compatibility

mythmon:
Eventlet > 0.17.4 has a problem with Python 3.6: https://github.com/eventlet/eventlet/issues/371

Eventlet is only used by Gunicorn when enabled by setting a environment variable.


mythmon:
I failed to update the hashes when changing the version. I forcepushed a change to update the hashes.

mythmon:
The build appears to have got stuck. I restarted it.

PR #550, opened 2017-02-23
Add User API endpoint and current user id to base template

### LINKED ISSUE #538: API endpoint for current user details
We need an endpoint that returns details for the current active user.

Maybe something like `/user/me/`

rehandalal:
Fixes #538 

r?

### DIFF
--- a/recipe-server/normandy/base/api/views.py
+++ b/recipe-server/normandy/base/api/views.py
@@ -1,0 +1,10 @@
+from django.contrib.auth.models import User
+
+from rest_framework.viewsets import ModelViewSet
+
+from normandy.base.api.serializers import UserSerializer
+
+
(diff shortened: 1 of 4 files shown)


recipe-server/normandy/base/api/views.py
```diff
@@ -0,0 +1,10 @@
+from django.contrib.auth.models import User
+
+from rest_framework.viewsets import ModelViewSet
+
+from normandy.base.api.serializers import UserSerializer
+
+
+class UserViewSet(ModelViewSet):
```
Osmose: This exposes user info publicly to anyone, including emails. We can't do that.

An `APIView` that returns data for the current user is more appropriate, that way the only user data you can access is your own.

PR #551, opened 2017-02-24
Recipe status workflow

### LINKED ISSUE #519: Recipe status workflow with peer approval
You should not be able to enable a recipe that is not approved.

Disabling a recipe should invalidate the old approvals by setting the `approved_revision` on the recipe to `None`.

rehandalal:
Fixes #519 

r?

### DIFF
--- a/recipe-server/normandy/recipes/api/views.py
+++ b/recipe-server/normandy/recipes/api/views.py
@@ -123,9 +123,8 @@ class RecipeViewSet(CachingViewsetMixin, UpdateOrCreateModelViewSet):
 
         try:
             recipe.save()
-        except Recipe.NotApproved:
-            return Response({'enabled': 'Cannot enable a recipe that has not been approved.'},
-                            status=status.HTTP_400_BAD_REQUEST)
+        except Recipe.NotApproved as e:
+            return Response({'enabled': str(e)}, status=status.HTTP_409_CONFLICT)
 
         return Response(status=status.HTTP_204_NO_CONTENT)
 
--- a/recipe-server/normandy/recipes/api/views.py
+++ b/recipe-server/normandy/recipes/api/views.py
@@ -120,7 +120,13 @@ class RecipeViewSet(CachingViewsetMixin, UpdateOrCreateModelViewSet):
     def enable(self, request, pk=None):
         recipe = self.get_object()
         recipe.enabled = True
-        recipe.save()
+
+        try:
+            recipe.save()
+        except Recipe.NotApproved:
+            return Response({'enabled': 'Cannot enable a recipe that has not been approved.'},
+                            status=status.HTTP_400_BAD_REQUEST)
+
         return Response(status=status.HTTP_204_NO_CONTENT)
 
     @reversion_transaction
--- a/recipe-server/normandy/recipes/tests/__init__.py
+++ b/recipe-server/normandy/recipes/tests/__init__.py
@@ -4,7 +4,7 @@ from django.utils import timezone
 
 import factory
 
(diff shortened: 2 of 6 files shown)


recipe-server/normandy/recipes/api/views.py
```diff
@@ -120,7 +120,13 @@ def history(self, request, pk=None):
     def enable(self, request, pk=None):
         recipe = self.get_object()
         recipe.enabled = True
-        recipe.save()
+
+        try:
+            recipe.save()
+        except Recipe.NotApproved:
+            return Response({'enabled': 'Cannot enable a recipe that has not been approved.'},
```
Osmose: Do we want to use the message from the exception object, instead of duplicating it here?

recipe-server/normandy/recipes/api/views.py
```diff
@@ -120,7 +120,13 @@ def history(self, request, pk=None):
...
+                            status=status.HTTP_400_BAD_REQUEST)
```
Osmose: IMO a 400 is more for malformed requests than for a request that isn't allowed by business logic. Maybe this is a 403, or a 409?

recipe-server/normandy/recipes/tests/__init__.py
```diff
@@ -108,6 +106,21 @@ def locales(self, create, extracted, **kwargs):
             for locale in extracted:
                 self.locales.add(locale)
 
+    @factory.post_generation
+    def enabled(self, create, extracted, **kwargs):
+        if extracted:
+            approval = ApprovalRequestFactory(revision=self.latest_revision)
```
Osmose: Would this overwrite an approval passed in to the factory? Do we need to check for that?

recipe-server/normandy/recipes/tests/__init__.py
```diff
@@ -108,6 +106,21 @@ def locales(self, create, extracted, **kwargs):
...
+            approval.approve(UserFactory(), 'r+')
```
Osmose: How would a test specify which user is approving a recipe? The implicit creation of a user feels unexpected to me.

Osmose: I think I'd be less surprised if this were in a helper function instead of being based off a factory attribute. The helper would be more explicit about creating the user and could allow you to override it more easily.

recipe-server/normandy/recipes/tests/__init__.py
```diff
@@ -108,6 +106,20 @@ def locales(self, create, extracted, **kwargs):
...
                 self.locales.add(locale)
 
+    # This should always be before `enabled`
+    @factory.post_generation
+    def approver(self, create, extracted, **kwargs):
+        if extracted:
+            approval = ApprovalRequestFactory(revision=self.latest_revision)
+            approval.approve(extracted, 'r+')
+
+    # This should always be after `approver` as we require approval to enable a recipe
```
Osmose: Can we guarantee that these two methods are executed in order? That seems a really finnicky thing to depend on.

I'm still not convinced that parameters to the factory are the best way to go about this. Do you have a reason to use them over something else?

rehandalal: Yes they are executed in order, as per:
http://factoryboy.readthedocs.io/en/latest/reference.html#post-generation-hooks

rehandalal: > I'm still not convinced that parameters to the factory are the best way to go about this. Do you have a reason to use them over something else?

I'm open to doing it another way, this just seemed the cleanest to me.

Issue #526, opened 2017-02-17
Use LDAP authentication


Osmose:
Splitting this off from #286, services ops security wants us to use LDAP for authentication, instead of storing passwords internally as we currently do.

There are no specific requirements as to whether we should build-in LDAP support, or have nginx handle it, and read in user data from an nginx-provided header.

Whichever method we use should integrate with our built-in permissions as they currently are.

mythmon:
CC @relud

mythmon:
With #569 merged and deployed, we are now using Auth0 and LDAP to authenticate users. We are still storing the most recent password the users used. I'll file a separate issue to remove those passwords later.

PR #552, opened 2017-02-24
Django 1.10

mythmon:
I've been meaning to to do this for a while, and some of the Auth0 stuff I'm looking at for #526 requires Django 1.10.

I stole d7e8fa8 from #510. That will probably merge before this. If so, I'll take that commit out of this branch. In the meantime, it is needed because DRF 3.3 isn't compatible with Django 1.10.

PS: The plan here is to update to Django 1.11 shortly after it comes out, and stay there for the foreseeable futures, since it is an LTS release.


mythmon:
I removed d7e8fa8, since #510 merged. This is ready to merge once CI passes.

Issue #489, opened 2017-02-07
Basic UI for approval requests


rehandalal:
This will involve creation of:
- [ ]  The approval request view
  - [ ] Show the user the revision that requires approval.
  - [x] This should allow the user to approve, reject or close an approval request
- [x] A button to create approval requests on recipes that have draft revisions

PR #554, opened 2017-02-24
Add basic UI for Review Requests

andymikulski:
Related to #489 

- Adds "action buttons" to recipe form based on current user/draft state (is user viewing old draft, is there a review open, is the user the one that opened the review, etc)
  - Adds `RecipeFormActions`, `FormButton` components
  - `RecipeFormActions` reports events to `RecipeForm`, which carries out the selected action
- Adds API calls for managing reviews
- Removes `?revisionId=...` query param business, replaces with proper route (`/recipe/:id/:revisionId`)
- Disables form fields when appropriate (form is actively submitting, viewing an in-review recipe)

Caveats:
- The app is often confused over what is the 'latest' draft. Loading a recipe directly (e.g. `/recipe/10`) is fine, but when loading a revision directly (`/recipe/10/e642b6...`) the app sets the recipe's "current revision" to whatever was loaded. *Note:* this will be resolved in another PR, in which revisions are saved to the redux store
- ~Tests are failing~
- ~Current user ID is not being looked up yet~

### DIFF
--- a/recipe-server/client/control/components/RecipeFormActions.js
+++ b/recipe-server/client/control/components/RecipeFormActions.js
@@ -1,0 +1,159 @@
+import React, {
+  PropTypes as pt,
+} from 'react';
+import { Link } from 'react-router';
+import cx from 'classnames';
+
+export const FormButton = ({
+    className,
+    label,
+    element = 'button',
+    type = 'button',
+    onClick,
+    display,
+    ...props,
+  }) => {
+  if (!display) {
+    return null;
+  }
+
+  // need titlecase for JSX
+  const Element = element;
+  return (
+    <Element
+      className={cx('button', className)}
+      onClick={onClick}
+      children={label}
+      type={type}
+      key={label + className}
+      {...props}
+    />
+  );
+};
+FormButton.propTypes = {
+  display: pt.bool.isRequired,
+  className: pt.string,
+  label: pt.string,
+  element: pt.any,
+  type: pt.string,
+  onClick: pt.func,
+};
+
+export default class RecipeFormActions extends React.Component {
+  static propTypes = {
+    onAction: pt.func.isRequired,
+    isUserViewingOutdated: pt.bool,
+    isPendingApproval: pt.bool,
+    isUserRequestor: pt.bool,
+    isAlreadySaved: pt.bool,
+    isFormPristine: pt.bool,
+    isCloning: pt.bool,
+    isFormDisabled: pt.bool,
+    recipeId: pt.number,
+  };
+
+  getActions({
+    isUserViewingOutdated,
+    isPendingApproval,
+    isUserRequestor,
+    isAlreadySaved,
+    isFormPristine,
+    isCloning,
+    isFormDisabled,
+    recipeId,
+  }) {
+    return [
+      // revert
+      <FormButton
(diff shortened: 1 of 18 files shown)


recipe-server/client/control/components/RecipeFormActions.js
```diff
@@ -0,0 +1,163 @@
...
+        label={'Revert to this Revision'}
+      />,
+      // cancel
+      <FormButton
+        display={!isUserViewingOutdated && isPendingApproval && isUserRequestor}
+        className={'action-cancel submit delete'}
+        onClick={this.createActionEmitter('cancel')}
+        label={'Cancel Review Request'}
+      />,
+      // back to latest
```
rehandalal: I think this should just be handled by the history sidebar.

recipe-server/client/control/components/RecipeFormActions.js
```diff
@@ -0,0 +1,163 @@
...
+      // reject
+      <FormButton
+        display={!isUserViewingOutdated && isPendingApproval && !isCloning && !isUserRequestor}
+        className={'action-reject submit delete'}
+        onClick={this.createActionEmitter('reject')}
+        label={'Reject'}
+      />,
+      // delete
+      <FormButton
+        display={!isUserViewingOutdated && !isPendingApproval && isAlreadySaved && !isCloning}
```
rehandalal: I think this should always be visible when editing

recipe-server/client/control/components/RecipeFormActions.js
```diff
@@ -0,0 +1,163 @@
...
+      <FormButton
+        display={!isUserViewingOutdated && !isPendingApproval && isAlreadySaved
+          && !isCloning && isFormPristine}
+        className={'action-request submit'}
+        onClick={this.createActionEmitter('request')}
+        label={'Request Approval'}
+      />,
+      // save
+      <FormButton
+        display={!isPendingApproval && isAlreadySaved && !isCloning && !isFormPristine
```
rehandalal: I think this or the new button should always be visible

rehandalal: If the form is pristine we should disable these buttons

recipe-server/client/control/routes.js
```diff
@@ -56,6 +56,13 @@ export default (
           component={DeleteRecipe}
           name="Delete"
         />
+        <Route
+          path=":revisionId/"
```
andymikulski: This will be changed when rebased onto #562

recipe-server/client/control/components/RecipeFormActions.js
```diff
@@ -0,0 +1,163 @@
...
+    isPendingApproval,
+    isUserRequestor,
+    isAlreadySaved,
+    isFormPristine,
+    isCloning,
+    isFormDisabled,
+    recipeId,
+  }) {
+    return [
+      // revert
```
andymikulski: Reverting is just taking the currently viewed revision and saving it as a new draft. Maybe `revert` isn't the correct language?

rehandalal: If we always show the save button including when viewing old revisions, people can just save an old revision as a new draft.

PR #555, opened 2017-02-25
Fix #544: Correctly send parameters passed to NormandyApi.

### LINKED ISSUE #544: Add-on executing disabled recipes
The system add-on is currently downloading and executing recipes that are disabled because it does not filter fetched recipes by their enabled status, nor does it add a filter for the enabled field to its request to the server. The add-on should probably use an enabled filter in the API request to only receive enabled recipes.

Osmose:
Updates NormandyApi.jsm to send request parameters correctly as query
string arguments for GET requests and as JSON-encoded data in the
request body for POST requests.

### DIFF
--- a/recipe-client-addon/lib/NormandyApi.jsm
+++ b/recipe-client-addon/lib/NormandyApi.jsm
@@ -24,7 +24,7 @@ this.NormandyApi = {
     method = method.toLowerCase();
 
     let body = undefined;
-    if (data !== undefined) {
+    if (data) {
       if (method === "get") {
         for (const key of Object.keys(data)) {
           url.searchParams.set(key, data[key]);
--- a/recipe-client-addon/lib/NormandyApi.jsm
+++ b/recipe-client-addon/lib/NormandyApi.jsm
@@ -1,7 +1,6 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-/* globals URLSearchParams */
 
 "use strict";
 
@@ -20,25 +19,23 @@ const prefs = Services.prefs.getBranch("extensions.shield-recipe-client.");
 let indexPromise;
 
 this.NormandyApi = {
-  apiCall(method, url, data = {}) {
+  apiCall(method, endpoint, data = {}) {
+    const url = new URL(endpoint);
     method = method.toLowerCase();
 
-    if (method === "get") {
-      if (data === {}) {
-        const paramObj = new URLSearchParams();
-        for (const key in data) {
-          paramObj.append(key, data[key]);
+    let body = undefined;
+    if (data !== undefined) {
+      if (method === "get") {
+        for (const key of Object.keys(data)) {
+          url.searchParams.set(key, data[key]);
         }
-        url += "?" + paramObj.toString();
+      } else if (method === "post") {
+        body = JSON.stringify(data);
       }
-      data = undefined;
     }
 
     const headers = {"Accept": "application/json"};
-    return fetch(url, {
-      body: JSON.stringify(data),
-      headers,
-    });
+    return fetch(url.href, {method, body, headers});
   },
 
   get(endpoint, data) {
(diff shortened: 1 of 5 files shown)


recipe-client-addon/lib/NormandyApi.jsm
```diff
@@ -10,35 +9,32 @@ Cu.import("resource://gre/modules/Services.jsm");
...
+    const url = new URL(`${api_url}/${endpoint}`);
     method = method.toLowerCase();
 
-    if (method === "get") {
-      if (data === {}) {
-        const paramObj = new URLSearchParams();
-        for (const key in data) {
-          paramObj.append(key, data[key]);
+    let body = undefined;
+    if (data !== undefined) {
```
gijsk: Is there any reason not to just use:

```js
if (data) {
  ...
}
```

?

recipe-client-addon/test/browser/test_server.sjs
```diff
@@ -1,8 +1,30 @@
...
   response.setHeader("Cache-Control", "no-cache", false);
   response.setHeader("Content-Type", "application/json", false);
-  response.write(JSON.stringify({test: "data"}))
+
+  // Read request body
+  const inputStream = new BinaryInputStream(request.bodyInputStream);
+  const bytes = [];
+  let available;
+  while ((available = inputStream.available()) > 0) {
+    Array.prototype.push.apply(bytes, inputStream.readByteArray(available));
```
gijsk: I think you can just use:

```js
bytes = bytes.concat(inputStream.readByteArray(available));
```
which is a little easier?

recipe-client-addon/test/browser/test_server.sjs
```diff
@@ -1,8 +1,30 @@
...
+  }
+  const body = String.fromCharCode.apply(null, bytes);
+
+  // Write response body
+  const data = {queryString: {}, body: body ? JSON.parse(body) : {}};
+  const params = request.queryString.split("&");
+  for (const param of params) {
+    const [key, value] = param.split("=");
+    data.queryString[key] = value;
```
gijsk: Should this unescape? I guess it's just a test, so maybe it doesn't matter...

recipe-client-addon/test/browser/browser_NormandyApi.js
```diff
@@ -15,7 +15,23 @@ add_task(function* () {
   });
 
   // Test that NormandyApi can fetch from the test server.
-  const response = yield NormandyApi.get("browser/test_server.sjs");
-  const data = yield response.json();
-  Assert.deepEqual(data, {test: "data"}, "NormandyApi returned incorrect server data.");
+  let response = yield NormandyApi.get("browser/test_server.sjs");
+  let data = yield response.json();
+  Assert.deepEqual(data, {queryString: {}, body: {}}, "NormandyApi returned incorrect server data.");
```
gijsk: Is this going to conflict with the other patch here?

recipe-client-addon/lib/NormandyApi.jsm
```diff
@@ -10,35 +9,32 @@ Cu.import("resource://gre/modules/Services.jsm");
...
+      if (method === "get") {
+        for (const key of Object.keys(data)) {
+          url.searchParams.set(key, data[key]);
```
gijsk: Sadfaces that there's no batch-set API here...

recipe-client-addon/test/browser/browser_NormandyApi.js
Osmose: I suspect it will. Can you use `sjs` files in xpcshell tests? All the docs I found on them were mochitest docs.

gijsk: Yes, I think the new test in #510 uses one.

mythmon: #510 doesn't use any `.sjs` files, but you should  be able to, according to [the httpd docs](https://developer.mozilla.org/en-US/docs/Mozilla/httpd.js/HTTP_server_for_unit_tests#SJS:_server-side_scripts). You might have to add `server.registerContentType("sjs", "sjs");`, but I wouldn't be surprised if that were the default.

recipe-client-addon/test/browser/test_server.sjs
Osmose: Yeah, I don't want to bother writing a robust fake server.

Osmose:
Addressed feedback and moved the tests to the xpcshell tests. I had to fix some issues with the existing xpcshell tests since they appeared to not be running properly, for some reason.

Issue #540, opened 2017-02-22
normandy.channel does not work for filtering


Osmose:
When I create a recipe locally with the filter `normandy.channel == 'aurora'`, it does not execute, despite me running Developer Edition.

Osmose:
Also tested on a local build with the latest master, `normandy.channel == 'default'` doesn't work there either.

Osmose:
Fixed by #557.

Issue #539, opened 2017-02-22
normandy.isDefaultBrowser does not work


Osmose:
When I create a recipe locally with the filter expression `normandy.isDefaultBrowser`, and DevEdition is set as my default browser, the recipe does not execute. When I invert the expression to `!normandy.isDefaultBrowser`, it gets executed.

Seems like it should be the opposite.

mythmon:
Could this be DevEdition's habit of trying to be a secondary browser? Is this specifically DevEdition? Does this happen on Nightly or Unbranded Beta?

Osmose:
I can reproduce this on a local build of mozilla-central with the latest code (both mozilla-central and our add-on).

PR #557, opened 2017-02-27
Add missing filter expression context to add-on.

Osmose:
So #539 and #540 only point out two of them, but the add-on is missing a significant amount of client environment info in the filter expression context. Luckily all of them are already implemented in the driver, so I just had to steal them from there.

When I started on this, I couldn't figure out where I wanted to add all this new data to the filter expression context, which led to the first three commits, which refactor things so that:

- `FilterExpressions.jsm` (formerly `EnvExpressions.jsm`) doesn't provide any default context, and only cares about evaluating expressions. Well, okay, there's still the filter functions, but we can handle that in another PR.
- The logic for figuring out info about the client environment is separated into the `ClientEnvironment.jsm` file.

The last commit actually adds the missing info to the filter expression context. I tried to add tests for most of them, but I'm stumped on how to mock some of them (like plugins). Others (like the appinfo stuff) don't seem to have a meaningful test. I'm pretty sure this PR still net-increases our test coverage, but suggestions on this are welcome.

One potential side effect of this PR is that it makes it so that clients don't hit the classify-client endpoint if there aren't any recipes enabled that use country or request_time filters. Which means a single database change can have a huge effect on our traffic. @relud: Are large swings in traffic to the classify endpoint something we specifically want to avoid? I can change this to consistently hit the server even if we're not using the data from it.

### DIFF
--- a/recipe-client-addon/lib/ClientEnvironment.jsm
+++ b/recipe-client-addon/lib/ClientEnvironment.jsm
@@ -22,14 +22,22 @@ this.EXPORTED_SYMBOLS = ["ClientEnvironment"];
 // Cached API request for client attributes that are determined by the Normandy
 // service.
 let _classifyRequest = null;
-const getClientClassification = Task.async(function *() {
-  if (!_classifyRequest) {
-    _classifyRequest = NormandyApi.classifyClient();
-  }
-  return yield _classifyRequest;
-});
 
 this.ClientEnvironment = {
+  /**
+   * Fetches information about the client that is calculated on the server,
+   * like geolocation and the current time.
+   *
+   * The server request is made lazily and is cached for the entire browser
+   * session.
+   */
+  getClientClassification: Task.async(function *() {
+    if (!_classifyRequest) {
+      _classifyRequest = NormandyApi.classifyClient();
+    }
+    return yield _classifyRequest;
+  }),
+
   /**
    * Test wrapper that mocks the server request for classifying the client.
    * @param  {Object}   data          Fake server data to use
@@ -69,11 +77,13 @@ this.ClientEnvironment = {
     });
 
     XPCOMUtils.defineLazyGetter(environment, "country", () => {
-      return getClientClassification().then(classification => classification.country);
+      return ClientEnvironment.getClientClassification()
+        .then(classification => classification.country);
     });
 
     XPCOMUtils.defineLazyGetter(environment, "request_time", () => {
-      return getClientClassification().then(classification => classification.request_time);
+      return ClientEnvironment.getClientClassification()
+        .then(classification => classification.request_time);
     });
 
     XPCOMUtils.defineLazyGetter(environment, "distribution", () => {
--- a/recipe-client-addon/lib/ClientEnvironment.jsm
+++ b/recipe-client-addon/lib/ClientEnvironment.jsm
@@ -5,6 +5,8 @@
 "use strict";
 
 const {classes: Cc, interfaces: Ci, utils: Cu} = Components;
+Cu.import("resource:///modules/ShellService.jsm");
+Cu.import("resource://gre/modules/AddonManager.jsm");
 Cu.import("resource://gre/modules/Preferences.jsm");
 Cu.import("resource://gre/modules/Services.jsm");
 Cu.import("resource://gre/modules/Task.jsm");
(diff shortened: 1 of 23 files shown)


recipe-client-addon/lib/ClientEnvironment.jsm
```diff
@@ -0,0 +1,173 @@
...
+
+"use strict";
+
+const {classes: Cc, interfaces: Ci, utils: Cu} = Components;
+Cu.import("resource:///modules/ShellService.jsm");
+Cu.import("resource://gre/modules/AddonManager.jsm");
+Cu.import("resource://gre/modules/Preferences.jsm");
+Cu.import("resource://gre/modules/Services.jsm");
+Cu.import("resource://gre/modules/Task.jsm");
+Cu.import("resource://gre/modules/TelemetryArchive.jsm");
```
gijsk: For this and a number of the other modules (basically, everything besides Task.jsm and XPCOMUtils), we should use lazy getters so as not to initialize these modules as soon as someone loads this jsm.

recipe-client-addon/lib/ClientEnvironment.jsm
```diff
@@ -0,0 +1,173 @@
...
+    });
+
+    XPCOMUtils.defineLazyGetter(environment, "isDefaultBrowser", () => {
+      return ShellService.isDefaultBrowser();
+    });
+
+    XPCOMUtils.defineLazyGetter(environment, "searchEngine", () => {
+      return Task.spawn(function* () {
+        const init = yield new Promise(resolve => Services.search.init(resolve));
+        if (Components.isSuccessCode(init)) {
```
gijsk: Nit: better name for the result from this promise?

gijsk: Also: I think you just want to use Task.async for the getter itself.

recipe-client-addon/lib/ClientEnvironment.jsm
```diff
@@ -0,0 +1,173 @@
...
+
+    XPCOMUtils.defineLazyGetter(environment, "request_time", () => {
+      return getClientClassification().then(classification => classification.request_time);
+    });
+
+    XPCOMUtils.defineLazyGetter(environment, "distribution", () => {
+      return Preferences.get("distribution.id", "default");
+    });
+
+    XPCOMUtils.defineLazyGetter(environment, "telemetry", () => {
```
gijsk: Incoming feature request: I think we'll want to be able to have a recipe that says "run me as soon as the *current* telemetry ping matches X". But that'll need a very different implementation.

recipe-client-addon/lib/ClientEnvironment.jsm
```diff
@@ -0,0 +1,173 @@
...
+   *
+   * Also note that, because filter expressions implicitly resolve promises, you
+   * can add getter functions that return promises for async data.
+   * @return {Object}
+   */
+  getEnvironment() {
+    const environment = {};
+
+    XPCOMUtils.defineLazyGetter(environment, "userId", () => {
+      let id = prefs.getCharPref("user_id");
```
gijsk: Why is everything else using `Preferences.jsm` and this isn't? Feels like this should use Preferences as well.

Osmose:
> I'm kind of assuming the tests are just being split up and are otherwise identical. Please let me know if there's anything there that I should look at in more detail.

Lots of them were just moved around, but new tests added include:

- [A test for ClientEnvironment.country and ClientEnvironment.request_time](https://github.com/mozilla/normandy/pull/557/files#diff-6569b73c5da8d51f3dc3f86ce37be38aR50)
- [ClientEnvironment.doNotTrack tests](https://github.com/mozilla/normandy/pull/557/files#diff-6569b73c5da8d51f3dc3f86ce37be38aR80)
- [A very small test of filter expression context working correctly](https://github.com/mozilla/normandy/pull/557/files#diff-e39c5a109140b65f857b784ce9b4c5bdR53)
- [Tests for properties we expect to be in the filter expression context when checking recipes, and a basic test of how executing a filter expression casts the return value and handles errors](https://github.com/mozilla/normandy/pull/557/files#diff-6169d799c871e62ee7841b8112cee917R89)

gijsk:
Thanks for the clarification! Those look good to me. :-)

recipe-client-addon/lib/ClientEnvironment.jsm
Osmose: Yeah, the idea of recipes being triggered by a condition instead of simply detecting it upon execution is one that's been floating around.

Osmose:
Feedback'd.

recipe-client-addon/lib/ClientEnvironment.jsm
```diff
@@ -5,21 +5,20 @@
...
-Cu.import("resource:///modules/ShellService.jsm");
-Cu.import("resource://gre/modules/AddonManager.jsm");
 Cu.import("resource://gre/modules/Preferences.jsm");
-Cu.import("resource://gre/modules/Services.jsm");
 Cu.import("resource://gre/modules/Task.jsm");
-Cu.import("resource://gre/modules/TelemetryArchive.jsm");
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
-Cu.import("resource://shield-recipe-client/lib/NormandyApi.jsm");
+
+XPCOMUtils.defineLazyModuleGetter(this, "Services", "resource:///modules/Services.jsm");
```
gijsk: Gah, sorry, I missed that you had Services.jsm as well - you can Cu.import that as it'll generally be loaded anyway.

gijsk: With that, r=me

Osmose:
After talking with @relud earlier today, we decided to make lazy-loading the API request to the classify endpoint a feature experiment behind a pref, as sudden shifts in server load due to filter changes has implications for scaling servers that we'd like to test out first.

@gijsk You might want to take a look at the new commits. They have some testing utils that may already exist elsewhere in Firefox that I'm not aware of.

recipe-client-addon/test/browser/Utils.jsm
```diff
@@ -27,4 +27,48 @@ this.Utils = {
...
+      object[propertyName] = patchValue;
+      yield testGenerator(patchValue);
+      object[propertyName] = original;
+    };
+  },
+
+  /**
+   * Creates a wrapper function that tracks whether it has been called.
+   */
+  createMock(wrapped) {
```
gijsk: I think other people have used `sinon` for this and similar tasks. I'm not super familiar with it, but at least for this particular util I believe `sinon.spy()` does what you want. See https://dxr.mozilla.org/mozilla-central/source/browser/components/preferences/in-content/tests/browser_advanced_siteData.js#9-10 (and further down) for an example.

Osmose: YAY WE HAVE SINON IN TREE

Like I wanted to include it myself but was hemming and hawwing about 50kb of JS for like one bit of its functionality.

PR #561, opened 2017-02-28
Implement Enable/Disable button

### LINKED ISSUE #527: UI for enabling/disabling recipes
We should get rid of the enabled radio field from the form and replace it with a Enable/Disable button on the recipe. 

We should not show a button if the recipe is not approved.

rehandalal:
Fixes #527.
Depends on #554.

5df8e02 onwards is relevant

r?


andymikulski:
Could `components/EnableRecipe` and `components/DisableRecipe` be consolidated into a single component? They're nearly identical.

recipe-server/client/control/components/DisableRecipe.js
```diff
@@ -0,0 +1,55 @@
...
+        <div className="fluid-8">
+          <form action="" className="crud-form">
+            <p>Are you sure you want to disable "{recipe.name}"?</p>
+            <div className="form-action-buttons">
+              <div className="fluid-2 float-right">
+                <input
+                  type="submit"
+                  value="Confirm"
+                  className="delete"
+                  onClick={this.disableRecipe}
```
andymikulski: nit: this could go in the `form`'s `onSubmit` - e.g. `<form onSubmit={this.disableRecipe} ...`. Currently trivial but would need to be changed if this form was expanded on in the future.

recipe-server/client/control/components/DisableRecipe.js
```diff
@@ -0,0 +1,55 @@
...
+                />
+              </div>
+            </div>
+          </form>
+        </div>
+      );
+    }
+
+    return null;
```
andymikulski: Why don't we do `if (!recipe) { return null; }` instead of wrapping the div in a big conditional?

andymikulski: (Same applies for `EnableRecipe`)

PR #562, opened 2017-02-28
Add `revision` property to Recipes store

andymikulski:
There's an issue in the front end code where the term `recipe` is used ambiguously. When a revision is selected, `recipe` objects totally inherit the revision, and from there it's tough to discern whether the user is looking at the most recent draft, if there's an approval request open, etc.

This patch:
  - Adds a `revisions` object to the Recipes redux store.
    - The `list` (`entries` on master) is the list of the most current recipe data, whereas the `revisions` is a collection of arrays keyed on recipe id.
  - Adds `recipe/:id/revision/:revisionId` route

ex:
 ```
const latestRecipeId = state.recipes.list[0].id;
const someRevisionId = 'abc123';

const selectedRevision = state.recipes.revisions[latestRecipeId][someRevisionId];
```

This is a concise example, but you get the idea.

Currently, there are **3** tests that are failing with this. I'm poking around to determine what the issue is but for now this is ready for feedback!

### DIFF
--- a/recipe-server/client/control/actions/ControlActions.js
+++ b/recipe-server/client/control/actions/ControlActions.js
@@ -189,10 +189,9 @@ function singleRecipeReceived(recipe) {
   };
 }
 
-function singleRevisionReceived({ revision, recipeId }) {
+function singleRevisionReceived({ revision }) {
   return {
     type: SINGLE_REVISION_RECEIVED,
-    recipeId,
     revision,
   };
 }
--- a/recipe-server/client/control/actions/ControlActions.js
+++ b/recipe-server/client/control/actions/ControlActions.js
@@ -3,6 +3,7 @@ export const REQUEST_COMPLETE = 'REQUEST_COMPLETE';
 
 export const RECIPES_RECEIVED = 'RECIPES_RECEIVED';
 export const SINGLE_RECIPE_RECEIVED = 'SINGLE_RECIPE_RECEIVED';
+export const SINGLE_REVISION_RECEIVED = 'SINGLE_REVISION_RECEIVED';
 
 export const SET_SELECTED_RECIPE = 'SET_SELECTED_RECIPE';
 export const SHOW_NOTIFICATION = 'SHOW_NOTIFICATION';
@@ -188,6 +189,14 @@ function singleRecipeReceived(recipe) {
   };
 }
 
+function singleRevisionReceived({ revision, recipeId }) {
+  return {
+    type: SINGLE_REVISION_RECEIVED,
+    recipeId,
+    revision,
+  };
+}
+
 function recipeAdded(recipe) {
   return {
     type: RECIPE_ADDED,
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: recipe-server/client/control/actions/ControlActions.js [hunk 1]
```diff
@@ -128,6 +129,14 @@ function singleRecipeReceived(recipe) {
   };
 }
 
+function singleRevisionReceived({ revision, recipeId }) {
+  return {
+    type: SINGLE_REVISION_RECEIVED,
+    recipeId,
```
rehandalal: do you need this? you should have this in `revision.recipe.id`

recipe-server/client/control/components/RecipeContainer.js
```diff
@@ -17,46 +22,88 @@ export default function composeRecipeContainer(Component) {
...
     }
 
+    getRecipeRevision(revisionId) {
+      const { dispatch } = this.props;
+
+      return dispatch(makeApiRequest('fetchSingleRevision', { revisionId }))
+        .then(revision => {
+          dispatch(singleRevisionReceived({
+            revision,
+            recipeId: revision.recipe.id,
```
rehandalal: this seems redundant

recipe-server/client/control/components/RecipeContainer.js
```diff
@@ -17,46 +22,88 @@ export default function composeRecipeContainer(Component) {
...
   }
 
   const mapStateToProps = (state, props) => {
     let recipeData = null;
-    if (state.recipes && state.recipes.list.length) {
+    let revisionData = null;
+    const selectedRecipeId = state.recipes && state.recipes.selectedRecipe;
+    let selectedRevisionId = props.routeParams && props.routeParams.revisionId;
+
+    const revReference = state.recipes.revisions[selectedRecipeId] || {};
```
rehandalal: nit: the name of this is a bit confusing, i think something like `recipeRevisions` would be better

recipe-server/client/control/components/RecipeContainer.js
```diff
@@ -17,46 +22,88 @@ export default function composeRecipeContainer(Component) {
...
+
     render() {
       return <Component {...this.props} {...this.state} />;
     }
   }
 
   const mapStateToProps = (state, props) => {
     let recipeData = null;
-    if (state.recipes && state.recipes.list.length) {
+    let revisionData = null;
```
rehandalal: nit: maybe just use `revision`

recipe-server/client/control/components/RecipeContainer.js
andymikulski: Do you mean attaching the `recipeId` prop? Or the actual `singleRevisionReceived` call?

rehandalal: sorry, i meant the `recipeId`. it's already in revision.

Issue #331, opened 2016-11-14
More customizable Heartbeat prompt


Osmose:
Original bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1294504

S&I wants to expand the capabilities of the `show-heartbeat` action:

- The icon should be customizable from a set of defaults.
- The engagement button option should allow for multiple buttons, each with separate URLs and thank-you messages. 
- There's interest in adding different ways for users to respond besides buttons and stars, such as [Likert scales](https://en.wikipedia.org/wiki/Likert_scale).

gregglind:
This is a fantastic list.

Do we need to talk about height of notification box / wrapping or other horrible topics?  RTL langs?

Claim: we DON'T need to embed images in the box, other than the icon?

gregglind:
Optional "more info" link as text?

rayborn:
I talked to Matt more about this yesterday.  I really like the list so far.

I'd add that I (and I've gotten a little feedback in our surveys from users about this) think that the transition between heartbeat and engagement URL can be a bit jarring.  I'd appreciate a UX person thinking about the best way to make that transition expected, but here's one idea:

1.  User is presented with opinion prompt (Likert, Heartbeat, etc)
2. User answers the question (and likely doesn't have expectations that they are about to be taken to a follow up)
3. [New] Heartbeat offers follow up text/actions instead of an engagement message.  I'm thinking that this will generally be: "Would you like to tell us more" "Yes" "No", but alternative options may be "Would you like to sign up to our mailing list" "Yes" "No" and other use cases targeted at the subgroups that we want to get into various funnels.

Essentially making Heartbeat interaction 2 stages (not N stages like some previous discussions) would be ideal to make sure the user knows what's happening.

Osmose:
> This is a fantastic list.
> 
> Do we need to talk about height of notification box / wrapping or other horrible topics? RTL langs?

I think the height is going to have to be a little taller for likert scales, and we'll probably want to test wrapping for very long strings. An RTL boolean option is probably good too. 

> Claim: we DON'T need to embed images in the box, other than the icon?

Agreed.

> Essentially making Heartbeat interaction 2 stages (not N stages like some previous discussions) would be ideal to make sure the user knows what's happening.

That's a rather heavy suggestion, but I agree that a UX pass that considers this question among others explicitly is worth doing.

Osmose:
So the finalized list of changes we want:

- The icon should be customizable from a set of defaults.
[...]

PR #568, opened 2017-03-01
Fix long show-heartbeat messages

Osmose:
The first commit fixes an issue with the mock-recipe server and how we were saving the URLs in the API index. I needed that to be able to run the long message test.

The second commit fixes the overflow issues by hiding overflow in the details element in the notification bar, which specifically does _not_ contain the close button. With this patch, the bar now looks like this:

screen shot 2017-03-01 at 3 03 00 pm

I'm not sure directly modifying the style attribute is the best way to fix this, but given that in #331 we're probably going to rewrite our implementation to not use XUL at all, I don't think it'll be an issue.

### DIFF
--- a/recipe-client-addon/lib/Heartbeat.jsm
+++ b/recipe-client-addon/lib/Heartbeat.jsm
@@ -167,6 +167,9 @@ this.Heartbeat = class {
       frag.appendChild(ratingContainer);
     }
 
+    const details = this.chromeWindow.document.getAnonymousElementByAttribute(this.notice, "anonid", "details");
+    details.style.overflow = "hidden";
+
     this.messageImage = this.chromeWindow.document.getAnonymousElementByAttribute(this.notice, "anonid", "messageImage");
     this.messageImage.classList.add("heartbeat", "pulse-onshow");
 


gijsk:
Is there an easy way to test this? I'm unsure what the 'details' element is off-hand, and just running a live example with the browser console is going to be the easiest way to check what this is doing. The comment calling this "odd" means I'd like to understand this better :-)

Osmose:
The easiest way to test it is using the [mock server](https://normandy-mock.dev.mozaws.net/); instructions are on the page. You're looking for the "HeartbeatLongMessages" test case.

gijsk:
Hrm. It's not amazing, but I guess it'll have to do.

This is all messy, because what normally happens is that the message wraps. This is what we do for normal notification bars. It seems that this is broken in the heartbeat notification. I'm not clear on why. I suspect the flex=0 setting on the messageText, but if I remove that my nightly OOMs (!). So uh, yeah. On the other hand, flex=1 looks like it would 'push' the stars away (if present), which the comments in the code say we're explicitly trying to avoid. So rock, meet hard place, I guess. The new flexbox model lets you deal with this in a saner way, I think...

Osmose:
> The new flexbox model lets you deal with this in a saner way, I think...

In the distance, you hear chants saying "Rewrite it in HTML, Rewrite it in HTML"

And they're getting closer.

PR #569, opened 2017-03-01
recipe-server: Optionally accept remote authentication

mythmon:
This enables Django's remote user authentication system, with some modifications. Specifically, we can customize it via settings instead of having certain values hard coded.

To test this: enable the OIDC settings, and then pass the `Remote-User: your_username` while making requests via the API. You should magically be authenticated. To test the web interface, you'll need to configure a proxy server to pass that header for you. Adding `proxy_set_header Remote-User mythmon;` to the Nginx config in the Compose setup worked nicely for me.

@relud Can you look over these docs and make sure this will work with your Auth0 things? I expect you'll need to modify the `DJANGO_OIDC_REMOTE_AUTH_HEADER` to match the environment.

Part of #526.

### DIFF
--- a/docs/ops/config.rst
+++ b/docs/ops/config.rst
@@ -222,12 +222,12 @@ in other Django projects.
 
    .. warning::
 
-      If this feature is enabled, the proxy server providing authentcation
-      *must* sanitize the headers passed along to Normandy. Specifcially, the
+      If this feature is enabled, the proxy server providing authentication
+      *must* sanitize the headers passed along to Normandy. Specifically, the
       header defined in :envvar:`DJANGO_OIDC_REMOTE_AUTH_HEADER` must not be
       passed on from the user.
 
-      Failing to do this will result from any client being able to authenticate
+      Failing to do this will result in any client being able to authenticate
       as any user, with no checks.
 
 .. envvar:: DJANGO_OIDC_REMOTE_AUTH_HEADER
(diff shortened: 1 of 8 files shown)


docs/ops/config.rst
```diff
@@ -192,19 +192,63 @@ in other Django projects.
...
+      as any user, with no checks.
+
+.. envvar:: DJANGO_OIDC_REMOTE_AUTH_HEADER
+
+   :default: ``HTTP_REMOTE_USER``
+
+   If :envvar:`DJANGO_USE_OIDC` is ``True``, this is the source of the user to
+   authenticate. This is subject to the Django header normalization, i.e.
+   the header will be capitalized, dashes replaced with underscores, and it will
+   be prefixed with ``HTTP_``.
```
Osmose: The wording "This is subject to" makes it sound like those changes will be automatically applied to the setting value, not that the setting must abide by those rules. 

Although actually I think those changes _should_ be automatically applied.

recipe-server/normandy/base/auth_backends.py
```diff
@@ -10,13 +10,19 @@
...
-    Model-backed authentication backend that logs the results of login
-    attempts.
+    Authentication backend mixin that logs the results of login attempts.
     """
-    def authenticate(self, username=None, **kwargs):
-        result = super().authenticate(username=username, **kwargs)
+    def authenticate(self, **kwargs):
+        result = super().authenticate(**kwargs)
+
+        username = None
```
Osmose: A method that subclasses override (like `get_username` or something) would be a cleaner solution than this if statement.

recipe-server/normandy/base/middleware.py
```diff
@@ -24,3 +26,31 @@ class RequestSummaryLogger(MiddlewareMixin, OriginalRequestSummaryLogger):
...
+    @property
+    def header(self):
+        """
+        Name of request header to grab username from.
+
+        This will be the key as used in the request.META dictionary. To
+        reference HTTP headers, this value should be all upper case and
+        prefixed with "HTTP_".
+        """
+        return settings.OIDC_REMOTE_AUTH_HEADER
```
Osmose: Could we have the setting be the real header name, and just apply the transformation here instead of making the person configuring the site do it?

relud: i would rather specify the header as `HTTP_OIDC_CLAIM_USER_EMAIL` than `Oidc-Claim-User-Email`, because that's how I specify it in openresty lua. It sounds like that would work either way though, and this would allow for more flexible input.

mythmon: I think I'd prefer to stick with the `HTTP_OIDC_CLAIM_USER_PROFILE_EMAIL` style. I'll clean up the docs, but I think enough tools use this format that it is a fine thing to do.

mythmon:
I made the logout button work with OIDC by reading a logout url from a setting. This would have been more complex to implemnet in the case that `OIDC_LOGOUT_IF_NO_HEADER=False`. Since we aren't going to use that setting right now, I removed it.

PR #570, opened 2017-03-02
mock-recipe-server: Add tests to validate generated data

mythmon:
I'm not sure if this will work in CI, and I'll work on that later, but this demonstrates a problem in our mock test cases (at least on my laptop).

While testing #556, I noticed that the signed and unsigned data for recipes was mixed up: the signed data was correct, but the unsigned data came from another test case. I'm not sure what's going on, but hopefully having this test in CI will help diagnose it and fix the issue.

### DIFF
--- a/mock-recipe-server/bin/ci/test
+++ b/mock-recipe-server/bin/ci/test
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+set -eu
+
+REPO_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && cd ../../.. && pwd)"
+
+compose () {
+    docker-compose \
+        -p mockrecipeserver \
+        -f "$REPO_DIR/compose/docker-compose.yml" \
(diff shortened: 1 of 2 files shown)


mythmon:
I haven't been able to verify this yet, but I bet this is because we cache the `/api/v1/recipe` endpoint, but we don't cache the `/api/v1/recipe/signed` endpoint.

Osmose:
> I haven't been able to verify this yet, but I bet this is because we cache the /api/v1/recipe endpoint, but we don't cache the /api/v1/recipe/signed endpoint.

Yeah, this is why #606 is happening; now that you added cache headers to the signed endpoint, the mock-server is caching the first test case it hits.

I put up #607 before remembering this PR existed. I don't have a test for comparing the signed and unsigned endpoints, and I don't do comparisons across all available testcases, so I'll see about integrating the work here into there.

Then the fix itself will be either disabling or bypassing the cache during mock-server runs.

PR #567, opened 2017-03-01
Approval/Rejection comments

andymikulski:
- Adds dialogs for comment entry when approving/rejecting reviews
- Adds `comment` data being sent to API
- Adds comment section on revisions where comments exist
- Fixes some minor bugs

PR #573, opened 2017-03-02
Revision history lookup fix

andymikulski:
~Depends on PR #567 .~

Changes the way in which recipe revisions are loaded. Currently, they are loaded as the user visits each revision (or, the whole history, if they go to the history page).

This patch switches it so that the initial recipe is loaded, followed by the revision history, regardless of which revision the user is viewing. This allows us to display data on the front end that requires knowledge about all revisions (e.g. latest approved, latest draft, revision approval/comments for each, etc).

r?

### DIFF
--- a/recipe-server/client/control/components/RecipeForm.js
+++ b/recipe-server/client/control/components/RecipeForm.js
@@ -17,8 +17,8 @@ import {
   recipeAdded,
   showNotification,
   setSelectedRecipe,
-  singleRecipeReceived,
   userInfoReceived,
+  revisionsReceived,
 } from 'control/actions/ControlActions';
 import {
   getLastApprovedRevision,
@@ -121,6 +121,7 @@ export class RecipeForm extends React.Component {
       route,
       routeParams = {},
       recipe = {},
+      revision = {},
       allRevisions = {},
       pristine,
       submitting,
@@ -130,7 +131,7 @@ export class RecipeForm extends React.Component {
         id: userId,
       },
     } = this.props;
-    const requestDetails = recipe && recipe.approval_request;
+    const requestDetails = revision && revision.approval_request;
     const currentUserID = userId;
     const isViewingLatestApproved = recipe && revision
       && revision.revision_id === recipe.approved_revision_id;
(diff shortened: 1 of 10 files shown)


andymikulski:
To manually test:
- Open a recipe page, then navigate to different revisions via the `History` page
- The draft status indicator should reflect the correct state for the revision
- Approved/Rejected revisions should have a comment section accompanying them

Osmose:
@chartjes Oh yeah, because you're out and because @stephendonner is still ramping up, I'm covering (and taking full responsibility for) QA reviews on PRs. Let me know if this is a major issue.

chartjes:
No objections to this at all while I am out.

andymikulski:
I've added reducer + selector tests, though I'm having a bit of trouble testing the RecipeContainer at the moment. I'll push those up when I get them working properly!

andymikulski:
I've added some new tests for the RecipeContainer business - the bulk of the testing was needed where the RecipeContainer picked a revision/recipe from the store. This function has been broken out into `getSelectedRevision` in `RecipesSelector.js`, and necessary tests have been written there.

andymikulski:
@rehandalal @Osmose r?

recipe-server/client/control/components/RecipeForm.js
```diff
@@ -121,25 +121,26 @@ export class RecipeForm extends React.Component {
...
       submitting,
       recipeId,
-      revision,
       user: {
         id: userId,
       },
     } = this.props;
-    const requestDetails = recipe && recipe.approval_request;
+
+    const requestDetails = revision.approval_request;
```
rehandalal: I get `Cannot read property 'approval_request' of null` on the Add New Recipe page.

recipe-server/client/control/components/RecipeForm.js
```diff
@@ -121,25 +121,26 @@ export class RecipeForm extends React.Component {
...
     const currentUserID = userId;
-    const isViewingLatestApproved = recipe && revision
+    const isViewingLatestApproved = recipe.approved_revision_id
```
rehandalal: i suspect i might have a similar issue here when the above issue is fixed.

PR #578, opened 2017-03-06
Fix #575: Add method for clearing index cache and mock-server tests.

### LINKED ISSUE #575: API index URLs are cached
The API index URLs are cached in `NormandyApi.jsm`. In normal operation this isn't a problem, because we only run once a session. This does, however, break the mock-server pattern of pasting code into the developer console to re-fetch recipes.

In the near future we're going to look at fetching recipes once every 24 hours if the user hasn't restarted, which will also make this an issue, especially if we expect a long-running session to not break if the server decides to change the API urls.

I'm not sure if we just want to not cache them at all, or add an easy way for the mock-server to clear the cache.

Osmose:
Long-term I'd like to do something like make `NormandyApi` a class, and use an instance of it during recipe runs (and the instance could then be discarded and GC'd to get rid of the index cache), but as a short-term fix, I added a method to clear the index cache that we can use in the mock-server testscases.

I also added the test method to `RecipeRunner` to avoid piling up even _more_ calls in the mock-server testcases. The unloading in particular helps get around an issue where running the mock-server testcases in the browser console, then updating the add-on without restarting, would leave the modules loaded in the browser console out-of-date.

I didn't add tests for these changes because the `testRun` method is only supposed to be run manually for testing, and `NormandyApi.clearIndexCache` is hard to unit-test, and will become irrelevant with the aforementioned refactoring that I'll work on soon; it's just a stopgap for now. But feel free to call me out as wrong on this. :D

### DIFF
--- a/mock-recipe-server/api_index.html
+++ b/mock-recipe-server/api_index.html
@@ -148,11 +148,9 @@
         let url = new URL(`/${testcase.dataset.name}/api/v1`, window.location);
         testcase.querySelector('.url').textContent = url.href;
         testcase.querySelector('.browser-code').value = (
-          `Cu.import('resource://shield-recipe-client/lib/Storage.jsm');\n` +
-          `Storage.clearAllStorage();\n` +
-          `Services.prefs.setCharPref("extensions.shield-recipe-client.api_url", "${url.href}");\n` +
+          `Cu.unload('resource://shield-recipe-client/lib/RecipeRunner.jsm');\n` +
           `Cu.import('resource://shield-recipe-client/lib/RecipeRunner.jsm');\n` +
-          `RecipeRunner.start();`
+          `RecipeRunner.testRun("${url.href}");\n`
         );
       }
 
--- a/mock-recipe-server/api_index.html
+++ b/mock-recipe-server/api_index.html
@@ -148,9 +148,8 @@
         let url = new URL(`/${testcase.dataset.name}/api/v1`, window.location);
         testcase.querySelector('.url').textContent = url.href;
         testcase.querySelector('.browser-code').value = (
-          `Cu.unload('resource://shield-recipe-client/lib/RecipeRunner.jsm');\n` +
-          `Cu.import('resource://shield-recipe-client/lib/RecipeRunner.jsm');\n` +
-          `RecipeRunner.testRun("${url.href}");\n`
+          `Cu.import('resource://shield-recipe-client/lib/RecipeRunner.jsm', {})` +
+          `.RecipeRunner.testRun("${url.href}");`
         );
       }
 
--- a/recipe-client-addon/bootstrap.js
+++ b/recipe-client-addon/bootstrap.js
@@ -41,6 +41,7 @@ const PREF_SELF_SUPPORT_ENABLED = "browser.selfsupport.enabled";
 const PREF_LOGGING_LEVEL = PREF_BRANCH + "logging.level";
 
 let shouldRun = true;
+let log = null;
 
 this.install = function() {
   // Self Repair only checks its pref on start, so if we disable it, wait until
@@ -62,6 +63,7 @@ this.startup = function() {
 
   // Setup logging and listen for changes to logging prefs
   LogManager.configure(Services.prefs.getIntPref(PREF_LOGGING_LEVEL));
+  log = LogManager.getLogger("bootstrap");
   Preferences.observe(PREF_LOGGING_LEVEL, LogManager.configure);
   CleanupManager.addCleanupHandler(
     () => Preferences.ignore(PREF_LOGGING_LEVEL, LogManager.configure));
@@ -90,9 +92,13 @@ this.shutdown = function(data, reason) {
     "lib/SandboxManager.jsm",
     "lib/Storage.jsm",
   ];
-  for (const module in modules) {
+  for (const module of modules) {
+    log.debug(`Unloading ${module}`);
     Cu.unload(`resource://shield-recipe-client/${module}`);
   }
+
+  // Don't forget the logger!
+  log = null;
 };
 
 this.uninstall = function() {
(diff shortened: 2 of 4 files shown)


recipe-client-addon/lib/RecipeRunner.jsm
```diff
@@ -187,4 +188,21 @@ this.RecipeRunner = {
...
+   */
+  testRun(baseApiUrl) {
+    const oldApiUrl = prefs.getCharPref("api_url");
+    prefs.setCharPref("api_url", baseApiUrl);
+
+    Storage.clearAllStorage();
+    NormandyApi.clearIndexCache();
+    this.start();
+
+    prefs.setCharPref("api_url", oldApiUrl);
```
mythmon: Can we wrap this up in a try/finally to make sure we reset the preferences and clear the cache, even if there is an error?

mock-recipe-server/api_index.html
```diff
@@ -148,11 +148,9 @@
         let url = new URL(`/${testcase.dataset.name}/api/v1`, window.location);
         testcase.querySelector('.url').textContent = url.href;
         testcase.querySelector('.browser-code').value = (
-          `Cu.import('resource://shield-recipe-client/lib/Storage.jsm');\n` +
-          `Storage.clearAllStorage();\n` +
-          `Services.prefs.setCharPref("extensions.shield-recipe-client.api_url", "${url.href}");\n` +
+          `Cu.unload('resource://shield-recipe-client/lib/RecipeRunner.jsm');\n` +
```
mythmon: Instead of unloading aggressively, could we `delete RecipeRunner;`, so the reference held by the console is dropped immediately? Then the normal add-on reload behavior should work correctly.

Alternatively, we could wrap this in an iife, so that all our references fall out of scope immediately.

Osmose: I don't think that addresses the issue. Somehow, despite us unloading the modules when the add-on is shut down for an update, the modules we import into the browser console are the versions from the old add-on until we restart. I can't quite figure out why.

Osmose:
Feedbacked!

mock-recipe-server/api_index.html
gijsk: I'm afraid I don't understand what problem this is trying to solve. Can you clarify what steps currently show an issue here?

At any point, there's supposed to be only one copy of any given jsm loaded in any one process, keyed by the URL you're using for the load. My understanding is that all of this code only runs in the parent process, and so does the code that updates add-ons. It's not clear to me how calling `Cu.unload` here makes a difference compared to `Cu.unload` being called with the exact same URL from the add-on's uninstall code. Are we sure that that uninstall code is running?

Osmose:
> I'm afraid I don't understand what problem this is trying to solve. Can you clarify what steps currently show an issue here?

Here's two add-ons:

- https://dl.dropboxusercontent.com/u/168762/shield-recipe-client-foo.xpi
- https://dl.dropboxusercontent.com/u/168762/shield-recipe-client-bar.xpi

Both have a function `RecipeRunner.experiment()` that logs an error message to the console; one logs `"FOO"`, the other logs `"BAR"`.

Install one of them, then restart the browser to get a fresh state. Run this in the browser console:

```js
Cu.import("resource://shield-recipe-client/lib/RecipeRunner.jsm");
RecipeRunner.example();
```

It should log the error message for the version you installed. Now, install the other XPI without restarting. In theory, `Cu.unload` gets called during shutdown, and so if you run the lines above again after the update, it should log a different message. In practice, for some reason, it logs the same message even though the code updated. If you restart Firefox, then re-run the code above, it logs the correct message.

Unloading from within the Browser Console fixes the issue, at least within the Browser Console.

> Are we sure that that uninstall code is running?

Yeah, I tested this by logging an error message from the `shutdown` method, and it logged successfully when upgrading the add-on.

gijsk:
Wat.

> RecipeRunner.example.toSource()
> "example() {
>    log.error("BAR");
>   }"
> RecipeRunner.example()
> undefined
> 1488971019944	extensions.shield-recipe-client.recipe-runner	ERROR	FOO

Excuse me while I go try to put my brain back together after that explosion.

gijsk:
OK, so, I think I know what's going on with the unload stuff, at least, even if I don't understand `toSource()` being out of sync. The bootstrap.js code is just broken. Evaluate this in a JS console of choice:

```js
for (const x in ['a', 'b']) { console.log(x) }
```

We should be using `for of` rather than `for in`. That will remove the requirement to mess with these instructions as then we'll actually unload the right modules, rather than unloading `resource://shield-recipe-client/0`, `resource://shield-recipe-client/1`, `resource://shield-recipe-client/2`, ... (which are all no-ops, to be clear).

recipe-client-addon/bootstrap.js
```diff
@@ -90,9 +92,13 @@ this.shutdown = function(data, reason) {
...
     "lib/Storage.jsm",
   ];
-  for (const module in modules) {
+  for (const module of modules) {
+    log.debug(`Unloading ${module}`);
     Cu.unload(`resource://shield-recipe-client/${module}`);
   }
+
+  // Don't forget the logger!
+  log = null;
```
mythmon: I don't think we need this. It will fall out of scope and be removed at the end of the function anyway.

Osmose: Even if it's global in the bootstrap module? Does that just get implicitly unloaded?

mythmon: Oh, I misread the diff, and didn't catch that `log` was a module global. Carry on.

mock-recipe-server/api_index.html
gijsk: OK, so with the bootstrap.js changes, I *believe* this is no longer necessary, or if there are still problems, @mythmon's earlier suggestion should work (ie delete the curent ref, then re-import).

Osmose:
@gijsk Updated the JS code to be a one-liner:

```js
Cu.import('resource://shield-recipe-client/lib/RecipeRunner.jsm', {}).RecipeRunner.testRun("https://localhost:8080/ErrorInvalidSignature/api/v1");
```

PR #579, opened 2017-03-06
recipe-server: Add migration to change username to emails

mythmon:
This is to prepare for OIDC (#569). The authentication parameter passed is an email address, so we need to migrate to that.


recipe-server/normandy/base/migrations/0001_copy_email_to_username.py
```diff
@@ -0,0 +1,39 @@
...
+    for user in User.objects.all():
+        if '@' in user.username:
+            user.username = user.username.split('@')[0]
+            user.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
```
Osmose: wat

mythmon: [This is how our other migrations did it](https://github.com/mythmon/normandy/blob/cf69f3050ff7ad52b06723c8065a06df4e78ac85/recipe-server/normandy/recipes/migrations/0027_auto_20160509_2225.py#L15)

PR #583, opened 2017-03-06
Update canonical site url to normandy.services.mozilla.com

mythmon:
This changes the system add-on and our docs to point to https://normandy.services.mozilla.com, which will be our new home in the future.

The main reason for this is that we can easily make this new URL be backed by the CDN in the ways we want, and the SSL certs are a lot cheaper than what we'd have to do to pull off the same trick for self-repair.mozilla.org.

@relud Did I get all that right? Did I tell any lies? Can you confirm that normandy.services.mozilla.com should be able to handle the load of our full userbase?

### DIFF
--- a/docs/qa/environments.rst
+++ b/docs/qa/environments.rst
@@ -9,7 +9,8 @@ Production
 Production is the live, user-facing instance of Normandy that is hit by every
 active user of Firefox.
 
-- **Server URL:** https://self-repair.mozilla.org/
+- **Server URL:** https://normandy.cdn.mozilla.net/
+- **Alternate URL (deprecated):** https://self-repair.mozilla.org/
 - **Control Interface:** https://normandy-admin.prod.mozaws.net/control/
 
 Preferences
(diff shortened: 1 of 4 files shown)


recipe-client-addon/bootstrap.js
```diff
@@ -29,7 +29,7 @@ const REASONS = {
 
 const PREF_BRANCH = "extensions.shield-recipe-client.";
 const DEFAULT_PREFS = {
-  api_url: "https://self-repair.mozilla.org/api/v1",
+  api_url: "https://normandy.services.mozilla.com/api/v1",
```
Osmose: Once we add an action for pref hotfixes, we should plan to send one out to update users (mostly Aurora/Nightly users) to the new URL who don't already have it as a default.

mythmon:
Fixed ^

docs/qa/environments.rst
```diff
@@ -19,7 +20,7 @@ production server:
 
 .. describe:: extensions.shield-recipe-client.api_url
 
-   ``https://self-repair.mozilla.org/api/v1``
+   ``https://normandy.services.mozilla.org/api/v1``
```
relud: missed one

PR #595, opened 2017-03-07
#316 - Filters on Recipe Form

andymikulski:
~Currently there's a lot of commits in this PR for some reason - I'm going to go through and cut down on unrelated changes that made their way into the code.~

Well that was a hoot. This is ready to be reviewed!

- Adds MultiPicker component
- Adds filter controls to recipe page
- Adds new data to API calls where appropriate (new, updating)

### DIFF
--- a/recipe-server/client/control/components/Fields.js
+++ b/recipe-server/client/control/components/Fields.js
@@ -50,20 +50,28 @@ buildControlField.propTypes = {
   children: pt.node,
 };
 
+export const CheckboxGroup = ({ name, onChange, options = [], value }) => {
+  /**
+   * Checkbox change event handler. Appends or removes the selected checkbox's
+   * value to the existing `value` prop, and reports the change up to redux-form.
+   *
+   * @param  {Event} onChange event object
+   */
+  const handleChange = ({ event: target }) => {
+    if (target.checked) {
+      // Use Set to remove any dupes from the new value array
+      const newValue = new Set(value.concat([target.value]));
+      // Report an array up to redux-form
+      onChange(Array.from(newValue));
+    } else {
+      // Remove this target's value from the array of values and return that array
+      onChange(value.filter(val => val !== target.value));
+    }
+  };
 
-export const CheckboxGroup = ({ name, options = [], value, onChange }) => {
-  const handleChange = option =>
-    event => {
-      const newValue = [...value];
-      if (event.target.checked) {
-        newValue.push(option.value);
-      } else {
-        newValue.splice(newValue.indexOf(option.value), 1);
-      }
-
-      return onChange(newValue);
-    };
-
+  /**
+   * Render the full list of inputs
+   */
   return (
     <div>
       {
@@ -71,13 +79,14 @@ export const CheckboxGroup = ({ name, options = [], value, onChange }) => {
           <label className="checkbox" key={index}>
             <input
               type="checkbox"
-              name={`${name}[${index}]`}
+              name={name}
               value={option.value}
               checked={value.includes(option.value)}
-              onChange={handleChange(option)}
+              onChange={handleChange}
             />
-            {option.label}
-          </label>)
+            <span>{option.label}</span>
+          </label>
+        )
       }
     </div>
   );
--- a/recipe-server/client/control/components/Fields.js
+++ b/recipe-server/client/control/components/Fields.js
@@ -24,14 +24,15 @@ export function buildControlField({
   className = '',
   InputComponent,
   children,
+  wrapper = 'label',
   ...args // eslint-disable-line comma-dangle
 }) {
-  const WrappingElement = label ? 'label' : 'div';
+  const WrappingElement = wrapper;
   return (
     <WrappingElement className={`${className} form-field`}>
       <span className="label">{label}</span>
       {error && <span className="error">{error}</span>}
-      <InputComponent {...input} {...args} input={input}>
+      <InputComponent {...input} {...args}>
         {children}
       </InputComponent>
     </WrappingElement>
@@ -43,40 +44,49 @@ buildControlField.propTypes = {
     error: pt.string,
   }).isRequired,
   label: pt.string,
+  wrapper: pt.oneOfType([pt.func, pt.string]),
   className: pt.string,
   InputComponent: pt.oneOfType([pt.func, pt.string]),
   children: pt.node,
 };
 
 
-export const CheckboxGroup = ({ name, options = [], input }) =>
-  <div>
-    {
-      options.map((option, index) =>
-        <div className="checkbox" key={index}>
-          <input
-            type="checkbox"
-            name={`${name}[${index}]`}
-            value={option.value}
-            checked={input.value.indexOf(option.value) !== -1}
-            onChange={event => {
-              const newValue = [...input.value];
-              if (event.target.checked) {
-                newValue.push(option.value);
-              } else {
-                newValue.splice(newValue.indexOf(option.value), 1);
-              }
+export const CheckboxGroup = ({ name, options = [], value, onChange }) => {
+  const handleChange = option =>
+    event => {
+      const newValue = [...value];
+      if (event.target.checked) {
+        newValue.push(option.value);
+      } else {
+        newValue.splice(newValue.indexOf(option.value), 1);
+      }
 
-              return input.onChange(newValue);
-            }}
-          />
-          {option.value}
-        </div>)
-    }
-  </div>;
+      return onChange(newValue);
+    };
+
+  return (
+    <div>
+      {
+        options.map((option, index) =>
+          <label className="checkbox" key={index}>
+            <input
+              type="checkbox"
+              name={`${name}[${index}]`}
+              value={option.value}
+              checked={value.includes(option.value)}
+              onChange={handleChange(option)}
+            />
+            {option.label}
+          </label>)
+      }
+    </div>
+  );
+};
 
 CheckboxGroup.propTypes = {
   name: pt.string.isRequired,
   input: pt.object.isRequired,
+  value: pt.oneOfType([pt.string, pt.array]),
+  onChange: pt.func.isRequired,
   options: pt.array,
 };
--- a/recipe-server/client/control/components/Fields.js
+++ b/recipe-server/client/control/components/Fields.js
@@ -55,7 +55,7 @@ export class CheckboxGroup extends React.Component {
   static propTypes = {
     name: pt.string.isRequired,
     onChange: pt.func.isRequired,
-    options: pt.array,
+    options: pt.array.isRequired,
     value: pt.arrayOf(pt.string),
   };
 
@@ -84,7 +84,6 @@ export class CheckboxGroup extends React.Component {
     if (target.checked) {
       newValue = uniq(value.concat([target.value]));
     } else {
-      // Remove this target's value from the array of values and return that array
       newValue = value.filter(val => val !== target.value);
     }
 
--- a/recipe-server/client/control/components/Fields.js
+++ b/recipe-server/client/control/components/Fields.js
@@ -31,7 +31,7 @@ export function buildControlField({
     <WrappingElement className={`${className} form-field`}>
       <span className="label">{label}</span>
       {error && <span className="error">{error}</span>}
-      <InputComponent {...input} {...args}>
+      <InputComponent {...input} {...args} input={input}>
         {children}
       </InputComponent>
     </WrappingElement>
--- a/recipe-server/client/control/components/Fields.js
+++ b/recipe-server/client/control/components/Fields.js
@@ -1,5 +1,6 @@
 import React, { PropTypes as pt } from 'react';
 import { Field } from 'redux-form';
+import { uniq } from 'underscore';
 
 /**
  * redux-form Field component that wraps the form input in a label and error
@@ -30,7 +31,7 @@ export function buildControlField({
   const WrappingElement = wrapper;
   return (
     <WrappingElement className={`${className} form-field`}>
-      <span className="label">{label}</span>
+      {label && <span className="label">{label}</span>}
       {error && <span className="error">{error}</span>}
       <InputComponent {...input} {...args}>
         {children}
@@ -63,6 +64,7 @@ export class CheckboxGroup extends React.Component {
     this.state = {};
 
     this.handleChange = ::this.handleChange;
+    this.renderOption = ::this.renderOption;
   }
 
   /**
@@ -77,40 +79,46 @@ export class CheckboxGroup extends React.Component {
       onChange,
     } = this.props;
 
-    if (target.checked) {
-      // Use Set to remove any dupes from the new value array
-      const newValue = new Set(value.concat([target.value]));
+    let newValue = [];
 
-      onChange(Array.from(newValue));
+    if (target.checked) {
+      newValue = uniq(value.concat([target.value]));
     } else {
       // Remove this target's value from the array of values and return that array
-      onChange(value.filter(val => val !== target.value));
+      newValue = value.filter(val => val !== target.value);
     }
+
+    onChange(newValue);
   }
 
-  render() {
+  renderOption(option, index) {
     const {
       name,
-      options = [],
       value = [],
     } = this.props;
 
     return (
-      <div>
-        {
-          options.map((option, index) =>
-            <label className="checkbox" key={index}>
-              <input
-                type="checkbox"
-                name={name}
-                value={option.value}
-                checked={value.includes(option.value)}
-                onChange={this.handleChange}
-              />
-              <span>{option.label}</span>
-            </label>
-          )
-        }
+      <label className="checkbox" key={index}>
+        <input
+          type="checkbox"
+          name={name}
+          value={option.value}
+          checked={value.includes(option.value)}
+          onChange={this.handleChange}
+        />
+        <span>{option.label}</span>
+      </label>
+    );
+  }
+
+  render() {
+    const {
+      options = [],
+    } = this.props;
+
+    return (
+      <div className="checkbox-list">
+        { options.map(this.renderOption) }
       </div>
     );
   }
--- a/recipe-server/client/control/components/Fields.js
+++ b/recipe-server/client/control/components/Fields.js
@@ -50,52 +50,68 @@ buildControlField.propTypes = {
   children: pt.node,
 };
 
-export const CheckboxGroup = ({ name, onChange, options = [], value }) => {
+export class CheckboxGroup extends React.Component {
+  static propTypes = {
+    name: pt.string.isRequired,
+    onChange: pt.func.isRequired,
+    options: pt.array,
+    value: pt.arrayOf(pt.string),
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = {};
+
+    this.handleChange = ::this.handleChange;
+  }
+
   /**
    * Checkbox change event handler. Appends or removes the selected checkbox's
    * value to the existing `value` prop, and reports the change up to redux-form.
    *
    * @param  {Event} onChange event object
    */
-  const handleChange = ({ event: target }) => {
+  handleChange({ target }) {
+    const {
+      value = [],
+      onChange,
+    } = this.props;
+
     if (target.checked) {
       // Use Set to remove any dupes from the new value array
       const newValue = new Set(value.concat([target.value]));
-      // Report an array up to redux-form
+
       onChange(Array.from(newValue));
     } else {
       // Remove this target's value from the array of values and return that array
       onChange(value.filter(val => val !== target.value));
     }
-  };
+  }
 
-  /**
-   * Render the full list of inputs
-   */
-  return (
-    <div>
-      {
-        options.map((option, index) =>
-          <label className="checkbox" key={index}>
-            <input
-              type="checkbox"
-              name={name}
-              value={option.value}
-              checked={value.includes(option.value)}
-              onChange={handleChange}
-            />
-            <span>{option.label}</span>
-          </label>
-        )
-      }
-    </div>
-  );
-};
+  render() {
+    const {
+      name,
+      options = [],
+      value = [],
+    } = this.props;
 
-CheckboxGroup.propTypes = {
-  name: pt.string.isRequired,
-  input: pt.object.isRequired,
-  value: pt.oneOfType([pt.string, pt.array]),
-  onChange: pt.func.isRequired,
-  options: pt.array,
-};
+    return (
+      <div>
+        {
+          options.map((option, index) =>
+            <label className="checkbox" key={index}>
+              <input
+                type="checkbox"
+                name={name}
+                value={option.value}
+                checked={value.includes(option.value)}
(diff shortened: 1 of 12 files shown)


recipe-server/client/control/components/MultiPicker.js
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  getDisplayedOptions() {
+    const {
+      options = [],
+      value,
+    } = this.props;
+
+    const selectedOptions = options.filter(option =>
+      value.indexOf(option.value) === -1);
```
Osmose: It's a bit hard for me to read some of these lines that are split like this (the functions, the variable unpacking, etc.). I think our eslint config uses 100 characters for the line length, so we should have plenty of room to keep this on one line.

Alternatively, we could split up the line like this:

```js
const selectedOptions = options.filter(
  option => value.indexOf(option.value) === -1
);
```

recipe-server/client/control/components/MultiPicker.js
```diff
@@ -0,0 +1,234 @@
...
+      value,
+      onChange,
+    } = this.props;
+
+    let selectedFilters = MultiPicker.getActiveSelectOptions(this.availableRef);
+
+    if (!selectedFilters || selectedFilters.length === 0) {
+      selectedFilters = this.getDisplayedOptions();
+      selectedFilters = selectedFilters.map(option => option.value);
+    }
```
Osmose: The behavior of moving _all_ the items if none are selected makes it really easy for a user to accidentally lose all of their changes. I think we should take it out.

andymikulski:
This should be ready for another pass ( @Osmose , @rehandalal )

recipe-server/client/control/components/RecipeForm.js
```diff
@@ -100,9 +118,11 @@ export class RecipeForm extends React.Component {
     }
 
     const isCloning = route && route.isCloning;
-
     const submitButtonCaption = recipeId && !isCloning ? 'Update Recipe' : 'Add New Recipe';
 
+    const filters = getFilterObject(this.props.filters.list);
```
andymikulski: Nope - this has been moved to `mapStateToProps`

recipe-server/client/control/components/MultiPicker.js
```diff
@@ -0,0 +1,234 @@
...
+    unit: pt.string.isRequired,
+    plural: pt.string.isRequired,
+    options: pt.array,
+    // from redux-form
+    value: pt.oneOfType([pt.array, pt.string]),
+    onChange: pt.func,
+  };
+
+  static getActiveSelectOptions({ options }) {
+    if (!options) {
```
andymikulski: Technically, it's not intended to ever be called without `options`, so you're correct that it's unnecessary. I've removed the explicit check and set a default `options` param, e.g. `..({ options = [] }) {...`

recipe-server/client/control/components/MultiPicker.js
```diff
@@ -0,0 +1,234 @@
...
+        (!val ? null : { ...this.convertValueToObj(val) }))
+      .filter(x => x);
+
+    const {
+      filterText,
+    } = this.state;
+
+    if (filterText) {
+      displayedOptions = displayedOptions.filter(option =>
+        JSON.stringify(option).indexOf(filterText) > -1);
```
andymikulski: I've switched to using `Object.values`, which allows us to search over _only_ the object's values.

recipe-server/client/control/components/MultiPicker.js
```diff
@@ -0,0 +1,234 @@
...
+      .concat(value || [])
+      .concat(selectedFilters);
+
+    // clear user input
+    this.availableRef.value = null;
+
+    onChange(newOptions.join(','));
+  }
+
+  handleRemoveOption(event) {
```
andymikulski: `handleApplyOption`/`handleRemoveOption` have been refactored down into one function, since they shared near-identical functionality.

As for splitting up the different sides of the MultiPicker, I'm unsure how that would work since we are reading values from `ref`s, and splitting out those components would lose those references. I will investigate.

recipe-server/client/control/components/MultiPicker.js
```diff
@@ -0,0 +1,234 @@
...
+
+    const selectedLabel = pickerValue
+      && (pickerValue.length === 0 || pickerValue.length > 1)
+      ? plural : unit;
+
+    return (
+      <div className="multipicker">
+
+        <div className="mp-frame mp-from">
+          {`Available ${availableLabel}`}
```
andymikulski: iirc this renders as one React 'text node', versus `Available {availableLabel}` which is two - one for 'available' and then one for the expression.

From the user's perspective, nothing is different. From the dev's, it's almost identical.

recipe-server/client/control/components/MultiPicker.js
```diff
@@ -0,0 +1,234 @@
...
+
+          <input
+            type="text"
+            placeholder={`Search ${plural}`}
+            onChange={this.onTextChange}
+          />
+
+          <select ref={ref => { this.availableRef = ref || this.availableRef; }} multiple>
```
andymikulski: I had run into previous trouble where `ref` was `null`, so I used this to counteract. I've removed it.

recipe-server/client/control/components/Fields.js
```diff
@@ -26,23 +26,57 @@ export function buildControlField({
...
   ...args // eslint-disable-line comma-dangle
 }) {
+  const WrappingElement = label ? 'label' : 'div';
   return (
-    <label className={`${className} form-field`}>
+    <WrappingElement className={`${className} form-field`}>
       <span className="label">{label}</span>
       {error && <span className="error">{error}</span>}
-      <InputComponent {...input} {...args}>
+      <InputComponent {...input} {...args} input={input}>
```
andymikulski: The `CheckboxGroup` code I was using had pulled values from `input` and for some reason I thought it had to be passed as a whole object. I've removed this extra prop.

recipe-server/client/control/components/MultiPicker.js
```diff
@@ -0,0 +1,234 @@
+import React, { PropTypes as pt } from 'react';
+
+export default class MultiPicker extends React.Component {
```
andymikulski: I'll investigate using a FieldArray instead.

recipe-server/client/control/components/MultiPicker.js
```diff
@@ -51,88 +44,65 @@ export default class MultiPicker extends React.Component {
...
+        onChange = () => {},
+      } = this.props;
+
+      // Get a list of all selected values.
+      let selection = [].concat(value);
+      // Pick up list of selected options from the dom ref
+      const refSelected = MultiPicker.getActiveSelectOptions(ref);
+
+      // Apply = take from the 'available' section and put in 'selected'
+      if (type === 'apply') {
```
Osmose: IMO A function that takes a parameter (typically, but not always, a string) that then uses it in an if statement to do one thing or another is a common code smell. It's harder to understand than alternative ways to achieve the same goal, like:

- Deferring to another method on the class and subclassing to customize the behavior.
- Accepting a callback so callers can customize the behavior in the middle of the function.

I'd suggest the class-based option, and that will probably mix in well with seeing if we can write a standalone component for the two lists in the MultiPicker.

recipe-server/client/control/components/MultiPicker.js
Osmose: Instead of refs, what if we use the parent component's state?

I'm thinking we have two components: MultiPicker, and MultiPickerList. MultiPickerList would handle the logic for searching/filtering the list, and would accept a callback for when the selected items changes. MultiPicker itself would use that callback to manage a list of what is selected to be added/removed when the appropriate button is pressed in its state.

Thoughts?

recipe-server/client/control/components/MultiPicker.js
Osmose: Is it that bad? I think doing it this way is less readable and two text nodes vs one won't really affect the performance of our app enough to matter.

recipe-server/client/control/components/MultiPicker.js
Osmose: Let me know if you need help or wanna pair on this. I've just recently used `FieldArray` so I'm currently a bit familiar with it.

recipe-server/client/control/components/MultiPicker.js
```diff
@@ -51,88 +44,65 @@ export default class MultiPicker extends React.Component {
...
-
     const {
       filterText,
     } = this.state;
 
-    if (filterText) {
-      displayedOptions = displayedOptions.filter(option =>
-        JSON.stringify(option).indexOf(filterText) > -1);
-    }
+    let selectedOptions = options.filter(option => value.indexOf(option.value) === -1);
```
andymikulski: `value` is actually a string from redux-form, so we're using `indexOf` to search over it. We could split the comma-delineated value and use `includes` on that, but I don't necessarily see the benefit of adding that extra step.

recipe-server/client/control/components/MultiPicker.js
```diff
@@ -51,88 +44,65 @@ export default class MultiPicker extends React.Component {
...
-
-    let selectedFilters = MultiPicker.getActiveSelectOptions(this.availableRef);
+    if (filterText) {
+      const searchText = filterText.toLowerCase();
 
-    if (!selectedFilters || selectedFilters.length === 0) {
-      selectedFilters = this.getDisplayedOptions();
-      selectedFilters = selectedFilters.map(option => option.value);
+      selectedOptions = selectedOptions.filter(option => {
+        const searchValues = JSON.stringify(Object.values(option)).toLowerCase();
```
andymikulski: The `option`s only have two fields - `label` (displayed name) and `value` (the abbreviation), e.g. `{ label: 'Albania', value: 'AL'}`, so I don't think it's necessary to manually pluck those values out.

As for the JSON business, you're correct that searching with quotes/commas would produce unintended search results. How do you feel about:
```
selectedOptions = selectedOptions.filter(option => {
  const searchValues = Object.values(option).join(' ').toLowerCase();
  return searchValues.indexOf(searchText) > -1;
});
```
It still looks over only the values, but becomes a concatenated string without any extraneous quotes/commas.

Osmose:
I keep forgetting to remove myself from the assignment after a review pass, my bad.

recipe-server/client/control/components/MultiPicker.js
Osmose: > The options only have two fields - label (displayed name) and value (the abbreviation), e.g. { label: 'Albania', value: 'AL'}, so I don't think it's necessary to manually pluck those values out.

I don't understand why plucking them out is a problem that we're avoiding. What's bad about it?

recipe-server/client/control/components/MultiPicker.js
andymikulski: Sorry, can you elaborate on `Deferring to another method on the class and subclassing to customize the behavior.` ? I'm unsure how this function could be split up, as both operations (applying/removing) need to perform mostly the same operations, in the same order. Any clarity would be great, thank you!

Osmose: Sure! Also, I was mistaken: upon further thought, making a standalone component for the option lists makes the subclassing idea unnecessary. But I'll describe it anyway.

The basic pattern is, wherever the behavior needs to be customized, instead of an `if` statement, you call another method on the class. Then, you create a subclass for each branch of the `if` that implements that method with the customized behavior. 

Directly applying that to this example (and simplifying the classes a bit):

```js
class Base {
  handleChangeOption() {
    const ref = this.getRef();
    if (!ref) {
      return;
    }

    const {
      value = [],
      onChange = () => {},
    } = this.props;

    // Get a list of all selected values.
    let selection = [].concat(value);
    // Pick up list of selected options from the dom ref
    const refSelected = MultiPicker.getActiveSelectOptions(ref);

    selection = this.modifySelection(selection, refSelected);

    // Clear user input.
    ref.value = null;

    // Report the updated selected filters to the parent.
    onChange(selection.join(','));
  };
}

class Apply extends Base {
  getRef() {
    return this.availableRef;
  }

  modifySelection(selection, refSelected) {
    return selection.concat(refSelected)
  }
}

class Remove extends Base {
  getRef() {
    return this.selectedRef;
  }

  modifySelection(selection, refSelected) {
    return selection.filter(val => refSelected.indexOf(val) === -1);
  }
}
```

That's not the whole story though, because you need to figure out where to actually use these two subclasses to get the customized behavior. With React, it's almost always by adding a new component. 

One option would be to make the add and remove buttons components that share a base class. You'd have to pass the `value` and `onChange` props down to them, but that's fine.  Because the buttons themselves don't contain the lists, meaning you can't get a ref to them, you'd have to pass that down as a ref as well. 

---

Stepping back, the core problem here is finding the right abstraction for what we want to do. `handleChangeOptions` does two things:

1. Fetches the list of options in one of the two selection boxes.
2. Modifies the form value of selected values by either removing or adding the selected options.

Actually, these are the two things that differ between adding and removing. The rest of the shared code is mostly plumbing. 

The first item, fetching the selected options, is actually the same for both lists, just with a different ref. That's one reason why [I suggested](https://github.com/mozilla/normandy/pull/595#discussion_r106298641) that we use a single component for both lists. The component can care all it wants about refs, and it just takes a callback that reports the selected items.

If the first item is handled, the second one is easier, because the only shared logic left is calling `onChange`, and one line of duplicated code is easy to read and understand.

---

That's a bit wordy. I've never really pair programmed before, but if my spiel above is still unclear, we can work together on writing the generic filter component so I can show you what I mean. :D

recipe-server/client/control/components/MultiPicker.js
andymikulski: I wanted to avoid hardcoding the values for it to search over. Does this suffice:

```
filteredOptions = filteredOptions.filter(({value, label}) =>{
  const searchValues = [value, label].join('   ').toLowerCase();
  return searchValues.indexOf(filterText) > -1;
});
```

or did you have other thoughts on how that should be searched over entirely?

andymikulski:
@Osmose can you review b5f59bb1636ffa379ee079d4aab000793430f967 and let me know if this is more in line with what you're thinking

recipe-server/client/control/components/MultiPicker.js
andymikulski: Yeah, I'm going to need your help with implementing FieldArrays, @Osmose . I've been attempting to implement it for the [CheckboxGroup component](https://gist.github.com/andymikulski/5a4497f97354f8340761453a8848cd1a), however I'm running into several issues:

- `fields` never has any getters, despite the [documentation](http://redux-form.com/6.5.0/docs/api/FieldArray.md/) indicating there should be `getAll`, `get`, etc
- Pushing values to `fields` is ignored. It's weird - if I do `fields.push('whatever')`, it simply sets `fields` to `channels[0]` (where `channels` is the field name). Pushing again adds `channels[1], channels[2], ...` and so on - always ignoring the actual values.

I have a feeling I'm just getting the whole thing wrong; there seems to be increased complexity with this approach, and things like the getters missing is a big red flag for me. Any direction you could give would be appreciated.

Osmose:
I gave feedback to @andymikulski in person.

andymikulski:
@rehandalal I'm going to hold off on having you review this PR as it needs considerable changes

recipe-server/client/control/components/Fields.js
```diff
@@ -50,34 +50,43 @@ buildControlField.propTypes = {
...
+  /**
+   * Checkbox change event handler. Appends or removes the selected checkbox's
+   * value to the existing `value` prop, and reports the change up to redux-form.
+   *
+   * @param  {Event} onChange event object
+   */
+  const handleChange = ({ event: target }) => {
+    if (target.checked) {
+      // Use Set to remove any dupes from the new value array
+      const newValue = new Set(value.concat([target.value]));
```
Osmose: Oh, a set is a rad idea! Maybe we should even just be storing the set itself.

recipe-server/client/control/components/Fields.js
```diff
@@ -50,34 +50,43 @@ buildControlField.propTypes = {
...
+export const CheckboxGroup = ({ name, onChange, options = [], value }) => {
+  /**
+   * Checkbox change event handler. Appends or removes the selected checkbox's
+   * value to the existing `value` prop, and reports the change up to redux-form.
+   *
+   * @param  {Event} onChange event object
+   */
+  const handleChange = ({ event: target }) => {
+    if (target.checked) {
+      // Use Set to remove any dupes from the new value array
```
Osmose: These comments are getting pretty verbose; every line in this function has a comment. We can probably cut down on them. I think all of the lines in this function are well-written enough to not need comments, and there's no weird side effects

recipe-server/client/control/components/Fields.js
```diff
@@ -50,34 +50,43 @@ buildControlField.propTypes = {
...
 };
 
+export const CheckboxGroup = ({ name, onChange, options = [], value }) => {
+  /**
+   * Checkbox change event handler. Appends or removes the selected checkbox's
+   * value to the existing `value` prop, and reports the change up to redux-form.
+   *
+   * @param  {Event} onChange event object
+   */
+  const handleChange = ({ event: target }) => {
```
Osmose: If we make a child component (`CheckboxGroupField` maybe?) and add this as a method on it, we can avoid unnecessary re-renders due to the function comparison issue.

recipe-server/client/control/components/MultiPicker.js
```diff
@@ -51,24 +42,16 @@ class PickerControl extends React.Component {
...
 
-      // For each option, create a string value to search over, and then determine
-      // if that value contains our search text.
-      newOptions = newOptions.filter(({ value, label }) => {
-        const searchValues = [value, label].join('   ').toLowerCase();
-        return searchValues.indexOf(filterValue) > -1;
-      });
+      // Determine if selected option properties contain the filterValue.
+      return options.filter(({ value, label }) =>
+        [value, label].find(str => str.toLowerCase().indexOf(filterValue) > -1));
```
Osmose: [`Array.prototype.some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) would be more appropriate here than just `find`, but they both return truthy responses, so it's just a nit.

recipe-server/client/control/components/Fields.js
andymikulski: Can you elaborate on this? `handleChange` interacts with the `value` prop passed into CheckboxGroup, and I'm unsure of how nesting subcomponents would avoid rerenders as a result.

For instance:

```
options.map((option, index) =>
  <CheckboxGroupField
    key={index}
    option={option}
    onChange={onChange}
    checked={value.includes(option.value)}
  />
)
```
Passing redux-form's `onChange` directly (as shown here) means that we would need to pass the `value` into the CheckboxGroupField as well, which would cause a rerender. Using `onChange={handleChange}`, where `handleChange` is defined in CheckboxGroup, would loop back around to the issue that `handleChange` is defined every render, which re-renders the subcomponents.

Furthermore, since these 'components' are not actual React Components, I'm fairly positive that the whole list is rendered regardless of what prop changes are passed in. Perhaps we should actually just use React.Component and define `handleChange` as a component function, which would avoid all of this.

Osmose: > Using onChange={handleChange}, where handleChange is defined in CheckboxGroup, would loop back around to the issue that handleChange is defined every render, which re-renders the subcomponents.

Which means `handleChange` should also be a method on a component class instead of a within a function. I think the general rule is that anonymous functions should be avoided in components, wherever they are.

> Furthermore, since these 'components' are not actual React Components, I'm fairly positive that the whole list is rendered regardless of what prop changes are passed in.

`render` (or functional components) always gets called, so there's no avoiding that. What's avoided is messing with the DOM, which is generally the much-slower part. Functional components like this aren't different from class-based components in that regard.

> Perhaps we should actually just use React.Component and define handleChange as a component function, which would avoid all of this. This code should be sufficient.

Using the value on the input is another fine way of doing this. I think I prefer it to my own suggestion, actually. 👍

andymikulski:
@rehandalal This is now ready for review!

andymikulski:
Just waiting on TC/CI tests to run before merging!

andymikulski:
Looks like there was a selenium issue on CircleCI - @mythmon should I try to trigger another test, or is there a known issue with CI at the moment?

mythmon:
This looks like the selenium failure we were hitting earlier. We've disabled this test on master. I looked at the test output, and that's the only issue. Instead of dancing with CI to get this to pass, I think we can merge this as is, if @Osmose's change request has been satisfied.

@Osmose can you update your review?

recipe-server/client/control/components/MultiPicker.js
```diff
@@ -209,9 +208,9 @@ class PickerControl extends React.Component {
 export default class MultiPicker extends React.Component {
   static propTypes = {
     unit: pt.string.isRequired,
+    onChange: pt.func.isRequired,
```
Osmose: Why'd you replace the redux form prop types?

andymikulski:
> some of the comments from my previous review were not addressed (one about verbose comments and another about storing the set itself).

I apologize, I thought all feedback had been taken care of a while ago. I'll go back through and address any that I've missed.

> Also, the new commit added after @rehandalal 's review

This is my fault - I had corrected Rehan's review fix and then amended the `Code review` commit to try and keep the commit count low. I didn't realize it'd make it look like everything was added after the review, but it makes sense - sorry for the confusion.

recipe-server/client/control/components/Fields.js
```diff
@@ -50,52 +50,68 @@ buildControlField.propTypes = {
...
+  static propTypes = {
+    name: pt.string.isRequired,
+    onChange: pt.func.isRequired,
+    options: pt.array,
+    value: pt.arrayOf(pt.string),
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = {};
```
andymikulski: We follow this convention throughout the rest of the front end - _if_ we use a constructor on a `React.Component` we define the state regardless if it's used or not.

If you'd prefer, this can be removed across the front end.

recipe-server/client/control/components/Fields.js
andymikulski: This specific bit has been replaced with `_.uniq`, which uses arrays, but if we _were_ to use Sets, I think that'd be worth opening a separate PR for. I'm not sure how `fetch` handles using Set as a query param, but we'd probably have to do some sort of array-ifing when interacting with the API. I will look into it.

recipe-server/client/control/components/MultiPicker.js
andymikulski: I was getting a warning in the console saying the given propType(s) were undefined, so I switched it out with a normal React type. I think the [redux-form propTypes](https://github.com/erikras/redux-form/blob/master/src/propTypes.js) are provided for the form component, not the fields.

recipe-server/client/control/components/Fields.js
Osmose: wfm 👍

PR #596, opened 2017-03-08
Fix recipe preview by including credentials in self-repair fetch

mythmon:
We are now using OIDC authentication on stage, and soon on prod. This introduced an error in recipe previews, because when we fetched actions (and other things), we did not include credentials (cookies), which meant that those requests got redirected to the auth0.

The fix for this is easy: include credentials in fetch for those requests.

The verification is harder: I rigged up the compose nginx to fake OIDC login, by requiring HTTP basic auth, and passing that as a `Remote-User` header. Now whatever is put in the HTTP basic auth prompt will be the user you get logged in, including creating a new user.

To test this: use compose. With only the first commit, the error from stage is reproduced (preview doesn't work). With both commits, it is fixed.

@relud can you r? the nginx changes here?

### DIFF
--- a/recipe-server/client/control/api/index.js
+++ b/recipe-server/client/control/api/index.js
@@ -19,7 +19,7 @@ const BASE_API_URL = '/api/v1/';
 
 // Default API request settings
 export const API_REQUEST_SETTINGS = {
-  credentials: 'include',
+  credentials: 'same-origin',
   headers: {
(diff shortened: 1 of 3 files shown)


mythmon:
The nginx changes caused a lot of test failures. It isn't needed for the fix, so I'm taking it out.

Issue #587, opened 2017-03-07
Preference Experiments Action


Osmose:
We're planning to run several experiments in the form of Firefox features hidden behind a preference. Normandy's job is to store the experiments we want to run as recipes, filter users by eligibility, flip the preferences on the client, and notify telemetry when users are or aren't enrolled in specific experiments.

The spec for changes we need to make to Normandy is at https://docs.google.com/document/d/1_ZAvAHK_weun8sp0tiwF4IE_7dwL2vKkjdI-QgI0GR4/edit#heading=h.rjz7v8qx2ryy.

## Arguments
- Experiment slug
- Link to experiment doc
- Preference name
- Preference type
- Feature Branches
  - Preference value
  - Number of buckets to apply to

## Action Behavior
When executed: 
1. If the user has never participated in the experiment, assign the user to a feature branch and register a new preference-change experiment.
   - See #592 for the preference experiment API.
   - See #586 for how feature branch bucketing works.
2. If the user is currently participating in the experiment, update the last-checked date of the experiment.
3. Store the experiment slug in the “current run” list (shared among all `preference-experiment` recipes)

Also, the action should register a cleanup function (#590):
   1. For any experiment slugs in the active experiment list that are not in the current run list, expire them.
   2. Empty the current run list.

Osmose:
#623 landed almost all of this, except for the post-recipe-execution bits, which are covered by #590. Otherwise, this is done.

PR #603, opened 2017-03-09
Basic preference experiment action

Osmose:
This adds a new `preference-experiment` action. It has the arguments that we've spec'd out, and forms to add them, but doesn't do anything with them (yet). That's why this targets a feature branch. It also refactors and adds some things to the RecipeForm to support what I wanted to do with the new action.

Because some of the details are still being worked out, I didn't add any tests for this yet. I expect to land most of the features on the feature branch, changing them as I go, and then working on tests as details solidify, but I can change that plan if we think it's a problem. If we're fine with that, then I'll probably merge this if it passes review, regardless of whether the tests pass, and we'll gate the final merge on tests passing.

This covers a fair amount of #587 but doesn't fully resolve it.

r?

### DIFF
--- a/recipe-server/client/control/components/Fields.js
+++ b/recipe-server/client/control/components/Fields.js
@@ -49,42 +49,50 @@ buildControlField.propTypes = {
   children: pt.node,
 };
 
-export function IntegerControlField(props) {
-  return (
-    <ControlField
-      component="input"
-      type="number"
-      step="1"
-      parse={value => {
-        try {
-          return Number.parseInt(value, 10);
-        } catch (err) {
-          return 0;
-        }
-      }}
-      {...props}
-    />
-  );
+export class IntegerControlField extends React.Component {
+  parse(value) {
+    return Number.parseInt(value, 10);
+  }
+
+  render() {
+    return (
+      <ControlField
+        component="input"
+        type="number"
+        step="1"
+        parse={this.parse}
+        {...this.props}
+      />
+    );
+  }
 }
 
-export function BooleanRadioControlField(props) {
-  return (
-    <ControlField
-      component="input"
-      type="radio"
-      className="radio-field"
-      parse={value => value === 'true'}
-      format={value => {
-        if (value) {
-          return 'true';
-        } else if (value !== undefined) {
-          return 'false';
-        }
-        return undefined;
-      }}
-      {...props}
-    />
-  );
+export class BooleanRadioControlField extends React.Component {
+  parse(value) {
+    return value === 'true';
+  }
+
+  format(value) {
+    if (value) {
+      return 'true';
+    } else if (value !== undefined) {
+      return 'false';
+    }
+    return undefined;
+  }
+
+  render() {
+    return (
+      <ControlField
+        component="input"
+        type="radio"
+        className="radio-field"
+        parse={this.parse}
+        format={this.format}
+        {...this.props}
+      />
+    );
+  }
 }
 
 export function ErrorMessageField(props) {
--- a/recipe-server/client/control/components/Fields.js
+++ b/recipe-server/client/control/components/Fields.js
@@ -23,13 +23,14 @@ export function buildControlField({
   label,
   className = '',
   InputComponent,
+  hideErrors = false,
   children,
   ...args // eslint-disable-line comma-dangle
 }) {
   return (
     <label className={`${className} form-field`}>
       <span className="label">{label}</span>
-      {error && <span className="error">{error}</span>}
+      {!hideErrors && error && <span className="error">{error}</span>}
       <InputComponent {...input} {...args}>
         {children}
       </InputComponent>
@@ -39,10 +40,65 @@ export function buildControlField({
 buildControlField.propTypes = {
   input: pt.object.isRequired,
   meta: pt.shape({
-    error: pt.string,
+    error: pt.oneOfType([pt.string, pt.array]),
   }).isRequired,
   label: pt.string.isRequired,
   className: pt.string,
   InputComponent: pt.oneOfType([pt.func, pt.string]),
+  hideErrors: pt.bool,
   children: pt.node,
+};
+
+export function IntegerControlField(props) {
+  return (
+    <ControlField
+      component="input"
+      type="number"
+      step="1"
+      parse={value => {
+        try {
+          return Number.parseInt(value, 10);
+        } catch (err) {
+          return 0;
+        }
+      }}
+      {...props}
+    />
+  );
+}
+
+export function BooleanRadioControlField(props) {
+  return (
+    <ControlField
+      component="input"
+      type="radio"
+      className="radio-field"
+      parse={value => value === 'true'}
+      format={value => {
+        if (value) {
+          return 'true';
+        } else if (value !== undefined) {
+          return 'false';
+        }
+        return undefined;
+      }}
+      {...props}
+    />
+  );
+}
+
+export function ErrorMessageField(props) {
+  return <Field component={buildErrorMessageField} {...props} />;
+}
+
+export function buildErrorMessageField({ meta: { error } }) {
+  if (error) {
+    return <span className="error">{error}</span>;
+  }
+  return null;
+}
+buildErrorMessageField.propTypes = {
+  meta: pt.shape({
+    error: pt.oneOfType([pt.string, pt.array]),
+  }).isRequired,
 };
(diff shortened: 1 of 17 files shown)


recipe-server/client/control/components/Fields.js
```diff
@@ -23,13 +23,14 @@ export function buildControlField({
   label,
   className = '',
   InputComponent,
+  hideErrors = false,
```
andymikulski: Technically the `= false` shouldn't be needed, but I've been a fan of setting defaults this way lately, so I think it's fine anyway.

recipe-server/client/control/components/Fields.js
```diff
@@ -39,10 +40,65 @@ export function buildControlField({
...
 };
+
+export function IntegerControlField(props) {
+  return (
+    <ControlField
+      component="input"
+      type="number"
+      step="1"
+      parse={value => {
+        try {
```
andymikulski: try/catches are slow in JS - could we use underscore's `isNumber` instead? ex:

`...parse={ value => _.isNumber(value) ? parseInt(value, 10) : 0 }...`

recipe-server/client/control/components/RecipeForm.js
```diff
@@ -215,6 +225,18 @@ export function initialValuesWrapper(Component) {
     if (location.state && location.state.selectedRevision) {
       initialValues = location.state.selectedRevision;
     }
+
+    // If we still don't have initial values, roll with the defaults.
+    if (!initialValues) {
+      initialValues = Object.assign({}, DEFAULT_FORM_VALUES);
```
andymikulski: The front end uses the spread operator instead of `Object.assign`, ex:

`initalValues = { ...DEFAULT_FORM_VALUES };`

recipe-server/client/control/components/RecipeForm.js
```diff
@@ -215,6 +225,18 @@ export function initialValuesWrapper(Component) {
...
+
+      // ActionField subclasses define their own initial values.
+      if (props.selectedAction) {
+        const ActionFields = RecipeForm.argumentsFields[props.selectedAction];
+        initialValues.arguments = Object.assign({}, ActionFields.initialValues);
```
andymikulski: Spread operator instead of Object.assign

recipe-server/client/control/components/action_fields/HeartbeatFields.js
```diff
@@ -1,99 +1,111 @@
...
-        type="text"
-      />
-      <ControlField
-        label="Post-Answer URL"
-        name="arguments.postAnswerUrl"
-        component="input"
-        type="text"
-      />
+export class HeartbeatFields extends ActionFields {
+  render({ recipeArguments }) {
```
andymikulski: This should default to an empty object, ex `({ recipeArguments = {} })`

recipe-server/client/control/components/Fields.js
```diff
@@ -39,10 +40,65 @@ export function buildControlField({
...
+export function BooleanRadioControlField(props) {
+  return (
+    <ControlField
+      component="input"
+      type="radio"
+      className="radio-field"
+      parse={value => value === 'true'}
+      format={value => {
+        if (value) {
+          return 'true';
```
andymikulski: The booleans-as-strings business seems funky to me, but I'm assuming it's a redux-form thing?

recipe-server/client/control/sass/control.scss
```diff
@@ -151,15 +155,32 @@
...
+
+    /* Checkbox/radio fields within fieldsets should be inline in a row. */
+    .checkbox-field, .radio-field {
+      display: inline-flex;
+      margin-right: 1.5em;
+    }
+  }
+
+  /* Checkbox/radio inputs should appear to the left of the label. */
+  .checkbox-field, .radio-field {
```
andymikulski: Has this been linted?

recipe-server/client/control/components/Fields.js
mythmon: When would this throw an error anyways? For non-number inputs, parseInt returns `NaN`.

Additionally, won't defining this function inline here cause unneeded re-renders of `ControlField`, since it will have a new value each time?

Osmose: I dunno where I got the idea that `parseInt` throws, but it's been stuck in my head for a long time. Someday I'll get it right.

recipe-server/client/control/components/Fields.js
Osmose: Redux-form reads the values from the input element, and input elements return strings as their values. The `parse` and `format` pairs here handle translating between that and a boolean value, which is what I want redux-form to actually store.

recipe-server/client/control/components/action_fields/HeartbeatFields.js
Osmose: This does need a default, but `initialValues` on the form is probably a better place for it.

recipe-server/client/control/sass/control.scss
Osmose: # CORNERS ARE FOR CUTTING

<sub><sup>No, I'll go and fix my linting errors. Sorry.</sup></sub>

Osmose:
Feedback'd, this is ready for another pass.

Osmose:
Reabsed on master. Pardon the opening and closing, Github was not picking up my changes to the branch for some reason.

recipe-server/client/control/components/RecipeForm.js
```diff
@@ -210,16 +221,32 @@ export const formConfig = {
...
+    if (!initialValues) {
+      initialValues = { ...DEFAULT_FORM_VALUES };
+
+      // ActionField subclasses define their own initial values.
+      if (selectedAction) {
+        const ActionFields = RecipeForm.argumentsFields[selectedAction];
+        initialValues.arguments = { ...ActionFields.initialValues };
+      } else {
+        // DEFAULT_FORM_VALUES wouldn't copy this, so we add it here instead.
+        initialValues.arguments = {};
```
andymikulski: Instead of `else`ing this, we could probably just do:

```
initialValues = { ...DEFAULT_FORM_VALUES, arguments: {} };
if (selectedAction) { ... }
```

recipe-server/client/control/components/action_fields/PreferenceExperimentFields.js
```diff
@@ -0,0 +1,179 @@
...
+            <ConnectedBranchFields
+              branch={branch}
+              onClickDelete={() => fields.remove(index)}
+            />
+          </li>
+        ))}
+        <li>
+          <a
+            className="button"
+            onClick={() => fields.push(Object.assign({}, DEFAULT_BRANCH_VALUES))}
```
andymikulski: Use spread operator (`{ ...DEFAULT_BRANCH_VALUES }`)

recipe-server/client/control/tests/components/test_RecipeForm.js
```diff
@@ -33,12 +33,14 @@ describe('<RecipeForm>', () => {
       expect(wrapper.find(ConsoleLogFields).length).toBe(1);
     });
 
-    it('should render the fields for the show-heartbeat action', () => {
-      const wrapper = shallow(
-        <RecipeForm selectedAction="show-heartbeat" {...propFactory()} />
-      );
-      expect(wrapper.find(HeartbeatFields).length).toBe(1);
-    });
+    // TODO(osmose): Figure out how to shallow-render and test connected components
```
andymikulski: I think you're looking for `mount` instead of `shallow`

Osmose: If I use mount, I get an error:

```
Error: Field must be inside a component decorated with reduxForm() in client/control/tests/index.js line 140 > eval (line 59)
```

The shallow render of `RecipeForm` elsewhere works because it's not a connected component (vs the default export, which I don't import during these tests), and the `Field` components never actually get rendered.

I suspect the long-term solution is either use mount and get better at mocking the state tree so that we can fully render connected components during tests, or do some wrapper around the heartbeat fields with a classname we can hook into. The former is definitely the more desirable option, but also conflicts with the upcoming state tree changes.

recipe-server/package.json
```diff
@@ -39,7 +39,7 @@
     "react-router-redux": "4.0.5",
     "reactable": "0.14.1",
     "redux": "3.6.0",
-    "redux-form": "6.1.0",
+    "redux-form": "6.5.0",
```
Osmose: So it _was_ in a PR!

I'm the worst.

Osmose:
Updated based on feedback. I also realized I was doing the function-comparison thing and refactored my components to be class-based components so I could use member functions to avoid the issue.

PR #604, opened 2017-03-10
Stop blocking framing of selfrepair via CSP

mythmon:
This changes our CSP policy to allow framing the self-repair view, since it gets loaded in an iframe in Firefox. This should only affect the `/en-US/repair` endpoint (and other locales). It doesn't change our global policy.

I also included two small changes to the contract tests to make them nicer to run, since I ran them a lot while making this.

### DIFF
--- a/recipe-server/normandy/settings.py
+++ b/recipe-server/normandy/settings.py
@@ -290,7 +290,7 @@ class Base(Core):
     APP_SERVER_URL = values.URLValue(None)
 
     # URL for the CSP report-uri directive.
-    CSP_REPORT_URI = values.Value(None)
+    CSP_REPORT_URI = values.Value('/__cspreport__')
 
     # Normandy settings
     ADMIN_ENABLED = values.BooleanValue(True)
--- a/recipe-server/normandy/settings.py
+++ b/recipe-server/normandy/settings.py
@@ -124,8 +124,13 @@ class Core(Configuration):
 
     # these don't fallback to default-src
     CSP_BASE_URI = "'none'"  # not using <base>
-    CSP_FRAME_ANCESTORS = "'none'"  # this page isn't iframed elsewhere
     CSP_FORM_ACTION = "'self'"  # we only submit forms to ourselves
+
+    # TODO(mythmon) Re-add this once either:
+    #   1) We know the domain we can use for Firefox's framing of the self-repair frame
+    #   2) We've deprecated the self-repair frame entirely
+    #   3) django-csp allows removing directives on an individual view (mozilla/django-csp#85)
+    # CSP_FRAME_ANCESTORS = "'none'"  # Block framing by default
 
     # Action names and the path they are located at.
     ACTIONS = {
--- a/recipe-server/contract-tests/test_api.py
+++ b/recipe-server/contract-tests/test_api.py
@@ -110,7 +110,7 @@ def test_recipe_signatures(conf, requests_session):
 
 
 def test_recipe_api_is_json(conf, requests_session):
-    r = requests_session.get(conf.getoption('server') + '/api/v1/recipe/')
+    r = requests_session.get(conf.getoption('server') + '/api/v1/recipe/?enabled=1')
     r.raise_for_status()
     data = r.json()
     assert isinstance(data, list)
--- a/recipe-server/contract-tests/test_performance.py
+++ b/recipe-server/contract-tests/test_performance.py
@@ -34,7 +34,7 @@ class TestHotPaths(object):
         r.raise_for_status()
         assert 'cookie' not in r.headers.get('vary', '').lower()
 
-    def test_cache_headers(self, conf, requests_session, path):
+    def test_cache_headers(self, conf, requests_session, path, only_readonly):
         r = requests_session.get(conf.getoption('server') + path)
         r.raise_for_status()
         cache_control = r.headers.get('cache-control')
--- a/recipe-server/contract-tests/test_performance.py
+++ b/recipe-server/contract-tests/test_performance.py
@@ -6,8 +6,8 @@ from pytest_testrail.plugin import testrail
 HOT_PATHS = [
     '/en-US/repair',
     '/en-US/repair/',
-    '/api/v1/recipe/',
-    '/api/v1/recipe/signed/',
+    '/api/v1/recipe/?enabled=1',
+    '/api/v1/recipe/signed/?enabled=1',
     '/api/v1/action/',
 ]
 
--- a/recipe-server/normandy/health/api/views.py
+++ b/recipe-server/normandy/health/api/views.py
@@ -1,3 +1,5 @@
+import json
+import logging
 import os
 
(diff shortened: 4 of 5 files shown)


recipe-server/contract-tests/test_selfrepair.py
```diff
@@ -0,0 +1,28 @@
...
+        # no CSP, no problem
+        return
+
+    csp_data = parse_csp(r.headers['Content-Security-Policy'])
+    if 'frame-src' in csp_data:
+        assert csp_data['frame-ancestors'] == '*'
+    else:
+        assert csp_data['default-src'] == '*'
+
+    if 'framce-ancestors' in csp_data:
```
Osmose: Whoops, typo. You mean `frame-ancestors` I think.

recipe-server/contract-tests/test_selfrepair.py
```diff
@@ -0,0 +1,28 @@
...
+            data = True
+        directives[directive] = data
+    return directives
+
+
+def test_selfrepair_csp_allows_framing(conf, requests_session):
+    r = requests_session.get(conf.getoption('server') + '/en-US/repair')
+    r.raise_for_status()
+    if 'Content-Security-Policy' not in r.headers:
+        # no CSP, no problem
```
jvehent: on the contrary, this should fail miserably

recipe-server/normandy/selfrepair/views.py
```diff
@@ -4,11 +4,14 @@
...
 from django.views.decorators.cache import cache_control
 
+from csp.decorators import csp_replace
 from rest_framework.reverse import reverse
 
 from normandy.base.decorators import short_circuit_middlewares
 
 
+# This page is primarily used by a hidden iframe in Firefox, so allow framing.
+@csp_replace(FRAME_SRC='*', FRAME_ANCESTORS='*')
```
jvehent: Is `*` needed here? It would be better to list trusted origins, such as `https://mozilla.org, https://*.mozilla.org` (just guessing, I don't know which origins are the right ones)

recipe-server/contract-tests/test_selfrepair.py
mythmon: I agree that not having CSP should be a test failure. However, this test is very narrowly scoped, only dealing with the bug I'm trying to fix here, which is critical functionality not loading due to an overly aggressive CSP policy.

recipe-server/normandy/selfrepair/views.py
mythmon: I'm not sure what the origin should be. We aren't being iframed in the traditional sense, but we are being [loaded in a "hidden browser"](https://dxr.mozilla.org/mozilla-central/source/browser/modules/SelfSupportBackend.jsm?q=path%3Aself+ext%3Ajsm&redirect_type=single#255-269). I'd be happy to restrict the origin to only that, if we can figure out what it is.

mythmon:
I learned a bit more about CSP and verified this better by actually running it against Firefox 52. Like @Osmose learned from the docs, `*` isn't a valid value for `frame-ancestors`. Setting it to `[]` makes `django-csp` delete the directive entirely, which is what we want for an "allow all" rule".

The "There was a problem fetching the SelfSupport URL" error we saw earlier was actually a CSP error, and after adding the CSP reporting endpoint, it was pretty easy to verify that. I can remove that if we don't think it is useful, it isn't needed for this fix, but it is quite helpful.

@jvehent I looked through the Firefox code and still wasn't able to figure out a origin that works here.  I tried `chrome:` and `resource:` based on [the code I found](http://searchfox.org/mozilla-central/source/xpfe/appshell/nsAppShellService.cpp#55), but they didn't work. This page doesn't render anything, so I don't think there is any risk of click jacking. Are there any other attacks we are worried about with any site being able to frame this page?

Osmose:
> @jvehent I looked through the Firefox code and still wasn't able to figure out a origin that works here. I tried chrome: and resource: based on the code I found, but they didn't work.

We ran into something like this with snippets on about:home, it ended up having a blank origin and we couldn't find a way to work around it.

recipe-server/normandy/settings.py
```diff
@@ -124,7 +124,7 @@ def CSP_DEFAULT_SRC(self):
 
     # these don't fallback to default-src
     CSP_BASE_URI = "'none'"  # not using <base>
-    CSP_FRAME_ANCESTORS = "'none'"  # this page isn't iframed elsewhere
+    # CSP_FRAME_ANCESTORS = "'none'"  # this site isn't, generally, iframed elsewhere
```
Osmose: We should probably just remove this.

mythmon: ...That change was not supposed to be in this patch. Oops.

mythmon:
Fixed the mistake @Osmose pointed o ut.

mythmon:
Turns out that I was wrong about passing `[]` making django-csp remove the directive. Looking at the code, any falsey value will make it fallback to the default. I don't think there is a way to have one page override to remove a directive. I'll make a PR to django-csp to fix this.

jvehent:
@mozfreddyb: Do you know which origin should be used in a CSP `frame-ancestors` to allows `about` pages to iframe `self-repair.mozilla.org` ?

mozfreddyb:
> @mozfreddyb: Do you know which origin should be used in a CSP frame-ancestors to allows about pages to iframe self-repair.mozilla.org ?

Our CSP implementation [gives about: URIs a pass](http://searchfox.org/mozilla-central/source/dom/security/nsCSPService.cpp#77). Are you sure your frmaing page has a proper URL?

On the other hand, if your framing element is a XUL `<browser>`, I'd be surprised if it was suspect to CSP at all.

mozfreddyb:
In general, for those odd, Firefox-specific things, the best way is to find out *shrugs*

jvehent:
> Our CSP implementation gives about: URIs a pass.

So, if we set `frame-ancetors: None;`, about: pages will still be allowed to frame the self-repair site but no one else will? 

@mythmon: Any chance you could test that behavior?

jvehent:
After discussion with @mozfreddyb, we realized that `frame-ancestors` gets applied to local resources, but the rest of the CSP does not.

We will have to find the proper source to put in that directive, or leave it empty to allow anyone to frame this site. Investigation continues, as I'm genuinely interested in finding the Right Way™️ of doing this.

jvehent:
Related: https://bugzilla.mozilla.org/show_bug.cgi?id=1346774

mythmon:
Since we aren't getting any traction on the blocking issues here, I've disabled the CSP `frame-ancestors` directive entirely, for the entire site. This is no worse than what we have today, and this PR is blocking a deployment of several other security issues.

mythmon:
@Osmose r?

mozfreddyb:
For the time being, you can use X-FRAME-OPTIONS: DENY, this should get a carve-out for chrome resources.

PR #607, opened 2017-03-11
Mock server testcases are all copies of ConsoleLogBasic

Osmose:
WIP. I'm having trouble replicating the issue locally and want to test if the test I wrote fails on CI, which _can_ replicate the issue.

### DIFF
--- a/recipe-server/normandy/base/decorators.py
+++ b/recipe-server/normandy/base/decorators.py
@@ -1,6 +1,8 @@
 from functools import wraps
 
+from django.conf import settings
 from django.db import transaction
+from django.views.decorators.cache import cache_control
 
 from reversion import revisions
 
@@ -25,3 +27,23 @@ def reversion_transaction(view_func):
         with transaction.atomic(), revisions.create_revision(manage_manually=True):
             return view_func(*args, **kwargs)
     return wraps(view_func)(wrapped_view)
+
+
+def api_cache_control(**kwargs):
+    """
+    Adds cache headers to a view using our API cache header defaults.
+    """
+    if settings.API_CACHE_ENABLED:
+        directives = {
+            'public': True,
+            'max_age': settings.API_CACHE_TIME,
+        }
+    else:
+        directives = {
+            'no_cache': True,
+            'no_store': True,
+            'must_revalidate': True,
+        }
+
+    directives.update(kwargs)
+    return cache_control(**directives)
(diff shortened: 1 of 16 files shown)


Osmose:
Righto, this is ready for review!

recipe-server/normandy/base/decorators.py
```diff
@@ -25,3 +27,23 @@ def wrapped_view(*args, **kwargs):
...
+            'max_age': settings.API_CACHE_TIME,
+        }
+    else:
+        directives = {
+            'no_cache': True,
+            'no_store': True,
+            'must_revalidate': True,
+        }
+
+    directives.update(kwargs)
```
mythmon: I think that doing this in all cases will cause problems when `API_CACHE_ENABLED=False`. The actions view would still be cache, even if the cache is disabled, right?

recipe-server/normandy/settings.py
```diff
@@ -297,6 +297,7 @@ def RAVEN_CONFIG(self):
     ACTION_IMPLEMENTATION_CACHE_TIME = values.IntegerValue(60 * 60 * 24 * 365)
     NUM_PROXIES = values.IntegerValue(0)
     API_CACHE_TIME = values.IntegerValue(30)
+    API_CACHE_ENABLED = values.BooleanValue(True)
```
mythmon: We need docs for this setting.

recipe-server/normandy/base/decorators.py
mythmon: I think that doing this in all cases will cause problems when API_CACHE_ENABLED=False. The actions view would still be cache, even if the cache is disabled, right?

recipe-server/normandy/settings.py
mythmon: We need docs for this setting.

recipe-server/normandy/base/decorators.py
Osmose: I don't think so, because it would still send `no-cache` and `no-store`, which (according to the [docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control)) forces the cache to submit to the origin server for revalidation.

I just tested this via compose, and with `DJANGO_API_CACHE_ENABLED=false` in the environment, all my requests to the action implementation view have `X-Cache: MISS` in the headers. Without it, they all have `X-Cache: HIT`.

mythmon: Yeah, I guess that makes sense. A little weird, but not bad.

recipe-server/normandy/settings.py
Osmose: Done!

PR #613, opened 2017-03-14
Merge peer-signing feature branch into master

rehandalal:
ec829a6 is the relevant merge commit.

r?


rehandalal:
@andymikulski could you look into the `recipes.list` issue?

andymikulski:
I've gone through and manually tested the front end (filters, edit recipe page, recipe history, etc) and fixed the `recipes.list` issue, as well as relevant tests (a few were using `list` when they should have been using `entries`). I also cleaned up the warnings in the console (mostly relating to changed propTypes).

npm tests pass and functionality checks out manually, though I think the Selenium tests still need updates.

rehandalal:
@mythmon this should be ready for review now

Osmose:
I ran this locally and noticed a few issues:

1. The enabled and draft indicators are misaligned:

screen shot 2017-03-27 at 11 07 51 am

2. When I save as a draft, the save goes through successfully, but then the edit page goes back to the loading indicator and stays there. From what I can tell the PATCH request went through fine, seems like a state issue. If I refresh, it fixes itself.

3. If I request approval, it says it successfully created an approval request and changes the buttons below, but if I refresh the page, I see the approval buttons briefly, and then they reset back to the "Request Approval" buttons. When in this state, I can't create a new approval request because one already exists.

4. If I try to disable a recipe, nothing happens and I get the following error in the console:

```
TypeError: _ControlActions.makeApiRequest is not a function
```

andymikulski:
I've fixed @Osmose 's review points:

1. I've abstracted the general "status indicator" component to normalize how those appear across the site (recipe form, history, preview)

2 + 3. These issues seem to be fixed with a correction to how recipes are loaded/'selected' in the state tree. Not sure if it was a merge issue or what, but it should be set now.

4. I think this was a merge issue - there were a couple of things still being imported from `ControlActions` that should have been imported from other `-Actions` files

rehandalal:
Hm... Selenium tests are passing for me locally but refuse to pass on CI.

rehandalal:
@mythmon merged master ^ 

it seems to have only shown the conflicts on this one which is nice!

Osmose:
> I've fixed @Osmose 's review points:

Thanks, this does seem to fix the issues I was seeing before! However, I'm now seeing new issues:

1. After creating a new recipe (or editing an existing recipe), the "Request Approval" button remains disabled until I refresh the page. If I save the draft, then edit the form back to it's original state before I made any edits, the "Save Draft" button is disabled and "Request Approval" is enabled again, even though the version saved on the server doesn't match the form.

2. If I submit a recipe for approval, and reject it, the "Request Approval" button disappears. But then, if I save a new draft, the button doesn't appear again, and the rejection information remains even though I've created a new revision (the rejection information says specifically "this revision has been rejected", which is why I was confused by this). Refreshing fixes this.

3. If I approve a recipe, and then save a new draft before enabling it, I get shown the loading indicator, and the devtools shows a successful PATCH request, but the returned recipe data doesn't match my edit. If I refresh the page, the recipe doesn't have my saved draft and is waiting for me to enable it again.

4. If I approve a recipe, then enable it, then save a new draft, I get shown the loading indicator. Refreshing the page shows that my new draft wasn't saved. If I disable the recipe, then the edit form updates and now has my saved draft from earlier.

Let me know if I can provide any more info or better replication steps.

mythmon:
The taskcluster failure had the line `virtual memory exhausted: Cannot allocate memory`, but it passed on the "push" build. I suspect this is a one-off thing that will work fine next time.

andymikulski:
1 + 2. There was an issue with the form not updating the recipes in memory - though the changes are sent to API and saved, the 'selected revision' was not updated locally, so the form didn't know that it needed to update.

3. You're right that the PATCH isn't saving the data passed to it - @rehandalal will need to check that out. There was also a front end issue where the page grabbed the `is approved` and `is enabled` value from the wrong place, which has been fixed.

4. The loading indicator should be fixed with the patch for 1 + 2, though with the non-working PATCH, the form appears with incorrect information. Fixing the API should fix all the 'outdated info' issues.

rehandalal:
@Osmose I believe my last commit should fix 3.

Osmose:
Here's the issues with the latest revision:

1. If I open the recipe edit page, make a change, and save a new draft, the approval button disappears entirely. The URL looks like it's a revision-specific page instead of the general recipe editing page. I expected to stay on the recipe edit page when saving new drafts.

2. If I save a new recipe, request approval, and approve it, I see an "Enable" and "Save Draft" button. If I then save a new draft, I no longer hit a loading indicator, but get redirected to the revision page as in issue 1. If I go back to the recipe edit page, it shows me the approved revision and not the new drafts. If I then enable the recipe, it appears that those drafts are gone.

3. If I try to save a new draft for an enabled recipe, the recipe edit view will only show me the approved version and not the new drafts. If I then disable the recipe, my edits show up again.

Osmose:
> The latest revision works consistently in every flow I tried, except for an odd bug where sometimes approving a revision will hang after submitting the approval, but @rehandalal is already working on fixing that.

Also he landed the fix for that immediately before I submitted my approval, and it appears to have fixed the issue. Hooray!

rehandalal:
@mythmon do you want to do a quick pass over the changes since b015026375e3f292767532f5a4e306c4b5be39a1?

mythmon:
...you know, once CI finishes.

PR #618, opened 2017-03-16
Disable newrelic browser monitoring

relud:
needs testing to confirm that it works


mythmon:
I have no way of testing that this has the desired affect, but it does seem like it should, and this shouldn't break anything. What can we do to test this?

relud:
maybe I can build a docker container and push it to docker hub, and I can run it on stage, and then curl a page and look for the script tag we are supposed to be disabling ?

mythmon:
That would work.

relud:
okay, confirmed, it does what we expect on stage, and it doesn't break newrelic either

mythmon:
Cool, thanks for testing it.

PR #619, opened 2017-03-17
recipe-server: Make whitenoise recognize our webpack'd files as immutable

mythmon:
This fixes a regression where our static files only had a cache length of 60 seconds. The upgrade of whitenoise also gets us support for [`Cache-Control: immutable`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#Revalidation_and_reloading), which should save us some traffic in prod.

I added the SRI test here, because last time we fiddled with the static loaders, we broke SRI. I wanted to make sure it stayed fixed.

### DIFF
--- a/recipe-server/normandy/settings.py
+++ b/recipe-server/normandy/settings.py
@@ -28,6 +28,7 @@ class Core(Configuration):
         'django.contrib.contenttypes',
         'django.contrib.sessions',
         'django.contrib.messages',
+        'whitenoise.runserver_nostatic',
         'django.contrib.staticfiles',
     ]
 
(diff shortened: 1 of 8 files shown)


recipe-server/webpack.config.js
```diff
@@ -10,11 +10,14 @@ var childProcess = require('child_process');
 const BOLD = '\u001b[1m';
 const END_BOLD = '\u001b[39m\u001b[22m';
 const production = process.env.NODE_ENV === 'production';
+const cssNamePattern = production ? '[name].[hash].css' : '[name].css';
```
Osmose: This is great. Someday I'd like to clean up this file and have something similar to django-configurations for specifying environmental settings for our webpack config.

recipe-server/normandy/settings.py
```diff
@@ -185,7 +186,7 @@ def MIDDLEWARE(self):
     def LOGGING(self):
         return {
             'version': 1,
-            'disable_existing_loggers': True,
+            'disable_existing_loggers': False,
```
Osmose: Why change this?

mythmon: This makes logging while running in Gunicorn work correctly, which helped greatly in debugging this.

PR #623, opened 2017-03-21
Preference experiment driver API

Osmose:
This adds `PreferenceExperiments.jsm` to the add-on, which implements the API for starting and stopping preference experiments, persists experiment state to the profile, modifies preferences, and starts watchers to end the experiment if the user changes the preference. This also includes a new `SandboxManager.wrapAsync` method for wrapping privileged async methods for use in the sandbox.

It also updates the action code to use the API and to assign the user to a feature branch correctly according to the branch ratios.

This is (intentionally, for now) missing:
- Checking for preference changes on a new session startup
- Starting preference watchers on a new session startup
- Removing expired preference experiments

(Including @chartjes on this for test coverage review of non-addon bits)

### DIFF
--- a/recipe-client-addon/lib/PreferenceExperiments.jsm
+++ b/recipe-client-addon/lib/PreferenceExperiments.jsm
@@ -75,6 +75,26 @@ CleanupManager.addCleanupHandler(() => PreferenceExperiments.stopAllObservers())
 
 this.PreferenceExperiments = {
   /**
+   * Test wrapper that temporarily replaces the stored experiment data with fake
+   * data for testing.
+   */
+  withMockExperiments(testGenerator) {
+    return function* inner(...args) {
+      const oldPromise = storePromise;
+      const mockExperiments = {};
+      storePromise = Promise.resolve({
+        data: mockExperiments,
+        saveSoon() { },
+      });
+      try {
+        yield testGenerator(...args, mockExperiments);
+      } finally {
+        storePromise = oldPromise;
+      }
+    };
+  },
+
+  /**
    * Clear all stored data about active and past experiments.
    */
   async clearAllExperimentStorage() {
--- a/recipe-client-addon/lib/PreferenceExperiments.jsm
+++ b/recipe-client-addon/lib/PreferenceExperiments.jsm
@@ -58,7 +58,7 @@ const EXPERIMENT_FILE = "shield-preference-experiments.json";
  * Asynchronously load the JSON file that stores experiment status in the profile.
  */
 let storePromise;
-function loadStorage() {
+function ensureStorage() {
   if (storePromise === undefined) {
     const path = OS.Path.join(OS.Constants.Path.profileDir, EXPERIMENT_FILE);
     const storage = new JSONFile({path});
@@ -70,7 +70,7 @@ function loadStorage() {
 const log = LogManager.getLogger("preference-experiments");
 
 // List of active preference observers. Cleaned up on shutdown.
-let experimentObservers = {};
+const experimentObservers = new Map();
 CleanupManager.addCleanupHandler(() => PreferenceExperiments.stopAllObservers());
 
 this.PreferenceExperiments = {
@@ -98,7 +98,7 @@ this.PreferenceExperiments = {
    * Clear all stored data about active and past experiments.
    */
   async clearAllExperimentStorage() {
-    const store = await loadStorage();
+    const store = await ensureStorage();
     store.data = {};
     store.saveSoon();
   },
@@ -116,7 +116,7 @@ this.PreferenceExperiments = {
   async start(experimentName, branch, preferenceName, preferenceValue) {
     log.debug(`PreferenceExperiments.start(${experimentName}, ${branch})`);
 
-    const store = await loadStorage();
+    const store = await ensureStorage();
     if (experimentName in store.data) {
       throw new Error(`A preference experiment named "${experimentName}" already exists.`);
     }
@@ -160,7 +160,7 @@ this.PreferenceExperiments = {
   startObserver(experimentName, preferenceName, preferenceValue) {
     log.debug(`PreferenceExperiments.startObserver(${experimentName})`);
 
-    if (experimentName in experimentObservers) {
+    if (experimentObservers.has(experimentName)) {
       throw new Error(
         `An observer for the preference experiment ${experimentName} is already active.`
       );
@@ -174,7 +174,7 @@ this.PreferenceExperiments = {
         }
       },
     };
-    experimentObservers[experimentName] = observerInfo;
+    experimentObservers.set(experimentName, observerInfo);
     Preferences.observe(preferenceName, observerInfo.observer);
   },
 
@@ -187,13 +187,13 @@ this.PreferenceExperiments = {
   stopObserver(experimentName) {
     log.debug(`PreferenceExperiments.stopObserver(${experimentName})`);
 
-    if (!(experimentName in experimentObservers)) {
+    if (!experimentObservers.has(experimentName)) {
       throw new Error(`No observer for the preference experiment ${experimentName} found.`);
     }
 
-    const {preferenceName, observer} = experimentObservers[experimentName];
+    const {preferenceName, observer} = experimentObservers.get(experimentName);
     Preferences.ignore(preferenceName, observer);
-    delete experimentObservers[experimentName];
+    experimentObservers.delete(experimentName);
   },
 
   /**
@@ -201,10 +201,10 @@ this.PreferenceExperiments = {
    */
   stopAllObservers() {
     log.debug("PreferenceExperiments.stopAllObservers()");
-    for (const {preferenceName, observer} of Object.values(experimentObservers)) {
+    for (const {preferenceName, observer} of experimentObservers.values()) {
       Preferences.ignore(preferenceName, observer);
     }
-    experimentObservers = {};
+    experimentObservers.clear();
   },
 
   /**
@@ -217,7 +217,7 @@ this.PreferenceExperiments = {
   async markLastSeen(experimentName) {
     log.debug(`PreferenceExperiments.markLastSeen(${experimentName})`);
 
-    const store = await loadStorage();
+    const store = await ensureStorage();
     if (!(experimentName in store.data)) {
       throw new Error(`Could not find a preference experiment named "${experimentName}"`);
     }
@@ -239,7 +239,7 @@ this.PreferenceExperiments = {
   async stop(experimentName, resetValue = true) {
     log.debug(`PreferenceExperiments.stop(${experimentName})`);
 
-    const store = await loadStorage();
+    const store = await ensureStorage();
     if (!(experimentName in store.data)) {
       throw new Error(`Could not find a preference experiment named "${experimentName}"`);
     }
@@ -274,7 +274,7 @@ this.PreferenceExperiments = {
    */
   async get(experimentName) {
     log.debug(`PreferenceExperiments.get(${experimentName})`);
-    const store = await loadStorage();
+    const store = await ensureStorage();
     if (!(experimentName in store.data)) {
       throw new Error(`Could not find a preference experiment named "${experimentName}"`);
     }
@@ -290,7 +290,7 @@ this.PreferenceExperiments = {
    */
   async has(experimentName) {
     log.debug(`PreferenceExperiments.has(${experimentName})`);
-    const store = await loadStorage();
+    const store = await ensureStorage();
     return experimentName in store.data;
   },
 };
--- a/recipe-client-addon/lib/PreferenceExperiments.jsm
+++ b/recipe-client-addon/lib/PreferenceExperiments.jsm
@@ -114,7 +114,7 @@ this.PreferenceExperiments = {
    *   for the given preference is active.
    */
   async start(experimentName, branch, preferenceName, preferenceValue) {
-    log.debug(`PreferenceExperiments.start(${experimentName})`);
+    log.debug(`PreferenceExperiments.start(${experimentName}, ${branch})`);
 
     const store = await loadStorage();
     if (experimentName in store.data) {
--- a/recipe-client-addon/lib/PreferenceExperiments.jsm
+++ b/recipe-client-addon/lib/PreferenceExperiments.jsm
@@ -1,0 +1,276 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * Preference Experiments temporarily change a preference to one of several test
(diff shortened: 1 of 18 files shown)


recipe-client-addon/test/unit/test_SandboxManager.js
```diff
@@ -0,0 +1,95 @@
...
+    manager.addGlobal("resolve", result => resolve(result));
+    manager.evalInSandbox(`
+      // Unwrapped privileged promises are not accessible in the sandbox
+      try {
+        const privilegedResult = driver.privileged().then(() => false);
+        ok(false, "The sandbox could not use a privileged Promise");
+      } catch (err) { }
+
+      // Wrapped functions return promises that the sandbox can access.
+      const wrappedResult = driver.wrapped();
```
mythmon: It is confusing that both this value and the one on line 23 are both called `wrappedResult`. I tried for a while to see how they were related. Can one of them be a different name?

recipe-server/client/actions/preference-experiment/index.js
```diff
@@ -1,8 +1,59 @@
...
+ */
 export default class PreferenceExperimentAction extends Action {
-  async execute() {
-    this.normandy.log(this.recipe.arguments.slug, 'info');
+  execute() {
+    // Babel's async transform handles the promises in such a way that it tries
+    // to call `then` on Promises from a privileged context. I'm not entirely
+    // sure why, but it works fine if we use Promises directly.
+    //
+    // Once we remove self-repair support, we should be save to use native
```
mythmon: You said "save". I think you mean "able"?

recipe-server/client/actions/tests/test_preference-experiment.js
```diff
@@ -0,0 +1,136 @@
+import PreferenceExperimentAction from '../preference-experiment/';
+import { recipeFactory } from '../../tests/utils.js';
+
+function preferenceExperimentFactory({ ...args }) {
```
mythmon: What does `{ ...args }` in a parameter list do? How is this different than `function preferenceExperimentFactory(args) {`?

recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -0,0 +1,296 @@
...
+      );
+    }
+
+    PreferenceExperiments.stopObserver(experimentName);
+
+    if (resetValue) {
+      const {preferenceName, previousPreferenceValue} = experiment;
+      Preferences.reset(preferenceName);
+      if (previousPreferenceValue !== undefined) {
+        Preferences.set(preferenceName, previousPreferenceValue);
```
Osmose: Reset will remove a preference that doesn't have a default value. This lets us support removing a preference from about:config entirely if it didn't exist there before, with `undefined` being the difference between an empty default value, like an empty string, and a preference that wasn't even there.

recipe-client-addon/test/unit/test_PreferenceExperiments.js
```diff
@@ -0,0 +1,306 @@
...
+  equal(experiment.name, "test", "start saved the experiment name");
+  equal(experiment.branch, "branch", "start saved the experiment branch");
+  equal(experiment.expired, false, "start saved the expired status");
+  equal(experiment.preferenceName, "fake.preference", "start saved the preference name");
+  equal(experiment.preferenceValue, "newvalue", "start saved the preference value");
+  equal(
+    experiment.previousPreferenceValue,
+    "oldvalue",
+    "start saved the previous preference value",
+  );
```
Osmose: Deep equal would compare the `lastSeen` date (or at least I thought it would), which I don't want for this test. I could extract the properties I do want to compare, though. Do you prefer that?

recipe-server/client/actions/preference-experiment/index.js
Osmose: I think I meant safe, but able works just as well.

recipe-server/client/actions/tests/test_preference-experiment.js
Osmose: Nothing. I had it as `(...args)` at first, realized that didn't work like `**kwargs`, and slapped the `{}` on without thinking too hard. I'll fix this.

recipe-client-addon/test/unit/test_PreferenceExperiments.js
Osmose: I went ahead and made this change.

recipe-client-addon/test/unit/xpc_head.js
```diff
@@ -17,3 +17,10 @@ if (!extensionDir.exists()) {
   extensionDir.append(EXTENSION_ID + ".xpi");
 }
 Components.manager.addBootstrappedManifestLocation(extensionDir);
+
+// Load mocking/stubbing library, sinon
+// docs: http://sinonjs.org/docs/
+Cu.import("resource://gre/modules/Timer.jsm");
+const self = {}; // eslint-disable-line no-unused-vars
```
gijsk: Why do we need the `self` object?

recipe-client-addon/lib/Sampling.jsm
```diff
@@ -130,4 +130,35 @@ this.Sampling = {
 
     return Sampling.isHashInBucket(inputHash, wrappedStart, end, total);
   }),
+
+  /**
+   * Sample over a list of ratios such that, over the input space, each ratio
+   * has a number of matches in correct proportion to the other ratios.
```
gijsk: I'm not actually that familiar with the statistics involved here. What I *think* this means is that if you have these ratios:
```
[1,2,3]
```
half of all input will return 2, a third will return 1, and one sixth of input will return 0. Is that right? It feels like the comment would benefit from an example (either this or some other one).

recipe-client-addon/lib/SandboxManager.jsm
```diff
@@ -65,6 +65,43 @@ this.SandboxManager = class {
       }
     });
   }
+
+  /**
+   * Wraps a function that returns a Promise from a privileged (e.g. chrome)
```
gijsk: English/Latin nitpickery nit: I think you mean "i.e." ("id est", so "that/it is"), not "e.g." ("exempli gratia", so "for example"). In this case, 'privileged' and 'chrome' are the same thing, so one is not an example of the other.

recipe-client-addon/lib/SandboxManager.jsm
```diff
@@ -65,6 +65,43 @@ this.SandboxManager = class {
...
+   *   sandbox before being resolved by the sandbox Promise. Without this, the
+   *   result will be Xray-wrapped.
+   * @param {boolean} [options.cloneArguments=false]
+   *   If true, the arguments passed to wrappedFunction will be cloned into the
+   *   privileged chrome context. If wrappedFunction holds a reference to any of
+   *   its arguments, you will need this to avoid losing access to the arguments
+   *   when the sandbox they originate from is nuked.
+   * @return {Function}
+   */
+  wrapAsync(wrappedFunction, options = {cloneInto: false, cloneArguments: false}) {
```
gijsk: I wonder if gecko should have a builtin sandbox API for this...

recipe-client-addon/lib/RecipeRunner.jsm
```diff
@@ -176,11 +174,11 @@ this.RecipeRunner = {
       sandboxManager.addGlobal("actionFailed", err => {
         Cu.reportError(err);
 
-        // Error objects can't be cloned, so we just copy the message
-        // (which doesn't need to be cloned) to be somewhat useful.
-        const message = err.message;
+        // Copy the error before calling `removeHold`, which will nuke
+        // the sandbox and break the wrapper around the error.
+        const {message, fileName, lineNumber} = err.message;
```
gijsk: I think this should be assigning from `err`, not from `err.message` ?

recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -0,0 +1,296 @@
...
+
+this.EXPORTED_SYMBOLS = ["PreferenceExperiments"];
+
+const EXPERIMENT_FILE = "shield-preference-experiments.json";
+
+/**
+ * Asynchronously load the JSON file that stores experiment status in the profile.
+ */
+let storePromise;
+function loadStorage() {
```
gijsk: Nit: because this doesn't always actually do a file load, I think calling it `ensureStorage` or similar would make more sense.

recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -0,0 +1,296 @@
...
+ * They are single-depth objects to simplify cloning.
+ * @typedef {Object} Experiment
+ * @property {string} name
+ *   Unique name of the experiment
+ * @property {string} branch
+ *   Experiment branch that the user was matched to
+ * @property {boolean} expired
+ *   If false, the experiment is active.
+ * @property {string} lastSeen
+ *   ISO-formatted date string of when the experiment was last seen from the
```
gijsk: Why is this using an ISO-formatted date string instead of either a Date object or a unix timestamp?

recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -0,0 +1,296 @@
...
+  withMockExperiments(testGenerator) {
+    return function* inner(...args) {
+      const oldPromise = storePromise;
+      const mockExperiments = {};
+      storePromise = Promise.resolve({
+        data: mockExperiments,
+        saveSoon() { },
+      });
+      try {
+        yield testGenerator(...args, mockExperiments);
```
gijsk: Should this be an async function that uses `await` here?

recipe-client-addon/lib/PreferenceExperiments.jsm
gijsk: Sure, but we could do:

```js
if (previousPreferenceValue !== undefined) {
  Preferences.set(preferenceName, previousPreferenceValue);
} else {
  Preferences.reset(preferenceName);
}
```

recipe-client-addon/lib/Sampling.jsm
gijsk: Also, I wonder if this would be simpler as its own PR - it doesn't seem strictly related to the preferences API?

gijsk:
Also, should we check what type a preference is and explicitly reject values that don't match the existing type from the recipe runner API side?

recipe-client-addon/test/unit/xpc_head.js
Osmose: sinon tries to find a global object in a few different ways, one of which is looking for `self`. It throws an error if you don't have this, and [the other places that load sinon during xpcshell tests](https://dxr.mozilla.org/mozilla-central/search?q=sinon-1.16&=mozilla-central) that I could find all have something similar (That's where I copied this from, in fact).

This could use an explaining comment, though.

recipe-client-addon/lib/SandboxManager.jsm
Osmose: > 'privileged' and 'chrome' are the same thing,

I thought they were subtly different things (are there other non-chrome contexts that can be considered privileged?), hence the e.g. I'll fix this.

gijsk: "It's complicated". "system principal" is typically what we understand "privileged" or "chrome privileged" to mean, and there's only 1 of those.

However, the code here would also use for 2 non-privileged scopes (so if you had sandboxed bar.com code that cared about a promise from foo.com). So maybe just 'privilege' isn't necessarily the best term here, although of course really the only consumers here care about privileged code <-> sandboxed code interaction...

recipe-client-addon/lib/PreferenceExperiments.jsm
Osmose: Because it's what `Date.prototype.toJSON` returns, and these objects get persisted to a JSON file.

Right now we simply update the date, so storing it as a string isn't inconvenient, but a future PR will be comparing it to the current date, which would require converting it. I figured that converting it to a date at that one point wouldn't be an issue. 

Is there an established preference for storing dates in Firefox?

recipe-client-addon/lib/Sampling.jsm
Osmose: > Is that right?

Yep. I ran some tests in the browser console to confirm the behavior:

screen shot 2017-03-20 at 11 56 51 pm

> Also, I wonder if this would be simpler as its own PR - it doesn't seem strictly related to the preferences API?

Yeah, in retrospect that's a good cut point. At the time what I did was write the action code, then write the system add-on APIs to support it, and choosing a branch seemed critical.

recipe-client-addon/lib/PreferenceExperiments.jsm
gijsk: "It's complicated". :-)

We've been known to do things like using the unix timestamp and then either rounding to seconds to store in prefs (which don't support the accuracy of ms-based unix timestamps), or storing them as strings in prefs (yes, yes, I know). 

I hadn't seen the `toJSON` usage when I wrote that comment. I guess given the JSON factor it makes sense to keep it like this and you should ignore my comment. Sorry!

Osmose:
> I've already left some comments. Otherwise, is the goal that extant recipes get the power to trigger their own preference experiments? 

I'm not quite sure what you mean by "extant recipes". All the recipes will have access to the methods for starting them, but only the `preference-experiment` action will use the API. Having non-preference-experiment recipes use the API isn't a specific goal.

>     should we limit what preferences they can toggle? Some are more... problematic... than others.

I think limiting what preferences can be toggled is up to the approval process that RelMan comes up with for approving these recipes to go out. Even for scary things like update-server URLs, I can see us using Normandy to run a test of moving, like, 1% of users to a new update server to see if it can handle the load or break things for those users. This PR doesn't have it, but we plan to add a timeout so that if the browser can't reach the recipe server after some amount of days, we revert the experiment and preference, which makes this feature ideal for testing things like that.

>     can they coordinate setting more than one pref per recipe?

Nope, it's explicitly not something we want to support. Features that want to change multiple preferences for an experiment will be required to add a new preference that triggers those changes. Apparently changing multiple preferences has implications for experiment design and reporting that the working group didn't want to deal with.

Osmose:
> Also, should we check what type a preference is and explicitly reject values that don't match the existing type from the recipe runner API side?

I like that idea, I'll run it by product to make sure it's not a problem.

recipe-client-addon/lib/SandboxManager.jsm
Osmose: Thanks! Although I wonder where we'd put it, as AFAICT the sandbox class is in C++ and adding more sandbox functions to `Components.utils` seems kinda messy.

Osmose:
Put up fixes from feedback, including removing `ratioSample`, which will be in a small followup PR instead. Only thing that should still be waiting on me is checking about preference type-checking, which probably will land in a followup PR anyway instead of this one.

Osmose:
Pulling in the already-approved fix to our TC builds to get this passing CI.

PR #624, opened 2017-03-22
Refactor all uses of Task.jsm to use native async/await instead.

Osmose:
Because @Mossop gave me the side eyes earlier today about using `Task.async`.

### DIFF
--- a/recipe-client-addon/lib/Storage.jsm
+++ b/recipe-client-addon/lib/Storage.jsm
@@ -10,7 +10,6 @@ Cu.import("resource://shield-recipe-client/lib/LogManager.jsm");
 
 XPCOMUtils.defineLazyModuleGetter(this, "JSONFile", "resource://gre/modules/JSONFile.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "OS", "resource://gre/modules/osfile.jsm");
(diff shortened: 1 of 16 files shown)


recipe-client-addon/lib/Storage.jsm
```diff
@@ -21,10 +20,10 @@ function loadStorage() {
   if (storePromise === undefined) {
     const path = OS.Path.join(OS.Constants.Path.profileDir, "shield-recipe-client.json");
     const storage = new JSONFile({path});
-    storePromise = Task.spawn(function* () {
-      yield storage.load();
+    storePromise = (async function () {
```
gijsk: Can't this just do:

```js
storePromise = storage.load();
```

directly?

Osmose: Nah, [JSONFile.load doesn't return anything](https://dxr.mozilla.org/mozilla-central/source/toolkit/modules/JSONFile.jsm#183,211,216) and we want to return the storage object.

On another PR I think I might've refactored this to

```js
storePromise = storage.load().then(() => storage);
```

But I figured for this PR that keeping the translation to async dumb was the way to go.

Mossop:
:eyes:

PR #626, opened 2017-03-22
recipe-server: Use more eslint rules from eslint-plugin-react

mythmon:
The main rule that I wanted to add here was `react/jsx-no-bind`, which would have caught several review comments that have come up recently. While I was reading about the rules, I also noticed that there were several rules in `eslint-plugin-react` that seemed useful that we weren't using. Some of them even prevent straight up errors.

I'm happy to remove any of these extra rules if they aren't wanted.

### DIFF
--- a/recipe-server/client/control/components/RecipeFilters.js
+++ b/recipe-server/client/control/components/RecipeFilters.js
@@ -128,7 +128,7 @@ export class RecipeFilters extends React.Component {
     return (
       <div className="fluid-8">
         <div id="secondary-header" className="fluid-8">
-          <div className="header-search" className="fluid-2">
+          <div className="header-search fluid-2">
             <RecipeCombobox
               onFilterSelect={this.handleAddFilter}
               availableFilters={availableFilters}
--- a/recipe-server/client/control/sass/control.scss
+++ b/recipe-server/client/control/sass/control.scss
@@ -90,7 +90,6 @@
   h3 {
     color: $darkBrown;
     display: inline-block;
-    float: left;
     font-size: 14px;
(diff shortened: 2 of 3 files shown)


recipe-server/client/control/components/HistoryItem.js
```diff
@@ -0,0 +1,71 @@
+import React, { PropTypes as pt } from 'react';
```
mythmon: This is extracting the HistoryItem component from `RecipeHistory` into its own file.

recipe-server/client/control/components/RecipeFilters.js
```diff
@@ -128,7 +128,7 @@ export class RecipeFilters extends React.Component {
     return (
       <div className="fluid-8">
         <div id="secondary-header" className="fluid-8">
-          <div className="header-search" className="fluid-2">
+          <div className="header-search fluid-2">
```
mythmon: This changes the markup, because it used to render as just `class="fluid-2"`. I tested this, and fixed an issue that came up, relating to the display of the filter items when being hovered.

recipe-server/client/control/sass/control.scss
```diff
@@ -90,7 +90,6 @@
   h3 {
     color: $darkBrown;
     display: inline-block;
-    float: left;
```
mythmon: This was causing the first item in a list of possible filters to be much larger than it should have. Removing it didn't seem to cause any other issues.

mythmon:
I set the `react/no-multi-comp` back to not being required, and undid the changes I made in support of it.

PR #628, opened 2017-03-23
Bug 1348885: Do not call Cu.reportError on errors from the sandbox.

Osmose:
See [bug 1348885](https://bugzilla.mozilla.org/show_bug.cgi?id=1348885) for details. This is the simple way to avoid a hard-looking issue. :P

Only asking for one reviewer since this is a one-liner.

### DIFF
--- a/recipe-client-addon/lib/RecipeRunner.jsm
+++ b/recipe-client-addon/lib/RecipeRunner.jsm
@@ -174,8 +174,6 @@ this.RecipeRunner = {
         resolve(clonedResult);
       });
       sandboxManager.addGlobal("actionFailed", err => {
-        Cu.reportError(err);
-
         // Error objects can't be cloned, so we just copy the message
         // (which doesn't need to be cloned) to be somewhat useful.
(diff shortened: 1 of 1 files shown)


recipe-client-addon/lib/RecipeRunner.jsm
```diff
@@ -174,8 +174,6 @@ this.RecipeRunner = {
         resolve(clonedResult);
       });
       sandboxManager.addGlobal("actionFailed", err => {
-        Cu.reportError(err);
```
gijsk: I expect you could still report the fake-cloned error you create a few lines lower down? Which might still be helpful, but eh, I don't really mind too much. Up to you whether you want to bother with that or not.

Osmose: It does get reported, although I think it gets reported via the "you didn't handle a promise rejection and it happened, shame on you" mechanism. I'll followup with another PR to properly handle that rejection.

PR #629, opened 2017-03-23
Bug 1349596: Validate JEXL on the server instead of the client.

Osmose:
Validating JEXL on the client isn't ideal, because the JEXL library doesn't
have a public API for checking if a statement is valid without evaluating
it, which means certain valid expressions won't pass validation if they
reference data in the context that we don't provide in the admin interface.

PyJEXL is a Python library that parses and evaluates JEXL, and it includes
a validation API that is more suitable for our validation.


Osmose:
FWIW I ran some of the more complex filters from prod against this and they all validated fine.

recipe-server/normandy/recipes/api/serializers.py
```diff
@@ -96,6 +97,20 @@ def create(self, validated_data):
 
         return self.update(recipe, validated_data)
 
+    def validate_extra_filter_expression(self, value):
+        jexl = JEXL()
+
+        # Add mock transforms for validation
+        jexl.add_transform('date', lambda x: x)
+        jexl.add_transform('stableSample', lambda x: x)
+        jexl.add_transform('bucketSample', lambda x: x)
```
mythmon: This isn't very satisfying, but I don't know how we can do better. Maybe we can put a link to the docs that say what transforms need to be available?

recipe-server/requirements/default.txt
```diff
@@ -120,3 +120,6 @@ django-csp==3.2 \
 html5lib==1.0b10 \
     --hash=sha256:08a3efc117a4fc8c82c3c6d10d6f58ae266428d57ed50258a1466d2cd88de745 \
     --hash=sha256:0d5fd54d5b2b79b876007a70c033a4023577768d18022c15681c00561432a0f9
+pyjexl==0.1.2 \
```
mythmon: This should be pyjexl 0.1.3 now.

Osmose:
Test failures are unrelated and fixed by #632.

mythmon:
Ok, can you rebase this to pick up the changes in #632? Then this should be ready to merge.

Osmose:
Rebased!

PR #631, opened 2017-03-24
Enable artifact builds in TaskCluster.

Osmose:
Maybe it's this simple!


Osmose:
Welp, the log viewer has been frozen for 10 minutes now... so let's assume it's not simple.

mythmon:
For what it's worth, this is exactly the same problem I ran into before. The builds just hang, with no indication as to why.

PR #632, opened 2017-03-24
Fix TaskCluster builds.

Osmose:
@Mossop says installing libgl1-mesa-dev will fix the build errors due to webrender. Let's see!


Osmose:
Looks like this did the trick!

PR #633, opened 2017-03-24
addon: Add Storage durability checks (#433)

### LINKED ISSUE #433: Add storage durability checks to add-on
Self-repair checks that storage is durable before testing it. The add-on driver should do the same thing.

andymikulski:
Fixes #433 

I think this is ready for feedback, though I had a few questions:

- Should non-durable storage throw errors?
- Right now I have it set up so that when Storage is created, it seeds the durability check, and then every read/write after that it checks the durability.
  - Does it matter if the durability is set to 0 for each new Storage? Should it retain the existing value for any reason?
  - Is it overkill to run a durability check on each read/write? On selfrepair we check on every read, fwiw.
- Is `store.saveSoon()` necessary to save the durability key update? I'm assuming so.
- How should tests for this be written? There already exists a bunch of `browser_Storage` tests that cover storing different types of data (objects, strings, numbers, etc). 
  - Could it be as simple as creating two stores with the same namespace, and testing if they can interact with the same values?
  - On the self repair side, we simply set a value and then check if the store can read it - which doesn't necessarily seem to be a complete 'durability' check, but maybe I'm wrong.
- When a Storage object is created, I see a call to `Cu.cloneInto(...)` - does the durability seed need to happen after that step?

### DIFF
--- a/recipe-client-addon/bin/make-xpi.sh
+++ b/recipe-client-addon/bin/make-xpi.sh
@@ -23,7 +23,7 @@ rm "${DEST}/install.rdf.in"
 cp "${BASE_DIR}/install.rdf" "${DEST}"
 rm "${DEST}/jar.mn"
 cp "${BASE_DIR}/chrome.manifest" "${DEST}"
-rm "${DEST}/test" -r
+rm -r "${DEST}/test"
 
 pushd $DEST
 zip -r shield-recipe-client.xpi *
--- a/recipe-client-addon/lib/NormandyDriver.jsm
+++ b/recipe-client-addon/lib/NormandyDriver.jsm
@@ -26,16 +26,14 @@ const log = LogManager.getLogger("normandy-driver");
 const actionLog = LogManager.getLogger("normandy-driver.actions");
 
 this.NormandyDriver = function(sandboxManager) {
+
   if (!sandboxManager) {
     throw new Error("sandboxManager is required");
   }
   const {sandbox} = sandboxManager;
 
-  Storage.seedDurability(sandbox);
-
   return {
     testing: false,
-    skipDurabilityCheck: false,
 
     get locale() {
       return Cc["@mozilla.org/chrome/chrome-registry;1"]
@@ -120,10 +118,10 @@ this.NormandyDriver = function(sandboxManager) {
       return ret;
     },
 
-    async createStorage(keyPrefix) {
+    async createStorage(keyPrefix, skipDurabilityCheck) {
       let storage;
       try {
-        if (!this.skipDurabilityCheck) {
+        if (!skipDurabilityCheck) {
           await Storage.checkDurability(sandbox);
         }
 
--- a/recipe-client-addon/lib/NormandyDriver.jsm
+++ b/recipe-client-addon/lib/NormandyDriver.jsm
@@ -31,8 +31,11 @@ this.NormandyDriver = function(sandboxManager) {
   }
   const {sandbox} = sandboxManager;
 
+  Storage.seedDurability(sandbox);
+
   return {
     testing: false,
+    skipDurabilityCheck: false,
 
     get locale() {
       return Cc["@mozilla.org/chrome/chrome-registry;1"]
@@ -117,9 +120,13 @@ this.NormandyDriver = function(sandboxManager) {
       return ret;
     },
 
-    createStorage(keyPrefix) {
+    async createStorage(keyPrefix) {
       let storage;
       try {
+        if (!this.skipDurabilityCheck) {
+          await Storage.checkDurability(sandbox);
+        }
+
         storage = Storage.makeStorage(keyPrefix, sandbox);
       } catch (e) {
         log.error(e.stack);
--- a/recipe-client-addon/lib/Storage.jsm
+++ b/recipe-client-addon/lib/Storage.jsm
@@ -14,6 +14,8 @@ XPCOMUtils.defineLazyModuleGetter(this, "OS", "resource://gre/modules/osfile.jsm
 this.EXPORTED_SYMBOLS = ["Storage"];
 
 const log = LogManager.getLogger("storage");
+
+const STORAGE_DURABILITY_KEY = '_storageDurability';
 let storePromise;
 
 function loadStorage() {
@@ -36,6 +38,56 @@ this.Storage = {
 
     const storageInterface = {
       /**
+       * Sets the initial 'durability values' for this store to increase durability
+       * confidence in later checks. Should only be called on instantiation.
+       *
+       * @returns  {Promise}
+       * @resolves When the store durability is set successfully.
+       * @rejects  Javascript exception.
+       */
+      seedStoreDurability() {
+        return new sandbox.Promise((resolve, reject) =>
+          loadStorage()
+            .then(store => {
+              store.data[prefix] = store.data[prefix] || {};
+              store.data[prefix][STORAGE_DURABILITY_KEY] = 0;
+              store.saveSoon();
+            })
+            .catch(err => reject(new sandbox.Error()))
+        );
+      },
+
+      /**
+       * Checks the durability of the current store
+       *
+       * @returns  {Promise}
+       * @resolves When the store durability is set successfully.
+       * @rejects  Javascript exception.
+       */
+      checkStoreDurability(store) {
+        store.data[prefix] = store.data[prefix] || {};
+        const localStore = store.data[prefix];
+
+        // Determine if the store has an existing durability key, and if so,
+        // detect if it is an actual number or not.
+        let durability = parseInt(localStore[STORAGE_DURABILITY_KEY], 10);
+        const isDurabilityInvalid = isNaN(durability);
+
+        // If `durability` is invalid (doesn't exist, is NaN), there is a problem
+        // with the storage's integrity.
+        if(isDurabilityInvalid) {
+          log.warn('Storage durability unconfirmed');
+        } else {
+          // If we're still here, set the new durability value into storage.
+          store.data[prefix][STORAGE_DURABILITY_KEY] = durability + 1;
+          store.saveSoon();
+        }
+
+        // Return the store so this method can be chained into promises.
+        return store;
+      }
+
+      /**
        * Sets an item in the prefixed storage.
        * @returns {Promise}
        * @resolves With the stored value, or null.
@@ -44,6 +96,7 @@ this.Storage = {
       getItem(keySuffix) {
         return new sandbox.Promise((resolve, reject) => {
           loadStorage()
+            .then(this.checkStoreDurability)
             .then(store => {
               const namespace = store.data[prefix] || {};
               const value = namespace[keySuffix] || null;
@@ -65,6 +118,7 @@ this.Storage = {
       setItem(keySuffix, value) {
         return new sandbox.Promise((resolve, reject) => {
           loadStorage()
+            .then(this.checkStoreDurability)
             .then(store => {
               if (!(prefix in store.data)) {
                 store.data[prefix] = {};
@@ -89,6 +143,7 @@ this.Storage = {
       removeItem(keySuffix) {
         return new sandbox.Promise((resolve, reject) => {
           loadStorage()
+            .then(this.checkStoreDurability)
             .then(store => {
               if (!(prefix in store.data)) {
                 return;
@@ -113,6 +168,7 @@ this.Storage = {
       clear() {
         return new sandbox.Promise((resolve, reject) => {
           return loadStorage()
+            .then(this.checkStoreDurability)
             .then(store => {
               store.data[prefix] = {};
               store.saveSoon();
@@ -125,6 +181,9 @@ this.Storage = {
         });
       },
     };
+
+    // Prepare this store for future durability checks.
+    storageInterface.seedStoreDurability();
 
     return Cu.cloneInto(storageInterface, sandbox, {
       cloneFunctions: true,
--- a/recipe-client-addon/lib/Storage.jsm
+++ b/recipe-client-addon/lib/Storage.jsm
@@ -15,7 +15,6 @@ this.EXPORTED_SYMBOLS = ["Storage"];
 
 const log = LogManager.getLogger("storage");
 
-const STORAGE_DURABILITY_KEY = '_storageDurability';
 let storePromise;
 
 function loadStorage() {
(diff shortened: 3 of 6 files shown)


mythmon:
> * Should non-durable storage throw errors?

In self-repair, it does by default and there is an option to override this when creating a storage object. I think we should keep this behavior for now.

> - Right now I have it set up so that when Storage is created, it seeds the durability check, and then every read/write after that it checks the durability.
>  - Does it matter if the durability is set to 0 for each new Storage? Should it retain the existing value for any reason?
>  - Is it overkill to run a durability check on each read/write? On selfrepair we check on every read, fwiw.

I think there might be some confusion here about what "durability" means in this context. Which is entirely fair. I haven't really explained it, and the code is rather vague. Durability here means that data written to the storage objects *persists after Firefox restarts*. This is specifically to prevent a bug we had where due to a bug in the storage driver, the entire country of Australia got a heartbeat prompt every time they started the browser for about a week. (This incident was all pre-Normandy).

So having independent durability for each prefix is going to be problematic. It means that the first time every heartbeat runs, it won't consider storage to be durable, and will likely fail. It means that if a recipe only runs once, it will never work.

In self-repair we *read* the durability values on every call to `getItem`, but we only update the durability when we create a new driver object, which we do once per page load of self-repair. This should probably be called by the driver, and shouldn't be available to the actions.

> - Is `store.saveSoon()` necessary to save the durability key update? I'm assuming so.

`store` is a [`JSONFile` object](https://dxr.mozilla.org/mozilla-central/source/toolkit/modules/JSONFile.jsm). It gives us a JS object that it will occasionally write to disk. Calling `saveSoon` hints that it should probably write something to disk, and we do it after every modification, to avoid losing data. It is probably a good idea to call it here too.

> - How should tests for this be written? There already exists a bunch of `browser_Storage` tests that cover storing different types of data (objects, strings, numbers, etc). 
>   - Could it be as simple as creating two stores with the same namespace, and testing if they can interact with the same values?
>   - On the self repair side, we simply set a value and then check if the store can read it - which doesn't necessarily seem to be a complete 'durability' check, but maybe I'm wrong.

The read/write tests of various types are checking that the storage works at all, not testing its durability. We have [two tests specifically about durability](https://github.com/mozilla/normandy/blob/master/recipe-server/client/selfrepair/tests/test_normandy_driver.js#L277-L295) in the self-repair driver

> - When a Storage object is created, I see a call to `Cu.cloneInto(...)` - does the durability seed need to happen after that step?

I don't think it matters either way, but we should test this.

recipe-client-addon/test/browser/browser_NormandyDriver.js
```diff
@@ -1,5 +1,38 @@
...
+  const fakeSandbox = { Promise };
+  const driver = new NormandyDriver({ sandbox: fakeSandbox });
+
+  const store = Storage.makeStorage(Storage.DURABILITY_NAMESPACE, fakeSandbox);
+  await store.setItem(Storage.DURABILITY_KEY, -1);
+  let hadError = false;
+
+  try {
+    // Create storage with the 'skipDurabilityCheck' flag set to true
+    await driver.createStorage('test', true);
```
mythmon: I don't see where `Driver::createStorage` takes a second parameter for skipDurabilityCheck. Is this test stale?

recipe-client-addon/lib/Storage.jsm
```diff
@@ -29,6 +30,34 @@ function loadStorage() {
 }
 
 this.Storage = {
+  DURABILITY_NAMESPACE: '_storageDurability',
+  DURABILITY_KEY: 'durable',
+  async seedDurability(sandbox) {
+    const globalDurabilityStore = Storage.makeStorage(this.DURABILITY_NAMESPACE, sandbox);
+    let durability = await globalDurabilityStore.getItem(this.DURABILITY_KEY);
+
+    durability = parseInt(durability, 10);
```
mythmon: The storage driver [stores simple values directly](http://normandy.readthedocs.io/en/latest/dev/driver.html#Storage) instead of converting them to strings like localstorage would. Can we assume it returns numbers instead of doing the parseInt/isNaN dance?

recipe-client-addon/lib/NormandyDriver.jsm
```diff
@@ -31,8 +31,11 @@ this.NormandyDriver = function(sandboxManager) {
   }
   const {sandbox} = sandboxManager;
 
+  Storage.seedDurability(sandbox);
```
mythmon: Unfortunately, we [create one driver instances per recipe we run](https://github.com/mozilla/normandy/blob/b589f6c3eaa37f73025b87976cc08b78d9f94ca6/recipe-client-addon/lib/RecipeRunner.jsm#L165), so this will count only the first recipe run the first time as non-durable. It would probably be good to call seedDurability from the context of RecipeRunner, so that it only runs once.

recipe-client-addon/lib/Storage.jsm
```diff
@@ -29,6 +30,34 @@ function loadStorage() {
 }
 
 this.Storage = {
+  DURABILITY_NAMESPACE: '_storageDurability',
+  DURABILITY_KEY: 'durable',
+  async seedDurability(sandbox) {
+    const globalDurabilityStore = Storage.makeStorage(this.DURABILITY_NAMESPACE, sandbox);
```
mythmon: What if an action tries to calls `makeStorage` with the same prefix? Is that something we want to guard against? Or should we just accept that a prefix named "durability" is pretty clear about what it is doing?

recipe-client-addon/lib/Storage.jsm
andymikulski: Oh, okay! That makes sense. We should be able to remove the parsing/isNaN-ing in that case.

recipe-client-addon/test/browser/browser_NormandyDriver.js
andymikulski: D'oh! I didn't commit the `NormandyDriver` file 😑  sorry!

recipe-client-addon/lib/Storage.jsm
andymikulski: Technically, if an action tried to `makeStorage('_storageDurability')` then yes - the data would collide. In my opinion I don't think that's a huge thing to worry about - but I could be wrong. Should we specify it further, something like `normandy_storageDurability`?

While I'm on the subject - I think the term `prefix` should be swapped to `namespace` in  the storage code. Though we use prefixes within the actual `Storage`s, to outside modules the term `prefix` is a little vague. It makes sense once you open up the Storage file and see that prefixes effectively namespace the data, but at first the terminology didn't click for me. 

Also, if we switch to the term `namespace`, outside modules won't have to be updated if we ever decide to change how `Storage`s internally handle the data. Working with stores then would become a little less coupled.

Just a thought!

recipe-client-addon/lib/NormandyDriver.jsm
andymikulski: > this will count only the first recipe run the first time as non-durable

I'm not sure why this is would be the case - the durability value is seeded to `1` (if none is previously set) and then the driver checks that value when it creates its Storage. In theory, the first recipe/run should have a `durability` of `1` and should run as expected (assuming the store is durable).

However, In selfrepair, the durability check looks for a value of `2` instead, which didn't really add up to me. Maybe there's something about the recipe life cycle that I'm missing.

 > It would probably be good to call seedDurability from the context of RecipeRunner

 I'm confused how this would work with regards to the sandboxes. Would we need to seed each action/sandbox in `RecipeRunner.executeAction`? Would the call to `Storage.checkDurability` then still exist in NormandyDriver?

recipe-client-addon/lib/Storage.jsm
mythmon: 👍 to both of these ideas. I think the key `_storageDurability` is unique enough and obvious enough what it is doing. I mostly wanted to make sure we were doing this on purpose, knowing the tradeoffs.

As for `prefix`/`namespace`: that's also a good idea. You're right that `prefix` is more a implementation detail, and `namespace` is really the concept we should be conveying. That change would be good in another PR, and I think we'll have to update actions as well.

recipe-client-addon/lib/NormandyDriver.jsm
mythmon: @andymikulski and I chatted in Vidyo about this. To summarize for the record, the main confusion is that when I said

> this will count only the first recipe run the first time as non-durable

What I was saying is that when storage has not proven itself to be durable, with the current code the first recipe to execute would see storage as not-proven-durable correctly, and all the other executed recipes would incorrectly see storage as durable.

The self-repair check looking for 2 was an implementation detail relating to how it handled retrieving and incrementing the number. It would have a value of 1 the first time, and never have a value of 0, so only a value of 2 would represent durability to Firefox restarting.

We also cleared up that sandboxes are annoying and confusing and complicate this a bit.

andymikulski:
@mythmon This is ready for review! I've removed the unnecessary parsing, moved the durability seeding to RecipeRunner, and bumped the 'durability value threshold' to 2, as in selfrepair. I think this should fix the issues you had pointed out re: non-durable recipes still executing.

andymikulski:
@mythmon Any idea as to why Taskcluster is failing on this? Locally, the addon tests pass fine, and looking at (what I think is) the logs, I don't really see any errors indicating the problem.

Osmose:
> Any idea as to why Taskcluster is failing on this? Locally, the addon tests pass fine, and looking at (what I think are) the logs, I don't really see any errors indicating the problem.

I think our TC builds in general might be busted, but the errors aren't consistent; I've seen three separate types of failures in the last 24 hours.

andymikulski:
@mythmon This finally passed TC, let me know what you think of the changes!

recipe-client-addon/lib/Storage.jsm
```diff
@@ -29,6 +30,35 @@ function loadStorage() {
 }
 
 this.Storage = {
+  DURABILITY_NAMESPACE: 'normandy_storageDurability',
+  DURABILITY_KEY: 'durable',
+  isDurabilityInvalid(value) {
+    return typeof value === 'undefined' || isNaN(value);
```
mythmon: The linter should be yelling about this, but Firefox internal code (which this add-on sort  of is) prefers double quotes.

recipe-client-addon/lib/Storage.jsm
```diff
@@ -29,6 +30,35 @@ function loadStorage() {
...
+
+    globalDurabilityStore.setItem(this.DURABILITY_KEY, durability + 1);
+  },
+
+  async checkDurability(sandbox) {
+      const globalDurabilityStore = Storage.makeStorage(this.DURABILITY_NAMESPACE, sandbox);
+
+      const durability = await globalDurabilityStore.getItem(this.DURABILITY_KEY)
+
+      const isDurabilityInvalid = this.isDurabilityInvalid(durability) || durability < 2;
```
mythmon: Nit: We could probably reduce the number of blank lines in this function.

mythmon:
I ran this locally by building an XPI and installing it in Firefox Nightly (55), and I got `TypeError: globalDurabilityStore.getItem is not a function   Storage.jsm:40:28`. I'm not sure what could cause that.

andymikulski:
Huh, super weird. Few things:

- The linter doesn't appear to be working for me at all locally - Therapist just marks it as [SKIPPED] (it may only be looking for `.js` files and not `.jsm`), and running it manually doesn't seem to use the local `.eslintrc` file. I'll continue to investigate.

- The `make-xpi` script was failing for me, but I found the issue and have committed the fix (a command flag was in the wrong spot)

- After finally getting the XPI to build and all that, I encountered the same error you did, however with this warning above it:

`
XrayWrapper denied access to property "getItem" (reason: value is callable). See https://developer.mozilla.org/en-US/docs/Xray_vision for more information. Note that only the first denied property access from a given global object will be reported.
`

Looking into it, it seems like a sandboxing issue, though I have no idea why it thinks that object is global (removing the 'global' part of the variable name did not do anything, either). Any thoughts?

andymikulski:
After discussing it with @Osmose , I've gone and simply added the `Cu.waiveXrays` fix. Ultimately, we should investigate removing the need for `sandbox`es to be passed to `Storage`s.

@mythmon This is ready for review

recipe-client-addon/bin/make-xpi.sh
```diff
@@ -23,7 +23,7 @@ rm "${DEST}/install.rdf.in"
 cp "${BASE_DIR}/install.rdf" "${DEST}"
 rm "${DEST}/jar.mn"
 cp "${BASE_DIR}/chrome.manifest" "${DEST}"
-rm "${DEST}/test" -r
```
mythmon: I'm guessing this broke on the BSD version of `rm`?

andymikulski: Yep, I think OSX uses BSD. Does this change conflict with the GNU implementation? Should I reverse it?

mythmon: No, this change is fine. The way you have it is easier to read anyways. It just took me a moment to realize why it was needed, so I wanted to double check.

Osmose:
@mythmon @andymikulski Did this get reviewed by @gijsk before merging? If not, we should probably revert this and get a Firefox peer to review it.

Osmose:
For context, because we probably didn't ever explain it for @andymikulski: Firefox has a [code review policy](https://wiki.mozilla.org/Firefox/Code_Review) that ties in with the [module system](https://www.mozilla.org/en-US/about/governance/policies/module-ownership/) we use for governance of Mozilla the organization (vs Mozilla the company). In particular, Firefox code (including our system add-on) must be reviewed by a peer of the Firefox module, and no one on our team is one. Luckily, @gijsk is one, has previous experience with the system SHIELD is replacing, and is an all-around great person, and graciously accepted our request to handle code reviews.

For add-on code reviews, we generally have two reviewers: @gijsk, and someone from the Normandy team. We still do the reviews on Github, though. Occasionally, @rhelmer handles reviews as well, mostly for stuff related to our use of the AddonManager in Firefox, which he maintains.

Osmose:
Also, this updates `NormandyDriver.makeStorage` to be async, but does not update the docs, the `show-heartbeat` recipe, and the self-repair driver. We should address that in a followup PR or in a new PR if we decide to revert.

gijsk:
`waiveXrays` is scary and we should try not to use it.

At a glance, it seems to me like this checks for storage durability for every sandbox we create from a driver/recipe. It seems like we should be able to have a shared state and don't need to use the sandbox for this, which would mean we wouldn't need to use waiveXrays and the overall implementation will likely be simpler. I'm happy to help look at this in more detail once you decide when/where/how to do this (not sure from the earlier comments if this will get reopened or we get a new issue or something else, also because I don't know about github flow).

Finally, I would expect more tests to fail because of this change - effectively, unless you seed durability from a test file, AIUI on the first run of Firefox (which is what tests get, because clean profile), nothing will be in storage, and (the real) init() should only be called once, so we seed durability only once, so it will always be <2, so we don't run things. So more tests should fail. I'm probably missing something, but I would like to understand what, exactly. :-)

PR #634, opened 2017-03-24
Bug 1349576: Remove use of services.sync.numClients preference.

Osmose:
It seems that sync is no longer maintaining this preference, so instead we
simply add the mobile and desktop client counts to get the same value.

### DIFF
--- a/recipe-client-addon/lib/NormandyDriver.jsm
+++ b/recipe-client-addon/lib/NormandyDriver.jsm
@@ -78,11 +78,11 @@ this.NormandyDriver = function(sandboxManager) {
         syncSetup: Preferences.isSet("services.sync.username"),
         syncDesktopDevices: Preferences.get("services.sync.clients.devices.desktop", 0),
         syncMobileDevices: Preferences.get("services.sync.clients.devices.mobile", 0),
-        syncTotalDevices: Preferences.get("services.sync.numClients", 0),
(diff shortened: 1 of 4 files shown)


recipe-client-addon/lib/NormandyDriver.jsm
```diff
@@ -78,11 +78,11 @@ this.NormandyDriver = function(sandboxManager) {
         syncSetup: Preferences.isSet("services.sync.username"),
         syncDesktopDevices: Preferences.get("services.sync.clients.devices.desktop", 0),
         syncMobileDevices: Preferences.get("services.sync.clients.devices.mobile", 0),
-        syncTotalDevices: Preferences.get("services.sync.numClients", 0),
         plugins: {},
         doNotTrack: Preferences.get("privacy.donottrackheader.enabled", false),
         distribution: Preferences.get("distribution.id", "default"),
       };
+      appinfo.syncTotalDevices = appinfo.syncDesktopDevices + appinfo.syncMobileDevices;
```
mythmon: Modifying the "shape" of an object after first creation is sub-optimal. Can we include the key in `appinfo` right away, probably by just having `syncTotalDevices: null,`?

Osmose: Fixed.

PR #641, opened 2017-03-27
recipe-client-addon: Use nsIUpdateTimerManager to run every 24 hours

### LINKED ISSUE #588: Fetch and execute recipes once every 24 hours if no restart
We should be running recipes once every 24 hours for users who haven't restarted the browser in that time. Executing recipes should reset the time, so that a user with the browser open for an entire week should fetch and execute new recipes 7 times (unless their computer is asleep or some other state that stops Firefox from running; we should default to whatever behavior the updater uses in that case).

mythmon:
This makes the add-on *not* run automatically on start up and instead run once every 24 hours. It re-uses the same timer that the update manager uses (though they aren't synced up, and in fact the timer prevents them from both running at once).

Fixes #588.

### DIFF
--- a/recipe-client-addon/lib/RecipeRunner.jsm
+++ b/recipe-client-addon/lib/RecipeRunner.jsm
@@ -4,14 +4,13 @@
 
 "use strict";
 
-const {utils: Cu, classes: Cc, interfaces: Ci} = Components;
+const {utils: Cu} = Components;
 Cu.import("resource://gre/modules/Services.jsm");
 Cu.import("resource://shield-recipe-client/lib/LogManager.jsm");
 Cu.import("resource://shield-recipe-client/lib/NormandyDriver.jsm");
 Cu.import("resource://shield-recipe-client/lib/FilterExpressions.jsm");
 Cu.import("resource://shield-recipe-client/lib/NormandyApi.jsm");
 Cu.import("resource://shield-recipe-client/lib/SandboxManager.jsm");
-Cu.import("resource://shield-recipe-client/lib/Storage.jsm");
 Cu.import("resource://shield-recipe-client/lib/ClientEnvironment.jsm");
 Cu.import("resource://shield-recipe-client/lib/CleanupManager.jsm");
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
@@ -73,6 +72,9 @@ this.RecipeRunner = {
     return true;
   },
 
+  /**
+   * Watch for preference changes from Services.pref.addObserver.
+   */
   observe(changedPrefBranch, action, changedPref) {
     if (action === "nsPref:changed" && changedPref === "run_interval_seconds") {
       this.updateRunInterval();
--- a/recipe-client-addon/lib/RecipeRunner.jsm
+++ b/recipe-client-addon/lib/RecipeRunner.jsm
@@ -4,7 +4,7 @@
 
 "use strict";
 
-const {utils: Cu} = Components;
+const {utils: Cu, classes: Cc, interfaces: Ci} = Components;
 Cu.import("resource://gre/modules/Services.jsm");
 Cu.import("resource://gre/modules/Timer.jsm"); /* globals setTimeout */
 Cu.import("resource://shield-recipe-client/lib/LogManager.jsm");
@@ -14,6 +14,7 @@ Cu.import("resource://shield-recipe-client/lib/NormandyApi.jsm");
 Cu.import("resource://shield-recipe-client/lib/SandboxManager.jsm");
 Cu.import("resource://shield-recipe-client/lib/Storage.jsm");
 Cu.import("resource://shield-recipe-client/lib/ClientEnvironment.jsm");
+Cu.import("resource://shield-recipe-client/lib/CleanupManager.jsm");
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
 Cu.importGlobalProperties(["fetch"]); /* globals fetch */
 
@@ -24,6 +25,7 @@ this.EXPORTED_SYMBOLS = ["RecipeRunner"];
 
 const log = LogManager.getLogger("recipe-runner");
 const prefs = Services.prefs.getBranch("extensions.shield-recipe-client.");
+const TIMER_NAME = "recipe-client-addon-run";
 
 this.RecipeRunner = {
   init() {
@@ -34,15 +36,25 @@ this.RecipeRunner = {
     const durabilityManager = new SandboxManager();
     Storage.seedDurability(durabilityManager.sandbox);
 
-    let delay;
     if (prefs.getBoolPref("dev_mode")) {
-      delay = 0;
-    } else {
-      // startup delay is in seconds
-      delay = prefs.getIntPref("startup_delay_seconds") * 1000;
+      // Run right now in dev mode
+      this.run();
     }
 
-    setTimeout(this.start.bind(this), delay);
+    // Run once every `runInterval` wall-clock seconds.
+    // This is managed by setting a "last ran" timestamp, and running if it is
+    // more than `runInterval` seconds ago.
+    const runInterval = prefs.getIntPref("run_interval_seconds");
+    const timerManager = Cc["@mozilla.org/updates/timer-manager;1"]
+                         .getService(Ci.nsIUpdateTimerManager);
+    timerManager.registerTimer(TIMER_NAME, () => this.run(), runInterval);
+    if (timerManager.unregisterTimer) {
+      CleanupManager.addCleanupHandler(() => timerManager.unregisterTimer(TIMER_NAME));
+    } else {
+      // Since we can't unregister the timer yet (Bug 1350471), re-register
+      // with a period so long as to never actually run
+      timerManager.registerTimer(TIMER_NAME, () => {}, 1e10); // about 300 years
+    }
   },
 
   checkPrefs() {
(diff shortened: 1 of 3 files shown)


recipe-client-addon/lib/RecipeRunner.jsm
```diff
@@ -23,22 +24,33 @@ this.EXPORTED_SYMBOLS = ["RecipeRunner"];
...
     }
 
-    setTimeout(this.start.bind(this), delay);
+    // Run once every `runInterval` wall-clock seconds.
+    // This is managed by setting a "last ran" timestamp, and running if it is
+    // more than `runInterval` seconds ago.
+    const runInterval = prefs.getIntPref("run_interval_seconds");
+    const timerManager = Cc["@mozilla.org/updates/timer-manager;1"]
+                         .getService(Ci.nsIUpdateTimerManager);
+    timerManager.registerTimer(TIMER_NAME, () => this.run(), runInterval);
```
Osmose: We should register an observer for the pref that re-registers the timer with the new delay when it changes.

recipe-client-addon/lib/RecipeRunner.jsm
```diff
@@ -23,22 +24,33 @@ this.EXPORTED_SYMBOLS = ["RecipeRunner"];
...
-      // startup delay is in seconds
-      delay = prefs.getIntPref("startup_delay_seconds") * 1000;
+      // Run right now in dev mode
+      this.run();
     }
 
-    setTimeout(this.start.bind(this), delay);
+    // Run once every `runInterval` wall-clock seconds.
+    // This is managed by setting a "last ran" timestamp, and running if it is
+    // more than `runInterval` seconds ago.
```
Osmose: Does this mean that there's some minimum delay from the timer system we're using? When I tested locally and set the timeout to 5 seconds, it didn't repeat every 5 seconds, but at seemingly semi-random but frequent intervals.

recipe-client-addon/lib/RecipeRunner.jsm
```diff
@@ -23,22 +24,33 @@ this.EXPORTED_SYMBOLS = ["RecipeRunner"];
...
+    const runInterval = prefs.getIntPref("run_interval_seconds");
+    const timerManager = Cc["@mozilla.org/updates/timer-manager;1"]
+                         .getService(Ci.nsIUpdateTimerManager);
```
Osmose: Much like how it was recommended we lazy-load modules when possible, we should probably do this for services too. It looks like [`XPCOMUtils.defineLazyServiceGetter`](https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/XPCOMUtils.jsm#defineLazyServiceGetter()) can do this for us. 

Also, can this be in the import section instead of within this function?

recipe-client-addon/lib/RecipeRunner.jsm
mythmon: It is hard to put succintly, but there [are](https://dxr.mozilla.org/mozilla-central/source/toolkit/components/timermanager/nsUpdateTimerManager.js#84-87) [several](https://dxr.mozilla.org/mozilla-central/source/toolkit/components/timermanager/nsUpdateTimerManager.js#128-130) [reasons](https://dxr.mozilla.org/mozilla-central/source/toolkit/components/timermanager/nsUpdateTimerManager.js#164-169) why the timer might not run exactly on time, or on very short intervals. I'd expect timers of about 5 minutes to work well, and any shorter would start to exhibit weird behavior.

Osmose: Cool. Mentioning that minimum number in the comment would be very helpful. :D

Osmose:
FYI https://github.com/mozilla/normandy/pull/659 includes some tests for `run` that cross over with what was implemented in this PR, and I'm fine if we defer testing those bits to the other PR.

mythmon:
I've fixed @Osmose's review comments, and added some light testing.

@rhelmer To ease integration with mozilla-central, we have Firefox peers review add-on code here on GitHub. Usually we have Gijs do this, but since you recently worked on the timer, I figured you'd be a better reviewer here. Can you take a look at this?

rhelmer:
@mythmon sure I feel comfortable with the changes here - I should be able to get the review done today or tomorrow.

recipe-client-addon/lib/RecipeRunner.jsm
```diff
@@ -34,15 +41,26 @@ this.RecipeRunner = {
...
+    }
+
+    this.updateRunInterval();
+    if (timerManager.unregisterTimer) {
+      CleanupManager.addCleanupHandler(() => timerManager.unregisterTimer(TIMER_NAME));
     } else {
-      // startup delay is in seconds
-      delay = prefs.getIntPref("startup_delay_seconds") * 1000;
+      // Since we can't unregister the timer yet (Bug 1350471), re-register
+      // with a period so long as to never actually run
```
Osmose: Yesterday we decided not to ship to 52, which was previously a problem with not handling this, so I think this is safe to remove now.

recipe-client-addon/lib/RecipeRunner.jsm
```diff
@@ -4,26 +4,33 @@
...
+XPCOMUtils.defineLazyModuleGetter(this, "Storage",
+                                  "resource://shield-recipe-client/lib/Storage.jsm");
+XPCOMUtils.defineLazyServiceGetter(this, "timerManager",
+                                   "@mozilla.org/updates/timer-manager;1",
+                                   "nsIUpdateTimerManager");
 
 this.EXPORTED_SYMBOLS = ["RecipeRunner"];
 
+const PREF_BASE_NAME = "extensions.shield-recipe-client.";
+const PREF_RUN_INTERVAL = PREF_BASE_NAME + "run_interval_seconds";
```
mythmon: I didn't know about `prefs.addObserver()`. That's convenient for us.

mythmon:
Updated with @rhelmer's feedback.

recipe-client-addon/lib/RecipeRunner.jsm
```diff
@@ -34,15 +39,17 @@ this.RecipeRunner = {
...
+      // Run right now in dev mode
+      this.run();
     }
 
-    setTimeout(this.start.bind(this), delay);
+    this.updateRunInterval();
+    CleanupManager.addCleanupHandler(() => timerManager.unregisterTimer(TIMER_NAME));
+
+    // Watch for the run interval to change, and re-register the timer with the new value
+    prefs.addObserver("run_interval_seconds", this, false);
```
Osmose: Should `init` throw if it gets called more than once?

recipe-client-addon/lib/RecipeRunner.jsm
```diff
@@ -66,7 +73,24 @@ this.RecipeRunner = {
     return true;
   },
 
-  async start() {
+  observe(changedPrefBranch, action, changedPref) {
```
Osmose: A short doc comment would be nice. I missed at first that `RecipeRunner` itself was the preference observer.

recipe-client-addon/lib/RecipeRunner.jsm
mythmon: I don't think so. We don't have that kind of guard around any of our other listeners, and it would complicate tests, which do call init multiple times. I don't think that registering observers multiple times with the exact same arguments will be a problem.

Osmose:
@rhelmer Can you take another look at this soon? #659 depends on this PR merging. Thanks!

recipe-client-addon/lib/RecipeRunner.jsm
rhelmer: The string `run_interval_seconds` is used in multiple places - this should be a `const` declared once instead. Otherwise a typo in one place will break things in a hard-to-detect way.

recipe-client-addon/test/browser/browser_RecipeRunner.js
```diff
@@ -152,17 +154,45 @@ add_task(async function testStart() {
...
+  await SpecialPowers.pushPrefEnv({set: [["extensions.shield-recipe-client.dev_mode", false]]});
+  RecipeRunner.init();
+  ok(!runStub.called, "RecipeRunner.run is not called immediately when not in dev mode");
+  ok(addCleanupHandlerStub.called, "A cleanup function is registered when not in dev mode");
+  ok(updateRunIntervalStub.called, "A timer is registered when not in dev mode");
+
+  runStub.restore();
+  addCleanupHandlerStub.restore();
+  updateRunIntervalStub.restore();
+});
```
rhelmer: Have you tested manually that the timer is working the way you want, that there aren't unexpected changes to other timers, that unregister works etc? I think changing the system clock in a VM would be enough, or maybe just fiddling with the pref.

I can't imagine why there would be a problem from looking at the timer manager code and it has tests, but it's probably something to confirm, especially if we're going to uplift this so it won't spend a lot of time on Nightly.

This shouldn't block merging the code, I think it's worth at least manually double-checking.

mythmon: I tested that the timer works the way I want by fiddling the pref. I also tested that uninstalling the add-on didn't cause any weird errors, including when the timer would have fired. I haven't tested that there aren't unexpected changes to other times, but I'd expect that to be out of scope of Normandy's test cases.

mythmon:
Fixed.

recipe-client-addon/test/browser/browser_RecipeRunner.js
rhelmer: :+1:

PR #642, opened 2017-03-28
Add normandy.recipe to filter expressions in the add-on.

Osmose:
I noticed that we missed adding `normandy.recipe` to filter expressions, partially because we never documented it in the first place. This should fix that.

### DIFF
--- a/docs/user/filter_expressions.rst
+++ b/docs/user/filter_expressions.rst
@@ -222,6 +222,21 @@ filter expressions.
 
    Boolean specifying whether the user has enabled Do Not Track.
 
+.. js:attribute:: normandy.recipe
+
+   Object containing information about the recipe being checked. Only documented
+   attributes are guaranteed to be available.
+
+   .. js:attribute:: normandy.recipe.id
+
+      Unique ID number for the recipe.
+
+   .. js:attribute:: normandy.recipe.arguments
+
+      Object containing the arguments entered for the recipe. The shape of this
+      object varies depending on the recipe, and use of this property is only
+      recommended if you are familiar with the argument schema.
+
 Transforms
 ----------
 This section describes the transforms available to filter expressions, and what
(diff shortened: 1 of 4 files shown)


docs/user/filter_expressions.rst
```diff
@@ -222,6 +222,21 @@ filter expressions.
 
    Boolean specifying whether the user has enabled Do Not Track.
 
+.. js:attribute:: normandy.recipe
+
+   Object containing information about the recipe being checked. Only documented
+   attributes are guaranteed to be available.
+
+   .. js:attribute:: normandy.recipe.id
```
gijsk: So, looking at the test, we don't *only* include this and `arguments`... from the perspective of "I can accurately predict what will happen when you load this recipe", it would make more sense to me if we only included the properties we "needed" and/or if those matched what was documented... is that difficult to do here?

Osmose: Nah I can do that. It makes it mildly more difficult in case we want to use another property of the recipe (I have one future use case of this in mind in particular, simplifying our country/locale/channel filtering), but that's partly the point.

Osmose:
Updated with feedback.

recipe-client-addon/lib/RecipeRunner.jsm
```diff
@@ -99,9 +99,14 @@ this.RecipeRunner = {
     }
   },
 
-  getFilterContext() {
+  getFilterContext(recipe) {
     return {
-      normandy: ClientEnvironment.getEnvironment(),
+      normandy: Object.assign(ClientEnvironment.getEnvironment(), {
```
andymikulski: When using `Object.assign`, I tend to assign everything to an empty object to start - e.g. `Object.assign({}, ClientEnvironment.getEnvironment(), ...)`. Realistically, it probably doesn't make a difference.

Osmose: That's a good point. `getEnvironment` returns a new object so it doesn't matter in this case, but if it didn't this'd be a problem. I'll leave it for now.

PR #650, opened 2017-03-29
Fix #586: Apply a bucket sample to preference experiments before enrolling users.

### LINKED ISSUE #586: Bucket sampling for preference experiments
This is related to #584 in that we require per-action server logic. In this case, we need to make the server compute a custom filter expression for recipes that are a certain type of action, and append it to the other filters on the recipe already.

`preference-experiment` actions will use this to filter by sampling buckets. Users will choose the number of buckets they want, and the server will compute a filter expression that:

- Splits the userbase into 100,000 buckets, and reserve even chunks of that space for each branch.
- Uses the user's unique ID and the experiment slug for the hash
- At each of those start points, matches users for X buckets, where X is defined in the recipe's arguments per-branch.

Osmose:
r?

### DIFF
--- a/recipe-client-addon/lib/NormandyDriver.jsm
+++ b/recipe-client-addon/lib/NormandyDriver.jsm
@@ -148,6 +148,7 @@ this.NormandyDriver = function(sandboxManager) {
 
     // Sampling
     ratioSample: sandboxManager.wrapAsync(Sampling.ratioSample),
+    bucketSample: sandboxManager.wrapAsync(Sampling.bucketSample),
 
     // Preference Experiment API
     preferenceExperiments: {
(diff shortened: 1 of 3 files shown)


mythmon:
Couldn't this be done by adding a `bucketSample` call in the filter expression? This allows us to make a simpler UI, but I think it is more complex overall and hard codes logic that could be expressed easily in JEXL expressions.

Osmose:
> Couldn't this be done by adding a bucketSample call in the filter expression? This allows us to make a simpler UI, but I think it is more complex overall and hard codes logic that could be expressed easily in JEXL expressions.

It could. I forgot that, since we added `ratioSample`, that the recipe doesn't require the `bucketCount` argument to do branch selection anymore.

We might want to add a simple-filter-esque field for this kind of sampling to make the UX better in the future, but for the first few preference experiments, this will be fine.

Issue #610, opened 2017-03-13
Merge peer approval feature branch


rehandalal:
The [feature branch for peer approval](https://github.com/mozilla/normandy/tree/peer-signing) has conflicts that need to be addressed.

Osmose:
Fixed in https://github.com/mozilla/normandy/pull/613.

PR #651, opened 2017-03-29
recipe-server: Add per-action validation of recipe arguments

### LINKED ISSUE #599: Recipe validation depending on its action
@gregglind wants some requirements in place for editing preference experiments:

-  Once published, users should be unable to modify the branches of a feature experiment. If they want to add new feature branches, they need to launch a new experiment.
- Feature branch slugs should be unique within an experiment.
- Experiment slugs should be unique.

To support this, we need to be able to validate recipes different depending on their action type.

mythmon:
Fixes #599

### DIFF
--- a/recipe-server/normandy/recipes/models.py
+++ b/recipe-server/normandy/recipes/models.py
@@ -1,6 +1,7 @@
 import hashlib
 import json
 import logging
+from collections import defaultdict
 
 from django.contrib.auth.models import User
 from django.core.exceptions import ImproperlyConfigured, ValidationError
@@ -9,6 +10,7 @@ from django.utils import timezone
 from django.utils.functional import cached_property
 
 from dirtyfields import DirtyFieldsMixin
+from rest_framework import serializers
 from rest_framework.reverse import reverse
 from reversion import revisions as reversion
 
@@ -324,7 +326,11 @@ class RecipeRevision(models.Model):
         return hashlib.sha256(data.encode()).hexdigest()
 
     def save(self, *args, **kwargs):
-        self.action.validate_arguments(self.arguments)
+        if self.parent:
+            old_arguments = self.parent.arguments
+        else:
+            old_arguments = None
+        self.action.validate_arguments(self.arguments, old_arguments)
 
         if not self.created:
             self.created = timezone.now()
@@ -345,13 +351,13 @@ class Action(models.Model):
 
     errors = {
         'duplicate_branch_slug': (
-            'Feature branch slugs must be unique within an experiment: "{slug}" is duplicated.'
+            'Feature branch slugs must be unique within an experiment'
         ),
         'duplicate_branch_value': (
-            'Feature branch values must be unique within an experiment: "{value}" is duplicated.'
+            'Feature branch values must be unique within an experiment'
         ),
         'duplicate_experiment_slug': (
-            'Experiment slugs must be globally unique: "{slug}" is duplicated'
+            'Experiment slugs must be globally unique'
         ),
     }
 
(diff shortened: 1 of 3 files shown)


recipe-server/normandy/recipes/models.py
```diff
@@ -341,6 +346,18 @@ class Action(models.Model):
 
     arguments_schema_json = models.TextField(default='{}', validators=[validate_json])
 
+    errors = {
+        'duplicate_branch_slug': ('Feature branch slugs must be unique within an '
+                                  'experiment: "{slug}" is duplicated.'),
```
Osmose: Do we do hanging indents in Python? We may not have explicitly decided on this. I personally prefer to avoid them, and I think it's uncommon in our codebase.

recipe-server/normandy/recipes/models.py
```diff
@@ -373,6 +390,45 @@ def save(self, *args, **kwargs):
...
+            if old_arguments is None:
+                # This is a newly created revision, so its slug should not be in the DB
+                experiment_slugs.add(arguments['slug'])
+            for recipe in experiment_recipes:
+                if recipe.arguments['slug'] in experiment_slugs:
+                    raise ValidationError(self.errors['duplicate_experiment_slug']
+                                          .format(slug=arguments['slug']))
+
+            # Once published, branches of a feature experiment cannot be modified.
+            if old_arguments is not None and arguments['branches'] != old_arguments['branches']:
```
Osmose: This doesn't check whether the experiment has been published, only that it has been saved. That may or may not be suitable for this check.

@gregglind Do we want branches to be modifiable after creating a recipe, but before it is enabled and sent out to anyone?

recipe-server/normandy/recipes/models.py
```diff
@@ -250,7 +250,7 @@ def save(self, *args, **kwargs):
         super().save(*args, **kwargs)
 
 
-class RecipeRevision(models.Model):
+class RecipeRevision(DirtyFieldsMixin, models.Model):
```
Osmose: We're not using the dirty fields stuff, so no need to include this.

recipe-server/normandy/recipes/models.py
gregglind: The use case for notifiable after creation is OH NO I MADE A TYPO.  I lean toward modifable until "attempt to send" but pick whichever is more convenient.  +1 for whichever is lazy.

mythmon: Turns out that "published" isn't a concept we have until #610 is merged. I'm going to completely take out this check, and file an issue to add it back in once we have peer-review landed.

mythmon:
Error messages now appear in the form next to the appropriate fields.

recipe-server/normandy/recipes/models.py
```diff
@@ -373,6 +393,52 @@ def save(self, *args, **kwargs):
...
+                if branch['value'] in branch_values:
+                    msg = self.errors['duplicate_branch_value']
+                    errors['branches'][i]['value'] = msg
+
+                branch_slugs.add(branch['slug'])
+                branch_values.add(branch['value'])
+
+            # Experiment slugs should be unique.
+            experiment_recipes = Recipe.objects.filter(latest_revision__action=self)
+            experiment_slugs = set()
```
Osmose: We never add the slugs that we've seen to this set, so this doesn't actually validate for conflicting slugs. I'm pretty sure I tested this on my first pass and it worked, but I can't explain why it might've worked then.

recipe-server/normandy/recipes/tests/test_models.py
```diff
@@ -44,6 +45,50 @@ def test_recipes_used_by_empty(self):
...
+            action.validate_arguments(arguments)
+        error = action.errors['duplicate_branch_value']
+        assert exc_info.value.detail == {'arguments': {'branches': {2: {'value': error}}}}
+
+    def test_validate_arguments_preference_experiments_unique_experiment_slug(self):
+        action = ActionFactory(name='preference-experiment')
+        arguments = {'slug': 'duplicate', 'branches': []}
+        RecipeFactory(action=action, arguments=arguments)
+        with pytest.raises(serializers.ValidationError) as exc_info:
+            action.validate_arguments(arguments)
```
Osmose: This test passes, but it's incorrect. We only create one recipe, so its slug has nothing to conflict. But, because we don't pass `old_arguments`, it treats the experiment as brand new, and compares itself to its own slug, which conflicts, and passes the test.

We'll probably want to have two tests, one for the newly-created case, and one for the existing-recipe case.

recipe-server/normandy/recipes/models.py
```diff
@@ -373,6 +393,52 @@ def save(self, *args, **kwargs):
         self.implementation_hash = self.compute_implementation_hash()
         super().save(update_fields=['implementation_hash'])
 
+    def validate_arguments(self, arguments, old_arguments=None):
+        """
+        Test if `arguments` follows all action-specific rules.
+
+        If there is a previous version of the arguments, they should
+        be passed as `old_arguments`. Some validation rules depend on
+        the history of the arguments.
```
Osmose: Technically, they don't. They only rely on knowing whether the recipe arguments being validated are for a new or existing recipe. Although I know the intent is to eventually use `old_arguments` in that way.

Earlier we talked about this being a method on the `Recipe` class and using the dirty mixin for getting old values for comparison, which would give you access to the info from `old_arguments` without having to pass it in. Why'd you move this to the action class instead?

mythmon: We can't use the dirty mixin because `Recipe` doesn't store any data. All the data is in revisions, which don't change over time. Given that, it made more sense for `Action` instances to validate their own arguments.

PR #655, opened 2017-03-30
taskcluster: Use custom base image for TC builds

mythmon:
This should speed up our TC build times a little


mythmon:
This is ready to be reviewed now. To summarize: this moves some CI setup into a docker image so we don't have to download a bunch of packages every time.

mythmon:
The testing is [in Taskcluster](https://tools.taskcluster.net/task-group-inspector/#/L-ntt6IjTcutBDzKtqsroA) successfully running with the new base image.

PR #658, opened 2017-04-03
Fix ordering on recipe history

rehandalal:
r?

### DIFF
--- a/recipe-server/contract-tests/test_api.py
+++ b/recipe-server/contract-tests/test_api.py
@@ -134,6 +134,6 @@ def test_recipe_history(conf, requests_session):
 
         last_date = datetime.now()
         for revision in history:
-            created = datetime.strptime(revision)
+            created = datetime.strptime(revision['date_created'], '%Y-%m-%dT%H:%M:%S.%fZ')
             assert created < last_date
             last_date = created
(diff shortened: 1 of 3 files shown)


mythmon:
Two questions:

1. Isn't there a way to make this the default sort order for the class? That might be a good idea.
2. Can we have a test for this? Maybe a contract test that fetches data from the API and asserts it is in the right order?

rehandalal:
@mythmon i added a contract test but not sold on whether i did that correctly.

recipe-server/contract-tests/test_api.py
```diff
@@ -114,3 +116,24 @@ def test_recipe_api_is_json(conf, requests_session):
...
+
+    for item in data:
+        endpoint = '/api/v1/recipe/{}/history/'.format(item['id'])
+        r = requests_session.get(conf.getoption('server') + endpoint)
+        r.raise_for_status()
+        history = r.json()
+
+        last_date = datetime.now()
+        for revision in history:
+            created = datetime.strptime(revision)
```
mythmon: I ran these tests against prod, and I got an error here:

```
$ pytest contract-tests --server https://normandy.cdn.mozilla.net -vvv --tb=short
...
contract-tests/test_api.py::test_recipe_history
contract-tests/test_api.py:137: in test_recipe_history
    created = datetime.strptime(revision)
E   TypeError: strptime() takes exactly 2 arguments (1 given)
```

rehandalal:
D'oh! Fixed ^

PR #659, opened 2017-04-04
Stop preference experiments when the server doesn't send them

Osmose:
__NOTE:__ This depends on https://github.com/mozilla/normandy/pull/641. I don't expect major changes on that PR so this is probably fine for review, but we should wait for that to land before merging this.

This _should_ be the last major refactor for preference experiments. The problem is that we want to stop any experiments that weren't reported by the server or matched to the user, as disabling a recipe simply stops sending it to users. But how do we run logic for preference experiments that don't have action code running? Or, if there's no experiments going out, how do run any action code at all?

My answer is to add pre-execution and post-execution hooks that run before any recipes are executed, as well as after all of them are. Each action can register a pre- and post-execution hook. These hooks are executed regardless of whether any recipes were run with the action, which covers stopping the last active experiment. And, if the pre-execution hook fails, we skip recipes and post-execution hooks for the action that failed, to avoid running them with potentially-invalid state.

Since that meant there were now 3 different types of callables being registered inside the sandbox, it made sense to generalize the pattern. I left the old `registerAction` API in so that we don't immediately have to rewrite our actions. 

There was also enough logic to prepping the sandbox that I split it into a new `ActionSandboxManager` class, but I kept the `SandboxManager` class as a separate base class. In my head, `SandboxManager` is our general object-oriented wrapper for sandboxes, while `ActionSandboxManager` is where Normandy-specific sandbox prep goes.

Each commit is pretty standalone, and the shift-click trick for choosing commits to include in the diff will probably come in handy too. I think the bits not from https://github.com/mozilla/normandy/pull/641 are ready for review. r?


recipe-client-addon/lib/RecipeRunner.jsm
```diff
@@ -81,6 +81,20 @@ this.RecipeRunner = {
       await ClientEnvironment.getClientClassification();
     }
 
+    // Run pre-execution hooks. If a hook fails, we don't run recipes with that
+    // action to avoid inconsistencies.
+    const actions = await NormandyApi.fetchActions();
+    for (const action of Object.values(actions)) {
+      try {
+        await this.runActionCallback(action, "preExecution");
```
gijsk: We should make very clear in the docs that preExecution hooks, if any, will run even on builds where filters don't match, as that might surprise people who didn't expect to have to deal with backward compat in the future, or with windows environments when they wrote something for mac users, or...

recipe-client-addon/lib/Utils.jsm
```diff
@@ -0,0 +1,31 @@
...
+   * > keyBy(list, "foo") == {bar: {foo: "bar"}, baz: {foo: "baz"}}
+   *
+   * @param  {Array} list
+   * @param  {String} key
+   * @return {Object}
+   */
+  keyBy(list, key) {
+    return list.reduce((map, item) => {
+      if (!(key in item)) {
+        throw new Error(`List item was missing key "${key}".`);
```
gijsk: What happens if this happens? That is, where is this error caught? There should probably be a test that that doesn't break the world.

recipe-client-addon/lib/ActionSandboxManager.jsm
```diff
@@ -0,0 +1,88 @@
...
+  constructor() {
+    super();
+
+    // Prepare the sandbox environment
+    const driver = new NormandyDriver(this);
+    this.cloneIntoGlobal("sandboxedDriver", driver, {cloneFunctions: true});
+    this.evalInSandbox(`
+      // Shim old API for registering actions
+      function registerAction(name, Action) {
+        this.ActionClass = Action;
```
gijsk: I don't see this (`ActionClass`) being used, and you removed the tests for `registerAction`, it seems? We should probably have tests that the shim works correctly.

recipe-client-addon/test/browser/browser_Storage.js
```diff
@@ -44,3 +45,24 @@ add_task(async function() {
...
+  const manager2 = new SandboxManager();
+  const store1 = Storage.makeStorage("testSandboxValueStorage", manager1.sandbox);
+  const store2 = Storage.makeStorage("testSandboxValueStorage", manager2.sandbox);
+  manager1.addGlobal("store", store1);
+  manager2.addGlobal("store", store2);
+  manager1.addHold("testing");
+  manager2.addHold("testing");
+
+  await manager1.evalInSandbox("store.setItem('foo', {foo: 'bar'});");
+  manager1.removeHold("testing");
```
gijsk: Can we assert somehow that this does in fact nuke the sandbox?

Osmose:
> Hm, I didn't realize that we were implementing the expiry system as part of a recipe. It seems like it would more properly live inside the add-on, as it's the client that should expire the recipe... is it too late to change this?

It's not too late, but I think it's better in the long run. The general rule I've been following is to prefer logic within the sandbox to logic outside of it, because we can ship updates and changes to it way faster than the system add-on, and the impact of breakage is reduced. I would've put _most_ of the preference experiment logic within the sandbox, but the need to persist event handlers past the lifetime of the recipe sandbox made that difficult.

I also think that having the logic for when to start and when to stop experiments in the same place is a better design, although there are expiry cases where we can't really do that cleanly (e.g. when the user modifies the preference).

Does that sound reasonable?

recipe-client-addon/lib/RecipeRunner.jsm
Osmose: Yes. I intend to do a documentation pass once we get the MVP for experiments ready for mozilla-central.

recipe-client-addon/lib/Utils.jsm
Osmose: It's caught when running filter expressions, and not when fetching actions. Thinking more, I think I want this to log a warning and skip instead of throwing. This PR adds graceful handling of missing actions, which is where this wouldn't catch previously, and I think running other actions if only one of them is malformed from the server is better.

gijsk:
It does sound reasonable. I guess I'm wary because of some of the issues we've had in the past where the restrictions of recipes / self-support meant that it was tricky to ensure things do/don't run in particular situations, and at a very high level this seems vaguely similar-ish.

I suppose the difference is that we're actively maintaining the side pulling in recipes here, and so if we do run into limitations we can address those in the add-on, and with go-faster it should be possible to do a reasonably quick roll-out.

mythmon:
Reading through this, it seems odd to create `ActionSandboxManager` instances for every stage of every action. I expected that one `ActionSandboxManager` would be creating for each action, and then that would be kept and re-used for all of the stages. Something like this (though I've left out all the error and sanity checking)

```
let actionManagers = actions.map(action => new ActionManager(action))
for (const manager of actionManagers) {
  manager.runActionMethod('preExecution')
}
for (const manager of actionManagers) {
  manager.runActionMethod('action')
}
for (const manager of actionManagers) {
  manager.runActionMethod('postExecution')
}
```

This has less object churn, and it lets actions keep state in memory between stages of execution (maybe that isn't a good thing though?). It might also let us move more of the action execution management into the manager, which may simplify things. For example, if a stage fails, the manager would just refuse to run any further actions, centralizing the error handling.

What do you think? Do you prefer the short lived action managers?

Osmose:
> it lets actions keep state in memory between stages of execution (maybe that isn't a good thing though?).

That was why I didn't, but I'm starting to doubt that. Now that I think about it, moving the persistence to the storage object instead of, say, persisting stuff in the sandbox, doesn't actually strongly prevent actions from messing with eachother, and maybe the likelihood of one recipe accidentally breaking things for other recipes isn't as high as I thought.

The error handling improvements are pretty compelling. I'd love for `run` to be simpler.

I'll ponder it over lunch and figure out if I want to update the PR.

Osmose:
Updated with the suggested refactor. I ended up not moving the "disable if pre-execution-hook failed" logic to `ActionSandboxManager` because it is tied to a specific kind of async handler (individual recipes can fail without stopping other recipes or the post-execution-hook) and fits better with `RecipeRunner`'s purpose anyway. Reusing the managers should help with memory/performance, and I like the new API for `ActionSandboxManager` anyway.

recipe-client-addon/test/unit/test_ActionSandboxManager.js
```diff
@@ -2,79 +2,101 @@
...
       return "log" in normandy;
     });
-  `, "testCallback");
-  ok(logInDriver, "runAsyncCallbackFromScript passes a driver as the first parameter");
-}));
+  `;
+
+  await withManager(script, async manager => {
+    const logInDriver = await manager.runAsyncCallback("testCallback");
+    ok(logInDriver, "runAsyncCallback passes a driver as the first parameter");
```
gijsk: Wonder if we can assert something stronger here, as pretty much anything would pass this check, not just the driver...

Osmose: Only other thing I can think of that would work with the driver (since it's cloned into the sandbox) is some special attribute like `_isNormandyDriver` or something. Thoughts?

gijsk: You can clone it back, right? Then you can check that it has driver-y methods, and/or potentially (depends on structured clone) whether its `obj.constructor.name` matches.

Osmose:
- The TaskCluster build failure seemed intermittent, so I needed to bump the build.
- This PR had merge conflicts, so I rebased and fixed them.
- Rebasing on the latest preference-experiments branch ended up mixing async-style and generator-style tests in the PreferenceExperiments.jsm tests, so I did the conversion over to async. This should've been done here anyway, because @mythmon found out that the mixup between async and generator tests was causing test failures to be ignored.

The end result is that:

1. There's a new commit with the async test refactor and one new test.
2. #681 is probably going to conflict after this merges since I did the async refactor in this PR.

Thank you for your patience!

Osmose:
Actually, I'm removing the review requests, let's wait until #641 lands before we do another review cycle on this.

Osmose:
Okay, review requests put in. One last time!

Also, we decided earlier today that, for preference experiment changes, we're going to merge regardless of Taskcluster results, and will depend on the try push before we sync the changes to mozilla-central until we fix the flakyness of our TC builds.

PR #660, opened 2017-04-04
Fix #612: Add ability to filter recipes by active and expired experiments.

### LINKED ISSUE #612: Experiments in filter expressions
#592 is adding native support for preference experiments. Along with that, we'll need the ability to access active and inactive experiments in filter expressions to target recipes based on the experiments that a user has participated in.

Osmose:
What's this? Only 107 additions?

### DIFF
--- a/docs/user/filter_expressions.rst
+++ b/docs/user/filter_expressions.rst
@@ -222,6 +222,39 @@ filter expressions.
 
    Boolean specifying whether the user has enabled Do Not Track.
 
+.. js:attribute:: normandy.experiments
+
+   Object with several arrays containing the unique slugs for experiments that
+   the user has participated in. Currently, this is limited to preference
+   experiments.
+
+   .. js:attribute:: normandy.experiments.all
+
+      Array of experiment slugs for every experiment that the user has  enrolled
+      in, whether currently active or expired.
+
+   .. js:attribute:: normandy.experiments.active
+
+      Array of experiment slugs for active experiments that the user is enrolled
+      in.
+
+   .. js:attribute:: normandy.experiments.expired
+
+      Array of experiment slugs for expired experiments that the user has
+      enrolled in.
+
+   .. code-block:: javascript
+
+      // Target clients that have ever participated in the "australis"
+      // experiment, including clients that are currently running it
+      "australis" in normandy.experiments.all
+
+      // Target clients that are currently running the "quantum" experiment
+      "quantum" in normandy.experiments.active
+
+      // Target clients that ran the "photon" experiment, and have finished it
+      "photon" in normandy.experiments.expired
+
 Transforms
 ----------
 This section describes the transforms available to filter expressions, and what
--- a/recipe-client-addon/lib/ClientEnvironment.jsm
+++ b/recipe-client-addon/lib/ClientEnvironment.jsm
@@ -180,12 +180,18 @@ this.ClientEnvironment = {
     });
 
     XPCOMUtils.defineLazyGetter(environment, "experiments", async () => {
-      const experiments = await PreferenceExperiments.getAll();
-      return {
-        all: experiments.map(e => e.name),
-        active: experiments.filter(e => !e.expired).map(e => e.name),
-        expired: experiments.filter(e => e.expired).map(e => e.name),
-      };
+      const names = {all: [], active: [], expired: []};
+
+      for (const experiment of await PreferenceExperiments.getAll()) {
+        names.all.push(experiment.name);
+        if (experiment.expired) {
+          names.expired.push(experiment.name);
+        } else {
+          names.active.push(experiment.name);
+        }
+      }
+
+      return names;
     });
 
     return environment;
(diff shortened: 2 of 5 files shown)


docs/user/filter_expressions.rst
```diff
@@ -222,6 +222,39 @@ filter expressions.
...
+   .. js:attribute:: normandy.experiments.expired
+
+      Array of experiment slugs for expired experiments that the user has
+      enrolled in.
+
+   .. code-block:: javascript
+
+      // Target clients that have ever participated in the "australis"
+      // experiment, including clients that are currently running it
+      "australis" in normandy.experiments.all
```
gijsk: You can't do this in JS, ie:

```js
"foo" in ["foo"]
```

returns false.

Also, because order isn't important here and there should be no duplicates, technically Sets would be better data structures to use here. However, I don't know if jexl can deal with that. If so, that would be the best, IMO. Otherwise, I guess these examples should say:

```js
normandy.experiments.all.includes("australis")
```
etc.

(for sets, substitute "has" for "includes").

gijsk: Though, in fact, I don't know if jexl supports ES6 for that, either... or if the 'in' syntax is jexl specific?

recipe-client-addon/lib/ClientEnvironment.jsm
```diff
@@ -174,6 +179,15 @@ this.ClientEnvironment = {
       return Preferences.get("privacy.donottrackheader.enabled", false);
     });
 
+    XPCOMUtils.defineLazyGetter(environment, "experiments", async () => {
+      const experiments = await PreferenceExperiments.getAll();
+      return {
+        all: experiments.map(e => e.name),
+        active: experiments.filter(e => !e.expired).map(e => e.name),
+        expired: experiments.filter(e => e.expired).map(e => e.name),
```
gijsk: Technically, it'd be more efficient to just have 1 for loop and add the names to the right arrays. Meh. Up to you whether you do something about this, I guess.

recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -285,6 +285,15 @@ this.PreferenceExperiments = {
   },
 
   /**
+   * Get a list of all stored experiment objects.
+   * @resolves {Experiment[]}
+   */
+  async getAll() {
+    const store = await ensureStorage();
+    return Object.values(store.data).map(experiment => Object.assign({}, experiment));
```
gijsk: Is the Object.assign stuff just so the original data isn't touched? This only works one level deep... so I wonder if using `cloneInto` would be more future-proof here.

docs/user/filter_expressions.rst
mythmon: JEXL has special support for the `in` operator, and it does what is intended here:

```js
> const jexl = require('jexl')
undefined
> jexl.eval('"foo" in ["foo"]').then(res => console.log(res))
Promise { <pending> }
true
```

Additionally, JEXL doesn't support function or method calls, like `normandy.experiments.all.includes("australis")`.

Osmose: We _could_ add JEXL transforms that call Set methods, and have something like:

```js
normandy.experiments.all|includes("australis")
```

But that would stand out as weird compared to all the other APIs in filter expressions, which use lists and work with the `in` operator. 

The filter expression doesn't really care about enforcing uniqueness or lack of order in the list, so maybe `PreferenceExperiments.getAll()` should return a set, and then the code in `ClientEnvironment` should convert that to a list?

Although it'd be nice to patch JEXL to support `in` for sets. We'd just have to add it here: https://github.com/TechnologyAdvice/Jexl/blob/master/lib/grammar.js#L53

recipe-client-addon/lib/PreferenceExperiments.jsm
Osmose: It is, I thought I had a comment here. Must've forgot, whoops!

The typedef for experiment objects at the top explicitly calls out the one-level-deep requirement, so it's better than not being documented. We don't use it, but I also like that cloning with `Object.assign` is possible within sandboxes as well as outside of them, vs `cloneInto`, which would require writing a deep copy for cloning outside of chrome code.

I don't really feel strongly one way or another. You think we might as well use `cloneInto`?

gijsk: If we've documented this is supposed to be one-level deep, then I'm happy with the current state, though a comment here saying we're returning a fresh copy to avoid consumers altering our internal state would still be nice. :-)

Osmose:
Feedback'd. RE: the `in` operator, I think adding set support to JEXL is the right answer, but I want to punt on that for this PR. I think in the medium term we're on a path to fork JEXL, given that

1. It seems relatively unmaintained.
2. I'm already maintaining a Python implementation that we're using.
3. JEXL has some weird edge cases that we don't like, like filter expressions (not the expressions that Normandy use) or implicitly using undefined for attribute values, but not when using square brackets.
4. We have more changes we'd like to make.

recipe-client-addon/lib/ClientEnvironment.jsm
```diff
@@ -174,6 +179,21 @@ this.ClientEnvironment = {
       return Preferences.get("privacy.donottrackheader.enabled", false);
     });
 
+    XPCOMUtils.defineLazyGetter(environment, "experiments", async () => {
```
mythmon: These lazy getters run once on first use, and then get cached forever, yeah? What happens when we run every 24 hours, do these lists never update? Maybe these should be normal getters, instead of lazy getters?

Osmose: These are being set on the return value of `getEnvironment`, which is called once per filter expression we evaluate (we actually should probably optimize that to run once per call to `run` instead). So they aren't shared between filter expression evals, nor between calls to `run` every 24 hours.

mythmon: Awesome

PR #661, opened 2017-04-04
#351 Add Preference filter expressions

### LINKED ISSUE #351: Preferences in filter expressions
There's interest in being able to target recipes by preference values. The syntax of this would probably look something like:

```js
normandy.prefs['xpinstall.signatures.required'] == false
```

@mythmon Thoughts on the syntax?

@MattGrimes What are the use cases here? I suspect we'll need privacy approval?

andymikulski:
Fixes #351 

- Adds `preferenceValue`, `preferenceIsUserSet`, `preferenceExists` filter expressions to JEXL
- Adds relevant tests for new filters

This was suspiciously easy, so I have a feeling I'm missing something, but this is ready for feedback!

### DIFF
--- a/recipe-client-addon/test/browser/browser_FilterExpressions.js
+++ b/recipe-client-addon/test/browser/browser_FilterExpressions.js
@@ -1,6 +1,5 @@
 "use strict";
 
-Cu.import("resource://gre/modules/Preferences.jsm");
 Cu.import("resource://shield-recipe-client/lib/FilterExpressions.jsm", this);
 
 // Basic JEXL tests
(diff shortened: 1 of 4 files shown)


recipe-client-addon/test/browser/browser_FilterExpressions.js
```diff
@@ -47,8 +61,32 @@ add_task(async function() {
...
+});
 
-  // Test that it reads from the context correctly.
-  val = await FilterExpressions.eval("first + second + 3", {first: 1, second: 2});
-  is(val, 6, "context is available to filter expressions");
+// Preference tests
+add_task(async function() {
+  let val;
+  // Compare the value of the preference
+  Preferences.set("normandy.test.value", 3);
```
mythmon: This should use [`SpecialPowers`](https://github.com/mozilla/normandy/blob/6e96cacba67546909c37c7e574c694042b0267d0/recipe-client-addon/test/browser/browser_ClientEnvironment.js#L31) to set the preference, which makes sure the preference gets cleaned up after the test.

mythmon:
The Taskcluster failure here is unrelated; it looks like there was some glitch pulling the commit from GitHub. It should go away on future runs.

PS: To test actually running recipes on the system-addon, you'll need to make sure to have Autograph set up to sign recipes. The easiest way to do this is with docker-compose the `compose` directory.

Osmose:
Swapping myself out for @gijsk as a review, since all add-on reviews need a Firefox peer to review them.

recipe-client-addon/test/browser/browser_FilterExpressions.js
```diff
@@ -1,7 +1,9 @@
 "use strict";
 
+Cu.import("resource://gre/modules/Preferences.jsm");
```
gijsk: You don't seem to use this import here?

recipe-client-addon/test/browser/browser_FilterExpressions.js
```diff
@@ -47,8 +61,32 @@ add_task(async function() {
...
+  // Compare if the preference is user set
+  val = await FilterExpressions.eval('"normandy.test.isSet"|preferenceIsUserSet != true');
+  is(val, true, "preferenceIsUserSet expression determines if preference is even set");
+
+  val = await FilterExpressions.eval('"normandy.test.value"|preferenceIsUserSet == true');
+  is(val, true, "preferenceIsUserSet expression determines if user set preference themselves");
+
+  // Compare if the preference has _any_ value, whether it's user-set or default,
+  val = await FilterExpressions.eval('"normandy.test.nonexistant"|preferenceExists == false');
+  is(val, true, "preferenceExists expression determines if preference exists at all");
```
gijsk: Every single one of the filter expressions in this task succeeds/passes/returns true. We should have corresponding tests for the cases where they don't.

recipe-client-addon/test/browser/browser_FilterExpressions.js
andymikulski: D'oh, forgot to take that out after mythmon's review!

recipe-client-addon/test/browser/browser_FilterExpressions.js
andymikulski: I've added some 'negative' tests for when return values would be false. Let me know if you think theses tests suffice, or if there's a particular case I'm missing!

gijsk: Oh, bother, I meant to make another comment - can you use `ok(val)` and `ok(!val)` instead of `is(val, true)` and `is(val, false)`? :-)

PR #665, opened 2017-04-05
#663 Fix revision page buttons

### LINKED ISSUE #663: Title bar buttons link to the wrong URL on revision pages
Bugzilla: [Bug 1353747](https://bugzilla.mozilla.org/show_bug.cgi?id=1353747)
Example: https://normandy-admin.stage.mozaws.net/control/recipe/65/revision/3e2bfd54e13d3a760f1ac335c82fd55e162363f6d42a3e268d87603de313fef3

The title bar buttons (Clone, Preview, History) all link to URLs without the recipe ID in them (e.g. https://normandy-admin.stage.mozaws.net/control/recipe/preview/), which has the effect of requiring them to be clicked twice to work. This only applies to the revision-specific pages; recipe pages (https://normandy-admin.stage.mozaws.net/control/recipe/65/) function as expected.

andymikulski:
Fixes #663 

- Adds required trailing slash on revision pages (fixes navbar buttons to clone/history/etc)
- Noticed the 'cloning' page had a couple elements showing up that shouldn't be there (draft/enabled statuses), so I added `isCloning` checks when displaying those elements

I'll dig into adding a functional test or two for this.

Ready for feedback/review!


functional_tests/tests/test_basic.py
```diff
@@ -174,6 +177,55 @@ def test_create_recipe_show_heartbeat(selenium):
...
+    form_page.thanks_message = 'Thanks!'
+    form_page.form.submit()
+
+    # Wait until we navigate away from the new recipe page
+    wait_for(selenium, lambda driver: '/control/recipe/new/' not in driver.current_url)
+
+    # Grab the 'base' URL with the recipe id
+    recipe_url = selenium.current_url
+
+    # Make an edit and save it to createa  new revision
```
mythmon: Typo: "createa  new"

functional_tests/tests/test_basic.py
```diff
@@ -174,6 +177,55 @@ def test_create_recipe_show_heartbeat(selenium):
     assert len(recipe_list_page.recipe_rows) == 0
 
 
+def test_revision_page_navbar(selenium):
+    """Test the navbar buttons on a recipe revision page."""
```
mythmon: This description is pretty vague. Can it (and the maybe the title) be changed to describe specifically what is being tested?

functional_tests/tests/test_basic.py
```diff
@@ -174,6 +177,55 @@ def test_create_recipe_show_heartbeat(selenium):
...
+    # Wait for the page to change to the revision URL
+    wait_for(selenium, lambda driver: recipe_url not in driver.current_url)
+
+    # Click the clone button and ensure it went to the right place
+    form_page.wait_for_element('clone_button')
+    form_page.clone_button.click()
+
+    clone_url = recipe_url + 'clone/'
+
+    assert clone_url in selenium.current_url
```
mythmon: Can you do an exact comparison here, instead of using `in`? That is, `assert clone_url == selenium.current_url`

recipe-server/client/control/routes.js
```diff
@@ -36,7 +36,7 @@ export default (
           ]}
         />
         <Route
-          path="revision/:revisionId"
+          path="revision/:revisionId/"
```
mythmon: How does this fix the bug that we have?

andymikulski: I noticed the bug only appeared when there was _not_ a trailing slash in the URL. Every URL on the site is set up with a trailing slash, and our `absolute-path` util was resolving incorrectly as a result.

I've corrected the route here, as well as another file which was pushing to a non-slash'd route, in order to maintain conformity across the site. However, since the issue actually lied within the `absolute-path` util, the real remedy was to either fix the util or change it out with something else.

After looking for alternative npm packages we could use, I realized we could just use the `resolve` method from node's `path` module! I've swapped that in and deleted the `absolute-path` util all together.

andymikulski: Oh, turns out `resolve` strips trailing slashes - fun fact!

mythmon:
I think this only fixed part of the problem: We now correctly go to the clone page on the first click, instead of requiring two clicks, but the clone page still doesn't include a revision id, like I would expect. This has the affect of always cloning the latest revision. If you go to an older revision, and click the clone button, you get the latest revision's data in the fields. Is this the expected behavior?

andymikulski:
Ahhh, I see.

As far as I'm aware, the intended functionality was to have `clone` work from the latest revision, and that's what's currently implemented. In theory, adding the ability to clone from a revision should be trivial... in theory.

Unless @Osmose or @rehandalal think it's unnecessary, I can add the ability to clone from a revision.

Osmose:
I think what a user would expect is that the clone button clones what they see in front of their eyes, and that's what we should do.

andymikulski:
Update: I've successfully added cloning from a revision, however it seems there's a weird state tree issue I'm running into. After cloning a revision, the recipe form cycles between a previous revision's form values and the new clone. Seeing that, I decided to focus on preference experiment addon stuff instead of spinning my wheels.

rehandalal:
I'm gonna close this for now pending the refactor.

PR #666, opened 2017-04-05
recipe-client-addon: Separate build and test in Taskcluster CI

mythmon:
This will make it easier to distinguish random build failures from actual problems with the tests. Hopefully this will also help us debug our builds in the future.

### DIFF
--- a/ci/taskcluster/bin/decision.py
+++ b/ci/taskcluster/bin/decision.py
@@ -7,15 +7,23 @@ from taskcluster.utils import stableSlugId, dumpJson
 
 tasks = [
     {
+        'name': 'recipe-client-addon:build',
+        'description': 'Build Firefox with recipe-client-addon',
+        'command': 'normandy/recipe-client-addon/bin/tc/build.sh',
+    },
+    {
         'name': 'recipe-client-addon:test',
         'description': 'Test recipe-client-addon with gecko-dev',
         'command': 'normandy/recipe-client-addon/bin/tc/test.sh',
+        'dependencies': ['recipe-client-addon:build'],
+        'artifacts_from': {
+            'BUILD_RESULT': {'task': 'recipe-client-addon:build', 'path': 'build.tar.gz'},
+        },
     },
     {
         'name': 'recipe-client-addon:make-xpi',
         'description': 'Build XPI for recipe-client-addon',
         'command': 'normandy/recipe-client-addon/bin/tc/make-xpi.sh',
-        'dependencies': ['recipe-client-addon:test'],
     },
 ]
 
@@ -29,6 +37,7 @@ def main():
     decisionTaskId = os.environ['TASK_ID']
     owner = os.environ['GITHUB_HEAD_USER_EMAIL']
     source = os.environ['GITHUB_HEAD_REPO_URL']
+    artifact_url_pattern = 'http://taskcluster/queue/v1/task/{}/artifacts/{}'
 
     with requests.Session() as session:
         for task in tasks:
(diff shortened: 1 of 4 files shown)


mythmon:
@Osmose the bash changes are pretty light. Mostly I copied `test.sh` to `build.sh`, and made them only do one part of the process.

ci/taskcluster/bin/decision.py
```diff
@@ -7,15 +7,23 @@
...
+        'dependencies': ['recipe-client-addon:build'],
+        'artifacts_from': {
+            'BUILD_RESULT': {'task': 'recipe-client-addon:build', 'path': 'build.tar.gz'},
+        },
     },
     {
         'name': 'recipe-client-addon:make-xpi',
         'description': 'Build XPI for recipe-client-addon',
         'command': 'normandy/recipe-client-addon/bin/tc/make-xpi.sh',
-        'dependencies': ['recipe-client-addon:test'],
```
mythmon: We don't technically need the tests to pass before making an XPI, and this may serve as an interesting debugging tool. I'd be happy changing this to depend on the build, tests, or both.

ci/taskcluster/bin/decision.py
```diff
@@ -7,15 +7,23 @@
...
+        'name': 'recipe-client-addon:build',
+        'description': 'Build Firefox with recipe-client-addon',
+        'command': 'normandy/recipe-client-addon/bin/tc/build.sh',
+    },
+    {
         'name': 'recipe-client-addon:test',
         'description': 'Test recipe-client-addon with gecko-dev',
         'command': 'normandy/recipe-client-addon/bin/tc/test.sh',
+        'dependencies': ['recipe-client-addon:build'],
+        'artifacts_from': {
```
Osmose: The `_from` at the end is kinda confusing for a mapping; I couldn't figure out what the key of the mapping was at first. Maybe name this `artifacts`, or use a list with `envvar` keys instead? This is a nitpick though.

ci/taskcluster/bin/decision.py
```diff
@@ -7,15 +7,23 @@
...
+            'BUILD_RESULT': {'task': 'recipe-client-addon:build', 'path': 'build.tar.gz'},
```
Osmose: `taskName` is a better name for this key, I think. I was confused as to why the `task` attribute could be passed to the ID maker.

recipe-client-addon/bin/tc/test.sh
```diff
@@ -4,20 +4,14 @@ set -eu
...
 
-# Creates gecko-dev-master
-echo 'Downloading gecko-dev...'
-curl -sL https://github.com/mozilla/gecko-dev/archive/master.tar.gz | tar xz
-
-pushd normandy/recipe-client-addon
-npm install
-./bin/update-mozilla-central.sh ../gecko-dev/
-popd
+# creates gecko-dev-master
```
Osmose: Instead of creates, How about "Fetches"?

recipe-client-addon/bin/tc/test.sh
```diff
@@ -4,20 +4,14 @@ set -eu
...
+echo 'Downloading build result'
+curl -sfL "$BUILD_RESULT" | tar xz
```
Osmose: Log messages declaring that we're running lints and tests like the other ones you added would be neat.

ci/taskcluster/bin/decision.py
mythmon: I'm not sure about this name. I started with `artifacts`, but it didn't sit right. To me `artifacts` would mean the artifacts this task exports. The actual intent of this is to declare what artifacts need to be imported from other tasks, and bind them to convenient names.

Making this a list with `envvar` keys feels good. What do you think about this?

```python
{
    'name': 'recipe-client-addon:test',
    ...
    'artifacts_from': [
        {
            'task_name': 'recipe-client-addon:build',
            'path': 'build.tar.gz',
            'envvar': 'BUILD_RESULT',
        },
    ],
}
```

`artifacts_from` is still a weird name, but I like it better than `artifacts`. I'd be happy to switch to something more apt.

Osmose: Yeah, that looks good to me.

Issue #645, opened 2017-03-28
Check for changed preference on startup


Osmose:
On startup, the add-on should, for each active preference experiment:

- Check if the preference changed from the experiment value. If it did, expire the experiment, but do not modify the preference back to its old value. 
- Register the preference observer that monitors for changes to the preference.

Osmose:
Fixed by #684.

PR #670, opened 2017-04-07
Fix #652: Add option for setting user or default prefs for experiments.

### LINKED ISSUE #652: Option to change default preference value for preference experiment
Currently, preference experiments write new values to preferences that are being experimented on. Historically, Firefox has had issues with experiments or fixes that write user preferences, because they persist between versions and are hard to detect vs user preference changes that we want to respect.

By default, preference experiments should be modifying the default value for preferences that are being tested. Because changes to defaults do not persist between sessions, the add-on will have to modify these values at startup as well.

We will still need the option to change the user value for a preference for things that read preferences before the add-on finishes initialization, as is probably likely for Quantum features in Firefox. This can be a toggle/option input in the admin interface, but should have a warning of some kind to make sure users know the risks of running a preference experiment that sets the user value of a preference vs the default value.

Osmose:
In which I continue to sidestep rewriting the entire PreferenceExperiments test suite to use async/await.

It turns out preference observers react to changes in the effective preference value, which covers default value changes, which helped cut down on a bunch of the code and tests I originally had here.

### DIFF
--- a/recipe-client-addon/bootstrap.js
+++ b/recipe-client-addon/bootstrap.js
@@ -45,6 +45,8 @@ const PREF_LOGGING_LEVEL = PREF_BRANCH + "logging.level";
 let shouldRun = true;
 let log = null;
 
+this.EXPORTED_SYMBOLS = ["install", "startup", "shutdown", "uninstall"];
+
 this.install = function() {
   // Self Repair only checks its pref on start, so if we disable it, wait until
   // next startup to run, unless the dev_mode preference is set.
@@ -56,7 +58,7 @@ this.install = function() {
   }
 };
 
-this.startup = function() {
+this.startup = async function() {
   setDefaultPrefs();
 
   // Setup logging and listen for changes to logging prefs
@@ -72,11 +74,13 @@ this.startup = function() {
 
   // Initialize experiments first to avoid a race between initializing prefs
   // and recipes rolling back pref changes when experiments end.
-  PreferenceExperiments.init().catch(err => {
+  try {
+    await PreferenceExperiments.init();
+  } catch (err) {
     log.error("Failed to initialize preference experiments:", err);
-  }).then(() => {
-    RecipeRunner.init();
-  });
+  }
+
+  await RecipeRunner.init();
 };
 
 this.shutdown = function(data, reason) {
--- a/recipe-client-addon/bootstrap.js
+++ b/recipe-client-addon/bootstrap.js
@@ -15,6 +15,8 @@ XPCOMUtils.defineLazyModuleGetter(this, "RecipeRunner",
   "resource://shield-recipe-client/lib/RecipeRunner.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "CleanupManager",
   "resource://shield-recipe-client/lib/CleanupManager.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "PreferenceExperiments",
+  "resource://shield-recipe-client/lib/PreferenceExperiments.jsm");
 
 const REASONS = {
   APP_STARTUP: 1,      // The application is starting up.
@@ -68,7 +70,13 @@ this.startup = function() {
     return;
   }
 
-  RecipeRunner.init();
+  // Initialize experiments first to avoid a race between initializing prefs
+  // and recipes rolling back pref changes when experiments end.
+  PreferenceExperiments.init().catch(err => {
+    log.error("Failed to initialize preference experiments:", err);
+  }).then(() => {
+    RecipeRunner.init();
+  });
 };
 
 this.shutdown = function(data, reason) {
--- a/recipe-client-addon/bootstrap.js
+++ b/recipe-client-addon/bootstrap.js
@@ -45,6 +45,7 @@ const PREF_LOGGING_LEVEL = PREF_BRANCH + "logging.level";
 let shouldRun = true;
 let log = null;
 
+// These are exported for testing purposes only.
 this.EXPORTED_SYMBOLS = ["install", "startup", "shutdown", "uninstall"];
 
 this.install = function() {
--- a/recipe-client-addon/lib/PreferenceExperiments.jsm
+++ b/recipe-client-addon/lib/PreferenceExperiments.jsm
@@ -37,6 +37,13 @@
  * @property {string|integer|boolean|undefined} previousPreferenceValue
  *   Value of the preference prior to the experiment, or undefined if it was
  *   unset.
+ * @property {PreferenceBranchType} preferenceBranchType
+ *   Controls how we modify the preference to affect the client.
+ *
+ *   If "default", when the experiment is active, the default value for the
+ *   preference is modified on startup of the add-on. If "user", the user value
+ *   for the preference is modified when the experiment starts, and is reset to
+ *   its original value when the experiment ends.
  */
 
 "use strict";
@@ -53,6 +60,16 @@ XPCOMUtils.defineLazyModuleGetter(this, "Preferences", "resource://gre/modules/P
 this.EXPORTED_SYMBOLS = ["PreferenceExperiments"];
 
 const EXPERIMENT_FILE = "shield-preference-experiments.json";
+const DefaultPreferences = new Preferences({defaultBranch: true});
+
+/**
+ * Enum storing Preference modules for each type of preference branch.
+ * @enum {Object}
+ */
+const PreferenceBranchType = {
+  user: Preferences,
+  default: DefaultPreferences,
+};
 
 /**
  * Asynchronously load the JSON file that stores experiment status in the profile.
@@ -74,6 +91,20 @@ const experimentObservers = new Map();
 CleanupManager.addCleanupHandler(() => PreferenceExperiments.stopAllObservers());
 
 this.PreferenceExperiments = {
+  /**
+   * Set the default preference value for active experiments that use the
+   * default preference branch.
+   */
+  async init() {
+    const store = await ensureStorage();
+    for (const experiment of Object.values(store.data)) {
+      const {expired, preferenceBranchType, preferenceName, preferenceValue} = experiment;
+      if (!expired && preferenceBranchType === "default") {
+        DefaultPreferences.set(preferenceName, preferenceValue);
+      }
+    }
+  },
+
   /**
    * Test wrapper that temporarily replaces the stored experiment data with fake
    * data for testing.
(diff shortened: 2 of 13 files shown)


recipe-client-addon/bootstrap.js
```diff
@@ -68,7 +70,13 @@ this.startup = function() {
   CleanupManager.addCleanupHandler(
     () => Preferences.ignore(PREF_LOGGING_LEVEL, LogManager.configure));
 
-  RecipeRunner.init();
+  // Initialize experiments first to avoid a race between initializing prefs
+  // and recipes rolling back pref changes when experiments end.
+  PreferenceExperiments.init().catch(err => {
+    log.error("Failed to initialize preference experiments:", err);
```
Osmose: I would've liked to test that `PreferenceExperiments.init` failing does not stop `RecipeRunner.init` from running, but I dunno how to test bootstrap stuff.

recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -254,11 +292,15 @@ this.PreferenceExperiments = {
...
+      const {preferenceName, previousPreferenceValue, preferenceBranchType} = experiment;
+      const preferences = PreferenceBranchType[preferenceBranchType];
       if (previousPreferenceValue !== undefined) {
-        Preferences.set(preferenceName, previousPreferenceValue);
+        preferences.set(preferenceName, previousPreferenceValue);
       } else {
-        Preferences.reset(preferenceName);
+        // This does nothing if we're on the default branch, which is fine. The
+        // preference will be reset on next restart, and most preferences should
+        // have had a default value set before the experiment anyway.
```
Osmose: See also [bsmedberg's email on dev-shield](https://mail.mozilla.org/pipermail/dev-shield/2017-April/000161.html).

gijsk:
General note, that might be worth bringing up on the mailing list: code can (and does) check whether the user has "modified" preferences (ie whether a pref has a value on the user branch, using the pref service's `prefHasUserValue` method) as a short cut in some cases. Changing the default value with an experiment in that case would in such cases potentially not yield the result you expect. I don't expect this to be a serious problem in practice, but it seems like it's worth flagging up for QE testing things or when coordinating with devs as something to look out for, as behaviour could potentially be different when comparing the preference experiment with QE setting the pref in about:config.

recipe-client-addon/bootstrap.js
gijsk: Yeah, I don't know how to do that either. I guess you could write a browser test that disables and re-enables the add-on (using add-on manager APIs)? I think that'd run this code, maybe?

recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -75,6 +92,20 @@ CleanupManager.addCleanupHandler(() => PreferenceExperiments.stopAllObservers())
...
 this.PreferenceExperiments = {
   /**
+   * Set the default preference value for active experiments that use the
+   * default preference branch.
+   */
+  async init() {
+    const store = await ensureStorage();
+    for (const experiment of Object.values(store.data)) {
+      const {expired, preferenceBranchType, preferenceName, preferenceValue} = experiment;
+      if (!expired && preferenceBranchType === "default") {
```
gijsk: I'm a bit surprised here - why do we not just set it for all of the pref experiments, including the ones using the user branch?

recipe-client-addon/test/unit/test_PreferenceExperiments.js
```diff
@@ -110,6 +165,25 @@ add_task(withMockPreferences(function* (mockPreferences) {
...
+  const stop = sinon.stub(PreferenceExperiments, "stop");
+  mockPreferences.set("fake.preference", "startvalue", "default");
+
+  PreferenceExperiments.startObserver("test", "fake.preference", "experimentvalue");
+
+  // Setting it to the experimental value should not trigger the call.
+  DefaultPreferences.set("fake.preference", "experimentvalue");
+  ok(!stop.called, "Changing to the experimental pref value did not trigger the observer");
+
+  // Setting it to something different should trigger the call.
```
gijsk: Not entirely sure what we're testing here, and if it's worth testing this again once setting it back to the experiment value from the new value? Specifically, is calling `startObserver` expected to set the preference immediately, or is it not set until the test here calls `DefaultPreferences.set` ? In the latter case, I suppose testing this again would make no difference, but a comment might be nice. Otherwise, testing the correct behaviour (whatever that is) when setting the thing back to the expected experiment value would make sense.

recipe-client-addon/bootstrap.js
mythmon: Since the return value of the bootstrap methods isn't used (afaik?), I bet you could make them async functions if you wanted.

recipe-server/client/control/components/action_fields/PreferenceExperimentFields.js
```diff
@@ -63,12 +69,33 @@ export default class PreferenceExperimentFields extends ActionFields {
...
+        <ControlField
+          label="Preference Branch Type"
+          name="arguments.preferenceBranchType"
+          component="select"
+        >
+          <option value="default">Default</option>
+          <option value="user">User</option>
+        </ControlField>
+        {preferenceBranchType === 'user' &&
+          <p className="field-warning">
```
andymikulski: Since this element isn't dynamic, we could add it as a static property of the class. That would have React "render" it once, and would let us write something like:
```
  static userPrefWarning = (<p className="field-warning">...</p>);

// in render
{ preferenceBranchType === 'user' && PreferenceExperimentFields.userPrefWarning }
```

Minor nit, I just prefer to keep `render` concise when possible.

recipe-server/client/control/tests/components/test_PreferenceExperimentFields.js
```diff
@@ -0,0 +1,24 @@
...
+    const wrapper = () => {
+      shallow(<PreferenceExperimentFields preferenceBranchType="default" />);
+    };
+
+    expect(wrapper).not.toThrow();
+  });
+
+  it('should render a warning if using the user preference branch type', () => {
+    const wrapper = shallow(<PreferenceExperimentFields preferenceBranchType="user" />);
+    const warning = wrapper.find('.field-warning');
```
andymikulski: re: adding the warning as a static class member - I _thiiiink_ we could do something like `wrapper.find(PreferenceExperimentFields.userPrefWarning)` here.

Osmose:
> General note, that might be worth bringing up on the mailing list: code can (and does) check whether the user has "modified" preferences (ie whether a pref has a value on the user branch, using the pref service's prefHasUserValue method) as a short cut in some cases. Changing the default value with an experiment in that case would in such cases potentially not yield the result you expect. I don't expect this to be a serious problem in practice, but it seems like it's worth flagging up for QE testing things or when coordinating with devs as something to look out for, as behaviour could potentially be different when comparing the preference experiment with QE setting the pref in about:config.

I dunno if it should live in our docs (probably?), but a short guide on designing and implementing a preference experiment on the Firefox for developers seems like it'd be a nice place to put this.

recipe-client-addon/lib/PreferenceExperiments.jsm
Osmose: It's super edge case-y: If we're doing a user pref experiment (and the pref doesn't have a default value already), and the experiment ends, the default will retain the experimental value for the rest of the session. This check prevents setting the default when unnecessary. Although the reason I added it in the first place was more that it made sense only to set the pref when necessary, rather than because I thought that edge case mattered.

I'm fine with removing it, although I don't see the harm in keeping it.

recipe-client-addon/test/unit/test_PreferenceExperiments.js
Osmose: Testing after setting it back doesn't really matter, because by that point `stop` has already been called, which (when not mocked, like in this test) would remove the observer.

`startObserver` doesn't actually modify the preference, it only sets up the preference watcher, so the preference isn't set to the experimental value until the first `DefaultPreferences.set` call. #645 isn't finished yet, but it will need to call this function on startup for the active experiments, and I figured it'd be easier to reason about how this behaves without setting the preference than it was to figure out if setting the experiment preferences on startup was an issue.

I'll add a comment clarifying that `startObserver` does not modify the preference to these tests.

recipe-client-addon/bootstrap.js
Osmose: @rhelmer Does ^^ sound right?

Osmose: @rhelmer Are there any better ways to test add-on startup code than ^^?

recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -75,6 +92,20 @@ CleanupManager.addCleanupHandler(() => PreferenceExperiments.stopAllObservers())
 
 this.PreferenceExperiments = {
   /**
+   * Set the default preference value for active experiments that use the
+   * default preference branch.
+   */
+  async init() {
+    const store = await ensureStorage();
+    for (const experiment of Object.values(store.data)) {
```
Osmose: It ends up only saving a few characters. I think we use `for/of` in most other places over `forEach` as well, but I might be misremembering.

recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -105,15 +136,16 @@ this.PreferenceExperiments = {
...
+   * @param {string} branch
+   * @param {string} preferenceName
+   * @param {string|integer|boolean} preferenceValue
+   * @param {PreferenceBranchType} preferenceBranchType
    * @rejects {Error}
    *   If an experiment with the given name already exists, or if an experiment
    *   for the given preference is active.
    */
-  async start(experimentName, branch, preferenceName, preferenceValue) {
+  async start(experimentName, branch, preferenceName, preferenceValue, preferenceBranchType) {
```
Osmose: Nah, that's usually the right thing. I often wish JS had keyword arguments in the same way Python does.

recipe-server/client/control/tests/components/test_PreferenceExperimentFields.js
Osmose: It works! I used `wrapper.contains` instead of `find`, but I like this a lot better than the old test. Thanks!

Osmose:
Feedback'd, waiting on responses before making more changes.

recipe-client-addon/bootstrap.js
rhelmer: Yes you can make bootstrap methods async, and the return values are not used by AOM AFAICT.

rhelmer: Chatted about this in IRC - using the public `AddonManager` API won't be great for this since built-in system add-ons cannot be uninstalled or disabled, so re-running `bootstrap.js` `startup()` isn't really doable.

I'd suggest either simplifying`bootstrap.js` and moving the code you want to test into a different JSM that you write a test for, or if you want to have more control over the `AddonManager` lifecycle we have convenience methods for setting up and restarting AddonManager in http://searchfox.org/mozilla-central/source/toolkit/mozapps/extensions/internal/AddonTestUtils.jsm

Another option would be to use mochitest and restart the whole browser, but that's a bit more heavy-handed.

Osmose:
Alright, updated with one last commit for testing bootstrap.js, sorry for invalidating ya'lls existing review.

recipe-client-addon/bootstrap.js
```diff
@@ -45,6 +45,8 @@ const PREF_LOGGING_LEVEL = PREF_BRANCH + "logging.level";
 let shouldRun = true;
 let log = null;
 
+this.EXPORTED_SYMBOLS = ["install", "startup", "shutdown", "uninstall"];
```
gijsk: Please add a comment here saying we export these for testing purposes.

PR #671, opened 2017-04-07
Update locales/channel/countries with recipe API

### LINKED ISSUE #609: Update Recipe API doesn't update filters
New filters (added in #595 ) are not being saved when creating/editing a recipe, though are properly being passed from the front end. Fields include `locales`, `channels`, and `countries`.

You can use the `mozilla/recipe-page-filters` feature branch for your PR.

rehandalal:
Fixes #609

r?

@andymikulski I added you just to confirm if this fixes the issues you are having.

### DIFF
--- a/recipe-server/normandy/recipes/api/serializers.py
+++ b/recipe-server/normandy/recipes/api/serializers.py
@@ -91,39 +91,11 @@ class RecipeSerializer(serializers.ModelSerializer):
         ]
 
     def update(self, instance, validated_data):
-        if 'enabled' in validated_data:
-            instance.enabled = validated_data.pop('enabled')
-
-        instance.save()
-
-        if 'action' in validated_data:
-            validated_data.update({
-                'action': Action.objects.get(name=validated_data['action'])
-            })
-
-        if 'channels' in validated_data:
-            validated_data.update({
-                'channels': Channel.objects.filter(slug__in=validated_data['channels'])
-            })
-
-        if 'countries' in validated_data:
-            validated_data.update({
-                'countries': Country.objects.filter(code__in=validated_data['countries'])
-            })
-
-        if 'locales' in validated_data:
-            validated_data.update({
-                'locales': Locale.objects.filter(code__in=validated_data['locales'])
-            })
-
         instance.revise(**validated_data)
-
         return instance
 
     def create(self, validated_data):
         recipe = Recipe.objects.create()
-        recipe.save()
-
         return self.update(recipe, validated_data)
 
     def validate_extra_filter_expression(self, value):
--- a/recipe-server/normandy/recipes/tests/test_api.py
+++ b/recipe-server/normandy/recipes/tests/test_api.py
@@ -499,34 +499,37 @@ class TestRecipeAPI(object):
         assert r.action == a
 
     def test_update_recipe_locale(self, api_client):
-        r = RecipeFactory()
-        l = LocaleFactory(code='en-US')
+        l1 = LocaleFactory(code='fr-FR')
+        l2 = LocaleFactory(code='en-US')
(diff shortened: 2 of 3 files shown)


recipe-server/normandy/recipes/api/serializers.py
```diff
@@ -91,39 +91,11 @@ class Meta:
         ]
 
     def update(self, instance, validated_data):
-        if 'enabled' in validated_data:
-            instance.enabled = validated_data.pop('enabled')
-
-        instance.save()
```
rehandalal: I ditched lines 94-97 because enabled is `read_only` now.

recipe-server/normandy/recipes/api/serializers.py
```diff
@@ -91,39 +91,11 @@ class Meta:
...
-
-        if 'countries' in validated_data:
-            validated_data.update({
-                'countries': Country.objects.filter(code__in=validated_data['countries'])
-            })
-
-        if 'locales' in validated_data:
-            validated_data.update({
-                'locales': Locale.objects.filter(code__in=validated_data['locales'])
-            })
```
rehandalal: since these are all `SlugRelatedField` now we don't need this any more.

recipe-server/normandy/recipes/api/serializers.py
```diff
@@ -91,39 +91,11 @@ class Meta:
...
-
         instance.revise(**validated_data)
-
         return instance
 
     def create(self, validated_data):
         recipe = Recipe.objects.create()
-        recipe.save()
-
         return self.update(recipe, validated_data)
```
Osmose: Does anything else call update? If not, can we just call `revise` here directly and get rid of `update`?

recipe-server/normandy/recipes/tests/test_serializers.py
```diff
@@ -168,6 +168,42 @@ def test_validation_with_valid_data(self):
         }
         assert serializer.errors == {}
 
+    def test_update_action(self):
```
Osmose: I think the view tests are enough, this is straying a bit too far into testing DRF itself.

recipe-server/normandy/recipes/tests/test_api.py
```diff
@@ -488,6 +488,46 @@ def test_list_filter_text(self, api_client):
...
+
+    def test_update_recipe_locale(self, api_client):
+        r = RecipeFactory()
+        l = LocaleFactory(code='en-US')
+
+        res = api_client.patch(f'/api/v1/recipe/{r.pk}/', {'locales': ['en-US']})
+        assert res.status_code == 200
+
+        r.refresh_from_db()
+        assert r.locales.filter(pk=l.pk).exists()
```
Osmose: Could we assert against a list of locales instead of just checking if the one exists? As it is this will pass if other countries remain in the locale list, which isn't what we want, I think.

Might even want to set up another locale as already assigned to the recipe, to be sure that it's removing the old locales as well as adding the new ones.

recipe-server/normandy/recipes/api/serializers.py
rehandalal: it is used by DRF views

recipe-server/normandy/recipes/tests/test_serializers.py
rehandalal: Yeah sure, I wrote these when I was trying to isolate where the issue was occuring. I just kept them in cause they were already written. But I can ditch 'em.

recipe-server/normandy/recipes/tests/test_api.py
rehandalal: Good point, I'll make the change.

rehandalal:
@chartjes does this look okay to you?

PR #673, opened 2017-04-10
#635 - Add pref experiment type checks

### LINKED ISSUE #635: Check preference type before modifying existing preference in a preference experiment
As [suggested by @gijsk](https://github.com/mozilla/normandy/pull/623#issuecomment-288384084):

> Also, should we check what type a preference is and explicitly reject values that don't match the existing type from the recipe runner API side?

Yes. Yes we should.

(This is blocked by the linked-to PR merging)

andymikulski:
Fixes #635 

- Adds type checks when setting preference experiment values

---

I'm having trouble getting the test(s) for this feature to work. It doesn't seem to catch errors as it should, and manually throwing hard errors in `start` doesn't bubble up to the test runner, either.  I'm wondering if I have the `yield` syntax incorrect or something.

@Osmose @mythmon If you could offer some feedback on this, that'd be super appreciated!

### DIFF
--- a/recipe-client-addon/lib/PreferenceExperiments.jsm
+++ b/recipe-client-addon/lib/PreferenceExperiments.jsm
@@ -203,16 +203,16 @@ this.PreferenceExperiments = {
       preferenceBranchType,
     };
 
-    const existingPreferenceType = Services.prefs.getPrefType(preferenceName);
-    const givenType = PREFERENCE_TYPE_MAP[preferenceType];
+    const prevPrefType = Services.prefs.getPrefType(preferenceName);
+    const givenPrefType = PREFERENCE_TYPE_MAP[preferenceType];
 
-    if (!preferenceType || !givenType) {
+    if (!preferenceType || !givenPrefType) {
       throw new Error(`Invalid preferenceType provided (given "${preferenceType}")`);
     }
 
-    if (existingPreferenceType && existingPreferenceType !== givenType) {
+    if (prevPrefType !== Services.prefs.PREF_INVALID && prevPrefType !== givenPrefType) {
       throw new Error(
-        `Previous preference value is of type "${existingPreferenceType}", but was given "${givenType}" (${preferenceType})`
+        `Previous preference value is of type "${prevPrefType}", but was given "${givenPrefType}" (${preferenceType})`
       );
     }
 
--- a/recipe-client-addon/lib/PreferenceExperiments.jsm
+++ b/recipe-client-addon/lib/PreferenceExperiments.jsm
@@ -34,11 +34,15 @@
  *   Name of the preference affected by this experiment.
  * @property {string|integer|boolean} preferenceValue
  *   Value to change the preference to during the experiment.
+ * @property {string} preferenceType
+ *   Type of the preference value being set.
  * @property {string|integer|boolean|undefined} previousPreferenceValue
  *   Value of the preference prior to the experiment, or undefined if it was
  *   unset.
  * @property {PreferenceBranchType} preferenceBranchType
  *   Controls how we modify the preference to affect the client.
+ * @rejects {Error}
+ *   If the given preferenceType does not match the existing stored preference.
  *
  *   If "default", when the experiment is active, the default value for the
  *   preference is modified on startup of the add-on. If "user", the user value
@@ -62,6 +66,13 @@ XPCOMUtils.defineLazyModuleGetter(this, "TelemetryEnvironment", "resource://gre/
 this.EXPORTED_SYMBOLS = ["PreferenceExperiments"];
 
 const EXPERIMENT_FILE = "shield-preference-experiments.json";
+
+const PREFERENCE_TYPE_MAP = {
+  boolean: Services.prefs.PREF_BOOL,
+  string: Services.prefs.PREF_STRING,
+  integer: Services.prefs.PREF_INT,
+};
+
 const DefaultPreferences = new Preferences({defaultBranch: true});
 
 /**
@@ -176,7 +187,7 @@ this.PreferenceExperiments = {
 
     const preferences = PreferenceBranchType[preferenceBranchType];
     if (!preferences) {
-      throw new  Error(`Invalid value for preferenceBranchType: ${preferenceBranchType}`);
+      throw new Error(`Invalid value for preferenceBranchType: ${preferenceBranchType}`);
     }
 
     /** @type {Experiment} */
@@ -192,32 +203,17 @@ this.PreferenceExperiments = {
       preferenceBranchType,
     };
 
-    // If there's a previous preference value set, we need to ensure that the
-    // incoming value is of the same type.
-    if (preferences.has(preferenceName)) {
-      // get the previous type in PREF_ form
-      const previousType = Services.prefs.getPrefType(preferenceName);
+    const existingPreferenceType = Services.prefs.getPrefType(preferenceName);
+    const givenType = PREFERENCE_TYPE_MAP[preferenceType];
 
-      let argumentValueType = preferenceType;
-      const typeConversions = {
-        boolean: Services.prefs.PREF_BOOL,
-        string: Services.prefs.PREF_STRING,
-        integer: Services.prefs.PREF_INT,
-      };
+    if (!preferenceType || !givenType) {
+      throw new Error(`Invalid preferenceType provided (given "${preferenceType}")`);
+    }
 
-      // Default to an INVALID pref type, unless there is a key in the typeConversions
-      // that matches the given argument type
-      let givenType = Services.prefs.PREF_INVALID || 0;
-      if(typeConversions.hasOwnProperty(argumentValueType)){
-        givenType = typeConversions[argumentValueType];
-      }
-
-      // Compare both PREF_'d value types
-      if (previousType !== givenType) {
-        throw new Error(
-          `Previous preference value is of type "${previousType}", but was given "${givenType}" (${preferenceType})`
-        );
-      }
+    if (existingPreferenceType && existingPreferenceType !== givenType) {
+      throw new Error(
+        `Previous preference value is of type "${existingPreferenceType}", but was given "${givenType}" (${preferenceType})`
+      );
     }
 
     preferences.set(preferenceName, preferenceValue);
--- a/recipe-client-addon/lib/PreferenceExperiments.jsm
+++ b/recipe-client-addon/lib/PreferenceExperiments.jsm
@@ -187,10 +187,10 @@ this.PreferenceExperiments = {
       lastSeen: new Date().toJSON(),
       preferenceName,
       preferenceValue,
+      preferenceType,
       previousPreferenceValue: preferences.get(preferenceName, undefined),
       preferenceBranchType,
     };
-
 
     // If there's a previous preference value set, we need to ensure that the
     // incoming value is of the same type.
@@ -209,7 +209,7 @@ this.PreferenceExperiments = {
       // that matches the given argument type
       let givenType = Services.prefs.PREF_INVALID || 0;
       if(typeConversions.hasOwnProperty(argumentValueType)){
-        givenType = typeConversions[givenType];
+        givenType = typeConversions[argumentValueType];
       }
 
       // Compare both PREF_'d value types
--- a/recipe-client-addon/lib/PreferenceExperiments.jsm
+++ b/recipe-client-addon/lib/PreferenceExperiments.jsm
@@ -51,6 +51,7 @@
 const {utils: Cu} = Components;
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
 
+XPCOMUtils.defineLazyModuleGetter(this, "Services", "resource://gre/modules/Services.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "CleanupManager", "resource://shield-recipe-client/lib/CleanupManager.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "JSONFile", "resource://gre/modules/JSONFile.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "OS", "resource://gre/modules/osfile.jsm");
@@ -155,7 +156,7 @@ this.PreferenceExperiments = {
    *   If an experiment with the given name already exists, or if an experiment
    *   for the given preference is active.
    */
-  async start({name, branch, preferenceName, preferenceValue, preferenceBranchType}) {
+  async start({name, branch, preferenceName, preferenceValue, preferenceBranchType, preferenceType}) {
     log.debug(`PreferenceExperiments.start(${name}, ${branch})`);
 
     const store = await ensureStorage();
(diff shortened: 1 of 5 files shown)


recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -142,6 +142,16 @@ this.PreferenceExperiments = {
       previousPreferenceValue: Preferences.get(preferenceName, undefined),
     };
 
+    if (experiment.previousPreferenceValue) {
+      const previousType = typeof experiment.previousPreferenceValue;
+      const givenType = typeof preferenceValue;
```
Osmose: Instead of using `typeof`, there's an API for getting the type of a preference that you'll want to use: [nsiPrefBranch.getPrefType](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsIPrefBranch#getPrefType()).

That interface is available on `Services.prefs`. `RecipeRunner.jsm` is an example of how to import and use that object.

mythmon:
> I'm having trouble getting the test(s) for this feature to work. It doesn't seem to catch errors as it should, and manually throwing hard errors in start doesn't bubble up to the test runner, either. I'm wondering if I have the yield syntax incorrect or something.

I ran into this too, and Osmose and I tracked it down: it turns out that one of our `with...` helper methods doesn't work with generators, and only works with async/await functions. Converting the test to async/await made them work for me. It was as simple as replace `yield` with `await` and `function*` with `async function`.

recipe-client-addon/lib/PreferenceExperiments.jsm
andymikulski: @Osmose It doesn't seem that I can just give the method a value and have it spit the type of out; it takes a pref name and checks the current value of that pref.

Does that mean that in order to test the new given value, it needs to first set the pref, then check the type, then revert the value if necessary?

I looked through the docs and didn't see anything that would help cut out this weird set-then-reset business, but maybe I missed something. Thoughts?

Osmose: > @Osmose It doesn't seem that I can just give the method a value and have it spit the type of out; it takes a pref name and checks the current value of that pref.

Are you sure? I just ran this in my browser console and it worked fine:

```js
>> Services.prefs.getPrefType("asdf") // Boolean pref
128
>> Services.prefs.getPrefType("asdf") === Services.prefs.PREF_BOOL
true
>> Services.prefs.getPrefType("asdfasdf") // Doesn't exist
0
>> Services.prefs.getPrefType("asdfasdf") === Services.prefs.PREF_INVALID
true
```

Note that it's returning ints. The doc page mentioned that the return values were `PREF_BOOL` and the like, which I found as constants on the `Services.prefs` object. Sorry for not checking and explaining that earlier.

Osmose: Also, the docs for `getPrefType` doesn't mention it, but the `PREF_BOOL` return values are constants defined on `Services.prefs`. So `Services.prefs.PREF_BOOL` can be used to check if the preference is a boolean, and so forth.

andymikulski: Alright, I've corrected the approach and use the strings we use in the recipe arguments accordingly. _However,_ I'm having trouble figuring out how to actually read the recipe's arguments from within `PreferenceExperiments.jsm`. I've found that the RecipeRunner has access to the `recipe` object, but I couldn't figure out how to pass that into the PrefExperiments file.

Any suggestions?

andymikulski:
I was having a real hard time getting tests to run/fail accordingly, but I just rebased my branch on `mozilla/preference-experiment` and it all seems to be working properly now! Wish I had done that sooner.. This should be ready for review shortly.

andymikulski:
@Osmose can you take a look at 9837a97541206066534a4595f7020230fcf32aed and tell me if something's missing? The `start` test isn't receiving the `preferenceType` property I've added, but I'm not sure why. Thoughts?

Osmose:
> @Osmose can you take a look at 9837a97 and tell me if something's missing? The start test isn't receiving the preferenceType property I've added, but I'm not sure why. Thoughts?

The test that you added is actually working fine. When I run the tests locally, I see the error:

```
 0:01.58 LOG: Thread-9 ERROR Unexpected exception Error: Previous preference value is of type "32", but was given "0" (undefined) at resource://shield-recipe-client/lib/PreferenceExperiments.jsm:217
start@resource://shield-recipe-client/lib/PreferenceExperiments.jsm:217:15
async*@/Users/mkelly/Projects/mozilla-unified/obj-x86_64-apple-darwin16.4.0/_tests/xpcshell/browser/extensions/shield-recipe-client/test/unit/test_PreferenceExperiments.js:88:9
async*inner@utils.js:15:35
async*inner@resource://shield-recipe-client/lib/PreferenceExperiments.jsm:131:37
async*asyncFunction@resource://gre/modules/Task.jsm:241:18
Task_spawn@resource://gre/modules/Task.jsm:166:12
_run_next_test@/Users/mkelly/Projects/mozilla-unified/testing/xpcshell/head.js:1554:9
run@/Users/mkelly/Projects/mozilla-unified/testing/xpcshell/head.js:703:9
_do_main@/Users/mkelly/Projects/mozilla-unified/testing/xpcshell/head.js:211:5
_execute_test@/Users/mkelly/Projects/mozilla-unified/testing/xpcshell/head.js:546:5
@-e:1:1
```

Note the `test_PreferenceExperiments.jsm` line; it says line 88 is where it failed, which isn't inside the new test you added. It's in another test for `start`, which is not passing the new `preferenceType` parameter. You'll have to update the existing tests to pass the new parameter as well as adding a new test.

Also, there's a bug in `PreferenceExperiments.jsm` in that commit, on line 212 you're indexing into the `typeConversions` object using `givenType`, when the index should be `argumentValueType`.

Also, we just recently found out that the tests in `test_PreferenceExperiments.jsm` are broken because `PreferenceExperiments.jsm` now relies on Telemetry, which isn't compatible with xpcshell tests. Even if you fix these issues, your test runs should fail. @mythmon is, as we speak, working on a patch to convert the test suite to mochitests. In the meantime, if you can get the code working in a manual test, I can do a code review, until the test fix is done.

andymikulski:
Locally, I get completely different test results:

```
 0:00.96 LOG: Thread-9 ERROR Unexpected exception Error: Previous preference value is of type "32", but was given "0" (undefined) at resource://shield-recipe-client/lib/PreferenceExperiments.jsm:217
start@resource://shield-recipe-client/lib/PreferenceExperiments.jsm:217:15
_do_main@/Users/amikulski/ff-src/mozilla-central/testing/xpcshell/head.js:211:5
_execute_test@/Users/amikulski/ff-src/mozilla-central/testing/xpcshell/head.js:546:5
@-e:1:1
```

I will make the corrections you've mentioned in the above post, however I think I'm going to have to wipe everything and see if a clean 'install' fixes my issues.

Are there any docs on working with the `recipe-client-addon`? I've never actually seen anything that details how to get tests running etc.

Osmose:
> Are there any docs on working with the recipe-client-addon? I've never actually seen anything that details how to get tests running etc.

Looks like I'll be writing docs today. Which is to say, "No, and there should be."

andymikulski:
Wiping my mozilla-central and re-cloning seems to have fixed my issues. I should be able to effectively patch the remainder of this bug now!

andymikulski:
I've patched the remaining tests for pref type checking, however there appears to be an unrelated test failure (that even exists on the actual `preference-experiment` branch). Specifically, it relates to the `start did not modify the default preference` test - I don't know if this is a known issue or not.

Either way, this PR should be ready for review! @Osmose @mythmon

mythmon:
Since we need a Firefox peer to review changes to the add-on, I'm going to tag in @gijsk instead of me to review this.

recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -186,10 +187,39 @@ this.PreferenceExperiments = {
...
+      const typeConversions = {
+        boolean: Services.prefs.PREF_BOOL,
+        string: Services.prefs.PREF_STRING,
+        integer: Services.prefs.PREF_INT,
+      };
+
+      // Default to an INVALID pref type, unless there is a key in the typeConversions
+      // that matches the given argument type
+      let givenType = Services.prefs.PREF_INVALID || 0;
+      if(typeConversions.hasOwnProperty(argumentValueType)){
```
gijsk: Nit: spaces before `(` and after `)`

This should fail eslint (./mach eslint path/to/file inside m-c). Does it not?

recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -186,10 +187,39 @@ this.PreferenceExperiments = {
...
+      let argumentValueType = preferenceType;
+      const typeConversions = {
+        boolean: Services.prefs.PREF_BOOL,
+        string: Services.prefs.PREF_STRING,
+        integer: Services.prefs.PREF_INT,
+      };
+
+      // Default to an INVALID pref type, unless there is a key in the typeConversions
+      // that matches the given argument type
+      let givenType = Services.prefs.PREF_INVALID || 0;
```
gijsk: Shouldn't need to `||` this with 0.

recipe-client-addon/test/unit/test_PreferenceExperiments.js
```diff
@@ -165,6 +172,24 @@ add_task(withMockExperiments(withMockPreferences(async function (experiments, mo
   startObserver.restore();
 })));
 
+// start should detect if a new preference value type matches the previous value type
+add_task(withMockPreferences(async function (mockPreferences) {
+  mockPreferences.set("fake.type_preference", "oldvalue");
+
+  await Assert.rejects(
+    PreferenceExperiments.start({
```
gijsk: We should also have a test that the forwarded (wrapAsync) stuff on the NormandyDriver throws, too (so we don't accidentally swallow exceptions there).

More generally, we should make sure that the start() callers do the right thing with these exceptions. I tried to find them but github search is awful and this all seems to still live on a branch, maybe / sort of? So I couldn't figure out where the entrypoint is... 

In any case, we should make sure that if starting one of these experiments fails, it doesn't break running any/all experiments. Especially because, for preferences that don't live in modules/libpref/init/all.js and friends (ie appear in about:config on a default profile), it's possible for users to define those prefs with an unexpected type, and we wouldn't want users doing that to be breaking their browsers and/or all our code.

recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -186,10 +187,39 @@ this.PreferenceExperiments = {
...
       preferenceName,
       preferenceValue,
+      preferenceType,
       previousPreferenceValue: preferences.get(preferenceName, undefined),
       preferenceBranchType,
     };
 
+    // If there's a previous preference value set, we need to ensure that the
+    // incoming value is of the same type.
+    if (preferences.has(preferenceName)) {
```
gijsk: My understanding is that if:

1) a pref experiment uses a user-branch pref
2) there's a default branch value

this code won't detect it. Is that intentional? (I don't think it should be, but it's possible I'm missing something...)

recipe-client-addon/test/unit/test_PreferenceExperiments.js
```diff
@@ -148,11 +155,11 @@ add_task(withMockExperiments(withMockPreferences(async function (experiments, mo
     expired: false,
     preferenceName: "fake.preference",
     preferenceValue: "newvalue",
+    preferenceType: "string",
     previousPreferenceValue: "oldvalue",
     preferenceBranchType: "user",
   };
-  const experiment = {};
-  Object.keys(expectedExperiment).forEach(key => experiment[key] = experiments.test[key]);
+  const experiment = Object.assign({}, expectedExperiment);
```
Osmose: The original lines copied `experiments.test`, but only copied the keys from `experiment`. The intent was to only do a comparison of the values we care about, in case the experiment data ended up including more data later. The new line doesn't handle that, and should be changed back.

Although to be honest it's a confusing bit of code, and I should've left a comment mentioning that in the first place. My bad!

recipe-server/client/actions/preference-experiment/index.js
```diff
@@ -43,6 +43,7 @@ export default class PreferenceExperimentAction extends Action {
             preferenceName,
             preferenceValue: branch.value,
             preferenceBranchType,
+            preferenceType,
```
Osmose: CircleCI is failing partially for reasons unknown to me, but also because we need to update the tests for the action to assert that this parameter is passed as well.

recipe-client-addon/test/unit/test_PreferenceExperiments.js
```diff
@@ -165,6 +172,24 @@ add_task(withMockExperiments(withMockPreferences(async function (experiments, mo
...
+// start should detect if a new preference value type matches the previous value type
+add_task(withMockPreferences(async function (mockPreferences) {
+  mockPreferences.set("fake.type_preference", "oldvalue");
+
+  await Assert.rejects(
+    PreferenceExperiments.start({
+      name: "test",
+      branch: "branch",
+      preferenceName: "fake.type_preference",
+      preferenceBranchType: "invalid",
```
Osmose: You'll want to use a valid preference branch type (`user`, in this case), or else `start` will throw for an invalid branch type, not due to an invalid preference type.

It looks like [`Assert.throws` takes an `expected` argument](https://dxr.mozilla.org/mozilla-central/source/testing/modules/Assert.jsm#381) that we can use to assert that error message matches the cause for the exception that we expect. If you follow the code it looks like we can pass a regex object in to match against a portion of the error message. Someday we should refactor the rest of the throws tests to use this too, it looks useful!

recipe-client-addon/test/unit/test_PreferenceExperiments.js
Osmose: > We should also have a test that the forwarded (wrapAsync) stuff on the NormandyDriver throws, too (so we don't accidentally swallow exceptions there).

That's probably a test for a different PR, in the SandboxManager tests. I'd rather test that once for wrapAsync instead of testing every use of wrapAsync. Although there is the unfilled gap of an integration test for preference-experiment recipes.

> More generally, we should make sure that the start() callers do the right thing with these exceptions. I tried to find them but github search is awful and this all seems to still live on a branch, maybe / sort of? So I couldn't figure out where the entrypoint is...

The entrypoint is in the action code in `recipe-server` that gets run in the sandbox: https://github.com/andymikulski/normandy/blob/d74e71bc7c86d392e193d79a3cc819bb7d2427c6/recipe-server/client/actions/preference-experiment/index.js#L40

Feature branches are... well, they are _a_ solution to a real problem. It's still up for debate if they're a _good_ solution. :P

> In any case, we should make sure that if starting one of these experiments fails, it doesn't break running any/all experiments.

We have existing tests that [ActionSandboxManager throws when the sandbox code throws](https://github.com/andymikulski/normandy/blob/d74e71bc7c86d392e193d79a3cc819bb7d2427c6/recipe-client-addon/test/unit/test_ActionSandboxManager.js#L70). We're missing a test that `RecipeRunner`, which calls that method, handles the error appropriately and doesn't halt execution of other actions. I think that's separate from this PR though, and shouldn't block it.

recipe-client-addon/lib/PreferenceExperiments.jsm
andymikulski: It does! I had been hitting an issue with eslint not running properly, but after some fiddling it's picking this up appropriately now.

andymikulski:
I believe I've addressed all review feedback. I'm having tests fail reliably (again), even after rebasing onto `preference-experiment`. Really hoping I don't need to have to wipe my mozilla-central yet again, but I have a feeling I will.

Ready for feedback.

mythmon:
It looks like the only CI failures here are some lint problems in Circle, so we can probably merge this if it passes review.

If you get to the point where you need to wipe mozilla-central again, I'd like to have a vidyo call with you to see if we can work through the issue using less extreme methods. Maybe there is something easy we can figure out that won't require a full rebuild.

andymikulski:
Bummer. I thought I had fixed my eslint issue, but looks not. I'll correct now.

PR #674, opened 2017-04-10
Inform telemetry of active preference experiments

### LINKED ISSUE #646: Inform Telemetry about active preference experiments
The preference experiment driver API should inform Telemetry when a preference experiment is active. The API landed in [~~bug 1348748~~](https://bugzilla.mozilla.org/show_bug.cgi?id=1348748) and docs are currently available [here](http://searchfox.org/mozilla-central/source/toolkit/components/telemetry/docs/collection/index.rst). 

It's currently only available on Nightly, but it's on track to be uplifted, so I don't think we need to worry about the API not being available before the preference-experiments feature branch lands on master.

Telemetry doesn't persist this state between sessions, so we need to notify Telemetry both when experiments start and end, and on startup for any in-progress experiments.

mythmon:
This depends on #659 (which in turn depends on #641). The relevant commits are 2c4ee93 and after.

Fixes #646.

### DIFF
--- a/recipe-client-addon/lib/PreferenceExperiments.jsm
+++ b/recipe-client-addon/lib/PreferenceExperiments.jsm
@@ -108,7 +108,9 @@ this.PreferenceExperiments = {
       }
 
       // Notify Telemetry of experiments we're running, since they don't persist between restarts
-      TelemetryEnvironment.setExperimentActive(name, branch);
+      if (!expired) {
+        TelemetryEnvironment.setExperimentActive(name, branch);
+      }
     });
   },
 
--- a/recipe-client-addon/lib/PreferenceExperiments.jsm
+++ b/recipe-client-addon/lib/PreferenceExperiments.jsm
@@ -37,6 +37,13 @@
  * @property {string|integer|boolean|undefined} previousPreferenceValue
  *   Value of the preference prior to the experiment, or undefined if it was
  *   unset.
+ * @property {PreferenceBranchType} preferenceBranchType
+ *   Controls how we modify the preference to affect the client.
+ *
+ *   If "default", when the experiment is active, the default value for the
+ *   preference is modified on startup of the add-on. If "user", the user value
+ *   for the preference is modified when the experiment starts, and is reset to
+ *   its original value when the experiment ends.
  */
 
 "use strict";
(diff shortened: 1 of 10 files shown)


recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -97,11 +98,17 @@ this.PreferenceExperiments = {
...
-      const {expired, preferenceBranchType, preferenceName, preferenceValue} = experiment;
+      const {name, branch, expired, preferenceBranchType, preferenceName, preferenceValue} = experiment;
+
+      // Set experiment default preferences, since they don't persist between restarts
       if (!expired && preferenceBranchType === "default") {
         DefaultPreferences.set(preferenceName, preferenceValue);
       }
+
+      // Notify Telemetry of experiments we're running, since they don't persist between restarts
+      TelemetryEnvironment.setExperimentActive(name, branch);
```
Osmose: This needs to check if the experiment in question is disabled or not.

recipe-client-addon/test/unit/test_PreferenceExperiments.js
```diff
@@ -477,3 +478,45 @@ add_task(withMockExperiments(withMockPreferences(function* testInit(experiments,
...
+  let currentExperiments = TelemetryEnvironment.getActiveExperiments();
+  ok(!("experiment1" in currentExperiments), "experiment1 is not in telemetry");
+  ok(!("experiment2" in currentExperiments), "experiment2 is not in telemetry");
+
+  await PreferenceExperiments.init();
+
+  currentExperiments = TelemetryEnvironment.getActiveExperiments();
+  deepEqual(currentExperiments["experiment1"], {branch: "branch1"}, "experiment1 is in telemetry");
+  deepEqual(currentExperiments["experiment2"], {branch: "branch2"}, "experiment2 is in telemetry");
+}));
```
Osmose: Are the started experiments persisted across tests? We might want to clean up after ourselves so that other test cases aren't affected.

recipe-client-addon/test/unit/test_PreferenceExperiments.js
```diff
@@ -477,3 +478,45 @@ add_task(withMockExperiments(withMockPreferences(async function testInit(experim
     "init set the value for a default pref experiment",
   );
 })));
+
+// init should register telemetry experiments
+add_task(withMockExperiments(async function testInit(experiments) {
+  experiments["experiment1"] = experimentFactory({name: "experiment1", branch: "branch1"});
+  experiments["experiment2"] = experimentFactory({name: "experiment2", branch: "branch2"});
+
+  let currentExperiments = TelemetryEnvironment.getActiveExperiments();
```
gijsk: Would it make more sense to mock/stub TelemetryEnvironment with sinon and then just check that we're calling TelemetryEnvironment when the experiments get added?

gijsk: Also, given @Osmose's comment, I guess we should have tests that we don't set things as active in telemetry which really aren't.

mythmon:
Fixed.

PR #676, opened 2017-04-11
Merge master into preference-experiments branch

Osmose:
Time for another merge! This one was a little hairy.

As before, your best bet for reviewing this is pulling it down and running `git show 4350083b963e99442df3a832e12d5808fb6be005`. It'll still have some extra stuff beyond the merge fixes I made, but it's better than Github's diff.

Note that I didn't just make merge changes, but some fixes to stuff in master in order to make my local testing work. I suspect it's easiest to just land these in master when we merge this branch back in, but I'm happy to go through and port them to master sooner if that's preferred.

Interesting changes from the merge:

- I reverted the [storage checks](https://github.com/mozilla/normandy/pull/633) merge because I was having issues getting the test to pass (I think it had to do with how the test was creating the driver, but I don't quite remember), noticed that the PR was missing a peer review, and decided to punt on figuring it out myself. This is actually in a commit separate from the merge, in case you want to review it. The revert didn't conflict so I figured it's good.
- The [getAppLocaleAsLangTag](https://github.com/mozilla/normandy/pull/662/) change breaks the XPI outside of Nightly, as the API doesn't exist. So I added feature detection for it.
- `install.rdf.in` has [a comment of some kind at the top](https://github.com/mozilla/normandy/blob/master/recipe-client-addon/install.rdf.in#L3) which breaks the RDF parser and causes the add-on to be corrupt. I removed it.
- `_common.scss` has some CSS styles for `<a>` tags with the `:disabled` selector, which is invalid (I think). I needed it for the preference experiment fields, so I changed it to a class selector (`.disabled`).
- I added the logic for disabling form controls to `PreferenceExperimentFields.js`.
- `test_models.py` needed a small rename of `recipe.update()` to `recipe.revise()`.

As far as I can tell, all three recipe types work with the system add-on, the control interface functions as expected, and the tests and lints all pass.


rehandalal:
yikes!

Osmose:
@mythmon Looks like the TaskCluster build failed trying to download the build result, any ideas as to why?

mythmon:
I'm not sure why Taskcluster failed. I've raised the issue in #taskcluster, and I'll report if I hear anything more.

Osmose:
Updated. All the changes in the git show diff are now required for the merge.

mythmon:
The taskcluster failure is due to a timeout, and is likely unrelated. I suggest we merge this anyways.

PR #681, opened 2017-04-11
Inform telemetry pt 2

mythmon:
Recreating #674, since I accidentally merged it too early, and Github didn't pick up my back-out.

#674 was approved, and this is just waiting for #659 to merge before it can be rebased and merged.


mythmon:
I rebased this on preference-experiments now that #659 is merged. rs?

mythmon:
We figured out that the artifact problems we have been skipping simply never worked, despite the PR that introduced the feature having a green CI run. I'm pre-emptively ignoring Taskcluster on this PR, since I know it will fail in the way we agreed to ignore.

PR #683, opened 2017-04-12
Bump TC max run time to 8 hours and deadline to 1 day.

Osmose:
@djmitche recommended, as a temporary fix, that we bump up our build times to get our PRs green again.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -16,6 +16,9 @@ tasks:
     workerType: "{{ taskcluster.docker.workerType }}"
     scopes:
       - queue:create-task:aws-provisioner-v1/github-worker
+      # the decision task creates tasks that run on the gecko worker for speed
+      # and to ensure the same execution environment as "regular" gecko builds
+      - queue:create-task:aws-provisioner-v1/gecko-1-b-linux
     payload:
       env:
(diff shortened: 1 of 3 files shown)


Osmose:
Well, the good news is that the build took 4 and a half hours and finished successfully.

The bad news is that it still failed trying to pull the artifact from the build task into the testing task:

```
Downloading build result

gzip: stdin: unexpected end of file
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

@mythmon said that he asked the TaskCluster folks about this when we first saw it a few days ago and they said it was due to a deploy that day and had been resolved. That doesn't appear to be the case here.

djmitche:
Yeah, on further questioning that correspondance didn't seem to be accurate.  This is intermittent, right?  The download is direct from S3, and for some reason S3 seems to have this problem more frequently than most HTTP services, and you'll need retry-on-failure logic.  IIRC you're using curl, which has `--retry`.

mythmon:
@Osmose We pass a very limited set of scopes into the decision task, so even though our repo has the scope to create the worker type, the decision task didn't inherit it. I changed that for you.

Osmose:
TaskCluster passed! But then I added one more commit. That should be the last one, though.

PR #684, opened 2017-04-12
Watch experiment prefs

### LINKED ISSUE #645: Check for changed preference on startup

mythmon:
This is based on #674, to use the addition of the `init` method to `PreferenceExperiments`. The relevant commit is 13628fa.

Fixes #645.

The issue mentioned "Register the preference observer that monitors for changes to the preference." This was already done.

### DIFF
--- a/recipe-client-addon/lib/PreferenceExperiments.jsm
+++ b/recipe-client-addon/lib/PreferenceExperiments.jsm
@@ -88,7 +88,7 @@ function ensureStorage() {
 const log = LogManager.getLogger("preference-experiments");
 
 // List of active preference observers. Cleaned up on shutdown.
-const experimentObservers = new Map();
+let experimentObservers = new Map();
 CleanupManager.addCleanupHandler(() => PreferenceExperiments.stopAllObservers());
 
 this.PreferenceExperiments = {
@@ -97,21 +97,26 @@ this.PreferenceExperiments = {
    * default preference branch.
    */
   async init() {
-    const store = await ensureStorage();
-
-    Object.values(store.data).forEach(experiment => {
-      const {name, branch, expired, preferenceBranchType, preferenceName, preferenceValue} = experiment;
-
+    for (const experiment of await this.getAllActive()) {
       // Set experiment default preferences, since they don't persist between restarts
-      if (!expired && preferenceBranchType === "default") {
-        DefaultPreferences.set(preferenceName, preferenceValue);
+      if (experiment.preferenceBranchType === "default") {
+        DefaultPreferences.set(experiment.preferenceName, experiment.preferenceValue);
+      }
+
+      // Check that the current value of the preference is still what we set it to
+      if (Preferences.get(experiment.preferenceName, undefined) !== experiment.preferenceValue) {
+        // if not, stop the experiment, and skip the remaining steps
+        log.info(`Stopping experiment "${experiment.name}" because its value changed`);
+        await this.stop(experiment.name, false);
+        continue;
       }
 
       // Notify Telemetry of experiments we're running, since they don't persist between restarts
-      if (!expired) {
-        TelemetryEnvironment.setExperimentActive(name, branch);
-      }
-    });
+      TelemetryEnvironment.setExperimentActive(experiment.name, experiment.branch);
+
+      // Watch for changes to the experiment's preference
+      this.startObserver(experiment.name, experiment.preferenceName, experiment.preferenceValue);
+    }
   },
 
   /**
@@ -126,10 +131,14 @@ this.PreferenceExperiments = {
         data: mockExperiments,
         saveSoon() { },
       });
+      const oldObservers = experimentObservers;
+      experimentObservers = new Map();
       try {
         await testFunction(...args, mockExperiments);
       } finally {
         storePromise = oldPromise;
+        PreferenceExperiments.stopAllObservers();
+        experimentObservers = oldObservers;
       }
     };
   },
--- a/recipe-client-addon/test/browser/browser_PreferenceExperiments.js
+++ b/recipe-client-addon/test/browser/browser_PreferenceExperiments.js
@@ -126,7 +126,6 @@ add_task(withMockExperiments(withMockPreferences(async function(experiments, moc
 // start should modify the user preference for the user branch type
 add_task(withMockExperiments(withMockPreferences(async function(experiments, mockPreferences) {
   const startObserver = sinon.stub(PreferenceExperiments, "startObserver");
-  mockPreferences.set("fake.preference", "defaultvalue", "default");
   mockPreferences.set("fake.preference", "oldvalue", "user");
   mockPreferences.set("fake.preference", "olddefaultvalue", "default");
 
@@ -312,8 +311,6 @@ add_task(withMockExperiments(async function(experiments) {
 // preference value.
 add_task(withMockExperiments(withMockPreferences(async function(experiments, mockPreferences) {
   const stopObserver = sinon.spy(PreferenceExperiments, "stopObserver");
-  const hasObserver = sinon.stub(PreferenceExperiments, "hasObserver");
-  hasObserver.returns(Promise.resolve(true));
 
   mockPreferences.set("fake.preference", "experimentvalue", "default");
   experiments["test"] = experimentFactory({
@@ -335,7 +332,6 @@ add_task(withMockExperiments(withMockPreferences(async function(experiments, moc
     "stop reverted the preference to its previous value",
   );
 
-  hasObserver.restore();
   stopObserver.restore();
   PreferenceExperiments.stopAllObservers();
 })));
@@ -381,7 +377,6 @@ add_task(withMockExperiments(withMockPreferences(async function(experiments) {
 // stop should remove a preference that had no value prior to an experiment for user prefs
 add_task(withMockExperiments(withMockPreferences(async function(experiments, mockPreferences) {
   const stopObserver = sinon.stub(PreferenceExperiments, "stopObserver");
-  mockPreferences.set("fake.preference", "defaultvalue", "default");
   mockPreferences.set("fake.preference", "experimentvalue", "user");
   experiments["test"] = experimentFactory({
     name: "test",
--- a/recipe-client-addon/test/browser/browser_PreferenceExperiments.js
+++ b/recipe-client-addon/test/browser/browser_PreferenceExperiments.js
@@ -197,26 +197,6 @@ add_task(withMockExperiments(withMockPreferences(async function(mockExperiments,
   stop.restore();
 })));
 
-// startObserver should observe changes to the default preference value.
-add_task(withMockExperiments(withMockPreferences(async function(mockExperiments, mockPreferences) {
-  const stop = sinon.stub(PreferenceExperiments, "stop");
-  mockPreferences.set("fake.preference", "startvalue", "default");
-
-  // NOTE: startObserver does not modify the pref
-  PreferenceExperiments.startObserver("test", "fake.preference", "experimentvalue");
-
-  // Setting it to the experimental value should not trigger the call.
-  DefaultPreferences.set("fake.preference", "experimentvalue");
-  ok(!stop.called, "Changing to the experimental pref value did not trigger the observer");
-
-  // Setting it to something different should trigger the call.
-  DefaultPreferences.set("fake.preference", "newvalue");
-  ok(stop.called, "Changing to a different value triggered the observer");
-
-  PreferenceExperiments.stopAllObservers();
-  stop.restore();
-})));
-
 add_task(withMockExperiments(async function testHasObserver() {
   PreferenceExperiments.startObserver("test", "fake.preference", "experimentValue");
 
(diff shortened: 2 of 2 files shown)


recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -99,18 +99,23 @@ this.PreferenceExperiments = {
   async init() {
     const store = await ensureStorage();
 
-    Object.values(store.data).forEach(experiment => {
-      const {name, branch, expired, preferenceBranchType, preferenceName, preferenceValue} = experiment;
-
+    Object.values(store.data)
+    .filter(experiment => !experiment.expired)
```
gijsk: Firefox style would be indenting this and the next line.

Osmose:
> The issue mentioned "Register the preference observer that monitors for changes to the preference." This was already done.

We also need to register the observer on startup, since it doesn't persist and there's a gap between when we check the value on startup and when the recipes are executed.

mythmon:
I've made the requested changes, and re-based this on mozilla/preference-experiments, so it includes the changes from #674. This is ready for review, and if approved, ready for merge.

recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -99,19 +99,27 @@ this.PreferenceExperiments = {
   async init() {
     const store = await ensureStorage();
 
-    Object.values(store.data).forEach(experiment => {
-      const {name, branch, expired, preferenceBranchType, preferenceName, preferenceValue} = experiment;
+    Object.values(store.data)
+      .filter(experiment => !experiment.expired)
+      .forEach(experiment => {
```
Osmose: You can just use `PreferenceExperiments.getAllActive()` for this.

recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -99,19 +99,27 @@ this.PreferenceExperiments = {
...
+        }
 
-      // Set experiment default preferences, since they don't persist between restarts
-      if (!expired && preferenceBranchType === "default") {
-        DefaultPreferences.set(preferenceName, preferenceValue);
-      }
+        // Check that the current value of the preference is still what we set it to
+        if (Preferences.get(experiment.preferenceName, undefined) !== experiment.preferenceValue) {
+          // if not, stop the experiment, and skip the remaining steps
+          this.stop(experiment.experimentName, false);
```
Osmose: I think the `this` keyword is broken in functions wrapped with `wrapAsync`, but I have a commit lying around that fixes that: https://github.com/Osmose/normandy/commit/2bf562431a6a7e6d2036eefbdb723ab9c2a39894

recipe-client-addon/test/unit/test_PreferenceExperiments.js
```diff
@@ -584,10 +584,44 @@ add_task(withMockExperiments(async function testInit() {
...
+  });
+  mockPreferences.set("fake.preference", "changed value");
+  await PreferenceExperiments.init();
+  ok(experiments["test"].expired, "Experiment expires because value changed");
+  ok(mockPreferences.get("fake.preference"), "changed value", "Preference value was not changed");
+})));
+
+// init should register an observer for experiments
+add_task(withMockExperiments(withMockPreferences(async function testInitRegistersObserver(experiments, mockPreferences) {
+  dump("@@ starting test");
```
Osmose: Leftover debugging!

recipe-client-addon/test/unit/test_PreferenceExperiments.js
```diff
@@ -584,10 +584,44 @@ add_task(withMockExperiments(async function testInit() {
...
+  mockPreferences.set("fake.preference", "experiment value");
+  await PreferenceExperiments.init();
+
+  ok(
+    startObserver.calledWith("test", "fake.preference", "experiment value"),
+    "init registered an observer",
+  );
+
+  startObserver.restore();
+  ok(false);
```
Osmose: That ain't ok

recipe-client-addon/test/unit/test_PreferenceExperiments.js
```diff
@@ -584,10 +584,44 @@ add_task(withMockExperiments(async function testInit() {
...
+// Experiments should end if the preference has been changed when init() is called
+add_task(withMockExperiments(withMockPreferences(async function testInitChanges(experiments, mockPreferences) {
+  experiments["test"] = experimentFactory({
+    name: "test",
+    preferenceName: "fake.preference",
+    preferenceValue: "experiment value",
+  });
+  mockPreferences.set("fake.preference", "changed value");
+  await PreferenceExperiments.init();
+  ok(experiments["test"].expired, "Experiment expires because value changed");
```
Osmose: Maybe we should mock `stop` instead, in case its implementation changes later?

recipe-client-addon/test/unit/test_PreferenceExperiments.js
```diff
@@ -554,13 +559,25 @@ add_task(withMockExperiments(withMockPreferences(async function testInit(experim
...
+  mockPreferences.set("fake.pref", "experiment value");
+
+  experiments["test"] = experimentFactory({
+    name: "test",
+    branch: "branch",
+    preferenceName: "fake.pref",
+    preferenceValue: "experiment value",
+    expired: false,
+    preferenceBranchType: "default",
+  });
```
Osmose: Any reason why you need this extra test data?

recipe-client-addon/test/unit/test_PreferenceExperiments.js
```diff
@@ -594,20 +611,21 @@ add_task(withMockExperiments(async function testInitTelemetryExpired(experiments
...
   experiments["test"] = experimentFactory({
     name: "test",
     preferenceName: "fake.preference",
     preferenceValue: "experiment value",
   });
   mockPreferences.set("fake.preference", "changed value");
   await PreferenceExperiments.init();
-  ok(experiments["test"].expired, "Experiment expires because value changed");
+  ok(stopStub.calledWith("test"), "Experiment is stopped because value changed");
   ok(mockPreferences.get("fake.preference"), "changed value", "Preference value was not changed");
```
Osmose: This is unnecessary now that you're checking if stop was called.

recipe-client-addon/test/unit/test_PreferenceExperiments.js
```diff
@@ -594,20 +611,21 @@ add_task(withMockExperiments(async function testInitTelemetryExpired(experiments
...
+  const stopStub = sinon.stub(PreferenceExperiments, "stop");
   experiments["test"] = experimentFactory({
     name: "test",
     preferenceName: "fake.preference",
     preferenceValue: "experiment value",
   });
   mockPreferences.set("fake.preference", "changed value");
   await PreferenceExperiments.init();
-  ok(experiments["test"].expired, "Experiment expires because value changed");
+  ok(stopStub.calledWith("test"), "Experiment is stopped because value changed");
```
Osmose: This should fail because `false` was also passed, which you want to assert.

recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -106,8 +106,9 @@ this.PreferenceExperiments = {
       // Check that the current value of the preference is still what we set it to
       if (Preferences.get(experiment.preferenceName, undefined) !== experiment.preferenceValue) {
         // if not, stop the experiment, and skip the remaining steps
+        log.info(`Stopping experiment "${experiment.name}" because it's value changed`);
```
Osmose: s/it's/its/

recipe-client-addon/test/browser/browser_PreferenceExperiments.js
```diff
@@ -30,7 +28,7 @@ add_task(withMockExperiments(async function (experiments) {
   await PreferenceExperiments.clearAllExperimentStorage();
   ok(
     !(await PreferenceExperiments.has("test")),
-    "clearAllExperimentStorage removed all stored experiments",
+    "clearAllExperimentStorage removed all stored experiments"
```
Osmose: Don't we always want trailing commas?

recipe-client-addon/test/browser/browser_PreferenceExperiments.js
```diff
@@ -399,6 +397,7 @@ add_task(withMockExperiments(withMockPreferences(async function (experiments) {
 // stop should remove a preference that had no value prior to an experiment for user prefs
 add_task(withMockExperiments(withMockPreferences(async function (experiments, mockPreferences) {
   const stopObserver = sinon.stub(PreferenceExperiments, "stopObserver");
+  mockPreferences.set("fake.preference", "defaultvalue", "default");
```
Osmose: This negates the point of this test, which is to test that preferences that have no value are remove after the test.

recipe-client-addon/test/browser/browser_PreferenceExperiments.js
```diff
@@ -521,41 +521,48 @@ add_task(withMockExperiments(async function (experiments) {
...
+  is(
+    mockPreferences.get("user", undefined, "default"),
     false,
-    "init ignored an expired default pref experiment",
+    "init ignored a user pref experiment",
   );
-  equal(
-    DefaultPreferences.get("default"),
+  is(
+    mockPreferences.get("default", undefined, "default"),
```
Osmose: I'm not sure how I feel about using `mockPreferences` for reading values like this. Why is it better?

recipe-client-addon/test/unit/test_PreferenceExperiments.js
mythmon: `calledWith` only asserts about the arguments you pass. so

```js
const foo = sinon.stub();
foo(1, 2, 3);
ok(foo.calledWith(1));
```

would pass. The behavior you're thinking of is a part of `calledWithExact`. Which we should probably be using.

recipe-client-addon/test/browser/browser_PreferenceExperiments.js
mythmon: It seemed strange to deal with two different objects for interacting with preferences during tests. This way a test that uses prefs only needs one object, instead of two or three (in the case of using both default and non-default prefs). It also gives us a consistent API between setting and getting prefs, which is nice.

mythmon:
I accidentally closed this be deleting the branch. I'm not sure what happened, most like my automation scripts got a bit over-zealous.

recipe-client-addon/test/browser/browser_PreferenceExperiments.js
```diff
@@ -126,6 +126,7 @@ add_task(withMockExperiments(withMockPreferences(async function(experiments, moc
 // start should modify the user preference for the user branch type
 add_task(withMockExperiments(withMockPreferences(async function(experiments, mockPreferences) {
   const startObserver = sinon.stub(PreferenceExperiments, "startObserver");
+  mockPreferences.set("fake.preference", "defaultvalue", "default");
```
Osmose: Is this leftover from a rebase? We set the default below.

recipe-client-addon/test/browser/browser_PreferenceExperiments.js
```diff
@@ -214,9 +215,9 @@ add_task(withMockPreferences(async function(mockPreferences) {
 
   PreferenceExperiments.stopAllObservers();
   stop.restore();
-}));
+})));
 
-add_task(async function testHasObserver() {
+add_task(withMockExperiments(async function testHasObserver() {
```
Osmose: Are all these mocks necessary?

recipe-client-addon/test/browser/browser_PreferenceExperiments.js
```diff
@@ -331,6 +332,9 @@ add_task(withMockExperiments(async function(experiments) {
 // preference value.
 add_task(withMockExperiments(withMockPreferences(async function(experiments, mockPreferences) {
   const stopObserver = sinon.spy(PreferenceExperiments, "stopObserver");
+  const hasObserver = sinon.stub(PreferenceExperiments, "hasObserver");
```
Osmose: I don't think we need this mock anymore, as the test starts an observer itself now.

recipe-client-addon/test/browser/browser_PreferenceExperiments.js
```diff
@@ -397,6 +401,7 @@ add_task(withMockExperiments(withMockPreferences(async function(experiments) {
 // stop should remove a preference that had no value prior to an experiment for user prefs
 add_task(withMockExperiments(withMockPreferences(async function(experiments, mockPreferences) {
   const stopObserver = sinon.stub(PreferenceExperiments, "stopObserver");
+  mockPreferences.set("fake.preference", "defaultvalue", "default");
```
Osmose: This shouldn't be necessary, the test asserts using `isSet` now, which only checks user preferences.

recipe-client-addon/test/browser/browser_PreferenceExperiments.js
mythmon: I added these mocks because they make sure that observers get properly cleaned up. We don't strictly need them on every test, but they make the tests more consistent.

mythmon:
Removed the unneeded stuff from the tests.

PR #686, opened 2017-04-12
recipe-server: Update signatures when closing approval requests.

mythmon:
This fixes the issue we saw on prod today where closing an approval request broke signatures, and thus heartbeat.

Sorry about the diff here, I indented the whole flow test case, which is a big diff. The behavior of the old test didn't change, but I did extract some behavior out of it.

### DIFF
--- a/recipe-server/normandy/recipes/tests/test_api.py
+++ b/recipe-server/normandy/recipes/tests/test_api.py
@@ -9,7 +9,7 @@ from rest_framework.reverse import reverse
 
 from normandy.base.api.permissions import AdminEnabledOrReadOnly
 from normandy.base.tests import UserFactory, Whatever
-from normandy.base.utils import aware_datetime
+from normandy.base.utils import aware_datetime, canonical_json_dumps
 from normandy.recipes.models import ApprovalRequest, Recipe
 from normandy.recipes.tests import (
     ActionFactory,
@@ -18,6 +18,7 @@ from normandy.recipes.tests import (
     CountryFactory,
     LocaleFactory,
     RecipeFactory,
+    fake_sign,
 )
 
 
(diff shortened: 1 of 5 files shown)


recipe-server/normandy/recipes/tests/test_api.py
```diff
@@ -664,140 +665,179 @@ def test_sets_no_cookies(self, api_client):
...
+
+        # The change should be visible now, since it is approved
+        res = api_client.get('/api/v1/recipe/{}/'.format(recipe_data_1['id']))
+        assert res.status_code == 200
+        recipe_data_4 = res.json()
+        assert recipe_data_4['extra_filter_expression'] == 'counter == 2'
+        assert recipe_data_4['latest_revision_id'] == recipe_data_4['revision_id']
+        self.verify_signatures(api_client, expected_count=1)
+
+    def test_request_and_then_unrequest(self, api_client, mocked_autograph):
```
Osmose: `test_cancel_approval` might be a better name

rehandalal:
part of me is wondering whether this is better handled with signals or editing the save/delete methods for `ApprovalRequest`?

mythmon:
I thought about overriding the delete method for this, but I went with a cancel method to better represent the semantics of what is going on. Maybe Moving all the signature handling to a save and delete method, and keeping the semantics methods (but without that logic) would be better.

I'm not sure how we would do this with signals. I haven't worked with them a lot. What did you have in mind?

Osmose:
> Maybe Moving all the signature handling to a save and delete method, and keeping the semantics methods (but without that logic) would be better.

I strongly think we should be consistent one way or the other, and less strongly believe that we should avoid putting this logic in overridden save/delete handlers, or signals. 

Save and delete signals only get called when the save/delete methods are called on models, so the only difference is whether they're defined on the model as overridden methods, or in standalone functions. Over time I've found that overriding save/delete has some downsides (for the projects I've worked on, at least):

- They're not always called when the models are saved or deleted, like when updating via a queryset, which means you have to put the logic in multiple places anyway.
- Predicting where save/delete are called isn't always easy (like during tests). You often end up running your custom code at times when you don't expect to.
- It's not obvious external to the class that save/delete do anything more than save to the database, as the names don't convey what you're doing.

I like the `close` method added in this PR over that.

rehandalal:
@Osmose raised some good points. I am convinced now!

mythmon:
The Taskcluster failure is the artifact issue, which we've agreed to ignore for now.

PR #687, opened 2017-04-13
Add gecko-1-b-linux scope to the decision task

djmitche:


mythmon:
@djmitche I think we'd also have to change the type of task created in [`decision.py`](https://github.com/mozilla/normandy/blob/master/ci/taskcluster/bin/decision.py). Both of these changes are being used in #683, so I think this PR isn't needed.

djmitche:
Indeed, you're right, but at least the role seems to be set up correctly now

PR #688, opened 2017-04-13
recipe-server/contract-tests: Skip cache tests on api URLs.

mythmon:
To work around Firefox caching bugs, we are hiding the the cache headers on API views in Nginx. These cache headers are still used server side to cache requests on webheads, but sending them to Firefox currently (on Beta 53) causes it to serve very stale requests.

This is temporary, and should be removed when the Firefox bug is fixed.

CC @relud


mythmon:
To test: without these changes, the contract tests fail on stage. With these changes, they pass by skipping the failing tests.

mythmon:
Taskcluster failure is the artifact problem, which we've agreed to ignore for now.

PR #691, opened 2017-04-14
Do not break `this` in functions wrapped with SandboxManager.wrapAsync.

Osmose:
wrapAsync now returns a non-arrow function, and binds the `this` value within the wrapper to `this` in the wrapped function.

(SandboxManager tests still use generators... that will change. Soon. But not in this PR.)


mythmon:
The TC failure here is the bad artifact URL we fixed in #683. I think this is good to merge now.

PR #693, opened 2017-04-15
Refactor TaskCluster tasks and fix linting / tests

Osmose:
Let's see what TC does!

### DIFF
--- a/recipe-client-addon/bin/tc/fetch.sh
+++ b/recipe-client-addon/bin/tc/fetch.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+set -eu
+
+# mach wants this
+export SHELL=$(which bash)
+
+# Creates gecko-dev-master
+echo 'Downloading gecko-dev...'
+curl --location --fail --silent --show-error https://github.com/mozilla/gecko-dev/archive/master.tar.gz | tar xz
+
+echo 'Syncing recipe-client-addon to gecko-dev...'
+pushd normandy/recipe-client-addon
+npm install
+./bin/update-mozilla-central.sh ../../gecko-dev-master/
+popd
+
+echo 'Making build tarball artifact'
+tar czvf /artifacts/source.tar.gz gecko-dev-master
--- a/recipe-client-addon/bin/tc/fetch.sh
+++ b/recipe-client-addon/bin/tc/fetch.sh
@@ -5,6 +5,12 @@ set -eu
 echo 'Downloading gecko-dev...'
 curl --location --fail --silent --show-error https://github.com/mozilla/gecko-dev/archive/master.tar.gz | tar xz
 
+echo 'Pulling tags from mozilla/normandy repo on Github...'
+pushd normandy
+git remote add mozilla https://github.com/mozilla/normandy.git
+git fetch mozilla
+popd
+
 echo 'Syncing recipe-client-addon to gecko-dev...'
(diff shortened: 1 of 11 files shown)


Osmose:
Getting closer! I've refactored the TaskCluster tasks so that linting happens in it's own task, and fetching gecko-dev (and updating the in-tree shield-recipe-client with the changes in our repo) also happens in its own task. I thought this increase build times by a ton, but apparently not, as lints seem to be done within minutes of submitting a PR.

Interestingly, I found a bug in `build.sh` that got moved over to `fetch.sh` during my refactor: the path passed to `update-mozilla-central.sh` is incorrect and doesn't exist. It writes our updated add-on files to a directory outside of `gecko-dev-master`, where the build happens. I _think_ this means that our test runs have, for at least a while, not been testing our changes in Github. This is also why the existing linting was passing.

Now, the tests are failing. Any import to a `resource://shield-recipe-client` path is failing. I'm still not entirely sure why.

One thing that is wrong, but that I'm not sure would cause this issue, is that the `install.rdf.in` in teh built Firefox is missing a version number. This is because `git describe --dirty`, which is used to generate the version number in `update-mozilla-central.sh`, fails with the error (I tested with a one-off loaner):

```
root@7292caabb71c:~/normandy/recipe-client-addon# git -C "$(pwd)" describe --dirty                               
fatal: No names found, cannot describe anything.
```

There's only one remote, named `origin`:

```
root@7292caabb71c:~/normandy# git remote show origin
* remote origin
  Fetch URL: https://github.com/Osmose/normandy.git
  Push  URL: https://github.com/Osmose/normandy.git
  HEAD branch: master
  Remote branches:
    action-refactor                       tracked
    fix-tc-builds                         tracked
    improve-tc                            tracked
    linting                               tracked
    master                                tracked
    preference-experiment-disabled-fields tracked
    storage-refactor                      tracked
    wrapasync-this                        tracked
  Local branch configured for 'git pull':
    master merges with remote master
  Local ref configured for 'git push':
    master pushes to master (up to date)
```

Oh look, it's my fork! Which doesn't have any tags. Which is not surprising. I'm going to try falling back to the commit hash if the version isn't available, which should work for automation since we don't use these builds for anything except testing.

Osmose:
Looks like the missing version _was_ the culprit. Tests are passing, linting is now working (verified with a build earlier today), and I'm pretty happy with this PR.

> I thought this increase build times by a ton, but apparently not, as lints seem to be done within minutes of submitting a PR.

The tarball of the source code checkout pre-build is 300mb, and post-build is 1.9gb, which helps explain why transferring artifacts from the fetch tasks goes way faster than artifacts from the build task (it seems to scale linearly). 

A future PR could shave about 10 minutes off our build time (which took about 34 minutes, so about a 30% savings) by merging the test task with the built task, although that'd conflate the test and build logs again. It's worth considering.

recipe-client-addon/bin/tc/fetch.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+set -eu
+
+# mach wants this
+export SHELL=$(which bash)
```
mythmon: I don't think we need this anymore. I added it because running mach in m-c fails if `$SHELL` is not set. I don't think we're using mach here, so we can remove this.

Osmose:
Linting fix will come in a followup PR.

PR #694, opened 2017-04-17
Bug 1345650 - Backport preference fix

mythmon:
> [Bug 1345650](https://bugzilla.mozilla.org/show_bug.cgi?id=1345650) - use https://%(server)s/ for shield/self-support URLs to not break web-platform-tests on beta, and set shield-recipe-client.api_url to a working value in browser_RecipeRunner.js

This backports a change that was made during the beta merge cycle by @philor. Presumably it only needs `rs?` here.


mythmon:
My scripts got a bit over-zealous, I didn't mean to close this.

Issue #698, opened 2017-04-18
[Important] Heartbeat Prompt Repeating


rayborn:
We've had a few repeated prompt reports in the past (eg [this](https://bugzilla.mozilla.org/show_bug.cgi?id=1322425)), but unless my reading of Telemetry is very wrong (possible), this seems to be a more serious issue than I thought.

I noticed some repeated UUIDs in a non-repeated campaign, so I decided to generate a histogram of the number of dismissed prompt pings for the recipe on a per user bases.  Here's what I found:

| # Dismissals  |  # Users who dismissed N times  |  % Users who dismissed N times  |  % All Dismissals in Bucket  |
| -----  |  -----  |  -----  |  -----  |
| 1  |  8015  |  99.2%  |  96.0%  |
| 2  |  33  |  0.4%  |  0.8%  |
| 3  |  9  |  0.1%  |  0.3%  |
| 4  |  5  |  0.1%  |  0.2%  |
| 5  |  4  |  0.0%  |  0.2%  |
| 6  |  1  |  0.0%  |  0.1%  |
| 7  |  1  |  0.0%  |  0.1%  |
| 8  |  2  |  0.0%  |  0.2%  |
| 10  |  1  |  0.0%  |  0.1%  |
| 14  |  1  |  0.0%  |  0.2%  |
| 20  |  1  |  0.0%  |  0.2%  |
| 32  |  1  |  0.0%  |  0.4%  |
| 40  |  1  |  0.0%  |  0.5%  |
| 56  |  1  |  0.0%  |  0.7%  |

LMK if you have questions or would like to review the Gist.  Until investigated, this seems like a high priority issue.
I used this recipe 130 for this Spark job.

Osmose:
@rayborn A link to the gist would be very useful, as I'd like to see what events correspond do a dismissal in that table.

Osmose:
Rob confirmed via Slack that the issue is still happening after the merged fix.

Osmose:
@MattGrimes @rayborn So looking through the data dump of pings, we're not seeing anything that usefully correlates affected users. We still suspect it has something to do with the reliability of client-side storage.

I anticipate that we'll release the system add-on on the release channel on Wednesday (sampled, but hopefully by the end of the week we'll be at 100%). Given that the system add-on has a different mechanism for storage, the issue might be resolved by that.

Here's my proposal:

1. Launch the system add-on and check to see if we're still getting repeat heartbeats.
[...]

PR #699, opened 2017-04-18
#698: Treat malformed lastShown dates as valid in show-heartbeat.

Osmose:
Previously, we tried to parse the lastShown date when checking if the
user has ever seen a heartbeat prompt. If the parsing failed, we
treated the resulting value (NaN) as false, which triggered
reprompting the user.

With this change, we treat _any_ stored value for lastShown as meaning
that the user saw the prompt at least once. This avoids reprompting
users with buggy storage, as well as issues with our own code writing
incorrect values to localStorage.

### DIFF
--- a/recipe-server/client/actions/show-heartbeat/index.js
+++ b/recipe-server/client/actions/show-heartbeat/index.js
@@ -84,17 +84,16 @@ export default class ShowHeartbeatAction extends Action {
   }
 
   /**
-   * Checks when this survey was last shown,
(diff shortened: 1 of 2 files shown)


mythmon:
Taskcluster's failure is the lint issue fixed in #696.

Osmose:
> Taskcluster's failure is the lint issue fixed in #696.

Yeah, I didn't pull master before branching. My bad. This is good to merge.

PR #697, opened 2017-04-17
recipe-client-addon: Convert PreferenceExperiments tests to mochitest.

mythmon:
I'm having some trouble here. Most of the tests are passing, but a few aren't, and I haven't been able to figure out why.

Here are the failing tests when I run locally:

```
405 INFO TEST-UNEXPECTED-FAIL | browser/extensions/shield-recipe-client/test/browser/browser_PreferenceExperiments.js | start did not modify the default preference - "newvalue" != "newvalue" - JS frame :: chrome://mochitests/content/browser/browser/extensions/shield-recipe-client/test/browser/browser_PreferenceExperiments.js :: <TOP_LEVEL> :: line 156
Stack trace:
chrome://mochitests/content/browser/browser/extensions/shield-recipe-client/test/browser/browser_PreferenceExperiments.js:null:156

406 INFO TEST-UNEXPECTED-FAIL | browser/extensions/shield-recipe-client/test/browser/browser_PreferenceExperiments.js | stop removed an observer - 
Stack trace:
chrome://mochitests/content/browser/browser/extensions/shield-recipe-client/test/browser/browser_PreferenceExperiments.js:null:344

407 INFO TEST-UNEXPECTED-FAIL | browser/extensions/shield-recipe-client/test/browser/browser_PreferenceExperiments.js | stop did not bother to stop an observer that wasn't active - 
Stack trace:
chrome://mochitests/content/browser/browser/extensions/shield-recipe-client/test/browser/browser_PreferenceExperiments.js:null:387

408 INFO TEST-UNEXPECTED-FAIL | browser/extensions/shield-recipe-client/test/browser/browser_PreferenceExperiments.js | stop removed the preference that had no value prior to the experiment - 
Stack trace:
chrome://mochitests/content/browser/browser/extensions/shield-recipe-client/test/browser/browser_PreferenceExperiments.js:null:406
```

PR #700, opened 2017-04-19
Merge master into preference-experiment

Osmose:
This was actually a clean merge. Given that, I'll just land this once automation passes. Let me know if ya'll think that's not something we want to do going forward.

### DIFF
--- a/ci/taskcluster/bin/decision.py
+++ b/ci/taskcluster/bin/decision.py
@@ -9,9 +9,35 @@ BASE_URL = 'http://taskcluster/queue/v1'
 
 tasks = [
     {
+        'name': 'recipe-client-addon:fetch',
+        'description': 'Download gecko-dev and sync Normandy changes',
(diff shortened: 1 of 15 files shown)


Osmose:
Welp, TaskCluster failed. It failed linting and tests, likely due to the fact that these weren't working properly until #693. #697 includes test fixes along with the refactor to mochitest, so I merged that, and the tests passed locally. I also added a commit to fix the new linting errors. 9625c2c is the only code unique to this PR.

@mythmon @gijsk I'm leaving this up to see if automation passes with all the fixes, but I'm not sure quite the order to land this. Should we r+ #697, review the lint fix commit in this PR, and then land this? Should we land #697 with broken automation and then land this PR afterwards?

Osmose:
@gijsk The tests passed, so I merged #697 and remade this to just include the merge and lint fixes. This just needs a tiny review on the lint commit now. Thanks!

mythmon:
There's no button for it, but if you add `?w=1`to the url of the diff view, you'll get it.

PR #703, opened 2017-04-20
recipe-client-addon: Use our .eslintrc.js in mozilla-central

mythmon:
We didn't include our config in mozilla-central out of oversight. This patch includes our config into mozilla-central, so our linters agree, and also changes our lint config slightly to make it compatible with mozilla-central's eslint.

I also fixed a couple lint problems while I was at it.


mythmon:
Changes in upstream are an issue even without this that I'd like to figure out a fix for eventually anyways.

We run `./mach lint` using recent versions of mozilla-central as a part of our Taskcluster builds on every commit and PR, so if anyone changes the mozilla/recommended rules, we should pick up on those pretty quick. Hopefully that will be enough.

mythmon:
I'm having troubles here getting our other linting infra to accept the changes that were needed to make this run in mozilla-central.

@Osmose suggested not having repo wide linting, and making it per app. I think this is a good idea, and would make this situation potentially easier. I'm going to close this PR with the goal of doing that linting change and then re-applying this change.

PR #704, opened 2017-04-21
Implement refactored state tree

### LINKED ISSUE #689: Create normalized redux state tree
Create a new normalized state tree in redux as part of the JS refactor.

rehandalal:
Fixes #689 

r?


recipe-server/client/control/utils/apiFetch.js
```diff
@@ -0,0 +1,49 @@
+/* global document, fetch, Headers */
```
mythmon: These should be listed in `recipe-server/.eslintrc` instead.

recipe-server/.eslintrc
```diff
@@ -22,6 +22,7 @@ rules:
   no-underscore-dangle: [off]
   no-use-before-define: [warn, { functions: false }]
   prefer-const: [warn]
+  new-cap: [off]
```
mythmon: This is for Immutable, right? I think I'd prefer to use `new Map()` and `new List()` instead, and keep this rule. In my testing it works the same (though maybe I've missed something?)

recipe-server/client/control/state/actions/actions.js
```diff
@@ -0,0 +1,97 @@
...
+  ACTION_FETCH,
+  ACTION_FETCH_FAILURE,
+  ACTION_FETCH_SUCCESS,
+  ACTION_RECEIVE,
+  ACTIONS_FETCH,
+  ACTIONS_FETCH_FAILURE,
+  ACTIONS_FETCH_SUCCESS,
+} from '../action-types';
+
+import apiFetch from '../../utils/apiFetch';
```
andymikulski: Can this be absolute path'd via the webpack config's `resolve.alias` property?

recipe-server/client/control/state/actions/actions.js
```diff
@@ -0,0 +1,97 @@
...
+function fetchActionsFailure(dispatch, requestId, error) {
+  dispatch({
+    type: ACTIONS_FETCH_FAILURE,
+    error,
+    requestId,
+  });
+}
+
+
+export function fetchActions() {
```
andymikulski: We have `fetchAction` and `fetchActions`, could this change to `fetchAllActions` or something for bit more clarity?

recipe-server/client/control/state/action-types.js
```diff
@@ -0,0 +1,38 @@
+/* Keep this list alphabetized */
+
+export const ACTION_FETCH = 'ACTION_FETCH';
```
andymikulski: In the past, I've done something like this to achieve the same effect, but with a little brevity:

```
['HI', 'THERE', 'ETC'].forEach((eventName)=>{
	module.exports[eventName] = eventName;
});

// then you can pull all of them as an object
import ActionTypes from 'path/to/action-types';
// or individually
import {ACTION_FETCH} from 'path/to/action-types';
```

recipe-server/client/control/state/approvalRequests/actions.js
```diff
@@ -0,0 +1,97 @@
...
+  });
+}
+
+
+export function fetchApprovalRequest(pk) {
+  return (dispatch, getState) => {
+    const requestId = `fetch-${pk}`;
+    const request = getRequest(getState(), requestId);
+
+    if (request.loading) { return true; }
```
andymikulski: I'm not sure how `Immutable#getIn` works - will it always return an object? If not, we will need to safeguard this conditional (`if (request && request.loading){ ...`)

recipe-server/client/control/state/approvalRequests/actions.js
```diff
@@ -0,0 +1,97 @@
...
+function fetchApprovalRequestsFailure(dispatch, requestId, error) {
+  dispatch({
+    type: APPROVAL_REQUESTS_FETCH_FAILURE,
+    error,
+    requestId,
+  });
+}
+
+
+export function fetchApprovalRequests() {
```
andymikulski: Same issue I had with another function - we have `fetchApprovalRequest` and `fetchApprovalRequests`, which are nearly identical. Maybe it's just me, but I find functions with such similar names to be a little uncomfortable.

In general, I'm not sure if I'm totally on board with all of the separated `request/requests` logic. It's not super DRY, and in fact a little confusing, since the code is pretty much the same (except for a few `s`'s).

Couldn't we just set up the actions to handle single requests just the same as multiple? I don't think I see the benefit to handling singular vs multiple responses differently.

recipe-server/client/control/state/approvalRequests/reducers.js
```diff
@@ -0,0 +1,58 @@
...
+  APPROVAL_REQUEST_FETCH_FAILURE,
+  APPROVAL_REQUEST_FETCH_SUCCESS,
+  APPROVAL_REQUEST_RECEIVE,
+  APPROVAL_REQUESTS_FETCH,
+  APPROVAL_REQUESTS_FETCH_FAILURE,
+  APPROVAL_REQUESTS_FETCH_SUCCESS,
+} from '../action-types';
+
+
+function objects(state = Map({}), action) {
```
andymikulski: I'm not in love with using `objects` as a state key - it's a little vague, and tickles my `reserved keyword` sense, as it's so close to `Object`. Could this be named to something more descriptive?

recipe-server/client/control/state/approvalRequests/actions.js
andymikulski: After reading more code, I see that the reducers actually _do_ handle single/multiple responses the same way. Is there a reason to keep the actions separate?

recipe-server/client/control/state/actions/actions.js
```diff
@@ -0,0 +1,97 @@
...
+  ACTIONS_FETCH,
+  ACTIONS_FETCH_FAILURE,
+  ACTIONS_FETCH_SUCCESS,
+} from '../action-types';
+
+import apiFetch from '../../utils/apiFetch';
+
+
+function fetchActionSuccess(dispatch, requestId, action) {
+  dispatch({
```
rehandalal: yep makes sense.

recipe-server/.eslintrc
rehandalal: I'll make the change.

recipe-server/client/control/state/actions/actions.js
rehandalal: Possibly, I'll look into it.

recipe-server/client/control/state/actions/actions.js
rehandalal: I prefer the less verbose way, but I'll defer to everyone for their opinions.

recipe-server/client/control/state/approvalRequests/actions.js
rehandalal: if you look at the selector it returns an object with the loading property as a default, so no safeguard required

recipe-server/client/control/state/approvalRequests/actions.js
rehandalal: I'm going to go with @mythmon's suggestion and move all requests out into their own little branch of the state tree.

recipe-server/client/control/state/approvalRequests/reducers.js
rehandalal: I'm not tied to `objects` it's just what I used on Morgoth and it sort of matches the Django ORM, so it feels familiar to me, but I don't have strong feelings about this.

recipe-server/client/control/state/action-types.js
mythmon: I don't think we should use `module.exports`. It makes more typing, but it makes the code easier to understand, both for humans and for linters. For example, this proposal will break the import linting we have that detects if we typo a module import.

recipe-server/client/control/state/actions/actions.js
```diff
@@ -0,0 +1,43 @@
...
+      action,
+    });
+  };
+}
+
+
+export function fetchAction(name) {
+  const requestId = `fetch-${name}`;
+  return makeApiRequest(requestId, `action/${name}/`, { method: 'GET' })
+    .then(action => receiveAction(action));
```
mythmon: Is `receiveAction` used anywhere besides here? I think breaking this out into its own method reduces readability.

recipe-server/client/control/state/actions/actions.js
```diff
@@ -0,0 +1,43 @@
...
+      });
+    });
+  };
+}
+
+
+export function fetchAllActions() {
+  const requestId = 'fetch-all';
+  return makeApiRequest(requestId, 'action/', { method: 'GET' })
+    .then(actions => receiveAllActions(actions));
```
mythmon: I think this would be better written using async/await, like this:

```js
export async function fetchAllActions() {
  const requestId = 'fetch-actions-all';
  const actions = await makeApiRequests(requestId, 'action/', { method: 'GET' });

  for (const action of actions) {
    dispatch({
      type: ACTION_RECEIVE,
      action,
    });
  }
}
```

recipe-server/client/control/state/actions/actions.js
```diff
@@ -0,0 +1,43 @@
...
+    dispatch({
+      type: ACTION_RECEIVE,
+      action,
+    });
+  };
+}
+
+
+export function fetchAction(name) {
+  const requestId = `fetch-${name}`;
```
mythmon: This `requestId` is not part of a global namespace now, right? So should it be something  like `fetch-action-${name}`?

recipe-server/client/control/state/approvalRequests/selectors.js
```diff
@@ -0,0 +1,5 @@
+/* eslint import/prefer-default-export: "off" */
+
+export function getApprovalRequest(state, id, defaultsTo) {
+  return state.approvalRequests.items.get(id, defaultsTo);
```
mythmon: Should this include `newState, `like `state.newState.approvalRequests`?

recipe-server/client/control/utils/apiFetch.js
```diff
@@ -0,0 +1,47 @@
...
+
+  return fetch(`${API_ROOT}${url}${queryString}`, settings)
+    .then(response => Promise.all([response.ok, response.json()]))
+    .catch(error => [false, { message: error.message }])
+    .then(([ok, data]) => {
+      if (!ok) {
+        throw data;
+      }
+      return data;
+    });
```
mythmon: This would proabably be a bit cleaner with async/await.

recipe-server/client/control/state/actions/actions.js
rehandalal: d'oh! yes

recipe-server/client/control/state/actions/actions.js
andymikulski: I'm actually fine with abstracting into its own function - skimming the code, it's faster for me to parse what's going on, and enforces separation of concerns. I prefer the 'redirect' to another method over reading a few lines of an inlined function to determine what it's actually doing, but maybe that's just me. Also, refactoring or adding to `recieveAction` would be cleaner if we ever decided to do so in the future.

But, the line itself could be rewritten to just `.then(receiveAction)`, which I would suggest.

recipe-server/client/control/state/recipes/actions.js
```diff
@@ -23,11 +23,11 @@ function fetchRecipeSuccess(recipe) {
 export function fetchRecipe(pk) {
   const requestId = `fetch-${pk}`;
   return makeApiRequest(requestId, `recipe/${pk}/`, { method: 'GET' })
-    .then(recipe => fetchRecipeSuccess(recipe));
+    .then(recipe => receiveRecipe(recipe));
```
andymikulski: All the single-argument handlers like this can be rewritten to just `.then(whateverHandlerFunc)`

recipe-server/client/control/state/approvalRequests/selectors.js
rehandalal: oops! yes!

rehandalal:
@andymikulski i agree we should settle on one convention going forward. i think that should be async/await.

recipe-server/client/control/utils/apiFetch.js
```diff
@@ -35,13 +35,6 @@ export default function apiFetch(url, options = {}) {
...
-    .then(response => Promise.all([response.ok, response.json()]))
-    .catch(error => [false, { message: error.message }])
-    .then(([ok, data]) => {
-      if (!ok) {
-        throw data;
-      }
-      return data;
-    });
+  const response = await fetch(`${API_ROOT}${url}${queryString}`, settings);
+  return response.json();
```
andymikulski: Technically, the `await fetch(...)` business should be wrapped in a `try/catch` block to handle errors.

rehandalal: we want this to throw errors so we can catch them later:
https://github.com/mozilla/normandy/pull/704/files#diff-b2daff5c1e0fc92e2f54a8cd981c4d3aR29

recipe-server/client/control/state/actions/actions.js
```diff
@@ -7,8 +7,12 @@ import {
 } from '../requests/actions';
 
 
-function receiveAction(action) {
-  return dispatch => {
+export function fetchAction(name) {
+  return async dispatch => {
+    const requestId = `fetch-action-${name}`;
+    const response = dispatch(makeApiRequest(requestId, `action/${name}/`, { method: 'GET' }));
+    const action = await response;
```
mythmon: What's the benefit of storing the promise and then awaiting it, instead of doing it all at once?

```js
const action = dispatch(makeApiRequest(requestId, `action/${name}/`, { method: 'GET' }));
```

PS: Can we make `method: 'GET'` the default for makeApiRequest?

rehandalal: oh i guess there's no real benefit i was just trying to keep the line wrap down and this looked a little less cluttered to me. but i can ditch the extra line if you'd prefer to see it that way. i'm not tied to this.

rehandalal: and yeah, defaulting `method: 'GET'` makes sense. I'll do that now.

rehandalal:
@andymikulski if this is good for you we can go ahead and :ship: it!

andymikulski:
@rehandalal looks like there's a conflict with `recipe-server/normandy/recipes/tests/test_api.py`

recipe-server/normandy/recipes/api/serializers.py
```diff
@@ -87,9 +110,21 @@ class Meta:
             'filter_expression',
             'latest_revision_id',
             'approved_revision_id',
+            'latest_revision',
+            'approved_revision',
             'approval_request',
         ]
 
+    def __init__(self, *args, **kwargs):
```
Osmose: This feels like it should be a mixin.

recipe-server/normandy/recipes/api/serializers.py
```diff
@@ -67,6 +88,8 @@ class RecipeSerializer(serializers.ModelSerializer):
     filter_expression = serializers.CharField(read_only=True)
     latest_revision_id = serializers.CharField(source='latest_revision.id', read_only=True)
     approved_revision_id = serializers.CharField(source='approved_revision.id', read_only=True)
+    latest_revision = RecipeRevisionSerializer(read_only=True)
+    approved_revision = RecipeRevisionSerializer(read_only=True)
```
Osmose: I might by missing something, but it seems like we declare these to be denormalized, and then everywhere we use them, we replace the denormalized data with an ID. Could we make these `latest_revision_id` and `approved_revision_id` instead?

rehandalal:
@mythmon I reassigned you because I think there were enough changes since your last review to need a look over the changes.

recipe-server/client/control/tests/state/actions/test_reducers.js
```diff
@@ -0,0 +1,28 @@
...
+    expect(actionsReducer(undefined, {})).toEqual(INITIAL_STATE);
+  });
+
+  it('should handle ACTION_RECEIVE', () => {
+    expect(actionsReducer(undefined, {
+      type: ACTION_RECEIVE,
+      action: ACTION,
+    })).toEqual({
+      ...INITIAL_STATE,
+      items: INITIAL_STATE.items.set(ACTION.id, fromJS(ACTION)),
```
Osmose: This feels a bit like writing the same code twice as a test... but it tests the matching, and behavior over time, and the only other thing I can think of would be doing a `fromJS` over replacement initial state, which doesn't seem great.

I think this is just what feelings feel like. I'm with this.

recipe-server/client/control/tests/state/actions/test_selectors.js
```diff
@@ -0,0 +1,37 @@
...
+
+
+describe('getAction', () => {
+  const STATE = {
+    ...INITIAL_STATE,
+    newState: {
+      ...INITIAL_STATE.newState,
+      actions: {
+        ...INITIAL_STATE.newState.actions,
+        items: INITIAL_STATE.newState.actions.items.set(ACTION.id, fromJS(ACTION)),
```
Osmose: Meet [setIn](https://facebook.github.io/immutable-js/docs/#/Map/setIn). He understands your pain.

Except, wait, can you use this? The state is not immutable until you get to collections?

recipe-server/client/control/tests/state/revisions/test_reducers.js
```diff
@@ -0,0 +1,28 @@
...
+  REVISION,
+} from '.';
+
+
+describe('Revisions reducer', () => {
+  it('should return initial state by default', () => {
+    expect(revisionsReducer(undefined, {})).toEqual(INITIAL_STATE);
+  });
+
+  it('should handle REVISION_RECEIVE', () => {
```
Osmose: The test descriptions for reducers could use a bit more detail, like "`REVISION_RECEIVE` should add the revision to state.items".

recipe-server/client/control/tests/state/actions/test_selectors.js
rehandalal: yep! me and setIn are already best buds.

rehandalal: Making the whole tree immutable is a bit janky with the reducers from redux-form etc.

rehandalal:
unassigning @mythmon and assigning @andymikulski for a quick pass over the changes since the last review.

PR #705, opened 2017-04-24
recipe-server: Add build time to __version__ endpoint

mythmon:
This came from a desire to know when we deployed a version to prod. It isn't quite the same, as we build the docker images at least an hour before they actually get deployed. This would also show the same time on both stage and prod, and won't take into account re-deploys. This is better than nothing though.

The code here is starting to get a little unwieldy, but I think the best next step is to start using [python-dockerflow](https://github.com/mozilla-services/python-dockerflow) instead of continuing on our own custom thing.

@milescrabill Does this look like an ok thing to add to our version endpoint? Is it possible to get the actual deploy time as an envvar or a file injected into our image?

### DIFF
--- a/recipe-server/normandy/health/api/views.py
+++ b/recipe-server/normandy/health/api/views.py
@@ -21,6 +21,13 @@ logger = logging.getLogger(__name__)
 
 @lru_cache(maxsize=1)
 def get_version_info():
+    """
+    Gets data about the current build.
+
+    This pulls the data from files in a directory named __version__ in
+    the root of the project. The files in this directory are generated
+    during the CI build by the script `bin/ci/dependencies`.
+    """
 
     def fetch_key(name):
         path = os.path.join(settings.BASE_DIR, '__version__', name)
@@ -48,6 +55,9 @@ def get_version_info():
 
 @api_view(['GET'])
 def version(request):
+    """
+    Return data for developers and users to inspect the state of a server.
+    """
     repo_url = 'https://github.com/mozilla/normandy'
     version_info = get_version_info()
 
--- a/recipe-server/normandy/health/api/views.py
+++ b/recipe-server/normandy/health/api/views.py
@@ -1,6 +1,8 @@
 import json
 import logging
 import os
+from functools import lru_cache
+from datetime import datetime
 
 from django.conf import settings
 from django.core.checks.registry import registry as checks_registry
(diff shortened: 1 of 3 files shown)


milescrabill:
This mostly lgtm and is a good idea. Getting an actual deploy time injected as an env var or file is possible but requires some hoopjumping. If AMI build time would be sufficient, that would be easier (but would also be the same for stage and prod), and also similar to this number. Another option that might be worth exploring is [deployment tracking with New Relic](https://docs.newrelic.com/docs/apm/applications-menu/events/deployments-page) which can give a fairly fine-grained log of deployments/revisions/performance.

mythmon:
New Relic deploy monitoring would be good, but I still think it is valuable to have this info in __version__.

recipe-server/normandy/health/tests/test_api.py
```diff
@@ -42,39 +50,66 @@ def version_info_dir(self, tmpdir, settings):
...
+    @pytest.fixture
+    def build_time_file(self, version_info_dir):
+        f = version_info_dir.join('build_time')
+        ts = 1493051820
+        f.write('1493051820\n')
+        f.datetime = datetime.utcfromtimestamp(ts)
+        return f
+
+    @pytest.fixture
+    def clear_version_info(self):
```
Osmose: I think this would be better suited to a context manager than a pytest fixture, given that it does cleanup after running some code.

recipe-server/normandy/health/tests/test_api.py
```diff
@@ -42,39 +50,66 @@ def version_info_dir(self, tmpdir, settings):
...
+        """Clear the cache for get_version_info() after each test run"""
+        try:
+            yield
+        finally:
+            views.get_version_info.cache_clear()
+
+    def test_with_everything(self, commit_file, tag_file, build_time_file, clear_version_info):
```
Osmose: Would a [parameterized test](https://docs.pytest.org/en/latest/parametrize.html#parametrize-basics) fit well here?

recipe-server/normandy/health/tests/test_api.py
```diff
@@ -18,6 +25,7 @@ def test_version(client, mocker):
         'commit_link': 'https://github.com/mozilla/normandy/commit/<git hash>',
         'configuration': 'Test',
         'version': '<tag>',
+        'build_time': Whatever(),
```
Osmose: We could also just save the datetime above and compare it to `dt.isoformat()` here.

recipe-server/normandy/health/api/views.py
```diff
@@ -15,42 +17,51 @@
...
+        'commit': fetch_key('commit'),
+        'version': fetch_key('tag'),
+        'build_time': build_time,
+    }
+
+    return version_info
 
 
 @api_view(['GET'])
 def version(request):
```
Osmose: In a strong tradition of asking you to fix code you didn't touch, this could use a docstring explaining the purpose of the view.

recipe-server/normandy/health/api/views.py
```diff
@@ -15,42 +17,51 @@
 
 
 logger = logging.getLogger(__name__)
-_commit = None
-_tag = None
 
 
+@lru_cache(maxsize=1)
 def get_version_info():
```
Osmose: This could also use a docstring explaining that it pulls data about the current build. It'd be great to reference the filename of the script generating the files, too.

mythmon:
CC @relud

relud:
> Is it possible to get the actual deploy time as an envvar

Yeah, we can make a cloudformation parameter for it, and have ansible calculate it when that is deployed. We could also define it in jenkins and pass it on down the chain. Would be maybe a 10 line change on our end.

recipe-server/normandy/health/tests/test_api.py
mythmon: That would be nice. I tried it, but `commit_file` and friends are fixtures, and for now [you can't use fixtures and paramterize together](https://github.com/pytest-dev/pytest/issues/349).

:cry:

PR #714, opened 2017-04-25
recipe-client-addon: Lazy load applicable modules in RecipeRunner

mythmon:
This file is loaded at browser startup, so lazily loading modules that aren't needed immediately drastically decreases the amount of time we block startup. In my tests, this patch changed the time we blocked startup from 22ms to 6ms.

@kmaglione To ease integration, we require that changes to the add-on in this repo get reviewed by a Firefox peer. Can you review this as if it were a patch to m-c?

### DIFF
--- a/recipe-client-addon/lib/RecipeRunner.jsm
+++ b/recipe-client-addon/lib/RecipeRunner.jsm
@@ -6,22 +6,31 @@
 
 const {utils: Cu} = Components;
 Cu.import("resource://gre/modules/Services.jsm");
-Cu.import("resource://shield-recipe-client/lib/LogManager.jsm");
-Cu.import("resource://shield-recipe-client/lib/NormandyDriver.jsm");
-Cu.import("resource://shield-recipe-client/lib/FilterExpressions.jsm");
-Cu.import("resource://shield-recipe-client/lib/NormandyApi.jsm");
-Cu.import("resource://shield-recipe-client/lib/SandboxManager.jsm");
-Cu.import("resource://shield-recipe-client/lib/ClientEnvironment.jsm");
-Cu.import("resource://shield-recipe-client/lib/CleanupManager.jsm");
-Cu.import("resource://shield-recipe-client/lib/ActionSandboxManager.jsm");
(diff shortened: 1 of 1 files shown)


recipe-client-addon/lib/RecipeRunner.jsm
```diff
@@ -6,22 +6,31 @@
...
-XPCOMUtils.defineLazyModuleGetter(this, "Preferences", "resource://gre/modules/Preferences.jsm");
-XPCOMUtils.defineLazyModuleGetter(this, "Storage",
-                                  "resource://shield-recipe-client/lib/Storage.jsm");
 XPCOMUtils.defineLazyServiceGetter(this, "timerManager",
                                    "@mozilla.org/updates/timer-manager;1",
                                    "nsIUpdateTimerManager");
+XPCOMUtils.defineLazyModuleGetter(this, "Preferences", "resource://gre/modules/Preferences.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "Storage",
+                                  "resource://shield-recipe-client/lib/Storage.jsm");
+XPCOMUtils.defineLazyServiceGetter(this, "NormandyDriver",
```
Osmose: You want `defineLazyModuleGetter` instead of `defineLazyServiceGetter`.

mythmon: Doh. Fixed.

Osmose:
Out of curiosity, did you re-run the profiling after fixing the patch? Did we see the same savings?

mythmon:
[Profile without lazy loading](https://perfht.ml/2oMvX09)

[Profile with lazy loading](https://perfht.ml/2q8ieEw)

When searching for "shield" on the main process, the top level item in the "without fix" profile is 17ms. When doing the same search on the "with fix" profile, that is 3ms.

kmaglione:
> When searching for "shield" on the main process, the top level item in the "without fix" profile is 17ms. When doing the same search on the "with fix" profile, that is 3ms.

Nice. Thanks!

lgtm

PR #720, opened 2017-04-26
Add Rollout UI

### LINKED ISSUE #712: Add UI for Rollout Actions
Arguments:
- [x] Rollout name (string)
- [x] Preference name (string)
- [x] New value (string|int|bool)
- [x] Branch (string, 'default' or 'user')

andymikulski:
Fixes #712 

---

screen shot 2017-04-26 at 11 48 41 am

This is pretty simple, but figured the PR was ready for a quick pass before it gets any bigger. Some notes:

- When the user selects the `user` preference branch, a checkbox appears to confirm that the user understands the ramifications (`I understand this will override existing user-set preferences.`)

- The selfrepair'd action has just been stubbed out at this point. The driver needs a bunch of things added to it, which I'd consider out of scope for this PR (but maybe I'm wrong!)

### DIFF
--- a/recipe-server/client/actions/preference-rollout/package.json
+++ b/recipe-server/client/actions/preference-rollout/package.json
@@ -28,16 +28,17 @@
                     "type": ["string", "number", "boolean"],
                     "default": ""
                 },
+                "preferenceType": {
+                    "description": "Data type of the preference that controls this experiment",
+                    "type": "string",
+                    "enum": ["string", "integer", "boolean"],
+                    "default": "boolean"
+                },
                 "preferenceBranch": {
                     "descript": "Controls whether the default or user value of the preference is modified",
                     "type": "string",
                     "enum": ["user", "default"],
                     "default": "default"
-                },
-                "branchConfirmation": {
-                    "descript": "Controls whether the default or user value of the preference is modified",
-                    "type": "boolean",
-                    "default": false
                 }
             }
         }
--- a/recipe-server/client/actions/preference-rollout/package.json
+++ b/recipe-server/client/actions/preference-rollout/package.json
@@ -1,0 +1,45 @@
+{
+    "name": "preference-rollout",
+    "version": "0.0.1",
+    "private": true,
+    "main": "./index.js",
+    "normandy": {
+        "driverVersion": "1.x",
+        "argumentsSchema": {
+            "$schema": "http://json-schema.org/draft-04/schema#",
+            "title": "Change a preference for a portion of the population",
+            "type": "object",
+            "required": [
+                "message"
+            ],
+            "properties": {
+                "name": {
+                    "description": "Rollout name",
+                    "type": "string",
+                    "default": ""
+                },
+                "preferenceName": {
+                    "description": "Preference to change",
+                    "type": "string",
+                    "default": ""
+                },
+                "preferenceValue": {
+                    "description": "Value to set preference to.",
+                    "type": ["string", "number", "boolean"],
+                    "default": ""
+                },
+                "preferenceBranch": {
+                    "descript": "Controls whether the default or user value of the preference is modified",
+                    "type": "string",
+                    "enum": ["user", "default"],
+                    "default": "default"
+                },
+                "branchConfirmation": {
+                    "descript": "Controls whether the default or user value of the preference is modified",
+                    "type": "boolean",
+                    "default": false
+                }
+            }
+        }
+    }
+}
--- a/recipe-server/client/control/components/RecipeForm.js
+++ b/recipe-server/client/control/components/RecipeForm.js
@@ -37,10 +37,14 @@ import {
 import composeRecipeContainer from 'control/components/RecipeContainer';
 import { ControlField } from 'control/components/Fields';
 import RecipeFormActions from 'control/components/RecipeFormActions';
+
 import HeartbeatFields from 'control/components/action_fields/HeartbeatFields';
 import ConsoleLogFields from 'control/components/action_fields/ConsoleLogFields';
 import PreferenceExperimentFields from
   'control/components/action_fields/PreferenceExperimentFields';
+import RolloutFields from 'control/components/action_fields/RolloutFields';
+
+
 import RecipeStatus from 'control/components/RecipeStatus';
 import DraftStatus from 'control/components/DraftStatus';
 import BooleanIcon from 'control/components/BooleanIcon';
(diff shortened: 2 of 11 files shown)


recipe-server/client/control/components/RecipeForm.js
```diff
@@ -37,10 +37,14 @@ import {
...
 import { ControlField } from 'control/components/Fields';
 import RecipeFormActions from 'control/components/RecipeFormActions';
+
 import HeartbeatFields from 'control/components/action_fields/HeartbeatFields';
 import ConsoleLogFields from 'control/components/action_fields/ConsoleLogFields';
 import PreferenceExperimentFields from
   'control/components/action_fields/PreferenceExperimentFields';
+import RolloutFields from 'control/components/action_fields/RolloutFields';
+
+
```
mythmon: What's the pattern in the spacing here?

recipe-server/client/control/components/action_fields/RolloutFields.js
```diff
@@ -0,0 +1,78 @@
...
+    } = this.props;
+
+    return (
+      <div className="arguments-fields">
+        <p className="info">
+          Change a preference for a portion of the population.
+          <br /><br />
+          Supports gradually increasing the number of users that get a feature, multiple features
+          rolling out at once to in/dependent groups, and rolling back a deployment entirely
+          (if problems arise).
```
mythmon: Most of this is actually about filter expressions. I'm not sure we need it here.

recipe-server/client/control/components/action_fields/HeartbeatFields.js
```diff
@@ -117,6 +121,6 @@ export class HeartbeatFields extends ActionFields {
 
 export default connect(
   state => ({
-    recipeArguments: selector(state, 'arguments'),
+    recipeArguments: selector(state, 'arguments') || {},
```
mythmon: The changes in this file are unrelated to preference rollouts, right? I like the changes, but I don't think they belong in this PR. If nothing else, they are a source of conflict when we merge the feature branch in.

recipe-server/client/actions/preference-rollout/package.json
```diff
@@ -0,0 +1,45 @@
...
+                    "type": ["string", "number", "boolean"],
+                    "default": ""
+                },
+                "preferenceBranch": {
+                    "descript": "Controls whether the default or user value of the preference is modified",
+                    "type": "string",
+                    "enum": ["user", "default"],
+                    "default": "default"
+                },
+                "branchConfirmation": {
```
mythmon: Does this need to be part of the arguments? I don't think any consumer of this will need this value, and it is mostly a UI treatment. I think it would be better to not include it here if that is possible.

recipe-server/client/control/components/RecipeForm.js
andymikulski: Sorry - I've rearranged this to match what I was actually intending. I've grouped the general UI components together, and grouped the `ActionField`s, eg:

```
...
import DraftStatus from 'control/components/DraftStatus';
import BooleanIcon from 'control/components/BooleanIcon';

import HeartbeatFields from 'control/components/action_fields/HeartbeatFields';
import ConsoleLogFields from 'control/components/action_fields/ConsoleLogFields';
...
```

recipe-server/client/control/components/action_fields/RolloutFields.js
andymikulski: Gotcha. The UX part of me thinks it'd be a good idea to have a little blurb describing the action, especially now that there is Preference Rollout _and_ Preference Experiments. Perhaps it could be a short bit of copy explaining the difference between `Rollout`s and `Experiment`s?

recipe-server/client/control/components/action_fields/HeartbeatFields.js
andymikulski: I had gotten errors/warnings when testing the RecipeForm, so I felt compelled to fix 'em. I can pull the changes out into a different PR, though.

recipe-server/client/actions/preference-rollout/package.json
andymikulski: I was actually getting a weird redux-form error on the RecipeForm if it _wasn't_ an argument. From memory, the error said something wrt a non-numeric array index, and would actually throw a hard error and mess with redux-form. I can dig into it some more.

recipe-server/client/control/components/action_fields/RolloutFields.js
mythmon: That sounds like a good idea. We're going to need to write lots of docs for this, and this is a good place to put it. As long as the docs are actually about preference rollout, and not separate features that apply to all recipes. Maybe also including a link to docs on readthedocs would be good (that doesn't have to be in this PR)

recipe-server/client/control/components/action_fields/HeartbeatFields.js
mythmon: That would be good, especially for feature branches. Osmose and I ran into some trouble on the pref experiments branch because we would do unrelated fixes on the wrong branch.

recipe-server/client/actions/preference-rollout/package.json
mythmon: That's weird. It's fine to stay here if we can't figure it out. I wouldn't spend much time on it if you're getting stuck.

recipe-server/client/control/components/action_fields/HeartbeatFields.js
andymikulski: Sounds good. Would that PR be onto this same feature branch (`preference-rollout`), or onto `master`?

mythmon: I think this should go to master, since it isn't related to preference rollouts.

recipe-server/client/actions/preference-rollout/package.json
andymikulski: Something else I had considered was after the recipe is created, what happens with this confirmation checkbox? Do we require it to be checked before submitting any changes to a `user`-branch rollout? Or should it be a one-and-done deal?

mythmon: Yeah, I'm not sure about that. If it is stored in arguments, then we kind of get that for free. If we don't do that then things get tricky. I guess it could be only for new recipes?

Alternatively, can we just have a big warning box here, or something else instead of a check box?

mythmon:
This is marked as `f?` (feedback) and `wip`. What still needs done before we can remove the `wip` tag? Can we add a todo list to the top comment listing those things?

andymikulski:
This is ready for review. I've corrected the feedback points from mythmon, added tests, and corrected some experiment fields (removed confirmation field, added pref type field).


screen shot 2017-04-28 at 4 12 21 pm

Issue #718, opened 2017-04-26
Try build fails due to duplicate bootstrap.js


Osmose:
A [try build](https://treeherder.mozilla.org/#/jobs?repo=try&revision=cf86a39cdddd3f8915d4fcbf4f185b453e2c2275&selectedJob=94368141) of master synced to mozilla-central failed with this message in the logs:

```
[task 2017-04-26T02:30:53.854588Z] 02:30:53     INFO -  ERROR: The following duplicated files are not allowed:
[task 2017-04-26T02:30:53.854756Z] 02:30:53     INFO -  browser/features/shield-recipe-client@mozilla.org/bootstrap.js
[task 2017-04-26T02:30:53.854935Z] 02:30:53     INFO -  browser/features/shield-recipe-client@mozilla.org/chrome/content/bootstrap.js
[task 2017-04-26T02:30:53.855123Z] 02:30:53     INFO -  /home/worker/workspace/build/src/toolkit/mozapps/installer/packager.mk:39: recipe for target 'stage-package' failed
[task 2017-04-26T02:30:53.855274Z] 02:30:53     INFO -  gmake[7]: *** [stage-package] Error 1
```

Looks like this is related to [my changes to jar.mn](https://github.com/mozilla/normandy/commit/3e09dcea8fbeb336e1590b6d0fcd3f78348e3535#diff-7f782ee3d7cb6ae1101b946a214af6fc). I made that change so that `bootstrap.js` could be imported during tests. Apparently the step that finds and eliminates duplicates isn't being run during our TC builds; I can't find anything about it in the logs.

@gijsk Tomorrow I'm going to try and read up on the build system so I can understand what jar.mn actually does (I get that it helps build a jarfile, but I dunno what that jarfile is used for, or how it relates to system add-ons). If you have any links to useful docs, or any ideas on how to fix this failure, I'd appreciate it. :D

gijsk:
It looks like bootstrap.js is now packaged twice. We try to stop people from packaging the same file contents twice because, uh, well, it's wasteful. :-)

It should be possible to import the bootstrap file from a file protocol, I guess, instead of resource:///, but I haven't tested that.

PR #721, opened 2017-04-26
Fix try builds of the system add-on

Osmose:
There are failures with our add-on that can happen only during
packaging of Firefox, such as duplicate bootstrap.js files. We should
run that step in TaskCluster CI so we can catch those issues early.

This should catch the issue from #718 and fail. I'm still working on a fix, but want to run this in the meantime.

### DIFF
--- a/recipe-client-addon/bin/tc/package.sh
+++ b/recipe-client-addon/bin/tc/package.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+set -eu
+
+# mach wants this
+export SHELL=$(which bash)
+
+# Fetches build results, and creates ./gecko-dev-master/
+echo 'Downloading build result'
+curl --location --fail --silent --show-error "$BUILD_RESULT" | tar xz
+
+echo 'Setting up environment'
+pushd gecko-dev-master
+source /root/.cargo/env
+python2.7 ./python/mozboot/bin/bootstrap.py --no-interactive --application-choice=browser
+source /root/.cargo/env
+
+echo 'Packaging Firefox'
+./mach package
+popd
--- a/recipe-client-addon/bootstrap.js
+++ b/recipe-client-addon/bootstrap.js
@@ -4,95 +4,27 @@
 "use strict";
 
 const {utils: Cu} = Components;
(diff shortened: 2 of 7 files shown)


recipe-client-addon/bin/tc/package.sh
```diff
@@ -0,0 +1,19 @@
...
+curl --location --fail --silent --show-error "$BUILD_RESULT" | tar xz
+
+echo 'Setting up environment'
+pushd gecko-dev-master
+source /root/.cargo/env
+python2.7 ./python/mozboot/bin/bootstrap.py --no-interactive --application-choice=browser
+source /root/.cargo/env
+
+echo 'Packaging Firefox'
+./mach package
```
mythmon: Might as well put the results of this in an artifact?

Osmose: Sure, if we can find a good way to determine the name of that directory. I wouldn't consider it a blocker though.

Osmose:
Okay, this is ready. The diff looks big, but most of the changes are just copy-pastes from bootstrap.js to ShieldRecipeClient.jsm.

recipe-client-addon/bootstrap.js
```diff
@@ -4,95 +4,27 @@
...
-  CleanupManager.cleanup();
-
-  if (reason === REASONS.ADDON_DISABLE || reason === REASONS.ADDON_UNINSTALL) {
-    Services.prefs.setBoolPref(PREF_SELF_SUPPORT_ENABLED, true);
-  }
+  ShieldRecipeClient.shutdown(reason);
 
+  // Unload add-on modules. We don't do this in ShieldRecipeClient so that
+  // modules are not unloaded accidentally during tests.
+  const log = LogManager.getLogger("bootstrap");
```
gijsk: We may or may not have called LogManager.configure() here, per https://bugzilla.mozilla.org/show_bug.cgi?id=1351465#c5 . I don't know if that should be a serious concern, but I wanted to flag this up.

Osmose:
Updated. I moved the logging setup to before the early return. There's some opportunity for refactoring this to be better but I'm punting on that for now.

PR #723, opened 2017-04-28
recipe-client-addon: Ensure trailing slash on API root.

### LINKED ISSUE #719: getApiUrl omits trailing slashes
I see this during late-stage startup on my profile:

```
GET XHR https://self-repair.mozilla.org/api/v1 [HTTP/1.1 301 Moved Permanently 0ms]
GET XHR https://self-repair.mozilla.org/api/v1/ [HTTP/1.1 200 OK 104ms]
…
GET XHR https://self-repair.mozilla.org/api/v1/action/console-log [HTTP/1.1 301 Moved Permanently 0ms]
GET XHR https://self-repair.mozilla.org/api/v1/action/console-log/ [HTTP/1.1 200 OK 53ms]
```

The call stacks, with promise bumpf elided:

```
        apiCall resource://shield-recipe-client/lib/NormandyApi.jsm:42:12
	get resource://shield-recipe-client/lib/NormandyApi.jsm:46:12
	this.NormandyApi.fetchAction< resource://shield-recipe-client/lib/NormandyApi.jsm:136:23
	this.NormandyApi.fetchAction< resource://shield-recipe-client/lib/NormandyApi.jsm:132:30
	this.RecipeRunner.executeRecipe< resource://shield-recipe-client/lib/RecipeRunner.jsm:135:26
	this.RecipeRunner.start< resource://shield-recipe-client/lib/RecipeRunner.jsm:95:17
	this.RecipeRunner.start< resource://shield-recipe-client/lib/RecipeRunner.jsm:84:17
	this.RecipeRunner.start< resource://shield-recipe-client/lib/RecipeRunner.jsm:84:17
	this.RecipeRunner.start< resource://shield-recipe-client/lib/RecipeRunner.jsm:84:17
```

I think there's a missing trailing slash somewhere, causing hundreds of millions of surplus HTTP requests each day. Worth a fix?

mythmon:
This fixes one of the lack of trailing slashes identified in #719. The other one, related to actions, was redesigned away in c4690415ac6ace78f2f3202cbfc161ac728c64cf.

@rnewman Can you review this as if it were a patch against mozilla-central? Eventually we will merge this into mozilla-central, and having reviews here makes that merge process easier.

Fixes #719.

### DIFF
--- a/recipe-client-addon/lib/NormandyApi.jsm
+++ b/recipe-client-addon/lib/NormandyApi.jsm
@@ -63,8 +63,8 @@ this.NormandyApi = {
 
   async getApiUrl(name) {
     let apiBase = new URL(prefs.getCharPref("api_url"));
-    if (!apiBase.pathname.endsWith('/')) {
-      apiBase.pathname += '/';
+    if (!apiBase.pathname.endsWith("/")) {
+      apiBase.pathname += "/";
     }
     if (!indexPromise) {
       indexPromise = this.get(apiBase.toString()).then(res => res.json());
--- a/recipe-client-addon/lib/NormandyApi.jsm
+++ b/recipe-client-addon/lib/NormandyApi.jsm
@@ -62,7 +62,10 @@ this.NormandyApi = {
   },
 
   async getApiUrl(name) {
-    const apiBase = prefs.getCharPref("api_url");
+    let apiBase = prefs.getCharPref("api_url");
+    if (!apiBase.endsWith('/')) {
+      apiBase += '/';
+    }
     if (!indexPromise) {
       indexPromise = this.get(apiBase).then(res => res.json());
     }
--- a/recipe-client-addon/lib/NormandyApi.jsm
+++ b/recipe-client-addon/lib/NormandyApi.jsm
@@ -62,11 +62,11 @@ this.NormandyApi = {
   },
 
   async getApiUrl(name) {
-    let apiBase = new URL(prefs.getCharPref("api_url"));
-    if (!apiBase.pathname.endsWith("/")) {
-      apiBase.pathname += "/";
-    }
     if (!indexPromise) {
+      let apiBase = new URL(prefs.getCharPref("api_url"));
+      if (!apiBase.pathname.endsWith("/")) {
+        apiBase.pathname += "/";
+      }
       indexPromise = this.get(apiBase.toString()).then(res => res.json());
     }
     const index = await indexPromise;
--- a/recipe-client-addon/lib/NormandyApi.jsm
+++ b/recipe-client-addon/lib/NormandyApi.jsm
@@ -62,12 +62,12 @@ this.NormandyApi = {
   },
 
   async getApiUrl(name) {
-    let apiBase = prefs.getCharPref("api_url");
-    if (!apiBase.endsWith('/')) {
-      apiBase += '/';
+    let apiBase = new URL(prefs.getCharPref("api_url"));
+    if (!apiBase.pathname.endsWith('/')) {
+      apiBase.pathname += '/';
     }
     if (!indexPromise) {
-      indexPromise = this.get(apiBase).then(res => res.json());
+      indexPromise = this.get(apiBase.toString()).then(res => res.json());
     }
     const index = await indexPromise;
     if (!(name in index)) {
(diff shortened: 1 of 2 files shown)


Osmose:
Subbing out @gijsk for @rnewman because, while absolutely qualified to review this patch, he's not a Firefox peer (yet) and I dunno whether that's fine or not for our stuff that we sync. Feel free to correct me if I'm being silly.

recipe-client-addon/lib/NormandyApi.jsm
```diff
@@ -62,7 +62,10 @@ this.NormandyApi = {
   },
 
   async getApiUrl(name) {
-    const apiBase = prefs.getCharPref("api_url");
+    let apiBase = prefs.getCharPref("api_url");
```
rnewman: I think a perfectly legit question to ask is: why does this not have a guaranteed format? Can we make sure that this pref has a trailing slash?

[This search](https://github.com/mozilla/normandy/search?utf8=%E2%9C%93&q=api_url&type=) suggests a bunch of places where we might fix the string instead of doing this.

recipe-client-addon/lib/NormandyApi.jsm
```diff
@@ -62,7 +62,10 @@ this.NormandyApi = {
...
+    if (!apiBase.endsWith('/')) {
+      apiBase += '/';
```
rnewman: This is fine if `apiBase` is guaranteed to never include query parameters or fragments. And if you can guarantee that, you can guarantee that it ends in a slash, no?

mythmon:
In writing this patch, I was originally going to do what would make sense and simply change [the default pref](https://github.com/mozilla/normandy/blob/master/recipe-client-addon/lib/ShieldRecipeClient.jsm#L35). In doing so, I noticed that the [way we set default prefs](https://github.com/mozilla/normandy/blob/master/recipe-client-addon/lib/ShieldRecipeClient.jsm#L96-L104) is actually quite bad and doesn't allow us to update the pref later. So if it were to be fixed that way, any user that has already installed this add-on would not receive the update.

This is obviously not good, and I have a plan to fix it<sup>1</sup>. In the meantime however, we opted to make the code more robust to this sort of common variance in URLs.

As for query strings: I'm ok guaranteeing that the API base URL will not validly contain query strings. The server wouldn't even know what to do with that. If it *does* contain query strings (which the server doesn't expect or  understand), the client will happily add a `/` to the end and get 301ed to the correct with-slash url. It should work even in the case of users fiddling with preferences.

---

1. The plan to fix our preference system is two fold. First, update the code to set the default prefs for the add-on on the default branch, while still safely supporting older users. Second, use a Normandy preference rollout to move all the users with the bad value to have the good value. This is something we are actively developing now. Having a use case at hand for our team will make it easier to validate the feature after we've developed it. After all that, we can fix the code to expect the prefs to work correctly.

rnewman:
> I noticed that the way we set default prefs is actually quite bad…

Oh, yikes. Yeah, that doesn't buy into the usual prefs mechanisms 😞 

> In the meantime however, we opted to make the code more robust to this sort of common variance in URLs.

Great, rock on! Your plan sounds good to me.

recipe-client-addon/lib/NormandyApi.jsm
gijsk: If you use a URL constructor here to parse the URL, you get 2 bonuses over this approach:
1) it'll be correct in the face of querystrings / hashes (the latter of which I imagine we can just drop anyway?) without changing those.
2) it'll normalize things using our 'regular' URL parser and also allow us to formally catch invalid URLs before passing them to fetch (and throw an appropriate error).

However, that would be slightly more code. I'm fine with either approach if we're comfortable with this butchering query strings / hashes if any occur (and still having the 301 in that case).

gijsk:
I don't know why @rnewman isn't listed as a Firefox peer, but yes, he's perfectly qualified. :-)

mythmon:
I've added a test for this behavior, and switched to using a URL constructor. This is ready for another review.

recipe-client-addon/lib/NormandyApi.jsm
```diff
@@ -62,9 +62,12 @@ this.NormandyApi = {
   },
 
   async getApiUrl(name) {
-    const apiBase = prefs.getCharPref("api_url");
+    let apiBase = new URL(prefs.getCharPref("api_url"));
```
gijsk: Should we move all of this inside the `if (!indexPromise)` check?

mythmon:
I made the changes that @gijsk recommended. I think it is small enough to keep the r+ and merge this.

gijsk:
\o/

PR #724, opened 2017-05-01
Preference Rollout - Driver Registration

### LINKED ISSUE #715: Add driver method to register a preference rollout
This driver method will be called by the Rollout action every time the recipe is executed. It should update the list of stored rollouts, and change prefs as needed.

For user branch rollouts, the original preference value should be stored in Storage, with care to detect if no value is set. Aftewards, the rollout value should be set on the user branch.

For default branch rollouts, store the original default value in memory and set the pref on the default branch.

It shouldn't worry about un-enrolling old rollouts, that will be done in another issue.

--

- [x] Update list of stored rollouts when Rollout action is executed
- [x] User branch rollouts should store original pref value in Storage
  - [x] Rollout value should be set on user branch
- [ ] Default branch rollouts should store original default value in memory
  - [x] Rollout value should be set on default branch

andymikulski:
Fixes #715 
~Depends on PR #720~
~Depends on PR #773~

- [x] Add rollout registration to driver
- [x] Update list of stored rollouts when Rollout action is executed
- [x] User branch rollouts should store original pref value in Storage
- [x] Preference values should be set on the appropriate branch (user/default)
- [x] Tests
- [x] Comments

--

This is ready for feedback!

### DIFF
--- a/recipe-client-addon/lib/PreferenceExperiments.jsm
+++ b/recipe-client-addon/lib/PreferenceExperiments.jsm
@@ -61,72 +61,20 @@ XPCOMUtils.defineLazyModuleGetter(this, "LogManager", "resource://shield-recipe-
 XPCOMUtils.defineLazyModuleGetter(this, "Preferences", "resource://gre/modules/Preferences.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "TelemetryEnvironment", "resource://gre/modules/TelemetryEnvironment.jsm");
 
+XPCOMUtils.defineLazyModuleGetter(this, "PreferenceManagement", "resource://shield-recipe-client/lib/PreferenceManagement.jsm");
+
 this.EXPORTED_SYMBOLS = ["PreferenceExperiments"];
 
-const EXPERIMENT_FILE = "shield-preference-experiments.json";
-
-const PREFERENCE_TYPE_MAP = {
-  boolean: Services.prefs.PREF_BOOL,
-  string: Services.prefs.PREF_STRING,
-  integer: Services.prefs.PREF_INT,
-};
-
-const DefaultPreferences = new Preferences({defaultBranch: true});
-
-/**
- * Enum storing Preference modules for each type of preference branch.
- * @enum {Object}
- */
-const PreferenceBranchType = {
-  user: Preferences,
-  default: DefaultPreferences,
-};
-
-/**
- * Asynchronously load the JSON file that stores experiment status in the profile.
- */
-let storePromise;
-function ensureStorage() {
-  if (storePromise === undefined) {
-    const path = OS.Path.join(OS.Constants.Path.profileDir, EXPERIMENT_FILE);
-    const storage = new JSONFile({path});
-    storePromise = storage.load().then(() => storage);
-  }
-  return storePromise;
-}
-
+const Manager = PreferenceManagement("experiments");
 const log = LogManager.getLogger("preference-experiments");
 
-// List of active preference observers. Cleaned up on shutdown.
-let experimentObservers = new Map();
-CleanupManager.addCleanupHandler(() => PreferenceExperiments.stopAllObservers());
-
 this.PreferenceExperiments = {
-  /**
-   * Set the default preference value for active experiments that use the
-   * default preference branch.
-   */
   async init() {
-    for (const experiment of await this.getAllActive()) {
-      // Set experiment default preferences, since they don't persist between restarts
-      if (experiment.preferenceBranchType === "default") {
-        DefaultPreferences.set(experiment.preferenceName, experiment.preferenceValue);
-      }
+    Manager.applyAll();
+  },
 
-      // Check that the current value of the preference is still what we set it to
-      if (Preferences.get(experiment.preferenceName, undefined) !== experiment.preferenceValue) {
-        // if not, stop the experiment, and skip the remaining steps
-        log.info(`Stopping experiment "${experiment.name}" because its value changed`);
-        await this.stop(experiment.name, false);
-        continue;
-      }
-
-      // Notify Telemetry of experiments we're running, since they don't persist between restarts
-      TelemetryEnvironment.setExperimentActive(experiment.name, experiment.branch);
-
-      // Watch for changes to the experiment's preference
-      this.startObserver(experiment.name, experiment.preferenceName, experiment.preferenceValue);
-    }
+  async getAll() {
+    return Manager.getChangedPrefs();
   },
 
   /**
@@ -135,31 +83,20 @@ this.PreferenceExperiments = {
    */
   withMockExperiments(testFunction) {
     return async function inner(...args) {
-      const oldPromise = storePromise;
       const mockExperiments = {};
-      storePromise = Promise.resolve({
+      const storePromise = Promise.resolve({
         data: mockExperiments,
         saveSoon() { },
       });
-      const oldObservers = experimentObservers;
-      experimentObservers = new Map();
+
       try {
         await testFunction(...args, mockExperiments);
       } finally {
-        storePromise = oldPromise;
-        PreferenceExperiments.stopAllObservers();
-        experimentObservers = oldObservers;
+        Manager.stopAllObservers();
       }
-    };
-  },
 
-  /**
-   * Clear all stored data about active and past experiments.
-   */
-  async clearAllExperimentStorage() {
-    const store = await ensureStorage();
-    store.data = {};
-    store.saveSoon();
+      return storePromise;
+    };
   },
 
   /**
@@ -178,12 +115,12 @@ this.PreferenceExperiments = {
   async start({name, branch, preferenceName, preferenceValue, preferenceBranchType, preferenceType}) {
     log.debug(`PreferenceExperiments.start(${name}, ${branch})`);
 
-    const store = await ensureStorage();
-    if (name in store.data) {
+    const hasExistingExperiment = await Manager.has(name);
+    if (hasExistingExperiment) {
       throw new Error(`A preference experiment named "${name}" already exists.`);
     }
 
-    const activeExperiments = Object.values(store.data).filter(e => !e.expired);
+    const activeExperiments = await Manager.getActive();
     const hasConflictingExperiment = activeExperiments.some(
       e => e.preferenceName === preferenceName
     );
@@ -193,10 +130,7 @@ this.PreferenceExperiments = {
       );
     }
 
-    const preferences = PreferenceBranchType[preferenceBranchType];
-    if (!preferences) {
-      throw new Error(`Invalid value for preferenceBranchType: ${preferenceBranchType}`);
-    }
+    const preferences = Manager.getPreferenceBranch(preferenceBranchType);
 
     /** @type {Experiment} */
     const experiment = {
@@ -211,8 +145,8 @@ this.PreferenceExperiments = {
       preferenceBranchType,
     };
 
-    const prevPrefType = Services.prefs.getPrefType(preferenceName);
-    const givenPrefType = PREFERENCE_TYPE_MAP[preferenceType];
+    // Validation will throw
+    Manager.validateRecipe(experiment);
 
     if (!preferenceType || !givenPrefType) {
       throw new Error(`Invalid preferenceType provided (given "${preferenceType}")`);
@@ -225,100 +159,15 @@ this.PreferenceExperiments = {
       );
     }
 
+    // Save experiment data locally to be read from on startup later
+    Manager.saveRecipeData(experiment);
+    Manager.startObserver(experiment.name, experiment.preferenceName, experiment.preferenceValue);
+
+    // Actually alter the set preference (this is done automatically on startup)
     preferences.set(preferenceName, preferenceValue);
-    PreferenceExperiments.startObserver(name, preferenceName, preferenceValue);
-    store.data[name] = experiment;
-    store.saveSoon();
 
+    // Notify Telemetry
     TelemetryEnvironment.setExperimentActive(name, branch);
-  },
-
-  /**
-   * Register a preference observer that stops an experiment when the user
-   * modifies the preference.
-   * @param {string} experimentName
-   * @param {string} preferenceName
-   * @param {string|integer|boolean} preferenceValue
-   * @throws {Error}
-   *   If an observer for the named experiment is already active.
-   */
-  startObserver(experimentName, preferenceName, preferenceValue) {
-    log.debug(`PreferenceExperiments.startObserver(${experimentName})`);
-
-    if (experimentObservers.has(experimentName)) {
-      throw new Error(
-        `An observer for the preference experiment ${experimentName} is already active.`
-      );
-    }
-
-    const observerInfo = {
-      preferenceName,
-      observer(newValue) {
-        if (newValue !== preferenceValue) {
-          PreferenceExperiments.stop(experimentName, false);
-        }
-      },
-    };
-    experimentObservers.set(experimentName, observerInfo);
-    Preferences.observe(preferenceName, observerInfo.observer);
-  },
-
-  /**
-   * Check if a preference observer is active for an experiment.
-   * @param {string} experimentName
-   * @return {Boolean}
-   */
-  hasObserver(experimentName) {
-    log.debug(`PreferenceExperiments.hasObserver(${experimentName})`);
-    return experimentObservers.has(experimentName);
-  },
-
-  /**
-   * Disable a preference observer for the named experiment.
-   * @param {string} experimentName
-   * @throws {Error}
-   *   If there is no active observer for the named experiment.
-   */
-  stopObserver(experimentName) {
-    log.debug(`PreferenceExperiments.stopObserver(${experimentName})`);
-
-    if (!experimentObservers.has(experimentName)) {
-      throw new Error(`No observer for the preference experiment ${experimentName} found.`);
-    }
-
-    const {preferenceName, observer} = experimentObservers.get(experimentName);
-    Preferences.ignore(preferenceName, observer);
-    experimentObservers.delete(experimentName);
-  },
-
-  /**
-   * Disable all currently-active preference observers for experiments.
-   */
-  stopAllObservers() {
-    log.debug("PreferenceExperiments.stopAllObservers()");
-    for (const {preferenceName, observer} of experimentObservers.values()) {
-      Preferences.ignore(preferenceName, observer);
-    }
-    experimentObservers.clear();
-  },
-
-  /**
-   * Update the timestamp storing when Normandy last sent a recipe for the named
-   * experiment.
-   * @param {string} experimentName
-   * @rejects {Error}
-   *   If there is no stored experiment with the given name.
-   */
-  async markLastSeen(experimentName) {
-    log.debug(`PreferenceExperiments.markLastSeen(${experimentName})`);
-
-    const store = await ensureStorage();
-    if (!(experimentName in store.data)) {
-      throw new Error(`Could not find a preference experiment named "${experimentName}"`);
-    }
-
-    store.data[experimentName].lastSeen = new Date().toJSON();
-    store.saveSoon();
   },
 
   /**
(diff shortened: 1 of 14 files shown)


andymikulski:
@mythmon Ready for feedback on commit 4d11c30309ecd846b504410099dc5f71278c5cbf . Essentially, I abstracted the preference managing bits into `PreferenceManagement.jsm`, and left the `start`/`stop` functions in `Preference[Rollout|Experiments].jsm`.

Basically, you can instantiate a namespaced pref manager based on its purpose (rollouts get the `rollout` namespace, experiments `experiments`, etc), and then it exposes all the necessary functions to access those prefs.

Let me know if you think this is a more appropriate path. Tests are currently failing, but I wanted to get some feedback before tackling those.

recipe-client-addon/lib/PreferenceRollout.jsm
```diff
@@ -0,0 +1,75 @@
...
+    };
+
+    // 'User' branch prefs are saved locally, so in the event the user is withdrawn
+    // from the rollout, the pref can be reset to what the user had originally.
+    // (Default branch pref rollouts simply unset the value and let the browser
+    // set its baked-in default value, so we don't need to store it.)
+    // if (preferenceBranch === "user") {
+      // store.data[name] = prevValue;
+      // store.saveSoon();
+    // }
```
mythmon: We probably should either uncomment or remove this.

recipe-client-addon/lib/PreferenceRollout.jsm
```diff
@@ -0,0 +1,75 @@
...
+
+    // Validation will throw
+    Manager.validateRecipe(experiment);
+
+    // Save experiment data locally to be read from on startup later
+    Manager.saveRecipeData(experiment);
+    Manager.startObserver(experiment.name, experiment.preferenceName, experiment.preferenceValue);
+
+    // Actually alter the set preference (this is done automatically on startup)
```
mythmon: Can you explain this comment a bit more? I'm not sure what "this is done automatically on startup" is referring to.

recipe-client-addon/lib/PreferenceManagement.jsm
```diff
@@ -0,0 +1,305 @@
...
+ * show what experiments are active, and we use this data to measure the
+ * effectiveness of the preference change.
+ *
+ * Info on active and past experiments is stored in a JSON file in the profile
+ * folder.
+ *
+ * Active preference experiments are stopped if they aren't active on the recipe
+ * server. They also expire if Firefox isn't able to contact the recipe server
+ * after a period of time, as well as if the user modifies the preference during
+ * an active experiment.
```
mythmon: This comment talks about experiments, but this file is shared between both experiments and rollouts, right?

andymikulski:
- `PreferenceManagement` changed from being instanced by Experiments and Rollouts, to being inherited by them. The reason for this was that both modules needed the similar pref API, however Rollouts needed direct access to the `store` in PrefMan. Inheriting PrefMan's logic instead lets Rollouts access what it needs to without manually exposing the store, and nothing really changes in Experiments.
  - This also cleaned up some of the exports in `NormandyDriver`
- A majority of the `PreferenceExperiment` tests were split into a PrefMan-specific suite.
- There is still a test failing - for some reason, a function is losing context (`"this" is undefined`), even though the function that's being called is already being bound. I'm pretty sure @Osmose ran into this problem before, so I'll consult him on how to fix it.

@mythmon Can you review this before I toss it to FF QA?

mythmon:
I read over this, and I think I'm ok with the direction this is going.

Lets get that test fixed though. Merging master into this feature branch, and then rebasing this branch onto that might help.

recipe-client-addon/lib/PreferenceManagement.jsm
```diff
@@ -0,0 +1,395 @@
...
+  /**
+   * Returns a reference to either the Default or User preferences branch
+   *
+   * @param  {string} which Branch to return ('user' or 'default')
+   * @return {Object}       Selected preferences branch
+   */
+  getPreferenceBranch(which) {
+    const branch = PreferenceBranchType[which];
+    if (!branch) {
+      throw new Error(`Invalid value for preference branch: "${which}"`);
```
gijsk: Nit: this would be slightly simpler as:

```js
if (!PreferenceBranchType.hasOwnProperty(which)) {
  throw new Error(...);
}
return PreferenceBranchType[which];
```

recipe-client-addon/lib/PreferenceManagement.jsm
```diff
@@ -0,0 +1,395 @@
...
+    }
+
+    return branch;
+  },
+
+  /**
+   * Register a preference observer that stops an recipe when the user
```
gijsk: Nit: a recipe

recipe-client-addon/lib/PreferenceManagement.jsm
```diff
@@ -0,0 +1,395 @@
...
+   *
+   * @returns {Promise}
+   * @resolves When the operation is completed succesfully
+   * @rejects Javascript exception.
+   */
+  async getActiveChanges() {
+    log.debug("PreferenceManagement.getActiveChanges()");
+    const allPrefs = await this.getStoredRecipes();
+    // Return copies so mutating them doesn't affect the storage.
+    return allPrefs.filter(e => !e.expired).map(e => Object.assign({}, e));
```
gijsk: getStoredRecipes already copied this, so now we're just copying them a second time...

recipe-client-addon/lib/PreferenceManagement.jsm
```diff
@@ -0,0 +1,395 @@
...
+   *
+   * @param  {string}  recipeName
+   * @returns {Promise}
+   * @resolves When the operation is completed succesfully
+   * @rejects Javascript exception.
+   */
+  async expire(recipeName) {
+    log.debug(`PreferenceManagement.expire(${recipeName})`);
+    const store = await this.getStorage();
+    store.data[namespace][recipeName].expired = true;
```
gijsk: Should this throw something more descriptive if the `[namespace][recipeName]` object doesn't exist?

recipe-client-addon/lib/PreferenceRollout.jsm
```diff
@@ -0,0 +1,78 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+
+"use strict";
+
+const {utils: Cu} = Components;
+Cu.import("resource://gre/modules/XPCOMUtils.jsm");
+
+XPCOMUtils.defineLazyModuleGetter(this, "Services", "resource://gre/modules/Services.jsm");
```
gijsk: Most of these are also unused in this file.

recipe-client-addon/lib/PreferenceRollout.jsm
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /** @type {Experiment} */
+    const experiment = {
+      name,
+      preferenceName,
+      preferenceValue,
+      preferenceType,
+      preferenceBranch,
+      branch: preferenceBranch,
```
gijsk: Can we make this API use the same parameter names as the PreferenceExperiments one?

recipe-client-addon/lib/PreferenceRollout.jsm
```diff
@@ -0,0 +1,78 @@
...
+const log = LogManager.getLogger("preference-rollouts");
+
+const PREF_NAMESPACE = "rollouts";
+
+this.PreferenceRollout = Object.assign({}, PreferenceManagement(PREF_NAMESPACE), {
+  async init() {
+    this.applyAll();
+  },
+
+  async start({name, preferenceName, preferenceValue, preferenceBranch, preferenceType}) {
```
gijsk: Missing jsdoc.

recipe-client-addon/lib/PreferenceRollout.jsm
```diff
@@ -0,0 +1,78 @@
...
+
+    // Actually alter the set preference (later, this is done automatically on startup)
+    preferences.set(preferenceName, preferenceValue);
+  },
+
+  /**
+   * Stubbed withdraw function
+   */
+  async stop(rolloutName, resetValue = true) {
+    await new Promise(resolve => setTimeout(resolve, 1000));
```
gijsk: Uhhh... what is this, why do we need it, why does it have an unused argument, and what's with the setTimeout delay in having resolve() be called? Can we just rm this?

recipe-client-addon/test/browser/browser_PreferenceManagement.js
```diff
@@ -0,0 +1,166 @@
...
+Cu.import("resource://gre/modules/TelemetryEnvironment.jsm", this);
+Cu.import("resource://shield-recipe-client/lib/PreferenceManagement.jsm", this);
+
+// Save ourselves some typing
+const Manager = PreferenceManagement("test");
+const withMockData = Manager.withMockData.bind(Manager);
+
+const DefaultPreferences = new Preferences({defaultBranch: true});
+
+function experimentFactory(attrs) {
```
gijsk: Can this function live in head.js instead of being duplicated between test files?

recipe-client-addon/test/browser/browser_PreferenceManagement.js
```diff
@@ -0,0 +1,166 @@
+"use strict";
+
+Cu.import("resource://gre/modules/Preferences.jsm", this);
+Cu.import("resource://gre/modules/TelemetryEnvironment.jsm", this);
```
gijsk: TelemetryEnvironment seems to be unused here.

recipe-client-addon/test/browser/browser_PreferenceManagement.js
```diff
@@ -0,0 +1,166 @@
...
+    "markLastSeen threw because there was not a matching experiment",
+  );
+}));
+
+// markLastSeen should update the lastSeen date
+add_task(withMockData(async function(experiments) {
+  const oldDate = new Date(1988, 10, 1).toJSON();
+  experiments["test"] = experimentFactory({name: "test", lastSeen: oldDate});
+  await Manager.markLastSeen("test");
+  Assert.notEqual(
```
gijsk: Hmm, this is pre-existing, but can we enforce the date has actually increased? Also, given that:

```js
let x = new Date(100), y = new Date(100);
x == y; // false
```

because dates are objects, I wonder if this is testing anything particularly meaningful right now...

recipe-client-addon/test/browser/browser_PreferenceManagement.js
```diff
@@ -0,0 +1,166 @@
...
+
+// get should throw if no experiment exists with the given name
+add_task(withMockData(async function() {
+  await Assert.rejects(
+    Manager.get("neverexisted"),
+    "get rejects if no experiment with the given name is found",
+  );
+}));
+
+// get
```
gijsk: Nit: missing description?

recipe-client-addon/test/browser/browser_PreferenceManagement.js
```diff
@@ -0,0 +1,166 @@
...
+
+  const fetchedExperiment = await Manager.get("test");
+  Assert.deepEqual(fetchedExperiment, experiment, "get fetches the correct experiment");
+
+  // Modifying the fetched experiment must not edit the data source.
+  fetchedExperiment.name = "othername";
+  is(experiments["test"].name, "test", "get returns a copy of the experiment");
+}));
+
+add_task(withMockData(async function testGetActiveChanges(experiments) {
```
gijsk: Nit: please add a description here and below.

recipe-client-addon/test/browser/browser_PreferenceManagement.js
```diff
@@ -0,0 +1,166 @@
...
+    fetchedExperiment2,
+    experiment2,
+    "getActiveChanges returns a list with the correct experiments, including disabled ones",
+  );
+
+  fetchedExperiment2.name = "othername";
+  is(experiment2.name, "experiment2", "getActiveChanges returns copies of the experiments");
+}));
+
+add_task(withMockData(withMockPreferences(async function testGetActiveChanges(experiments) {
```
gijsk: Please make sure your functions have unique names, even if they're function expressions rather than globals. :-)

recipe-client-addon/lib/PreferenceManagement.jsm
```diff
@@ -0,0 +1,395 @@
...
+ * @enum {Object}
+ */
+const PreferenceBranchType = {
+  user: Preferences,
+  default: DefaultPreferences,
+};
+
+const log = LogManager.getLogger("preference-management");
+
+this.PreferenceManagement = (namespace) => ({
```
andymikulski: I landed on this approach of creating an object for recipes to extend since Experiments/Rollouts are both singletons as it is, and Rollouts require access to some of the internals of the PrefMan business. I also based most of the work off `PrefExperiments`, which doesn't use a class or  prototype, so I assumed a POJO would be alright. If it makes more sense, I can switch to using ES6 classes here and in Rollouts/Experiments!

recipe-client-addon/lib/PreferenceManagement.jsm
```diff
@@ -0,0 +1,395 @@
...
+   */
+  async clearStorage() {
+    const store = await this.getStorage();
+    store.data[namespace] = {};
+    store.saveSoon();
+  },
+
+
+  /**
+   * Returns the internal storage object, after properly ensuring the store and
```
andymikulski: You're right - this function was actually phased out during dev and should have been removed. Sorry for the noise.

recipe-client-addon/lib/PreferenceManagement.jsm
andymikulski: I would argue that `getActiveChanges` shouldn't be aware of whether or not `getStoredRecipes` returns originals or copies, since we could change how it works in the future. Perhaps that's a non-issue in this instance, though.

recipe-client-addon/lib/PreferenceManagement.jsm
andymikulski: Yes! I've added a new error if a recipe does not exist.

recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -61,105 +61,21 @@ XPCOMUtils.defineLazyModuleGetter(this, "LogManager", "resource://shield-recipe-
...
-
-/**
- * Enum storing Preference modules for each type of preference branch.
- * @enum {Object}
- */
-const PreferenceBranchType = {
-  user: Preferences,
-  default: DefaultPreferences,
-};
+XPCOMUtils.defineLazyModuleGetter(this, "PreferenceManagement", "resource://shield-recipe-client/lib/PreferenceManagement.jsm");
```
andymikulski: Could you suggest how I could make this lazier? We could do something like defer instantiating it to the recipe `init`s, though I don't think that would take place much longer after startup. Would it make more sense to make PrefMan a singleton?

recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -61,105 +61,21 @@ XPCOMUtils.defineLazyModuleGetter(this, "LogManager", "resource://shield-recipe-
...
-let experimentObservers = new Map();
-CleanupManager.addCleanupHandler(() => PreferenceExperiments.stopAllObservers());
+const PREF_NAMESPACE = "experiments";
 
-this.PreferenceExperiments = {
-  /**
-   * Set the default preference value for active experiments that use the
-   * default preference branch.
-   */
+this.PreferenceExperiments = Object.assign({}, PreferenceManagement(PREF_NAMESPACE), {
```
andymikulski: I always use fresh objects when using `Object.assign`, I think as a defensive measure. I'll correct this.

recipe-client-addon/lib/PreferenceRollout.jsm
andymikulski: This was just a stub for a function we're implementing after this PR lands. I've removed it for now.

recipe-client-addon/test/browser/browser_PreferenceManagement.js
andymikulski: The dates are converted `toJSON`, which formats the datetime to `YYYY-MM-DDTHH:mm:ss.sssZ`, so comparisons should be fine. I can look at enforcing the date check.

recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -211,117 +124,21 @@ this.PreferenceExperiments = {
       preferenceBranchType,
     };
 
-    const prevPrefType = Services.prefs.getPrefType(preferenceName);
-    const givenPrefType = PREFERENCE_TYPE_MAP[preferenceType];
+    // Validation will throw
```
Osmose: @erikrose has convinced me that punctuation is only required for complete sentences.

erikrose: "Validation will throw." is a complete sentence. :-) I often like to use a colon on comments like this, if you want to emphasize that they apply to the following thing (rather than the preceding). Not sure if that's the case here…

recipe-client-addon/lib/PreferenceManagement.jsm
gijsk: I think ES6 classes would make sense, though we'd presumably still want singleton instances for rollouts/experiments.

recipe-client-addon/lib/PreferenceExperiments.jsm
gijsk: Well, if we need to check this stuff on startup (to flip prefs) it will never be lazy, I think, so we should probably just make it more obvious that when someone imports this module, PreferenceManagement is imported sync (Cu.import() rather than defineLazyModuleGetter).

recipe-client-addon/lib/PreferenceManagement.jsm
gijsk: Well, both are public methods, and so I think the expectation that `getStoredRecipes` offers you copies will need to remain reasonable in the future even if we refactor or something. We also have tests for both of these methods returning copies that don't affect the originals, so I think just making the assumption is fine - tests will save us if someone inadvertently starts violating that assumption. If you update/add comments accordingly, it seems unlikely we will break it, and not copying again does avoid doing extra work for no benefit. :-)

andymikulski:
@mythmon Could you do a pass through this PR?

recipe-client-addon/lib/PreferenceManager.jsm
```diff
@@ -106,19 +109,19 @@ this.PreferenceManagement = (namespace) => ({
...
     this.mockData = {
-      [namespace]: {},
+      [this.namespace]: {}
     };
 
     return async (...args) => {
       this.storePromise = Promise.resolve({
         data: this.mockData,
-        saveSoon() {},
+        saveSoon() {}
```
gijsk: Nit: trailing comma was correct here and 6 lines up, unless eslint told you otherwise? Convention in Firefox code is usually to keep it, though I think devtools might have a different one...

recipe-client-addon/lib/PreferenceManager.jsm
```diff
@@ -106,19 +109,19 @@ this.PreferenceManagement = (namespace) => ({
...
       });
 
       try {
         // Tests directly edit store values, so for test functions, the relevant
         // namespace is exposed for brevity.
-        await testFunction(this.mockData[namespace], ...args)
+        await testFunction(this.mockData[this.namespace], ...args)
```
gijsk: Nit: missing semicolon (does eslint not pick this up?)

andymikulski: It did not get picked up - I wonder if the `await` throws the linter off.

andymikulski:
@mythmon Looks like TC stuff is failing due to the Rust version issue. Can this PR be landed or should we wait until TC tests pass?

mythmon:
I'm pretty confident in this patch, since it passed on a recent commit. Lets go ahead and merge it.

PR #725, opened 2017-05-01
Recipe page filters

andymikulski:


Osmose:
Review the UI locally, it's way better than the last time I ran it locally; I think I had a bad CSS build or something.

Let's go ahead and fix the conflicts, get those reviewed and merged, and then:

- Deploy to staging
- Ask S&I for feedback in case there's any blockers we need to resolve
- Notify dev-shield of the interface changes prior to deployment
- Deploy to prod

Nice work!

andymikulski:
Rebased the branch onto `master` to fix onflicts, everything looks good. I had to correct a handful of things I missed in the rebase (see 108d73a0594930e59f4e8e51305fcaeb712fbd02 ), so I reckon that may need to be reviewed, but for the most part the code in this PR should be good to go.

@Osmose @rehandalal If y'all want to review this before merging, that'd be awesome. Thanks!

Osmose:
TEXT SHADOW. It's interesting, although I think adding it gives us multiple simultaneous visual styles, which is a bit jarring. I defer to @rehandalal for visual design.

The layout of the status indicators is inconsistent with and without the review comment boxes. With comment:

screen shot 2017-05-10 at 13 48 13

Without comment:

screen shot 2017-05-10 at 13 49 02

My armchair-designer suggestion would be to nix the box around the name field (since there's no heading to that grouping, the grouping itself is unnecessary), and flexbox it with the two indicators on the right (vertically stacking them in a container div), giving them no flex so that they stay at their initial width, letting the name box flex to take the rest of the space. 

When the comment's there we can just push everything down, including the indicators. In a followup PR we should probably move this, maybe only showing it in a tooltip if you hover over the "Approved" indicator?

Osmose:
When a recipe is under review, the simple filter boxes are not disabled and can be modified.

andymikulski:
I've gone through and addressed feedback. To fix issues with the stacking indicators etc, I simply rearranged where some elements were in the DOM and didn't have to flex anything. @Osmose Let me know what you think:

screen shot 2017-05-17 at 1 50 22 pm

If you'd prefer to have it flex'd with the status indicators stacked and the name field flex'd and all that, I can do that, this approach just seemed a bit simpler.

This is ready for review.

recipe-server/client/control/selectors/FiltersSelector.js
```diff
@@ -132,7 +132,7 @@ export const isFilteringActive = state =>
  * Get all filters as an object keyed on their slug
  *
  * @param  {Array<Object>} groups All filter groups
```
Osmose: This still has the wrong parameter name.

Also, I'm still curious: Why change the parameter name to `list` instead of `groups`?

andymikulski: D'oh, sorry.

The reason for using `list` was that it was more descriptive of what the function expected (it wants `state.filters.list`). However, looking at the other selectors in that file, they all expect the general `state` object, and then pull out what they need from there. I've changed `getFilterObject` to now accept the `state` object to conform to conventions.

I also noticed that a bunch of the docstrings were inaccurate, so I corrected those while I was in there, as well.

andymikulski:
It _was_ missing the indicator fix. That's weird, sorry! I've corrected the issue. This is ready for review @Osmose

recipe-server/client/control/selectors/FiltersSelector.js
```diff
@@ -35,13 +35,13 @@ export const getSelectedFilterGroups = state => state.active.filter(isGroupSelec
 
 
 /**
- * Given a set of groups, finds the selected groups,
+ * Given the current filter state, finds the selected groups,
  * and then removes any NON-selected options.
  *
  * Essentially, returns an array of groups and their selected options.
  *
- * @param  {Array<Object>} groups All possible filter groups
```
Osmose: We should probably standardize on the spacing for JSDoc comments somday.

PR #729, opened 2017-05-02
Refactor Storage.jsm into a class.

Osmose:
This simplifies the implementation of Storage.jsm, as well as usage
outside of the sandbox, at the cost of adding work to the driver to
make the storage instance usable within the sandbox.
SandboxManager.wrapAsync handles the heavy lifting there.

### DIFF
--- a/recipe-client-addon/lib/NormandyDriver.jsm
+++ b/recipe-client-addon/lib/NormandyDriver.jsm
@@ -123,15 +123,19 @@ this.NormandyDriver = function(sandboxManager) {
       return ret;
     },
 
-    createStorage(keyPrefix) {
-      let storage;
-      try {
-        storage = Storage.makeStorage(keyPrefix, sandbox);
-      } catch (e) {
-        log.error(e.stack);
-        throw e;
+    createStorage(prefix) {
+      const storage = new Storage(prefix);
+
+      // Wrapped methods that we expose to the sandbox. These are documented in
+      // the driver spec in docs/dev/driver.rst.
+      for (const method of ["getItem", "setItem", "removeItem", "clear"]) {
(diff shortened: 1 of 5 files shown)


recipe-client-addon/lib/NormandyDriver.jsm
```diff
@@ -123,15 +123,19 @@ this.NormandyDriver = function(sandboxManager) {
...
+      // the driver spec in docs/dev/driver.rst.
+      for (const method of ["getItem", "setItem", "removeItem", "clear"]) {
+        storage[method] = sandboxManager.wrapAsync(storage[method], {
+          cloneArguments: true,
+          cloneInto: true,
+        });
       }
-      return storage;
+
+      return sandboxManager.cloneInto(storage, {cloneFunctions: true});
```
mythmon: Instead of modifying and passing the storage object directly, could we make an interface object to pass in?

```js
const storage  = new Storage(prefix);

const interface = {};
for (const method of ["getItem", "setItem", "removeItem", "clear"]) {
  interface[method] = sandboxManager.wrapAsync(storage[method], {
    cloneArguments: true,
    cloneInto: true,
  });
}

return sandboxManager.cloneInto(interface, {cloneFunctions: true});
```

I think this will help avoid exposing unintended functions to the sandbox. I also think that it would be better to not modify the storage object, so any `this.foo` method calls it makes will use unmodified versions of the methods. Would that work?

Osmose:
Updated with @mythmon's feedback about using an interface. I added a test to ensure that the methods on the interface object were bound properly; without it, `this.prefix` within the storage methods was null, and the prefix between all storage instances was shared.

PR #730, opened 2017-05-02
TaskCluster: Combine tasks and use artifact builds.

Osmose:
To speed up builds, we want to use artifact builds. Artifact builds
require Mercurial history, so we pull from mozilla-central (and update
the version of Mercurial we have in the Docker image) instead of
gecko-dev.

However, switching to mozilla-central brings up a new issue: the
Mercurial history takes up a lot of space, and the transfer of artifacts
between tasks takes up enough time that our builds are slower than
before, when we had no history to cart around. By combining the fetch,
build, test, and package steps into a single command, we drastically
reduce the amount of time spent running in automation.

We still want lints to be easily distinguishable from test/build
failures, so those remain in their own task.

Tar'ing and uploading the build directory as an artifact also takes
a few minutes, and we rarely ever use it, so that is also removed.


mythmon:
The build task for the most recent commit seems to be hung during the artifact build.

Osmose:
Previous runs took between 25 and 30 minutes. This PR took 4 minutes to finish fetching, building, running the tests, and packaging Firefox. I'm pretty happy with those results.

PR #755, opened 2017-05-11
Fix #739: Add placeholder action for opt-out studies.

### LINKED ISSUE #739: Create opt-out study action
This covers creating a basic opt-out study action that does nothing (logs to the console, maybe?). Actual functionality of the action will be added later.

Osmose:
We'll fill this in later with the actual implementation.

r?

### DIFF
--- a/recipe-server/client/control/components/action_fields/OptOutStudyFields.js
+++ b/recipe-server/client/control/components/action_fields/OptOutStudyFields.js
@@ -1,0 +1,29 @@
+import React, { PropTypes as pt } from 'react';
+
+import { ControlField } from 'control/components/Fields';
+import ActionFields from 'control/components/action_fields/ActionFields';
+
+/**
+ * Form fields for the opt-out-study action.
+ */
+export default class OptOutStudyFields extends ActionFields {
+  static propTypes = {
+    disabled: pt.bool,
+  }
+
+  render() {
+    const { disabled } = this.props;
+    return (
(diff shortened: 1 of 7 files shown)


Osmose:
TaskCluster seems to be failing while trying to clone mozilla-central. Weird.

In any case, this PR doesn't touch the add-on at all, so it's an acceptable failure. I filed #757 to investigate the failures.

recipe-server/client/control/components/action_fields/OptOutStudyFields.js
```diff
@@ -0,0 +1,29 @@
+import React, { PropTypes as pt } from 'react';
+
+import { ControlField } from 'control/components/Fields';
+import ActionFields from 'control/components/action_fields/ActionFields';
+
+/**
+ * Form fields for the opt-out-study action.
+ */
+export default class OptOutStudyFields extends ActionFields {
```
crankycoder: This is just a comment because i'm not too familiar with this code.  What does ActionFields do?  The class seems to just have initialValues and doesn't inherit from redux-form.Field.

Osmose: At this point it's more just there to document the expected fields on the class, even if they don't inherit, sort've like an interface. I also suspected that we'd add specific, inheritable methods to the class in the future and figured using a base class now wouldn't hurt.

Issue #757, opened 2017-05-12
TaskCluster builds intermittently failing while cloning mozilla-central


Osmose:
Here's [an example](https://tools.taskcluster.net/task-group-inspector/#/SO1tViRoRNC6Nd3Ljl-8bQ/WgIIY0UeSt-TSJHwNZRDIw?_k=uvv1jr) (with progress bars removed):

```
~
Downloading mozilla-central...
real URL is https://hg.mozilla.org/mozilla-central/
destination directory: mozilla-central
applying clone bundle from https://s3-external-1.amazonaws.com/moz-hg-bundles-us-east-1/mozilla-central/d8762cb967423618ff0a488f14745f60964e5c49.packed1.hg
325316 files to transfer, 1.64 GB of data
transferred 1.64 GB in 61.0 seconds (27.4 MB/sec)
finished applying clone bundle
searching for changes
adding changesets
adding manifests
adding file changes                                                                                
added 252 changesets with 6009 changes to 3322 files
updating to branch default
abort: No such file or directory: '/root/mozilla-central/services/sync/tps/extensions'
abort: No such file or directory: '/root/mozilla-central/testing/web-platform/tests/html/semantics/common-idioms/rel-up'
abort: No such file or directory: '/root/mozilla-central/services/sync/tps/extensions/mozmill/resource/modules/stack.js'
```

mythmon:
We ran into this during our initial switch to mercurial as well, so it isn't a new thing.

```irc
<mythmon> while cloning, "abort: No such file or directory: '/root/mozilla-central/media/omx-plugin/include/ics'"
<rhelmer> during the clone?! that's really weird
```

Osmose:
@indygreg Could you help us figure out why we're seeing these cloning failures in TaskCluster? We're getting them when just running `hg clone http://hg.mozilla.org/mozilla-central/`, and we're using the same workers as the normal TC Firefox builds use.

indygreg:
That's a pretty special error. Fortunately, it rings a bell. I'm pretty certain the problem is an AUFS filesystem bug courtesy of TaskCluster running Docker. (By default / in TaskCluster is an AUFS volume.) For performance reasons (bug 1291940) and consistency reasons, you'll want to do as little work on AUFS as possible and instead use "cache" volumes, which are just EXT4 filesystems bind mounted into the Docker container.

[...]

PR #766, opened 2017-05-18
Fix #764: Add PEER_APPROVAL_ENFORCED setting.

### LINKED ISSUE #764: Allow users to approve their own approval requests with a setting
To ease development, we want to add a default-on setting that controls whether users are allowed to approve their own approval requests. This will allow developers to save changes to recipes without logging in to a different account, as well as QAs to do so on staging.

Osmose:
The setting controls whether the rule that an approver on an approval
request must not be the same user who created it is enforced or not.

Disabling this setting makes local development (as well as QA in
testing environments) more convenient.

### DIFF
--- a/docs/ops/config.rst
+++ b/docs/ops/config.rst
@@ -255,6 +255,16 @@ in other Django projects.
    If :envvar:`DJANGO_USE_OIDC` is set to ``True``, this settings must be set to
    the URL that a user can visit to logout. It may be a relative URL.
 
+.. envvar:: DJANGO_PEER_APPROVAL_ENFORCED
+
+   :default: ``True``
+
+   If ``True``, approval requests for recipe changes can only be approved by a
+   different user than the one who created the request. If ``False``, approval
+   requests can be approved by the same user who created it.
+
+   This defaults to ``False`` for local developer instances.
+
 Gunicorn settings
 -----------------
 These settings control how Gunicorn starts, when the default command of the
--- a/recipe-server/client/control/components/RecipeForm.js
+++ b/recipe-server/client/control/components/RecipeForm.js
@@ -170,6 +170,7 @@ export class RecipeForm extends React.Component {
     return {
       isCloning: !!(route && route.isCloning),
       isUserRequester: requestAuthorID === currentUserID,
+      isPeerApprovalEnforced: document.documentElement.dataset.peerApprovalEnforced === 'true',
       isAlreadySaved: !!recipeId,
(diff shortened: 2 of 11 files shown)


docs/ops/config.rst
```diff
@@ -255,6 +255,16 @@ in other Django projects.
    If :envvar:`DJANGO_USE_OIDC` is set to ``True``, this settings must be set to
    the URL that a user can visit to logout. It may be a relative URL.
 
+.. envvar:: DJANGO_PEER_APPROVAL_ENFORCED
```
rehandalal: I wonder if this should be named something else. My expectation with this is it disables peer approval altogether when set to `True`.

recipe-server/client/control/components/RecipeForm.js
```diff
@@ -170,6 +170,7 @@ export class RecipeForm extends React.Component {
     return {
       isCloning: !!(route && route.isCloning),
       isUserRequester: requestAuthorID === currentUserID,
+      isPeerApprovalEnforced: document.documentElement.dataset.peerApprovalEnforced === 'true',
```
rehandalal: same naming issue.

recipe-server/normandy/recipes/models.py
```diff
@@ -423,12 +425,25 @@ class NotActionable(Exception):
     class CannotActOnOwnRequest(Exception):
         pass
 
+    def verify_approver(self, approver):
+        if approver == self.creator:
+            if settings.PEER_APPROVAL_ENFORCED:
+                raise self.CannotActOnOwnRequest()
+            else:
+                logger.warning(
+                    f'Bypassing peer approver verification because it is disabled.',
```
rehandalal: was this supposed to be an f-string?

docs/ops/config.rst
Osmose: My thinking is: If true, then we are enforcing the peer approval requirement; if false, we are not enforcing the peer approval requirement. Does that make sense?

recipe-server/normandy/recipes/models.py
Osmose: Whooops

docs/ops/config.rst
rehandalal: Yeah I don't love this name, but I also don't have any useful suggestions for changing it, so r+ to this name.

Osmose:
TC failure is #757. Going to merge this.

PR #770, opened 2017-05-19
Bug 1366006 - Move Heartbeat CSS into add-on.

mythmon:
This allows us to completely [remove the old heartbeat implementation](https://bugzilla.mozilla.org/show_bug.cgi?id=1366005) from mozilla-central. Plus, it makes us more self-sufficient.

I didn't add any tests here, but I'm also not sure what an appropriate test would be. I tested this manually by using the patch from bug 1366005. Without this PR, heartbeat is entirely unstyled, and with this PR it   looks normal.


recipe-client-addon/skin/shared/Heartbeat.css
```diff
@@ -0,0 +1,289 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
mythmon: Note that I just copy and pasted this from mozilla-central. I could look through it further if we want, but I don't think that is important right now.

recipe-client-addon/lib/Heartbeat.jsm
```diff
@@ -17,9 +19,32 @@ Cu.importGlobalProperties(["URL"]); /* globals URL */
...
+CleanupManager.addCleanupHandler(() => {
+  if (anyWindowsWithInjectedCss) {
+    const windowEnumerator = Services.wm.getEnumerator('navigator:browser');
+    while (windowEnumerator.hasMoreElements()) {
+      const window = windowEnumerator.getNext();
+      if (windowsWithInjectedCss.has(window)) {
+        const utils = window.QueryInterface(Ci.nsIInterfaceRequestor).getInterface(Ci.nsIDOMWindowUtils);
+        utils.removeSheet(HEARTBEAT_CSS_URI, window.AGENT_SHEET);
+        if (AppConstants.platform === 'macosx') {
+          utils.removeSheet(HEARTBEAT_CSS_URI, window.AGENT_SHEET);
```
gijsk: CSS_URI_OSX ?

recipe-client-addon/lib/Heartbeat.jsm
```diff
@@ -17,9 +19,32 @@ Cu.importGlobalProperties(["URL"]); /* globals URL */
 
 this.EXPORTED_SYMBOLS = ["Heartbeat"];
 
-const log = LogManager.getLogger("heartbeat");
 const PREF_SURVEY_DURATION = "browser.uitour.surveyDuration";
 const NOTIFICATION_TIME = 3000;
+const HEARTBEAT_CSS_URI = Services.io.newURI("resource://shield-recipe-client/skin/shared/Heartbeat.css");
+const HEARTBEAT_CSS_URI_OSX = Services.io.newURI("resource://shield-recipe-client/skin/osx/Heartbeat.css");
+
+const log = LogManager.getLogger("heartbeat");
```
gijsk: This seems to be unused?

mythmon: This line was moved from above (I re-arranged these to separate CONSTANTS from non-constants). This is used in existing parts of the code, but none of the new code.

gijsk: Oh, d'oh - sorry for missing that!

mythmon:
Taskcluster build failure is related. I've copied svg files from m-c, and the dupe checker found the files in both places. I'm not sure what to do about this. @gijsk any ideas? We'll be removing the originals with [bug 1366005](https://bugzilla.mozilla.org/show_bug.cgi?id=1366005). Maybe we just need to do these two things atomically together?

gijsk:
Yeah, just land the two things together on m-c resulting in only 1 copy of each of the files, and you should be fine. :-)

recipe-client-addon/lib/Heartbeat.jsm
```diff
@@ -17,9 +19,32 @@ Cu.importGlobalProperties(["URL"]); /* globals URL */
...
+const windowsWithInjectedCss = new WeakSet();
+let anyWindowsWithInjectedCss = false;
+
+// Add cleanup handler for CSS injected into windows by Heartbeat
+CleanupManager.addCleanupHandler(() => {
+  if (anyWindowsWithInjectedCss) {
+    const windowEnumerator = Services.wm.getEnumerator("navigator:browser");
+    while (windowEnumerator.hasMoreElements()) {
+      const window = windowEnumerator.getNext();
```
Osmose: I wonder if we want to wrap this in a generator to make it a bit more readable, like:

```js
for (const window of getAllWindows("navigator:browser")) {

}
```

May or may not be worth it.

mythmon: I think it is a pretty minor improvement. If we had to do this more than once I'd consider it, but for now I think it is better this way.

mythmon:
> or testing this, I think we can launch a few new windows, and check window.document.styleSheets to see if the Heartbeat CSS is present, then trigger a Heartbeat on that window and check again.

I tried writing this test, and it didn't work. Mossop tells me that `window.document.styleSheets` wouldn't be expected to have the style sheet when we load it this way. Because this is easy to test manually, I don't think I'm going to spend any more time on this.

@Osmose do you want to change you review now?

Here's the test I tried.

```js
// Testing CSS injection
add_task(async () => {
  const targetWindow = await BrowserTestUtils.openNewBrowserWindow();
  const bystanderWindow = await BrowserTestUtils.openNewBrowserWindow();

  function windowHasHeartbeatCSS(window) {
    for (const styleSheet of window.document.styleSheets) {
      if (styleSheet.href === 'resource://shield-recipe-client/skin/shared/Heartbeat.css') {
        return true;
      }
    }
    return false;
  }

  Assert.ok(!windowHasHeartbeatCSS(targetWindow), "Windows start without heartbeat CSS");
  Assert.ok(!windowHasHeartbeatCSS(bystanderWindow), "Windows start without heartbeat CSS");

  const hb = new Heartbeat(targetWindow, sandboxManager, {
    testing: true,
    flowId: "test",
    message: "test",
  });

  Assert.ok(windowHasHeartbeatCSS(targetWindow), "Heartbeat CSS gets added to window");
  Assert.ok(!windowHasHeartbeatCSS(bystanderWindow), "Heartbeat CSS does not get added to unrelated windows");
});
```

mythmon:
Here's my plan for the TC failure: After this get's approved, I'm going to jump through some hoops to land this on m-c, while also removing the old versions at the same time. That way we can avoid the dupe detection. It does mean that this is more complicated to land though.

PR #772, opened 2017-05-22
Update the versions of ESLint modules (especially eslint-plugin-mozilla) and cleanup some of the now unnecessary global definitions

Standard8:
ESLint-plugin-mozilla has had a lot of updates recently, especially for being used outside of mozilla-central.

As a result it handles globals and other items a lot better than it used to, and this can then be used to clean up Normandy's code.

We also have a mozilla/recommended configuration (as well as a couple of others), so I've removed duplicate rules for those.

Tests may fail until 1.0.1 of the normandy configuration is released.


Standard8:
@mythmon Thanks for the suggestions, I've updated the PR and got the tests working on CircleCI - needed to update the node version, plus some other modules, to avoid issues.

mythmon:
Awesome, this looks good to me. The TC failure is an intermittent we're tracking in #757 that we can ignore here.

PR #774, opened 2017-05-23
Gracefully handle rendering null attributes in the Recipe list view.

Osmose:
Staging is throwing the following error on the recipe listing: 

```
TypeError: r is null[Learn More] control.4b9008bfd6711a2221a7.js:59:27176
```

A snippet of the minified code around that line/column:

```js
"object"===("undefined"==typeof r?"undefined":s(r))&&(r=f["default"].createElement("ul",{className:"nested-list"},r.map(function(e,t){return f["default"].createElement("
```

Searching for `nested-list` is what landed me at `renderTableCell`, where I noticed the `typeof` check.

`typeof null === 'object'`, which means the check to see if a value is an object will pass for null attributes. However, calling .map on the value afterwards will throw if it is null. Instead, this fix uses the isObject helper from underscore.

This leaves the question of why we have null attributes in the first place, which may be related to this Sentry issue: https://sentry.prod.mozaws.net/operations/normandy-stage/issues/608396/

### DIFF
--- a/recipe-server/client/control/components/RecipeList.js
+++ b/recipe-server/client/control/components/RecipeList.js
@@ -2,6 +2,7 @@ import React, { PropTypes as pt } from 'react';
 import { connect } from 'react-redux';
 import { push } from 'react-router-redux';
 import { Table, Thead, Th, Tr, Td } from 'reactable';
+import { isObject } from 'underscore';
 
 import makeApiRequest from 'control/api';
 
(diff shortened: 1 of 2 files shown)


andymikulski:
The TC test failure looks unrelated. @Osmose do you want to check that out or is this safe to merge?

Osmose:
> The TC test failure looks unrelated. @Osmose do you want to check that out or is this safe to merge?

That failure is totally #757, so this is good. Thanks!

PR #781, opened 2017-05-25
Make recipe create/update use transactions

### LINKED ISSUE #779: Use transaction when creating new recipes
We found a recipe on staging that had no revisions (`latest_revision_id` was null). This caused issues with migrations, the recipe listing, etc. In general, the site assumes all recipes have a revision.

@rehandalal pointed out that the code that creates recipes does so by creating a new empty recipe, and then calling `recipe.revise` to create the first revision, and if `revise` fails, the recipe will still stick around. We should wrap this code in a database transaction so that the recipe is rolled back in case creating a revision fails for whatever reason.

rehandalal:
Fixes #779.

r?

### DIFF
--- a/recipe-server/normandy/base/decorators.py
+++ b/recipe-server/normandy/base/decorators.py
@@ -1,7 +1,6 @@
 from functools import wraps
 
 from django.conf import settings
-from django.db import transaction
 from django.views.decorators.cache import cache_control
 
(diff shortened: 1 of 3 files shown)


Osmose:
Build failure is #757, this doesn't touch the add-on and can be merged with the failure.

PR #786, opened 2017-05-26
Bring Mozjexl into the repo

mythmon:
There are a bunch of commits here, but the only important one is the last one. All the rest are just the git history of [TechnologyAdvice/jexl](https://github.com/TechnologyAdvice/Jexl).

For now this doesn't include any changes to Jexl, it just brings it into our repo. I've already published mozjexl v1.1.4, so these tests *should* pass.

Don't merge this with Github's tools: that would squash all the history into one commit, which isn't the goal here. After this gets approval, I'll land it by hand to preserve history.


mozjexl/LICENSE.txt
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2015 TechnologyAdvice
```
Osmose: I think we want to update this to something like:

Copyright for portions of mozJexl are held by TechnologyAdvice, 2015 as part of Jexl. All other copyright for  mozJexl are held by the Mozilla Foundation, 2017.

PR #778, opened 2017-05-24
initial version of an install and uninstall hook

crankycoder:
This tries to close off #748 by adding an install and uninstall hook into the normandy driver.

This patch does *not* currently pass test cases though - the uninstall of the XPI file causes an error somewhere down inside the Extensions.jsm module.

I could stand to use a second pair of eyes on this to see why this is breaking so horribly.

My fixture data for the XPI installer is the simplest possible webextension - just a manifest.json file with no JS code, icons or anything else.  Just the manifest zipped into a file with an XPI extension.

PR #787, opened 2017-05-31
Addons in driver

### LINKED ISSUE #747: Add driver function for reading currently-installed add-ons
The opt-out study action will need to abort early if a user already has the study add-on installed. The [AddonManager](https://developer.mozilla.org/en-US/Add-ons/Add-on_Manager/AddonManager) has `getAddonByID`, which we can use. I think this API will work:

```js
var addon = await normandy.addons.get('addon-id'); // Returns null if addon not found
```

We can't just return the [add-on object](https://developer.mozilla.org/en-US/Add-ons/Add-on_Manager/Addon) we get from `getAddonByID`, but here's the properties I'm interested in returning from it:

- `id`
- `name`
- `version`
- `installDate`
- `isActive`

@crankycoder Thoughts on the above?

mythmon:
Depends on #778, since the infrastructure for testing installing add-ons was so very useful. I'll rebase once that lands.

Fixes #747.

### DIFF
--- a/docs/dev/driver.rst
+++ b/docs/dev/driver.rst
@@ -162,6 +162,27 @@ The driver object contains the following attributes:
       String containing the distribution ID of Firefox. This value is
       ``undefined`` on Firefox versions older than 48.0.
 
+.. js:function:: addons.get(addonId)
+
+   Retrieves information about an installed add-on.
+
+   :returns: Promise that resolves with an add-on object.
+
+   The add-on object includes the following fields:
+
+
+   id
+      String containing the add-on's ID.
+   name
+      String containing the add-on's name.
+   version
+      String containing the version of the add-on.
+   installDate
+      A Date object of when the add-on was installed.
+   isActive
+      Boolean specifying whether is the add-on is active (true) or
+      disabled (false).
+
 Plugins
 -------
 .. js:class:: Plugin
--- a/recipe-client-addon/lib/NormandyDriver.jsm
+++ b/recipe-client-addon/lib/NormandyDriver.jsm
@@ -157,6 +157,29 @@ this.NormandyDriver = function(sandboxManager) {
     },
 
     addons: {
+      /**
+       * Get information about an installed add-on by ID.
+       *
+       * @param {string} addonId
+       * @returns {Object} Object containing id, installDate,
+       *                   isActive, name, type, and version of addon.
+       */
+      get: sandboxManager.wrapAsync(async function get(addonId) {
+        const addon = await AddonManager.getAddonByID(addonId);
+
+        if (!addon) {
+          return null;
+        }
+        return {
+          id: addon.id,
+          installDate: new Date(addon.installDate),
+          isActive: true,
+          name: addon.name,
+          type: addon.type,
+          version: addon.version,
+        };
+      }, {cloneInto: true}),
+
       /*
       * Return a promise that resolves to an Addon ID if installation is successful.
       */
--- a/recipe-client-addon/lib/NormandyDriver.jsm
+++ b/recipe-client-addon/lib/NormandyDriver.jsm
@@ -156,38 +156,40 @@ this.NormandyDriver = function(sandboxManager) {
       sandboxManager.removeHold(`setTimeout-${token}`);
     },
 
-    /*
-     * Return a promise that resolves to an Addon ID if installation is successful.
-     */
-    installAddon: sandboxManager.wrapAsync(async function installAddon(installUrl) {
-      const installObj = await AddonManager.getInstallForURL(installUrl, null, "application/x-xpinstall");
-      const result = new Promise((resolve, reject) => installObj.addListener({
-        onInstallEnded(addonInstall, addon) {
-          resolve(addon.id);
-        },
-        onInstallFailed(addonInstall) {
-          reject(`AddonInstall error code: [${addonInstall.error}]`);
-        },
-        onDownloadFailed() {
-          reject(`Download failed: [${installUrl}]`);
-        },
-      }));
-      installObj.install();
-      return result;
-    }),
+    addons: {
+      /*
+      * Return a promise that resolves to an Addon ID if installation is successful.
+      */
+      install: sandboxManager.wrapAsync(async function install(installUrl) {
+        const installObj = await AddonManager.getInstallForURL(installUrl, null, "application/x-xpinstall");
+        const result = new Promise((resolve, reject) => installObj.addListener({
+          onInstallEnded(addonInstall, addon) {
+            resolve(addon.id);
+          },
+          onInstallFailed(addonInstall) {
+            reject(`AddonInstall error code: [${addonInstall.error}]`);
+          },
+          onDownloadFailed() {
+            reject(`Download failed: [${installUrl}]`);
+          },
+        }));
+        installObj.install();
+        return result;
+      }),
 
-    /*
-     * Return a promise that resolves to a success messsage if
-     * addon uninstall is successful.
-     */
-    uninstallAddon: sandboxManager.wrapAsync(async function uninstallAddon(addonId) {
-      const addon = await AddonManager.getAddonByID(addonId);
-      if (addon === null) {
-        throw new Error(`No addon with ID [${addonId}] found.`);
-      }
-      addon.uninstall();
-      return null;
-    }),
+      /*
+      * Return a promise that resolves to a success messsage if
+      * addon uninstall is successful.
+      */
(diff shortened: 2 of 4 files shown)


docs/dev/driver.rst
```diff
@@ -162,6 +162,27 @@ The driver object contains the following attributes:
       String containing the distribution ID of Firefox. This value is
       ``undefined`` on Firefox versions older than 48.0.
 
+.. js:function:: addon.get(addonId)
```
Osmose: s/addon/addons/

docs/dev/driver.rst
```diff
@@ -162,6 +162,27 @@ The driver object contains the following attributes:
...
+
+   Retrieves information about an installed add-on.
+
+   :returns: Promise that resolves with a add-on object.
```
Osmose: s/a/an/

docs/dev/driver.rst
```diff
@@ -162,6 +162,27 @@ The driver object contains the following attributes:
...
+
+
+   id
+      String containing the add-on's ID.
+   name
+      String containing the add-on's name.
+   version
+      String containing the version of the add-on.
+   installDate
+      Date that the add-on was installed.
```
Osmose: Might want to be a bit more explicit that this returns a Date object.

recipe-client-addon/lib/NormandyDriver.jsm
```diff
@@ -156,10 +156,37 @@ this.NormandyDriver = function(sandboxManager) {
...
+      /**
+       * Get information about an installed add-on by ID.
+       *
+       * @param {string} addonId
+       * @returns {Object} Object containing id, installDate,
+       *                   isActive, name, type, and version of addon.
+       */
+      get(addonId) {
+        return new sandbox.Promise((resolve, reject) => {
+          AddonManager.getAddonByID(addonId, addon => {
```
Osmose: `getAddonByID` actually also returns a promise, so you can make this whole function async.

recipe-client-addon/lib/NormandyDriver.jsm
```diff
@@ -156,10 +156,37 @@ this.NormandyDriver = function(sandboxManager) {
...
+    addons: {
+      /**
+       * Get information about an installed add-on by ID.
+       *
+       * @param {string} addonId
+       * @returns {Object} Object containing id, installDate,
+       *                   isActive, name, type, and version of addon.
+       */
+      get(addonId) {
+        return new sandbox.Promise((resolve, reject) => {
```
Osmose: I think we should start using `wrapAsync` in the driver even though the entire file isn't converted over yet.

recipe-client-addon/test/browser/browser_NormandyDriver.js
```diff
@@ -90,3 +90,69 @@ add_task(withSandboxManager(Assert, async function testCreateStorage(sandboxMana
...
   `);
 }));
+
+add_task(withDriver(Assert, async function addons(driver) {
+  const ADDON_ID = "normandydriver@example.com";
+  const startDate = new Date();
+  const dir = getChromeDir(getResolvedURI(gTestPath));
+  dir.append("fixtures");
+  dir.append("normandy.xpi");
+  const xpiUrl = Services.io.newFileURI(dir).spec;
```
Osmose: Smells like a test helper waiting to be written. :P

recipe-client-addon/test/browser/browser_NormandyDriver.js
```diff
@@ -90,3 +90,69 @@ add_task(withSandboxManager(Assert, async function testCreateStorage(sandboxMana
...
+
+  let addon = await driver.addons.get(ADDON_ID);
+  Assert.equal(addon, null, "Add-on is not yet installed");
+
+  await driver.addons.install(xpiUrl);
+  addon = await driver.addons.get(ADDON_ID);
+
+  Assert.notEqual(addon, null, "Add-on object was returned");
+  // compare with seconds-level accuracy
```
Osmose: I'm not a huge fan of depending on the timing of this; maybe we should just assert the type of the date over asserting it's recentness?

recipe-client-addon/test/browser/browser_NormandyDriver.js
```diff
@@ -90,3 +90,69 @@ add_task(withSandboxManager(Assert, async function testCreateStorage(sandboxMana
...
+    type: "extension",
+  }, "Add-on is installed");
+
+  await driver.addons.uninstall(ADDON_ID);
+  addon = await driver.addons.get(ADDON_ID);
+
+  Assert.equal(addon, null, "Add-on has been uninstalled");
+}));
+
+add_task(withSandboxManager(Assert, async function testAddonsGetWorksInSandbox(sandboxManager) {
```
Osmose: Ohhhhhh, testing use inside a sandbox. I like this!

I wonder if we can make this easier for all driver functions? Should we be testing their use in a sandbox by default?

recipe-client-addon/test/browser/browser_NormandyDriver.js
mythmon: I originally had something like `Assert.ok(addon.installDate instanceof Date)`, but that never worked and I don't know why.

recipe-client-addon/test/browser/browser_NormandyDriver.js
mythmon: I tried to make a test helper for this, and it really ended up being more trouble than it was worth for now.

mythmon:
@Osmose this is ready for review again.

recipe-client-addon/test/browser/browser_NormandyDriver.js
```diff
@@ -14,11 +21,7 @@ add_task(withDriver(Assert, async function uuids(driver) {
 
 add_task(withDriver(Assert, async function installXpi(driver) {
   // Test that we can install an XPI from any URL
-  const dir = getChromeDir(getResolvedURI(gTestPath));
-  dir.append("fixtures");
-  dir.append("normandy.xpi");
-  const xpiUrl = Services.io.newFileURI(dir).spec;
-  var addonId = await driver.installAddon(xpiUrl);
+  var addonId = await driver.addons.install(xpiUrl);
```
gijsk: Err, is this meant to be testXpiUrl? Or something else?

docs/dev/driver.rst
```diff
@@ -162,6 +162,27 @@ The driver object contains the following attributes:
       String containing the distribution ID of Firefox. This value is
       ``undefined`` on Firefox versions older than 48.0.
 
+.. js:function:: addons.get(addonId)
```
gijsk: I'm a little confused why the context doesn't contain the docs for install/uninstall. Are they documented somewhere else? Naively, I would expect them to be in the same place.

mythmon: They should be here, yes. They should have been added in #778, but that PR got merged without docs for some reason.

recipe-client-addon/test/browser/browser_NormandyDriver.js
mythmon: Oops, yes.

docs/dev/driver.rst
Osmose: Yep, I missed that. My bad.

recipe-client-addon/test/browser/browser_NormandyDriver.js
```diff
@@ -120,3 +123,57 @@ add_task(withSandboxManager(Assert, async function testCreateStorage(sandboxMana
...
+      deepEqual(addon, {
+        id: "${ADDON_ID}",
+        name: "normandy_fixture",
+        version: "1.0",
+        installDate: addon.installDate,
+        isActive: true,
+        type: "extension",
+      }, "Add-on is accesible in the driver");
+    })();
+  `);
```
Osmose: Do we need to uninstall the add-on once this test is done so that it doesn't stay installed for other tests?

mythmon:
The test failures now are `data['Extension:Extensions'] is undefined`, which we have seen on a couple PRs. I believe @Osmose is looking into this.

recipe-client-addon/test/browser/browser_NormandyDriver.js
mythmon: Nice catch. Done.

mythmon:
I rebased this on master to pick up the `data["Extensions:Extensions"]` issues.

mythmon:
@gijsk I think this is good to go. Do you have any more issues here?

gijsk:
> @gijsk I think this is good to go. Do you have any more issues here?

Nope, this looks good, sorry for dropping this on the floor.

Issue #789, opened 2017-05-31
Cannot stop preference experiment "test" because it is already expired


mythmon:
I've been getting this on multiple branches both locally and in CI. I suspect either something in mozilla-central changed, or one of the PRs we merged without CI because of #757 was actually bad.

```
The following tests failed:
418 INFO TEST-UNEXPECTED-FAIL | browser/extensions/shield-recipe-client/test/browser/browser_PreferenceExperiments.js | A promise chain failed to handle a rejection: Cannot stop preference experiment "test" because it is already expired - rejection date: Wed May 31 2017 18:42:12 GMT+0000 (UTC) - stack: null - false == true - JS frame :: resource://testing-common/PromiseTestUtils.jsm :: assertNoUncaughtRejections :: line 257
Stack trace:
    resource://testing-common/PromiseTestUtils.jsm:assertNoUncaughtRejections:257
    chrome://mochikit/content/browser-test.js:Tester.prototype.nextTest<:445
    testScope/test_finish/<@chrome://mochikit/content/browser-test.js:1063:11
    run@chrome://mochikit/content/browser-test.js:1000:9
419 INFO TEST-UNEXPECTED-FAIL | browser/extensions/shield-recipe-client/test/browser/browser_PreferenceExperiments.js | A promise chain failed to handle a rejection: Cannot stop preference experiment "test" because it is already expired - rejection date: Wed May 31 2017 18:42:12 GMT+0000 (UTC) - stack: null - false == true - JS frame :: resource://testing-common/PromiseTestUtils.jsm :: assertNoUncaughtRejections :: line 257
Stack trace:
    resource://testing-common/PromiseTestUtils.jsm:assertNoUncaughtRejections:257
    chrome://mochikit/content/browser-test.js:Tester.prototype.nextTest<:445
    testScope/test_finish/<@chrome://mochikit/content/browser-test.js:1063:11
    run@chrome://mochikit/content/browser-test.js:1000:9
Buffered messages finished
```

PR #791, opened 2017-06-02
Fix bug 1368977: Remove es2015 and stage-0 preset from Babel.

Osmose:
The minimum Firefox version that we target for both the system add-on
and the website is Firefox 52, which includes support for most es2015
features, including native async/await syntax. This means we can safely
switch to using those features without transpiling them via Babel.

This removes both the es2015 and stage-0 presets from Babel, and
includes a few individual plugins that we still need support for.
This also adds a new Webpack plugin to enable it to parse async/await
syntax while resolving modules.

UglifyJS can't parse async/await syntax, even on the harmony branch, so
this switches to using Babili for minification. We use the webpack
plugin instead of including it in the Babel config so that code inside
node_modules, which is ignored by babel-loader, also gets minified.

This change has implications to our CDN request size, so we should be prepared to revert it if the deploy makes us push the average below the minimum level.

### DIFF
--- a/recipe-server/webpack.config.js
+++ b/recipe-server/webpack.config.js
@@ -4,6 +4,8 @@ var path = require('path');
 var webpack = require('webpack');
 var BundleTracker = require('webpack-bundle-tracker');
 var ExtractTextPlugin = require('extract-text-webpack-plugin');
+var AsyncAwaitPlugin = require('webpack-async-await');
+var BabiliPlugin = require('babili-webpack-plugin');
 var argv = require('yargs').argv;
 var childProcess = require('child_process');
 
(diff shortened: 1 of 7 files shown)


recipe-server/webpack.config.js
```diff
@@ -60,6 +62,9 @@ module.exports = [
         './client/control/sass/control.scss',
         './node_modules/font-awesome/scss/font-awesome.scss',
       ],
+      'console-log': './client/actions/console-log/index',
+      'show-heartbeat': './client/actions/show-heartbeat/index',
+      'preference-experiment': './client/actions/preference-experiment/index',
```
mythmon: What is this? Merging our two webpack configs? Or something else?

Should we remove the other webpack config?

Osmose: We should do that soon, but probably not in this PR. My mistake.

mythmon:
Taskcluster failures are both #757 (lints) and #789 (tests).

CircleCI failures are lint failures about module imports: it seems to have forgot that `control/...` is our own namespace.

Besides that, I like this. r+

> This change has implications to our CDN request size, so we should be prepared to revert it if the deploy makes us push the average below the minimum level.

I talked with @relud a bit, and we now have a plan that isn't reverting this change. If this change causes the overall CDN request size average to drop below the minimum size, we should be able to move to a different CDN without those requirements.

Osmose:
@mythmon I force-pushed before remembering that you set an approval on the previous commit, sorry!

I updated this with Dockerfile/package fixes for the lint task, which was failing in CI. I also rebased on master, which should fix the TC errors as well. We can trust automation again!

This probably needs another approval of the changes I made before we merge.

mythmon:
Taskcluster failed with issue from #757 again, despite the robustcheckout fix. The log shows robustcheckout resuming after each "file not found" error, making more progress each time. It didn't finish before the 3rd error though, and the default number of retries is 3.

Osmose:
Builds are fixed now. Got lucky with the TaskCluster build, and fixed the CircleCI build with a workaround in the linting Dockerfile (see the commit message and comment in the diff for more info).

PR #793, opened 2017-06-05
recipe-server: Cachebust the x5u url for recipe signatures.

mythmon:
We are seeing weird caching issues on this URL, and this should help fix them.

I also improved test coverage for `SignatureSerializer` made our `Whatever` test helper a little nicer to use.

### DIFF
--- a/recipe-server/normandy/base/tests/__init__.py
+++ b/recipe-server/normandy/base/tests/__init__.py
@@ -23,7 +23,7 @@ class Whatever(object):
 
     @classmethod
     def contains(cls, *values):
-        name_values = ''.join(str(v) for v in values)
+        name_values = ', '.join(str(v) for v in values)
         return cls(lambda s: all(value in s for value in values), name=f'contains {name_values}')
 
     @classmethod
--- a/recipe-server/normandy/base/tests/__init__.py
+++ b/recipe-server/normandy/base/tests/__init__.py
@@ -30,7 +30,7 @@ class Whatever(object):
     def iso8601(cls):
         def is_iso8601_date(s):
             # Will throw is it does not match
-            d = datetime.strptime(s, '%Y-%m-%dT%H:%M:%S.%fZ')
+            datetime.strptime(s, '%Y-%m-%dT%H:%M:%S.%fZ')
             return True
 
         return cls(is_iso8601_date, name='datetime')
--- a/recipe-server/normandy/base/tests/__init__.py
+++ b/recipe-server/normandy/base/tests/__init__.py
@@ -1,4 +1,6 @@
 import os
+import re
+from datetime import datetime
 
 from django.contrib.auth.models import User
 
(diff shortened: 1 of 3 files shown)


recipe-server/normandy/base/tests/__init__.py
```diff
@@ -7,20 +9,42 @@
...
 
     @classmethod
     def endswith(cls, suffix):
-        return cls(lambda s: s.endswith(suffix))
+        return cls(lambda s: s.endswith(suffix), name=f'endswith {suffix}')
 
     @classmethod
     def contains(cls, *values):
-        return cls(lambda s: all(value in s for value in values))
+        name_values = ''.join(str(v) for v in values)
```
Osmose: Maybe separate these with commas?

recipe-server/normandy/base/tests/__init__.py
```diff
@@ -7,20 +9,42 @@
...
+        return cls(lambda s: all(value in s for value in values), name=f'contains {name_values}')
+
+    @classmethod
+    def iso8601(cls):
+        def is_iso8601_date(s):
+            # Will throw is it does not match
```
Osmose: Rather than throw, this should return false for the comparison. Otherwise things like deep object diffs in assertions won't work with this.

recipe-server/normandy/recipes/api/v1/serializers.py
```diff
@@ -205,13 +209,25 @@ class ClientSerializer(serializers.Serializer):
...
         fields = ['timestamp', 'signature', 'x5u', 'public_key']
 
+    def get_x5u(self, signature):
+        x5u = signature.x5u
+        if x5u is None:
+            return None
+
+        # Add cachebust parameter to x5u URL with the current time
+        url_parts = list(urlparse.urlparse(x5u))
+        query = dict(urlparse.parse_qsl(url_parts[4]))
```
Osmose: Why not just `parse_qs`?

recipe-server/normandy/recipes/api/v1/serializers.py
```diff
@@ -205,13 +209,25 @@ class ClientSerializer(serializers.Serializer):
...
+        query['cachebust'] = str(int(time.time()))
```
Osmose: This might have the side effect of hammering the content-signature server with a bunch of traffic since the cache-bust changes every second, forcing the CDN to pass traffic through to the app server. We should check with @jvehent first and maybe limit it to changing less often.

recipe-server/normandy/recipes/tests/api/v1/test_serializers.py
```diff
@@ -220,3 +225,22 @@ def test_it_works_with_no_signature(self, rf):
...
+            'signature': Whatever.regex(r'[a-f0-9]{40}'),
+            'x5u': Whatever.startswith(signature.x5u),
+            'timestamp': Whatever.iso8601(),
+            'public_key': Whatever.regex(r'[a-zA-Z0-9/+]{160}')
+        }
+
+    def test_it_cachebusts_x5u(self):
+        signature = SignatureFactory()
+        serializer = SignatureSerializer(instance=signature)
+        assert 'cachebust=' in serializer.data['x5u']
```
Osmose: Perhaps assert that it's in the query string rather than simply in the URL at all?

recipe-server/normandy/base/tests/__init__.py
mythmon: The exception thrown here is something like "blah blah blah does not match string format %blah%blah" which was really useful while making this. I think it is better this way.

PR #795, opened 2017-06-07
recipe-server: Check for expired signing certificates.

### LINKED ISSUE #792: Check dates when validating signatures
We have a validator we use to monitor recipe signatures. It currently only checks that the data matches the signature, but not that the signature is valid. It should check that the certificate used to make the signature hasn't expired yet.

mythmon:
This would have caught the certificate expiring problem that happened recently.

Fixes #792.


docs/ops/config.rst
```diff
@@ -265,6 +265,11 @@ in other Django projects.
 
    This defaults to ``False`` for local developer instances.
 
+.. envvar:: DJANGO_CERTIFICATES_EXPIRE_EARLY_DAYS
+
+   If set, when checking certificates for validity, start failing
+   system checks this many days before the certificate would expire.
```
mythmon: @relud A new setting for you. Does this look good? It doesn't have a default.

relud:
:+1:

mythmon:
@rehandalal said his review queue was pretty light, so I'm going to switch this over to him.

PR #800, opened 2017-06-08
recipe-client-addon: Remove unnecessary globals definition for ESLint, since eslint-plugin-mozilla is working correctly now.

Standard8:
This was fixed in 0.3.2, which you already have.

### DIFF
--- a/recipe-client-addon/test/browser/.eslintrc.js
+++ b/recipe-client-addon/test/browser/.eslintrc.js
@@ -8,10 +8,4 @@ module.exports = {
   plugins: [
     "mozilla"
   ],
-
-  globals: {
-    // Bug 1366720 - SimpleTest isn't being exported correctly, so list
-    // it here for now.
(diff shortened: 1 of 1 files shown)


mythmon:
The tests failed (lints passed though). I'm not sure how your change could have caused the failures, but they don't exist in the parent commit. Do you have any ideas how this could have been caused? My bets are flakiness or that m-c changed on us, but I'm not sure.

Standard8:
@mythmon Hmm, I doubt my change would have caused that. It looks like it is either an intermittent issue, or something has broken. Maybe with aritfact builds?

mythmon:
That's possible. Can you re-push to this branch, to trigger Taskcluster to retry? You can easily do that by amending your commit (`git commit --amend`), saving the commit message without changes (this changes the commit date, which changes the hash), and re-pushing (`git push --force-with-lease`).

PR #801, opened 2017-06-08
#750 - Add about:studies

### LINKED ISSUE #750: Add about:studies
Users should be able to visit `about:experiments` to see a listing of the experiments that they have participated in. An outdated visual mockup based off the about:addons listing is available [here](https://mozilla.invisionapp.com/share/UC9XM7O9X#/screens/212795383_Experiments_Tab).

To simplify things, this issue only covers adding a blank page at that URL. Subsequent issues will cover the content.

andymikulski:
Fixes #750 

- Adds `about:studies` protocol handler
- Registers `about:studies` after Shield client has started
- Adds placeholder page

This needs some tests, but the meat of the issue is here and ready for review.


andymikulski:
TC failure seems to be unrelated - tests pass locally.

recipe-client-addon/bootstrap.js
```diff
@@ -10,14 +10,18 @@ XPCOMUtils.defineLazyModuleGetter(this, "LogManager",
...
 XPCOMUtils.defineLazyModuleGetter(this, "ShieldRecipeClient",
   "resource://shield-recipe-client/lib/ShieldRecipeClient.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "AboutStudiesProtocol",
+  "resource://shield-recipe-client/lib/AboutStudiesProtocol.jsm");
 
 this.install = function() {};
 
 this.startup = async function() {
   await ShieldRecipeClient.startup();
+  AboutStudiesProtocol.register();
```
mythmon: We shouldn't be modifying `bootstrap.js` if we can avoid it. We should treat `ShieldRecipeClient` as the entrypoint for anything in the add-on. This is because it is very hard to control when bootstrap runs in tests, and putting everything in another file makes that a lot easier.

recipe-client-addon/bootstrap.js
```diff
@@ -10,14 +10,18 @@ XPCOMUtils.defineLazyModuleGetter(this, "LogManager",
...
 };
 
 this.shutdown = function(data, reason) {
+  AboutStudiesProtocol.unregister();
   ShieldRecipeClient.shutdown(reason);
 
   // Unload add-on modules. We don't do this in ShieldRecipeClient so that
```
mythmon: On the other hand, we *do* have to add `AboutStudiesProtocol.jsm` and any other modules we add to this list of modules to unload.

recipe-client-addon/lib/AboutStudiesProtocol.jsm
```diff
@@ -0,0 +1,105 @@
...
+ */
+function StudiesProtocolHandler() {}
+StudiesProtocolHandler.prototype = {
+  uri: Services.io.newURI("resource://shield-recipe-client/lib/AboutStudies.xml"),
+  classDescription: "about:studies page module",
+  classID: Components.ID("c7c3dd48-c1cf-4bbf-a5df-69eaf6cb27d9"),
+  contractID: "@mozilla.org/network/protocol/about;1?what=studies",
+  QueryInterface: XPCOMUtils.generateQI([Ci.nsIStudiesProtocolHandler]),
+
+  newChannel: (aURI) => {
```
mythmon: The `a` prefix stands for "argument", and is generally going out of style in Firefox code. This is a nit, but can we not use it?

recipe-client-addon/lib/AboutStudiesProtocol.jsm
```diff
@@ -0,0 +1,105 @@
...
+this.EXPORTED_SYMBOLS = ["AboutStudiesProtocol"];
+
+/**
+ * Component definition for the about:studies protocol handler.
+ * Registers a component with the browser that establishes an `about:studies`
+ * protocol handler. Navigating to `about:studies` displays the
+ * `AboutStudies.xml` file.
+ */
+function StudiesProtocolHandler() {}
+StudiesProtocolHandler.prototype = {
```
mythmon: Can this be either a class or a plain object? What's the benefit of using the prototype way of doing things?

recipe-client-addon/lib/AboutStudies.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0"?>
+<!DOCTYPE html>
```
mythmon: Why is this file an xml file? Can we make it `AboutStudies.xml`?

recipe-client-addon/lib/AboutStudiesProtocol.jsm
```diff
@@ -0,0 +1,105 @@
...
+  newChannel: (aURI) => {
+    let chan;
+    try {
+      chan = Services.io.newChannelFromURI2(
+        StudiesProtocolHandler.prototype.uri,
+        null,
+        Services.scriptSecurityManager.getSystemPrincipal(),
+        null,
+        Ci.nsILoadInfo.SEC_ALLOW_CROSS_ORIGIN_DATA_IS_NULL,
+        Ci.nsIContentPolicy.TYPE_OTHER
```
mythmon: We probably shouldn't use `TYPE_OTHER` here. For one, [the docs say we should avoid it](https://dxr.mozilla.org/mozilla-central/source/dom/base/nsIContentPolicyBase.idl#36). I think `TYPE_DOCUMENT` would be more appropriate.

mythmon:
Oh, I also don't think this needs to land on a feature branch. Since it doesn't change any existing features, I think it is safe to land this on master, even with the filler content.

recipe-client-addon/lib/AboutStudiesProtocol.jsm
andymikulski: I wanted to keep the `uri`, `classID`, etc tied to the ProtocolHandler class, so I just defined everything in one go. Using es6 classes we have to do something like:
```
class Whatever { ... }
Whatever.prototype = { thing: 'woo' }
```

Alternatively, we could define the `uri` etc in the class constructor, however I suspect the `Cm.registerFactory` call looks up the handler info from the prototype. I will investigate and see what we can do.

andymikulski: I pulled out the data that was being stored in the prototype into its own separate object - turns out we _dont_ need those attributes on the class itself, just need them while registering the protocol handler.

recipe-client-addon/lib/AboutStudiesProtocol.jsm
mythmon: I think you missed this one.

andymikulski: You're right, sorry! I've corrected this.

andymikulski:
@mythmon @gijsk This is ready for review. Let me know if the tests need to be more in-depth, currently I just have it testing that navigating to `about:studies` works, and that the page contains the content we expect.

mythmon:
The taskcluster failure is #806, which has been fixing master. Can you rebase on master to pick that up fix?

mythmon:
I have no idea what the CircleCI errors are. Webpack seems grumpy? I bet we didn't pin some dependency and something updated.

mythmon:
> May have something to do with one of our Babel dependencies.

One of the unpinned deps that @andymikulski just fixed was babili, so that makes a lot of sense.

gijsk:
I am sad about this. Do we need to do this?

Some general points - I don't have time to do an in-depth review right now, but from a quick look some points need addressing anyway, so...

- AboutPocket.jsm should allow you to have a better example and get rid of the boilerplate. I don't know where the register/unregister/protocol factory code came from, but Cm and various other aliases/XPCOMUtils should make this a lot less work. Maybe it's just the this.instance checking - XPCOM should be taking care of this for you.
- URI flags don't "not do anything", they control the security restrictions and abilities (e.g. can/must they load in the child process or no) of the page... they're quite important, generally speaking. I don't know if about's protocol handler here has sane defaults by now, and how they apply to things that don't use the builtin protocol redirector, but there we are.
- please verify that the page has a null principal or similar non-privileged principal in the test
- enforce this in newChannel (see aboutpocket.jsm)
- use the loadInfo parameter to create the channel (again, see aboutpocket.jsm)
- this is a new about: page and ideally it should be forced to load in the child process and not the parent (to take advantage of sandboxing). I don't know how easy it is to lazily load this module in the child process (to make the URI available there), but ideally we should try. It will need the relevant URI flags and I guess maybe we should make newChannel throw if called in the parent, I'm not sure.

mythmon:
> I am sad about this. Do we need to do this?

What in particular are you sad about? We definitely need to add some sort of user visible page. I don't think it strictly needs to be an about: page, but that seemed like the best choice. Do you have another suggestion?

gijsk:
To elaborate on my sadness: we have a tremendous amount of about: pages that were once happy aspirations of projects, with 0 UI exposure, that we have to carry (or only recently stopped carrying) around because of inertia. That list contains, off the top of my head:

- about:performance
- about:healthreport
- about:permissions (RIP)
- about:profiles
- really, about:pocket-*
- about:logo
- about:providerdirectory
- about:sync-tabs (going away soon, hopefully, in favour of things that have UI...)
- about:socialerror
- about:credits / about:license (I don't know why we have 2)
- about:cache but also about:cache-entry and about:networking (why do we have 3 separate things?)

I just sometimes get the feeling people think their project doesn't matter unless it has its own about: page. I don't think that makes sense. I think we should think carefully before adding more. Is this the best way of exposing this thing (whatever it is)? Can we just make a button somewhere generate a srcdoc or data URI?

It's not really clear to me what needs to end up living on this page. But from reading the original issue, maybe this should just be a part of about:addons ?

MattGrimes:
@gijsk The connection to about:addons is a technical relationship since we are using add-ons as a delivery mechanism. It doesn't really make sense from a user perspective. We don't want users to be confused about what they have installed themselves and what studies they are actively participating in. I've had quite a few conversations with some of the product, legal, and UX folks about this already. We'd like a unified place where users can go to see if they are currently participating in a study and view any studies they may have been a part of in the past. I don't really care if it's an about page, but that does seem (to me at least) a strong candidate based on prior art.

gijsk:
@MattGrimes OK, how does this new page get exposed? How do users find it?

Osmose:
> But from reading the original issue, maybe this should just be a part of about:addons ?

The UI for about:addons is strongly tied in with the registration of new AddonTypes; I looked into this originally, and feedback from @rhelmer and others was "Don't try to add new stuff to about:addons if at all possible".

mythmon:
As for the background of what this page is for, it is supposed to show a list of studies the user is participating in, both opt-out and opt-in, past and present. It's a legal requirement for doing opt-out studies. I definitely agree that having no UI exposure and letting this languish would be a problem, but hopefully we can avoid that. I hope that having the code isolated in the shield-recipe-client add-on will help with the languishing part: at least it is self contained.

gijsk:
> As for the background of what this page is for, it is supposed to show a list of studies the user is participating in, both opt-out and opt-in, past and present. It's a legal requirement for doing opt-out studies.

Is there supposed to be UI for the user to remove/disable studies (or opt in/out of all studies, or something) ?

Osmose:
> @MattGrimes OK, how does this new page get exposed? How do users find it?

- Studies will link to the page occasionally with links in their UIs saying things like "Why am I seeing this?" to help explain that they're in a feature experiment.
- A SUMO page will exist that will probably link to the page
- The privacy policy may also link to the page? I'm not sure about this one.

I wonder if a `resource://studies/index.html` URL would work?

Osmose:
> Is there supposed to be UI for the user to remove/disable studies (or opt in/out of all studies, or something) ?

Not in this PR, but eventually yes, the page will be used for viewing in-progress studies, and giving users the option to disable them, as well as see studies they previously participated in.

gijsk:
> > @MattGrimes OK, how does this new page get exposed? How do users find it?
> 
> - Studies will link to the page occasionally with links in their UIs saying things like "Why am I seeing this?" to help explain that they're in a feature experiment.
> - A SUMO page will exist that will probably link to the page
> - The privacy policy may also link to the page? I'm not sure about this one.

Generally speaking, Firefox doesn't let web pages (like SUMO/privacy policy) link to about: pages, as that's a security risk. about:blank is the obvious exception. We definitely shouldn't let web pages link to about: pages that can *do* anything vaguely privilege-related. It may or may not be possible to have exceptions for select mozilla-controlled properties x page combinations, but that would need code changes in Firefox proper so it would be useful to have a clear idea about what the needs are (also because this might make it more attractive to put the list in, say, about:support).

I need to go sleep now, but this stuff (how do people get there, what does the page do) all ties into security warranties/hole-punches we will need for this thing, whatever it ends up being, so fleshing this out more before pushing ahead with this PR seems useful - even if I don't know that this PR is the best place to have the discussion.

mythmon:
Regarding the CI issues: This looks like it is affecting all PRs, not just this one. I filed #812 for that problem. Lets remove the alleged fix from this branch, and try and get just a fix for just it landed ASAP.

andymikulski:
Closing until some specs are landed on.

PR #808, opened 2017-06-12
mozjexl: Handle undefined with filter expressions well.

### LINKED ISSUE #767: Filter Expressions cannot use array syntax on undefined values
CC @raymak @MattGrimes : I had to disable Kamyar's add-on bug due to this issue.

In JEXL, if you type in `foo.bar.baz`, and `foo.bar` is undefined, then the entire expression `foo.bar.baz` is returned as undefined, which is desired behavior. However, if you type `foo.bar['baz']`, and `foo.bar` is undefined, JEXL throws an exception. Instead, we'd like it to return undefined as well.

We will commonly run into this with Telemetry filters that use keys that are invalid for dotted-path attributes, like `normandy.telemetry.main.environment.addons.activeAddons['{2b10c1c8-a11f-4bad-fe9c-1c11e82cac42}']`. See [bug 1366125](https://bugzilla.mozilla.org/show_bug.cgi?id=1366125) for an actual example of us running into this in production.

This is a bug in JEXL, and I think it's likely that we'll end up forking JEXL to fix this.

mythmon:
Fixes #767

### DIFF
--- a/lints/package.json
+++ b/lints/package.json
@@ -17,7 +17,7 @@
     "json-loader": "0.5.4",
     "localforage": "1.4.3",
     "moment": "2.14.1",
-    "mozjexl": "1.1.4",
+    "mozjexl": "1.1.5",
     "node-sass": "3.9.3",
     "node-uuid": "1.4.7",
     "nsp": "^2.6.2",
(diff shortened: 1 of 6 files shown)


mythmon:
> nit: the .js files appear to have a weird mix of tabs and spaces.

*shakes fist at emacs*

I'm planning on doing a pass to convert all the indentation of these files over to our standard format, but not yet.

Issue #647, opened 2017-03-28
Expire active preference experiments after 30 days if we can't connect to the service


Osmose:
As a last-resort in case a preference experiment breaks something about network access, or the user is away from the network for a significant amount of time, the add-on should expire any preference experiments with a `lastSeen` date older than 30 days. 

@gregglind I think we might actually want to bump this down to something like a week instead of 30 days, as we've almost certainly lost the user by that point if their network access is still broken. Is that okay?

gregglind:
14 days?  I worry 7 is too fast or there will be some weird racing effect
where this gets called first.  Convince me this won't be called by accident
:)

Osmose:
> weird racing effect

Idea: If we only perform this check after an attempt (successful or not) to contact the Normandy server, then we shouldn't ever hit races where we'd disable a recipe before the server told us it was still active unless the user is having network issues, which is the case we're protecting against.

mythmon:
Given the experience we've had with pref studies in the wild since March, I think that this feature isn't necessary anymore.

PR #814, opened 2017-06-16
Fix #811: Do not execute action hooks if fetching recipes fails.

### LINKED ISSUE #811: Post-recipe-execution hooks are not run when recipe signatures are invalid
If the signature check for recipes fails, `RecipeRunner` fails to run the post-execution hooks for each action. We should probably make sure that actions have a chance to clean up after themselves in the case of a recipe signature failure, or we should decide to not run anything if the recipe signatures fail. I'm not entirely sure which of the two we should choose, but I'm leaning towards not running them at all, as a signature failure should be treated IMO as a failure to reach the server at all, vs the server explicitly saying that there are no recipes to run.

Currently, the only action with pre- and post- execution handlers is preference experiments, and the main consequence of not running them vs running them is that, if we do run them, signature failures would cause users to be unenrolled from their active experiments. If we didn't run them, they'd remain in the experiment.

Osmose:
This explicitly prevents side-effects from running pre-execution hooks
and then returning before we run the post-execution hooks.

After thinking a bit more, I'm more confident this is the behavior we want. If we want users to be unenrolled from an experiment when a recipe fetch goes wrong, we should be expiring them after not hearing from the server instead of running pre- and post-execution hooks even when recipe fetches fail.

Also, it turns out the sinon upgrade changed how you create a stub with a fake function. The old form still works, but a spy is returned instead of a stub, which breaks some things.

### DIFF
--- a/recipe-client-addon/lib/RecipeRunner.jsm
+++ b/recipe-client-addon/lib/RecipeRunner.jsm
@@ -105,6 +105,16 @@ this.RecipeRunner = {
       await ClientEnvironment.getClientClassification();
     }
 
+    // Fetch recipes before execution in case we fail and exit early.
+    let recipes;
+    try {
+      recipes = await NormandyApi.fetchRecipes({enabled: true});
+    } catch (e) {
+      const apiUrl = prefs.getCharPref("api_url");
+      log.error(`Could not fetch recipes from ${apiUrl}: "${e}"`);
+      return;
+    }
+
     const actionSandboxManagers = await this.loadActionSandboxManagers();
     Object.values(actionSandboxManagers).forEach(manager => manager.addHold("recipeRunner"));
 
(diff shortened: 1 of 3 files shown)


gijsk:
Actually, thinking about this... what happens if you once fetch the recipes successfully and are offline the next time. Do we ignore the recipes we've already fetched?

Osmose:
> Actually, thinking about this... what happens if you once fetch the recipes successfully and are offline the next time. Do we ignore the recipes we've already fetched?

Yep. Otherwise mechanisms like #647 would not work since we'd be re-running old recipes.

gijsk:
So this perma-disables recipes while we're offline? I'm... not sure that's a good idea? It will lead to different Firefox app behaviour (esp. for pref experiments etc.) depending on whether Firefox can connect to the internet...

Osmose:
> So this perma-disables recipes while we're offline? I'm... not sure that's a good idea? It will lead to different Firefox app behaviour (esp. for pref experiments etc.) depending on whether Firefox can connect to the internet...

This change doesn't; that behavior is already normal. This just avoids running pre-execution hooks without running the post-execution ones as well.

Just because recipes don't run, that doesn't mean that the experiment is no longer active. Rather, if a user is offline, the recipe won't run, and the only side effect is that, if the user is offline for 7/14 days (whatever we decide in #647) then the experiment will disable itself (which we want; otherwise experiments could break internet connectivity and the user would never be able to get back). But the experiment doesn't _require_ the recipe to be run to stay active all the time.

Generally I think that recipes should be designed to run occasionally and not have large side effects if they miss an update or two. Also, not running any recipes is semantically different than running recipes and not seeing a specific recipe, which we use as a signal for rollback in several places.

PR #816, opened 2017-06-19
Fix #756: Add keys transform and intersection operator to filter expressions.

### LINKED ISSUE #756: implementing a way to get object keys
My use case: targeting users based on their addons. For instance, to target those who have one of the top privacy/security extensions, part of the JEXL would look like this:

normandy.telemetry.main.environment.addons.activeAddons['e67f8350-7edf-11e3-baa7-0800200c9a66@fri-gate.org'] || normandy.telemetry.main.environment.addons.activeAddons['proxytool@proxylist.co'] || normandy.telemetry.main.environment.addons.activeAddons['stealthyextension@gmail.com'] || normandy.telemetry.main.environment.addons.activeAddons['ipfuck@p4ul.info'] || normandy.telemetry.main.environment.addons.activeAddons['jid1-AVgCeF1zoVzMjA@jetpack'] || normandy.telemetry.main.environment.addons.activeAddons['{b749fc7c-e949-447f-926c-3f4eed6accfe}'] || normandy.telemetry.main.environment.addons.activeAddons['{9c51bd27-6ed8-4000-a2bf-36cb95c0c947}'] || normandy.telemetry.main.environment.addons.activeAddons['{455D905A-D37C-4643-A9E2-F6FEFAA0424A}'] || normandy.telemetry.main.environment.addons.activeAddons['useragentoverrider@qixinglu.com'] || normandy.telemetry.main.environment.addons.activeAddons['jid1-5Fs7iTLscUaZBgwr@jetpack'] || normandy.telemetry.main.environment.addons.activeAddons['{d40f5e7b-d2cf-4856-b441-cc613eeffbe3}'] || normandy.telemetry.main.environment.addons.activeAddons['jid0-9XfBwUWnvPx4wWsfBWMCm4Jj69E@jetpack'] || normandy.telemetry.main.environment.addons.activeAddons['{bb6bc1bb-f824-4702-90cd-35e2fb24f25d}'] || normandy.telemetry.main.environment.addons.activeAddons['clickclean@hotcleaner.com'] || normandy.telemetry.main.environment.addons.activeAddons['yesscript@userstyles.org'] || normandy.telemetry.main.environment.addons.activeAddons['requestpolicy@requestpolicy.com'] || normandy.telemetry.main.environment.addons.activeAddons['{9D6218B8-03C7-4b91-AA43-680B305DD35C}'] || normandy.telemetry.main.environment.addons.activeAddons['{2b10c1c8-a11f-4bad-fe9c-1c11e82cac42}'] || normandy.telemetry.main.environment.addons.activeAddons['jid1-dwtFBkQjb3SIQp@jetpack']

Maybe a better way to write this would be:

normandy.telemetry.main.environment.addons.activeAddons|keys|hasAny|['jid1-5Fs7iTLscUaZBgwr@jetpack', '{d40f5e7b-d2cf-4856-b441-cc613eeffbe3}', 'jid0-9XfBwUWnvPx4wWsfBWMCm4Jj69E@jetpack', ...  ]

Osmose:


recipe-client-addon/lib/FilterExpressions.jsm
```diff
@@ -41,3 +43,23 @@ this.FilterExpressions = {
...
+}
+
+/**
+ * Find all the values that are present in both lists.
+ * @param {Array} listA
+ * @param {Array} listB
+ * @return {Array}
+ */
+function operatorIntersect(listA, listB) {
+  return listA.filter(item => listB.includes(item));
```
andymikulski: This could be shortened to `return listA.filter(listB.includes);`

Osmose: I think this fails because `includes` isn't bound to `listB`. I tested in the web console:

```js
>> var list1 = [1, 2, 3]
undefined
>> var list2 = [2, 3, 4]
undefined
>> list1.filter(list2.includes)
TypeError: can't convert undefined to object
```

Osmose:
> Only tiny thing I can think of is adding a test to verify that keys on the prototype aren't included in the keys operator, but not sure how valuable testing that would be.

It wouldn't hurt to test that. 

> Would it be beneficial to test a case in which a non-object is passed into keys, or a non-array into intersect?

Yes, we should ensure that they have sensical failure modes. I suspect we should either return empty lists or `undefined` when things go bad.

Assigning back to myself to add tests for this stuff.

recipe-client-addon/lib/FilterExpressions.jsm
andymikulski: Huh, I'm surprised that's a thing!

Osmose: Yep, `this` is a funny thing. [It is set to the object that the function is called on, _not_ the object that it is defined on.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this#As_an_object_method). So from `filter`'s perspective, it's getting a standalone function and calling it directly; it doesn't know about `list2`, and so `this` isn't set unless we bind the function.

mythmon:
TC failure is #757. Since this is a add-on touching PR, we should re-run it.

recipe-client-addon/lib/FilterExpressions.jsm
gijsk: Yeah, `this` in JS is... interesting.

As an alternative, you could also pass `list2.includes.bind(list2)` which avoids explicitly declaring another function (but doesn't otherwise make much of a difference). <insert shruggy emoticon here>

Osmose:
Updated with tests and rebase. I updated the implementation as well, so this probably needs another review pass.

recipe-client-addon/lib/FilterExpressions.jsm
```diff
@@ -41,3 +43,32 @@ this.FilterExpressions = {
     return jexl.eval(onelineExpr, context);
   },
 };
+
+/**
+ * Return an array of the given object's own keys (specifically, its enumerable
+ * properties), or undefined if the argument isn't an object.
```
gijsk: Interesting choice - undefined rather than empty array?

gijsk: FWIW, I can see arguments both ways and don't really mind either way, so approved the PR. I'm also painfully aware that small changes like this sometimes get more scrutiny than much bigger ones, because humans, and that's not necessarily justified - so carry on! (see also: the now possibly-slightly-cult-ified bikeshed v. nuclear reactor example from Parkinson's law of triviality)

recipe-client-addon/lib/FilterExpressions.jsm
```diff
@@ -41,3 +43,32 @@ this.FilterExpressions = {
...
+ * @param {Object} obj
+ * @return {Array[String]|undefined}
+ */
+function keys(obj) {
+  if (typeof obj !== "object" || obj === null) {
+    return undefined;
```
andymikulski: Should this be `return null;`? I almost never use `undefined`, especially when returning a value from a method.

andymikulski: Alternatively, `return;` would be the same thing here.

recipe-client-addon/lib/FilterExpressions.jsm
Osmose: No prob bob.

JEXL uses undefined as a fallback value for property access that fails in lieu of throwing a key error (`foo.bar.baz`, if `foo` is undefined or a number or something, returns undefined), and the kind of comparisons we'd expect will fail against undefined, like `(normandy.telemetry.main.iforgottherestofthepath.addons|keys).length > 0`, so I figured it's a good stand-in for cases where, in JavaScript, we'd throw.

recipe-client-addon/lib/FilterExpressions.jsm
gijsk: Unfortunately `return` and `return undefined` behave differently for eslint's "you must either never return something or always return something" rules, which I guess is why `undefined` was made explicit here.

Osmose: [I chose undefined specifically](https://github.com/mozilla/normandy/pull/816#discussion_r123776562), and I think being explicit about that choice is best. Also, I think the linter complains when you have branches in a function that don't return an explicit value when other branches do, so `return;` would fail linting.

PR #818, opened 2017-06-19
Fix #802: Report recipe execution status to uptake telemetry.

### LINKED ISSUE #802: Implement uptake telemetry
@leplatrem [announced uptake telemetry](https://mail.mozilla.org/pipermail/firefox-dev/2017-June/005475.html) on a bunch of Firefox lists today. It helps measure uptake of update systems via Telemetry histograms. It looks rad.

This is totally a thing we want to know about Shield. We should use this.

@leplatrem With Shield, we deliver and execute several recipes when out client hits our service. Some may fail while others may succeed. Is uptake telemetry suitable for measuring the uptake of each individual recipe? Could we add a unique per-recipe identifier to the source name when calling the API? Or is it more suitable for measuring uptake of the service as a whole?

Also, for Telemetry I'm mostly familiar with ATMO and ReDash. How are we expected to view the data being sent?

CC @rhelmer since he might be more familiar with Shield and was involved in uptake telemetry.

Osmose:
This is ready for review, I think, although I don't want to merge this until I talk to the Telemetry folks about sending and analyzing this data.

Uptake Telemetry landed in 55, so we should be fine shipping this with opt-out studies if it lands.


Osmose:
On second thought, while I want feedback from @rhelmer, @gijsk should probably still be the reviewer on this.

recipe-client-addon/lib/RecipeRunner.jsm
```diff
@@ -174,13 +199,17 @@ this.RecipeRunner = {
...
       } catch (err) {
         log.info(`Could not run post-execution hook for ${actionName}:`, err.message);
+        Uptake.reportAction(actionName, Uptake.ACTION_POST_EXECUTION_ERROR);
       }
     }
 
     // Nuke sandboxes
     Object.values(actionSandboxManagers).forEach(manager => manager.removeHold("recipeRunner"));
+
+    Uptake.reportRunner(Uptake.RUNNER_SUCCESS);
```
gijsk: This seems slightly surprising in that it reports success even when there were errors in parts of the execution. Is that intentional?

recipe-client-addon/test/unit/test_NormandyApi.js
```diff
@@ -126,6 +142,73 @@ add_task(withMockApiServer(async function test_fetchRecipes() {
...
+      "signature": {"signature": "", "x5u": ""}
+    }
+  ]`);
+  const get = sinon.stub(NormandyApi, "get").resolves(response);
+
+  try {
+    await NormandyApi.fetchRecipes();
+    ok(false, "fetchRecipes did not throw for canonical JSON mismatch");
+  } catch (err) {
+    ok(err instanceof InvalidSignatureError, "Error was not an InvalidSignatureError");
```
gijsk: Nitty nit: personally, I prefer phrasing assertion messages as "X should Y" rather than "X is Y" or "X is not Y", because reading passing test messages is confusing in the "X is not Y" case (and "X is Y" is confusing in the failure case, though less so). That is, if I look at this test passing and it says:

```
TEST-PASS | Error was not an InvalidSignatureError
```
I would assume that the error not being an invalid signature error was the right thing. Especially if there are then failures later in the test, the negations cause me to make wrong assumptions about the state of the program prior to the actual test failures.

Anyway, I recognize this is potentially mostly personal preference, so if you want to leave it like this that wfm.

recipe-client-addon/test/unit/test_NormandyApi.js
```diff
@@ -126,6 +142,73 @@ add_task(withMockApiServer(async function test_fetchRecipes() {
...
+    ok(/signature/.test(err), "Error was not due to a validation error");
+  }
+
+  getApiUrl.restore();
+  get.restore();
+});
+
+const invalidSignatureServer = makeMockApiServer(do_get_file("invalid_recipe_signature_api"));
```
gijsk: I don't really understand this last test. It seems to do the same thing as the test before it, with the same assertions. What is it trying to do and how is it different from the previous test, which also checks for an invalid signature error result?

recipe-client-addon/lib/Errors.jsm
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Module of shared error subclasses used within the recipe client add-on.
+ */
+
+"use strict";
+
+this.EXPORTED_SYMBOLS = ["InvalidSignatureError"];
+
+class InvalidSignatureError extends Error {}
```
gijsk: JSMs are unfortunately not as cheap as they should be. Does this really need its own file? Maybe it could live as a property on either NormandyApi or RecipeRunner and be reused from there?

recipe-client-addon/test/unit/test_NormandyApi.js
Osmose: Signature validation can fail for two reasons:

- If the data passed in is malformed (like if the x5u cert chain isn't in pem format) it throws an exception. This is what the first test handles.
- If the data is in the correct format but the signature doesn't match the content, it returns false. This is what the second test handles.

Uptake has a single status for signature issues, and I used it for both of these without thinking about it too much; on one hand, we could mark malformed data problems as a server issue, but on the other hand, it's still a problem that was raised during signature verification.

One is done using mocking and the other isn't mostly because I wrote the mock one first, and when I wanted to write the second one, the test data was long enough that I didn't want to keep it in JS. If we care, I could have them both use a mock API server for consistency's sake.

recipe-client-addon/test/unit/test_NormandyApi.js
Osmose: I don't have much of a preference except that we should document a preference. Right now we're inconsistent across the codebase. I'll roll with your suggestion.

(Well, that's a lie; my preference is the style we use with pytest, where assertions generally use the auto-generated "blah != foo" message unless the assertion is tricky enough to need more context, but that relies heavily on the context that comes with pytest's output.)

recipe-client-addon/lib/RecipeRunner.jsm
Osmose: Yes. Runner status is meant to track whether the runner as a whole was successful, vs things that stop the runner entirely (like errors fetching the recipes, or uncaught exceptions). This would allow, for example, us to see what percentage of all recipe runs are affected by a fetch problem, even if, at the same time, an action is also throwing an error.

recipe-client-addon/lib/Errors.jsm
Osmose: /me sighs

Fiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiine

Osmose:
I re-summon @rhelmer, per Gijs' request for someone familiar with uptake telemetry

Osmose:
I checked with rweiss and we need to document this data and get a data review, so I'll also need to add that to this PR.

recipe-client-addon/lib/Errors.jsm
gijsk: I'm sorry! I would offer getting you drinks, but I'm missing SF, so it'll have to wait until Cancun... :-(

recipe-client-addon/test/unit/test_NormandyApi.js
gijsk: The existing state wfm, but I think it would be worth having 2 comments, one for each test, explaining specifically what these tests are testing. :-)

rhelmer:
@leplatrem is the author of UptakeTelemetry, I think I get the gist of it but I'd suggest him if you'd like additional reviewers :)

Osmose:
This is ready for another review pass.

@rjweiss r? on the documentation for the new uptake data we're reporting to telemetry: https://github.com/mozilla/normandy/pull/818/files#diff-40f88927acb0373dd06756244c2ee0ae

@georgf Sunah said I should ask one of the telemetry client folks about sending this uptake data, and I figured since you were on the original uptake bug, you'd be a good person to glance over this. f?

@leplatrem I can't assign you since you're not part of the Mozilla Github org, but if you want to double-check my usage of uptake telemetry here, that'd be rad. :D

docs/user/data_collection.rst
```diff
@@ -0,0 +1,89 @@
+Data Collection
+===============
+This document describes the types of data that Shield collects.
+
+Uptake
+------
+Shield monitors the execution of :ref:`recipes <recipes>` and reports to
+`Firefox Telemetry`_ the amount of successful and failed runs. This data is
+reported as a `main-ping histogram`_ under the
+``UPTAKE_REMOTE_CONTENT_RESULT_1`` key. The type and format of this histogram
```
gijsk: Nit: I was confused here for a second because I was looking for namespacing of shield here and didn't find any. It's documented further down, but maybe just mentioning there's further key-ing is helpful.

recipe-client-addon/lib/Uptake.jsm
```diff
@@ -0,0 +1,48 @@
...
+this.Uptake = {
+  // Action uptake
+  ACTION_NETWORK_ERROR: UptakeTelemetry.STATUS.NETWORK_ERROR,
+  ACTION_PRE_EXECUTION_ERROR: UptakeTelemetry.STATUS.CUSTOM_1_ERROR,
+  ACTION_POST_EXECUTION_ERROR: UptakeTelemetry.STATUS.CUSTOM_2_ERROR,
+  ACTION_SERVER_ERROR: UptakeTelemetry.STATUS.SERVER_ERROR,
+  ACTION_SUCCESS: UptakeTelemetry.STATUS.SUCCESS,
+
+  // Per-recipe uptake
+  RECIPE_ACTION_DISABLED: UptakeTelemetry.STATUS.CUSTOM_1_ERROR,
```
leplatrem: Nit: maybe you could use `UptakeTelemetry.STATUS.PREF_DISABLED` here

(https://dxr.mozilla.org/mozilla-central/rev/53477d584130945864c4491632f88da437353356/services/common/uptake-telemetry.js#30)

leplatrem:
> One slight concern is that i'm not sure if the contents (i.e. used keys) of this histogram will be discoverable.

Could you please elaborate on that, please? I can see that the different keys of remote settings were discovered properly on the telemetry measurement dashboard :)

> Another is that all the recipes will be mixed in with all other uptake users; having this in a separate probe would be cleaner and less fragile for processing.

That was part of our initial goal: have a central location where we can get every kind of uptake at a glance.

georgf:
>> One slight concern is that i'm not sure if the contents (i.e. used keys) of this histogram will be discoverable.

> Could you please elaborate on that, please? I can see that the different keys of remote settings were discovered properly on the telemetry measurement dashboard :)

Sure!
They will show up in dashboards, but i can't get a single list of all the values that we can expect:

- As a data steward, i can't see all the reported values in one place.
- As a data engineer, i can't tell easily which values are valid/expected and which not.
- As an implementing engineer or analyst, i can't easily tell which values are missing.

>> Another is that all the recipes will be mixed in with all other uptake users; having this in a separate probe would be cleaner and less fragile for processing.

> That was part of our initial goal: have a central location where we can get every kind of uptake at a glance.

Which is great. I think this works well where you want to measure uptake of single, not strongly related items.
For cases where multiple items of the same type are being reported, you have to use work-arounds though to e.g. get totals (string prefix parsing etc.).
This works, but is fragile (relies on conventions, not enforced semantics).
Having those kinds of values in separate histograms would be more clean.

But... maybe we should take this conversation out of this issue :)
I was mainly trying to summarize the offline conversation about options here.

rjweiss:
@Osmose I have a request for you in response to this data review request.  I will send you a link to a template data collection request form off-line.  Please try to answer the questions in the form here in this thread and I will use your responses to complete my review.

Osmose:
# Firefox Request for Data Collection

## 1) What are the motivating questions you want to answer with this data?

- What types of errors are happening client-side when Shield executes recipes, and how often?
- How often is an individual Shield recipe being matched to a client and executed? 

## 2) Why does Mozilla need answers to this question?  Are there benefits for users? Do we need this information to address business requirements? 

We need this information to measure the technical quality of Shield as a platform, as well as to measure the impact of filters used on Normandy recipes. Better understanding the impact of filters helps us target experiments with minimal impact to users while still hitting user populations necessary for analysis.

## 3) List all proposed measurements.  Refer to data collection types in Firefox and indicate types of measurements for each one.   General characterization of your collection will fall the highest type of data.

| High-level Description | Measurement type | Tracking Bug # |
| ---------------------- | ---------------- | -------------- |
| Measuring whether recipe execution on the client was successful, and if not, why it failed, on a per-recipe basis | Keyed histogram via [Uptake Telemetry][] | #802 |

## 4) How long will this data be collected?  Choose one of the following:

I want to permanently monitor this data. (Michael Kelly, Shield Dev Team)

## 5) What populations do you need to these measurements for?

All Desktop Firefox users: all release channels, countries, and locales.

## 6) Please provide a general description of how you will analyze this data.

We plan to aggregate this data by recipe and by action and make the aggregated counts available via the Normandy admin interface. We may also build alert-based monitoring for unexpected error counts. It's undecided if we will store the aggregated data within Normandy or pull it remotely on demand.

Longer-term, [Uptake Telemetry][] may have plans to add automated analysis of all update channels using it to report uptake.

## 7) What alternative methods did you consider to answer this question? Why were they not sufficient?

There aren't many alternatives for answering the question of how client-side recipe execution occurred. The main alternative was to only report error information to an existing error monitoring system, like Sentry, but other Firefox projects have had technical issues reporting errors from privileged Firefox code to Sentry.

## 8) Can current instrumentation answer this question?

Nope. Nothing else is currently monitoring the execution of Shield recipes.

## 9) Where do you intend to share the results of your analysis?

The Normandy admin interface. We have existing plans to add public dashboards to Normandy describing the content that is currently going out, and will evaluate whether these counts are appropriate to share on that dashboard later.

# Supplementary details

## Describe the analysis you will perform on the collected data to answer the motivating question.

- Other (please describe in extensive detail below)

The engineering team will integrate the data analysis with the Normandy admin UI, which will be monitored by engineering, product management, and other Shield users.

## What data platform are you using to collect this data?  Examples:

Unified Telemetry + custom admin integration

## If using Unified Telemetry, please indicate how your measurements are implemented according to the following:

Telemetry histogram (main ping) via [Uptake Telemetry][]

## Provide an example of the measurement artifact (e.g. JSON blob) that will be produced

Here's an example of the keyed histogram that is included with the main telemetry ping for uptake data:

```json
{
  "UPTAKE_REMOTE_CONTENT_RESULT_1": {
    "shield-recipe-client/runner": {
      "range": [
        1,
        50
      ],
      "bucket_count": 51,
      "histogram_type": 1,
      "values": {
        "0": 1,
        "1": 0
      },
      "sum": 0
    },
    "shield-recipe-client/action/show-heartbeat": {
      "range": [
        1,
        50
      ],
      "bucket_count": 51,
      "histogram_type": 1,
      "values": {
        "6": 0,
        "7": 1,
        "8": 0
      },
      "sum": 7
    },
    "shield-recipe-runner/recipe/17": {
      "range": [
        1,
        50
      ],
      "bucket_count": 51,
      "histogram_type": 1,
      "values": {
        "0": 1,
        "1": 0
      },
      "sum": 0
    },
  },
}
```

[Uptake Telemetry]: http://gecko.readthedocs.io/en/latest/toolkit/components/telemetry/telemetry/collection/uptake.html

recipe-client-addon/lib/Uptake.jsm
Osmose: `DISABLED` here doesn't mean that the user disabled anything, but rather than an error in executing part of the action code caused us to disable running any recipes using that action. Actions have pre- and post-execution hooks, and if the pre-execution hook fails, we don't run recipes using that action since they may rely on state that hasn't been initialized.

I could find any statuses that reflected this kind of partial execution error, so I went with the custom one (thank heavens for the custom ones!).

Osmose:
Rebased and feedbacked.

@mythmon We forgot to get a Normandy peer review, which should be the only remaining thing before landing this. Mind taking a look?

PR #819, opened 2017-06-20
XPI uploads

### LINKED ISSUE #740: Opt-out study action: Upload XPI files
One of the fields for the opt-out study action will be for uploading an XPI file to install. The file will be hosted by Normandy, cached heavily, and the argument value passed to the actions should be the URL where the file is hosted.

rehandalal:
Fixes #740

r?


recipe-server/normandy/settings.py
```diff
@@ -328,6 +329,11 @@ def RAVEN_CONFIG(self):
 
     PROD_DETAILS_DIR = values.Value(os.path.join(Core.BASE_DIR, 'product_details'))
 
+    # AWS settings
+    AWS_ACCESS_KEY_ID = values.Value()
+    AWS_SECRET_ACCESS_KEY = values.Value()
+    AWS_STORAGE_BUCKET_NAME = values.Value()
```
rehandalal: @relud FYI. In case these need to be set up.

relud: they will

recipe-server/client/control/components/ExtensionForm.js
```diff
@@ -0,0 +1,137 @@
...
+        await this.props.createExtension(data);
+      }
+    } catch (error) {
+      this.props.showNotification({
+        messageType: 'error',
+        message: 'Extension cannot be saved. Please correct any errors listed in the form below.',
+      });
+    }
+
+    this.props.showNotification({
```
Osmose: Shouldn't this be inside the try so that it doesn't show when there's a failure?

recipe-server/client/control/components/ExtensionForm.js
```diff
@@ -0,0 +1,137 @@
...
+    });
+  }
+
+  render() {
+    const { extensionId, handleSubmit } = this.props;
+
+    return (
+      <form className="recipe-form">
+        {
+          extensionId ?
```
Osmose: If you use `&&` here you can avoid the else clause, assuming `extensionId` is never `0`.

recipe-server/client/control/components/ExtensionForm.js
```diff
@@ -0,0 +1,137 @@
...
+
+        {
+          extensionId ?
+            <ControlField
+              label="XPI URL"
+              name="xpi"
+              component="input"
+              type="text"
+              disabled
+            />
```
Osmose: Why use a field to display this vs just a `<div>` or `<p>` or something?

recipe-server/client/control/components/ExtensionForm.js
```diff
@@ -0,0 +1,137 @@
...
+            : null
+        }
+
+        <ControlField
+          label="Upload New XPI"
+          name="xpiUpload"
+          component="input"
+          type="file"
+          accept=".xpi"
```
Osmose: TIL

recipe-server/client/control/components/ExtensionForm.js
```diff
@@ -0,0 +1,137 @@
...
+        />
+
+        <div className="form-actions">
+          <button
+            className="button action-save submit"
+            type="submit"
+            onClick={handleSubmit(this.handleSave)}
+          >Save</button>
```
Osmose: The text should be on it's own line, I think.

recipe-server/client/control/components/ExtensionList.js
```diff
@@ -0,0 +1,66 @@
...
+          >
+            <Thead>
+              <Th column="name">
+                <span>Name</span>
+              </Th>
+            </Thead>
+            {extensions.map(extension =>
+              <Tr
+                key={extension.get('id')}
+                onClick={() => this.showDetails(extension)}
```
Osmose: This creates an anonymous function once per render per extension. We probably instead want the row to be it's own component with the click handler as a method on the component. That'd avoid re-renders as well, since the anonymous functions don't compare as equal.

recipe-server/client/control/components/ExtensionForm.js
```diff
@@ -0,0 +1,137 @@
+import { Map } from 'immutable';
+import React, { PropTypes as pt } from 'react';
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
+import { initialize, reduxForm } from 'redux-form';
+
+import QueryExtension from './data/QueryExtension';
```
Osmose: Should we be using absolute import paths when possible?

recipe-server/client/control/utils/apiFetch.js
```diff
@@ -5,7 +5,9 @@ export default async function apiFetch(url, options = {}) {
 
   const headers = new Headers();
   headers.append('Accept', 'application/json');
-  headers.append('Content-Type', 'application/json');
+  if (!(options.body && options.body instanceof FormData)) {
```
Osmose: We already have `data` vs `body` for specifying different types of data in the request body, why not a `formData` one as well? Alternatively, we should be detecting the type of a single argument and making callers who pass a JSON string explicitly set their content type.

A large refactor of `apiFetch` is out of scope for this PR, though.

recipe-server/normandy/settings.py
```diff
@@ -384,6 +391,7 @@ class ProductionInsecure(Production):
     CSRF_COOKIE_SECURE = values.BooleanValue(False)
     SECURE_HSTS_SECONDS = values.IntegerValue(0)
     SESSION_COOKIE_SECURE = values.BooleanValue(False)
+    DEFAULT_FILE_STORAGE = values.Value('django.core.files.storage.FileSystemStorage')
```
Osmose: This is for the docker stuff, right? I feel like we should set this by environment variable there instead of including it in this config, as it doesn't really fit with this config profile.

recipe-server/normandy/settings.py
Osmose: We'll need to document these in the ops docs.

recipe-server/client/control/components/ExtensionForm.js
rehandalal: i didn't think we'd settled on a standard, but I'm happy to switch. I think we should probably make a decision one way or the other and stick with it.

recipe-server/client/control/components/ExtensionForm.js
rehandalal: d'oh yes!

recipe-server/client/control/components/ExtensionForm.js
rehandalal: I had tried that and was getting `NaN` printed when there was no ID. I guess React is trying to type cast `undefined` to a number.

recipe-server/client/control/components/ExtensionForm.js
rehandalal: /me nods. good call.

recipe-server/client/control/components/ExtensionList.js
rehandalal: bd

recipe-server/client/control/utils/apiFetch.js
rehandalal: yeah i just wanted to get this working but I think a refactor of this method makes sense.

recipe-server/normandy/settings.py
rehandalal: sure, i'll ditch this.

recipe-server/client/control/components/ExtensionList.js
rehandalal: reactable doesn't like making custom row components. it refuses to render them so i'm leaving this as-is for now. will get sorted out in UI change.

recipe-server/client/control/components/ExtensionForm.js
Osmose: Really? Huh. I'm a little doubtful, but leaving this as it is is fine too.

recipe-server/client/control/utils/apiFetch.js
rehandalal: to clarify... i think it makes sense, but not in this PR.

recipe-server/normandy/settings.py
rehandalal: @relud docs added: https://github.com/mozilla/normandy/pull/819/files#diff-b9426c0a74f27abcdd7237af16c8954bR282

relud: to clarify here, I will only be setting `DJANGO_AWS_STORAGE_BUCKET_NAME`, because not setting the other two should cause the backend to fall back to boto3's default ways of determining credentials, in my case ec2 instance profile credentials. You can verify that this will work by supplying aws credentials as `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` (without the `DJANGO_` prefix), because env vars is one of the fallbacks for boto3.

rehandalal: sure, we can try that. although i had trouble locally where it did not read those variables from the environment and i had to populate these settings. but i might have been doing something wrong.

recipe-server/client/control/components/ExtensionForm.js
```diff
@@ -2,18 +2,48 @@ import { Map } from 'immutable';
...
+import { ControlField } from 'control/components/Fields';
+import { showNotification } from 'control/actions/NotificationActions';
+import { createExtension, updateExtension } from 'control/state/extensions/actions';
+import { getExtension } from 'control/state/extensions/selectors';
+
+
+const adaptFileEventToValue = delegate =>
+  e => delegate(e.target.files[0]);
+
+const FileInput = ({
```
Osmose: This might be better suited for the fields modules instead of the extension form module.

recipe-server/client/control/tests/state/extensions/test_selectors.js
```diff
@@ -0,0 +1,37 @@
...
+} from '..';
+
+
+describe('getExtension', () => {
+  const STATE = {
+    ...INITIAL_STATE,
+    newState: {
+      ...INITIAL_STATE.newState,
+      extensions: {
+        ...INITIAL_STATE.newState.actions,
```
Osmose: Shouldn't this be `extensions` instead of `actions`?

rehandalal: derp! yes. fixing.

recipe-server/client/control/components/ExtensionForm.js
rehandalal: /me nods, i meant to move it and then forgot.

PR #822, opened 2017-06-21
Native async/await preference experiments / warn for conflicting preference names

Osmose:
This PR depends on #821; the last commit is the interesting bit. Once self-repair is removed, we're free to use async/await in the preference experiments action again.

### DIFF
--- a/recipe-server/client/actions/preference-experiment/index.js
+++ b/recipe-server/client/actions/preference-experiment/index.js
@@ -33,6 +33,18 @@ export default class PreferenceExperimentAction extends Action {
     // If the experiment doesn't exist yet, enroll!
     const hasSlug = await experiments.has(slug);
     if (!hasSlug) {
+      // If there's already an active experiment using this preference, abort.
+      const activeExperiments = await experiments.getAllActive();
+      const hasConflicts = activeExperiments.some(exp => exp.preferenceName === preferenceName);
+      if (hasConflicts) {
+        this.normandy.log(
+          `Experiment ${slug} ignored; another active experiment is already using the
+          ${preferenceName} preference.`, 'warn'
+        );
+        return;
+      }
+
+      // Otherwise, enroll!
       const branch = await this.chooseBranch(branches);
       await experiments.start({
         name: slug,
--- a/recipe-server/client/actions/preference-experiment/index.js
+++ b/recipe-server/client/actions/preference-experiment/index.js
@@ -16,13 +16,7 @@ export function resetAction() {
  * affects Firefox via Telemetry.
  */
 export default class PreferenceExperimentAction extends Action {
-  execute() {
-    // Babel's async transform handles the promises in such a way that it tries
-    // to call `then` on Promises from a privileged context. I'm not entirely
-    // sure why, but it works fine if we use Promises directly.
-    //
-    // Once we remove self-repair support, we should be able to use native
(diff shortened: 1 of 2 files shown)


Osmose:
Given that this change was mostly to support work on [bug 1368724](https://bugzilla.mozilla.org/show_bug.cgi?id=1368724), and that the two patches were small, I've added the commit for fixing that bug to this PR as well. Should still be an easy review over the two commits.

recipe-server/client/actions/preference-experiment/index.js
```diff
@@ -31,38 +25,35 @@ export default class PreferenceExperimentAction extends Action {
...
+        preferenceName,
+        preferenceValue: branch.value,
+        preferenceBranchType,
+        preferenceType,
       });
-    });
+    } else {
+      // If the experiment exists, and isn't expired, bump the lastSeen date.
+      const experiment = await experiments.get(slug);
+      if (experiment.expired) {
```
andymikulski: Would it be worth it to abstract this section out into its own `markExperimentSeen` function, to avoid nesting conditionals? I find it easier to read something like:

```
} else {
  await this.markExperimentSeen(slug);
}
```

This may be unnecessary, though.

Osmose: It might be worth it! I have opinions about this that I haven't really articulated before.

I run into situations like this a lot (a complex conditional that could split into a few standalone functions) and what I've found is that moving parts of the conditional into standalone functions is less comprehensible most of the time. The reason is that the purpose of the function containing the larger conditional is generally to dispatch to one or more smaller operations. 

Here, `execute`'s purpose can be viewed as "given the client state, what should we do about this preference experiment?". Moving part of the conditional into another function makes it harder to understand how this dispatch will end up, because now you need to look at two functions to determine what choice will be made. It's like a state machine where you can only see half of the nodes at a given time and need to swap between the two ends.

What I usually do (and tried to do here), is to factor out the action that you end up taking into standalone functions, but keep the conditional checking in the main function. Sometimes (and this might be an appropriate place) it's also worth it to factor out the condition into a standalone function while keeping the conditional itself in the main function. Like this:

```js
} else if (!(await this.isExpired(slug))) {
  await experiments.markLastSeen(slug);
} else {
  this.normandy.log(`Experiment ${slug} has expired, aborting.`, 'debug');
}
```

One less level of indentation makes it more readable, and it's pretty easy to name the function such that the conditional is understandable. I was kinda iffy on the `!` in front of the await though; I've never liked `!(await blah)` vs `!await blah`. I didn't give it much more thought and just went with what's here.

For the same reasons, I also generally prefer having a full `if` statement with an `else` even if the `if` has a return inside, but our linter currently doesn't like that so I don't do it as often.

What do you think? Should we go with the alternative, what's there, or do you still like your suggestion?

andymikulski: I'm personally not a fan of `!(await whatever())`, if anything I'd break that into its own variable before the conditional statement, e.g.:

```
const ok = await whatever();
if (!ok) { ... }
```
In general, I agree with your thoughts on breaking actions into standalone functions. I think a lot of it can be determined case-by-case, and this is a small piece that it may not be necessary to refactor - what's there is fine for me. If you want to replace it with the alternative, that makes sense to me, too.

PR #827, opened 2017-06-22
Fix bug 1368703: Validate preference experiment slugs.

Osmose:
If we wanted to auto-trim slugs, we'd want to do it server-side to
avoid issues with other clients submitting non-compliant slugs. We don't
yet have a good solution for server-side transformations on a per-action
basis. To avoid addressing that large issue, for now we can just validate
the slugs going forward. Long-term we'll want to address per-action
customization in a better way.

### DIFF
--- a/recipe-server/client/actions/preference-experiment/package.json
+++ b/recipe-server/client/actions/preference-experiment/package.json
@@ -19,7 +19,7 @@
                 "slug": {
                     "description": "Unique identifier for this experiment",
                     "type": "string",
-                    "default": ""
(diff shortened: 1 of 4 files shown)


Osmose:
Adding tests to hit the API was a little difficult, since none of our Django tests load up the actual actions. And since we already have test coverage that argument schema validation works on the server, I figured just testing the schema directly would be fine. It's a little less robust than testing the API itself, but appropriate for the short-term focus of this PR anyway.

Osmose:
@mythmon Rebased and added another commit to disable one of the Babili plugins to get the tests passing. Did you want to check that before we merge this?

PR #833, opened 2017-06-27
Switch JS packaging to use npm@5.

mythmon:


mythmon:
We figured out that hashes in npm@5 seems to have the resilience of a feather, and keep changing between Osmose and I. I don't think we should do this anymore.

Issue #765, opened 2017-05-18
Send backend settings to frontend via API instead of data attributes


Osmose:
Currently, there are a few bits of info that the frontend needs to fetch from the API, including the current user ID and CSRF token. This data is sent to the frontend via data attributes on the `<html>` element: https://github.com/mozilla/normandy/blob/master/recipe-server/normandy/control/templates/control/admin/base.html

In the long term, we probably want an API that returns this data, and frontend code to fetch and store the data in the state tree so that it's accessible to the frontend.

Not sure what we should name the endpoint. I'm not a fan of `settings` or `config`, they're a bit too generic. Maybe `service_config`?

PR #839, opened 2017-06-28
Add /api/v2/service_info/ API to replace html data elements.

mythmon:
This is half of what we need for #765. I'm going to switch to something else for now, and the new UI isn't actually using this data yet, as far as I can tell.

### DIFF
--- a/recipe-server/normandy/base/api/serializers.py
+++ b/recipe-server/normandy/base/api/serializers.py
@@ -17,3 +17,15 @@ class UserSerializer(serializers.ModelSerializer):
             'last_name',
             'email',
         ]
+
+
+class ServiceInfoSerializer(serializers.ModelSerializer):
+    user = UserSerializer();
+    peer_approval_enforced = serializers.BooleanField()
+
+    class Meta:
+        model = User
+        fields = [
+            'user',
+            'peer_approval_enforced',
+        ]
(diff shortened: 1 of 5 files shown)


recipe-server/normandy/base/api/serializers.py
```diff
@@ -17,3 +17,15 @@ class Meta:
             'last_name',
             'email',
         ]
+
+
+class ServiceInfoSerializer(serializers.ModelSerializer):
+    user = UserSerializer()
+    peer_approval_enforced = serializers.BooleanField()
+
+    class Meta:
```
Osmose: Do you need this meta section at all? I don't think this should be backed by user models.

recipe-server/normandy/base/api/views.py
```diff
@@ -54,6 +54,19 @@ def get(self, request):
         return Response(UserSerializer(request.user).data)
 
 
+class ServiceInfoView(APIView):
```
Osmose: A docstring mentioning that this is used for sending data to the frontend would be useful.

recipe-server/normandy/base/api/serializers.py
mythmon: Oooops. I copy/pasted this from the UserSerializer above, which is where I got this.

PR #840, opened 2017-06-28
Move lints to be per-project, not global.

mythmon:
The only lints that exist right now are for recipe-server. recipe-client-addon gets linting in Taskcluster, and I don't think we have much that we have easy lints for right now.


lints/bin/default_command
```diff
@@ -1,25 +0,0 @@
...
-        pushd $directory
-        # Temporarily disable nsp, as it fails builds if it can't reach the NSP servers
-        # See https://github.com/mozilla/normandy/issues/521
-        nsp check || true
-        if [[ $? -ne 0 ]]; then
-            num_errors=$((num_errors + 1))
-        fi
-        popd
-    fi
-done
```
Osmose: As per offline-discussion, can we re-enable NSP for package.json files that we care about as a lint step?

PR #841, opened 2017-06-28
CircleCI 2.0

mythmon:
I'll be having a conversation with Circle for a while. It probably isn't useful to review this yet.


mythmon:
Another false start. I'm going to need to think about this some more.

PR #847, opened 2017-06-28
New sidebar menu

andymikulski:
- Adds new collapsible `NavigationMenu` component to the sidebar.
- Provides example links to filtered listings (nonexistant at this point).

### DIFF
--- a/recipe-server/client/control_new/components/common/NavigationMenu.js
+++ b/recipe-server/client/control_new/components/common/NavigationMenu.js
@@ -6,73 +6,47 @@ import { Link } from 'redux-little-router';
 
 const { SubMenu, Item } = Menu;
 
-export class NavigationMenu extends React.Component {
-  static propTypes = {
-    router: PropTypes.object.isRequired,
-  };
+export function NavigationMenu({ router }) {
+  const { pathname, search } = router;
 
-  static menuList = {
-    Home: '/',
-    Recipes: {
-      'View All': '/recipes',
-      // These are just examples - later, the URLs will point to filtered listings.
-      Heartbeat: '/recipes?heartbeat',
-      'console-log': '/recipes?console-log',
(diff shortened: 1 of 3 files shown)


recipe-server/client/control_new/components/common/NavigationMenu.js
```diff
@@ -0,0 +1,82 @@
...
+  /**
+   * Given an object of title:url's, recursively builds a tree navigation UI.
+   * Steps through each nested menu and creates `SubMenu`s if needed, otherwise
+   * creates an `Item` with a link inside of it.
+   *
+   * @param  {Object} menu       Menu definition object, can contain nested menus.
+   * @param  {String} namespace? Name of the current menu being explored.
+   * @return {Node}   Compiled Item or SubMenu node.
+   */
+  generateMenuContents(menu, namespace = '') {
```
rehandalal: You could probably just make this a `MenuItem` and use `.map` in the `render()` function below.

andymikulski: Can you elaborate on this? To support nested menus we need to call this recursively and spit out `MenuItem`s an `SubMenu`s as necessary, I'm not sure how this suggestion fits in to that.

andymikulski: Oh wait, do you mean converting `menuList` into a bunch of React components?

andymikulski:
I've switched this over to using straight up React components. I don't _love_ this implementation, mostly because we have to duplicate values in a few places. Ant's `Menu` components apparently rely on React's `key` prop to determine what submenu/item is open/selected. Attempting to abstract out our own `NavItem` for example causes React to mess with `key`s, which ultimately breaks that functionality. As a result, I've had to drop a bunch of nested components into the main `render` function.

@rehandalal Check this out and let me know if it closer matches your vision.

PR #848, opened 2017-06-28
Add 404 page

andymikulski:
Adds simple `MissingPage` component to keep the 404 page in line stylistically with the rest of the site.


recipe-server/client/control_new/less/main.less
```diff
@@ -3,6 +3,7 @@
 @import 'ant-override';
 
 // Pages
+@import 'pages/404';
```
Osmose: Less files mirroring our components? Yesssssssssssssss

recipe-server/client/control_new/routes.js
```diff
@@ -47,14 +48,13 @@ export const {
...
-    return (
-      <App>
-        {router.result.component && <router.result.component />}
-      </App>
-    );
+    content = <router.result.component />;
   }
-  return <div>404</div>;
+
+  return (<App children={content} />);
```
Osmose: Unnecessary parenthesis, I think?

recipe-server/client/control_new/components/pages/MissingPage.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+export default function MissingPage() {
+  return (
+    <div className="page-404">
+      <h1>404</h1>
+      <h3>Page not found.</h3>
```
Osmose: Nit: [MDN recommends avoiding skipping heading levels](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements). Since this isn't really a heading anyway, this should probably be a `p` tag anyway.

PR #851, opened 2017-06-29
Fix #829: Add edit and create pages for extensions.

### LINKED ISSUE #829: Redesigned extensions form
The extensions create/update form needs to be updated to use Ant.

Osmose:


andymikulski:
Looks like this hit some eslint issues:

```
===============================================================================
FAILED: eslint
===============================================================================

/app/code/recipe-server/client/control_new/components/extensions/EditExtensionPage.js
  2:22  error  Unable to resolve path to module 'autobind-decorator'  import/no-unresolved

/app/code/recipe-server/client/control_new/components/common/NavigationCrumbs.js
  2:22  error  Unable to resolve path to module 'autobind-decorator'  import/no-unresolved

/app/code/recipe-server/client/control_new/components/extensions/ExtensionForm.js
  2:22  error  Unable to resolve path to module 'autobind-decorator'  import/no-unresolved

/app/code/recipe-server/client/control_new/state/router/selectors.js
  6:1  error  Prefer default export  import/prefer-default-export

/app/code/recipe-server/client/control_new/components/common/FormActions.js
  18:39  error  'children' is missing in props validation  react/prop-types
  29:53  error  'children' is missing in props validation  react/prop-types
  38:57  error  'children' is missing in props validation  react/prop-types

/app/code/recipe-server/client/control_new/components/common/BaseAntForm.js
  1:22  error  Unable to resolve path to module 'autobind-decorator'  import/no-unresolved

/app/code/recipe-server/client/control_new/components/extensions/CreateExtensionPage.js
  2:22  error  Unable to resolve path to module 'autobind-decorator'  import/no-unresolved
```

The `autobind-decorator` issue is popping up in every PR, so I think we can ignore those at this time, but there are a handful of other issues.

recipe-server/client/control_new/components/common/BaseAntForm.js
```diff
@@ -0,0 +1,94 @@
...
+    try {
+      const values = await this.validateFields();
+      this.props.onSubmit(values);
+    } catch (error) {
+      message.error('Could not validate form. Please correct the errors below.');
+    }
+  }
+
+  @autobind
+  validateFields() {
```
andymikulski: This can be `async`, especially since we `await` this function elsewhere.

recipe-server/client/control_new/components/common/BaseAntForm.js
```diff
@@ -0,0 +1,94 @@
...
+   *         <this.FormItem name="myfield" label="My Field" initialValue="foo">
+   *           <Input />
+   *         </this.FormItem>
+   *       </Form>
+   *     );
+   *   }
+   * }
+   */
+  @autobind
+  FormItem({ children, name, initialValue, rules, config = {}, ...customItemProps }) {
```
andymikulski: I understand what you're doing here, but I think this is a _little_ weird. Since this is essentially its own component (takes children, props, etc), my instinct is to split this off into its own class/file instead of nesting it here, but I feel like there's a better solution than that. I'll think about it some more.

recipe-server/client/control_new/components/extensions/CreateExtensionPage.js
```diff
@@ -0,0 +1,57 @@
...
+import React, { PropTypes as pt } from 'react';
+import { connect } from 'react-redux';
+import { push } from 'redux-little-router';
+
+import ExtensionForm from 'control_new/components/extensions/ExtensionForm';
+import { createExtension } from 'control_new/state/extensions/actions';
+
+
+@autobind
+export class _CreateExtensionPage extends React.Component {
```
andymikulski: We haven't used this notation elsewhere. IMO, I don't think we particularly need to denote if the class is private or not, since we have `export`/`export default`.

Reckon this is a good thing for @rehandalal and/or @mythmon to chime in on.

recipe-server/client/control_new/components/extensions/CreateExtensionPage.js
```diff
@@ -0,0 +1,57 @@
...
+  static propTypes = {
+    createExtension: pt.func.isRequired,
+    push: pt.func.isRequired,
+  }
+
+  constructor(props) {
```
andymikulski: We _should_ be able to just add `state` to the class, instead of using the constructor, e.g.:

```
class Whatever extends React.Component {
  static propTypes = {...};
  state = { formErrors: undefined, };

  // ..component functions..
}
```

recipe-server/client/control_new/components/extensions/EditExtensionPage.js
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }
+
+  render() {
+    const { extension, extensionPk } = this.props;
+    return (
+      <div>
+        <h2>Edit Extension</h2>
+        <QueryExtension pk={extensionPk} />
+        {extension &&
```
andymikulski: This might be a good opportunity to use an Ant `Spin` component - https://ant.design/components/spin/

recipe-server/client/control_new/components/extensions/EditExtensionPage.js
```diff
@@ -0,0 +1,76 @@
...
+          />
+        }
+      </div>
+    );
+  }
+}
+
+export default connect(
+  state => {
+    const extensionPk = Number.parseInt(getRouterParam(state, 'pk'), 10);
```
andymikulski: This feels strange, like it should be set aside as its own selector. @rehandalal How do you feel about this?

recipe-server/client/control_new/components/extensions/ExtensionForm.js
```diff
@@ -0,0 +1,140 @@
...
+  componentWillMount() {
+    const { extension } = this.props;
+    if (extension) {
+      this.setInitialValues(extension);
+    }
+  }
+
+  componentWillReceiveProps(newProps) {
+    // Reset form if the extension changed
+    if (!isEqual(newProps.extension, this.props.extension)) {
```
andymikulski: This probably needs a `newProps.extension && ...` added to the beginning of the conditional.

recipe-server/client/control_new/components/extensions/ExtensionForm.js
```diff
@@ -0,0 +1,140 @@
...
+    const { xpiName, xpiUrl } = this.state;
+    if (xpiName) {
+      return [{
+        uid: -1,
+        name: xpiName,
+        url: xpiUrl,
+      }];
+    }
+
+    return [];
```
andymikulski: To avoid having multiple `return`s, I'd suggest setting up a variable, filling if it necessary, and then returning it, e.g.:

```
let fileList = [];
if (xpiName){ fileList = [{ ... }]; }
return fileList;
```

The linter will occasionally flag `unreachable code` when there are multiple returns in a function, so I try to take this approach when possible.

recipe-server/client/control_new/less/main.less
```diff
@@ -47,3 +55,25 @@ body {
...
+  .primary {
+    float: right;
+  }
+
+  .secondary {
+    float: left;
+  }
+
+  .primary > * + *,
+  .secondary > * + * {
```
andymikulski: Star selectors are 'slow', is there any way we can make this more specific?

recipe-server/client/control_new/routes.js
```diff
@@ -22,16 +24,16 @@ const routes = {
         crumb: 'View Recipe',
       },
     },
-    '/extensions': {
+    '/extension': {
```
andymikulski: Two things:
- This needs a trailing slash to stay consistent with other routes.
- Links in the `Gateway` component point to `/extensions`, which should be fixed. (And in the nav menu, but that's in another open PR, so I'll fix that.)

recipe-server/client/control_new/utils/apiFetch.js
```diff
@@ -36,5 +36,28 @@ export default async function apiFetch(url, options = {}) {
...
+    } catch (error) {
+      message = error.message;
+    }
+
+    throw new ApiError(message, data);
+  }
   return response.json();
 }
+
+export class ApiError extends Error {
```
andymikulski: nit, but I'd move this to the top of the document since it's short and is used within this same file.

recipe-server/client/control_new/components/common/FormActions.js
```diff
@@ -0,0 +1,42 @@
...
+ * <FormActions>
+ *   <FormActions.Primary>
+ *     <Button>Save</Button>
+ *   </FormActions.Primary>
+ *   <FormActions.Secondary>
+ *     <Button>Delete</Button>
+ *   </FormActions.Secondary>
+ * </FormActions>
+ */
+export default function FormActions({ children }) {
```
andymikulski: I'm not sure this component is necessary at all. Here, it's basically is a div wrapper, whereas in control v1 it handles a bit more data to automatically show/hide necessary buttons. Is that the plan here?

andymikulski:
Also, a major problem I found while checking out this branch locally was when attempting to view an `Edit Extension` page, the browser would beachball and eventually ask if I want to stop the script. This isn't because of the code in this PR, but because of `QueryExtension`'s `componentWillRecieveProps`. Instead of opening a PR for the one-liner fix, I'll just put it here, since you'll probably need the fix for continued work on the Edit page:

In `QueryExtension#componentWillRecieveProps`, `if (pk !== nextProps.pk)` needs to become `if (nextProps.pk && pk !== nextProps.pk)`, and that's it!

recipe-server/client/control_new/components/common/BaseAntForm.js
```diff
@@ -0,0 +1,94 @@
...
+ * from the API.
+ */
+export default class BaseAntForm extends React.Component {
+  static propTypes = {
+    form: pt.object.isRequired,
+    onSubmit: pt.func.isRequired,
+    errors: pt.object,
+  }
+
+  @autobind
```
Osmose: I didn't know if using the autobind decorator on a base class would autobind methods on a class that inherits from it, and to which instance.

I just tested it, and it seems to only autobind the inherited method; methods that aren't on the base class, or overridden methods, do not appear to be autobound, so I'll fix this.

recipe-server/client/control_new/components/common/BaseAntForm.js
Osmose: We still need the Promise for `form.validateFields`, right? This is just to make it clear it's effectively async? Just want to make sure I'm not missing if the Promise isn't necessary.

recipe-server/client/control_new/components/extensions/CreateExtensionPage.js
Osmose: I swear this used to be the recommended style in ant-design's docs, but looking again, they name the wrapped component `WrappedThing` instead.

I went with what I thought was their style because having a default import named `CreateExtensionPage` as well as a member import named `CreateExtensionPage` that are in fact two different things is confusing; if you import both in, say, a test, you end up naming the default import something different than what the rest of the app calls it. Having a special name for the unwrapped non-default export gets around this.

I'm not partial to the underscore; `UnwrappedCreateExtensionPage`, `UnconnectedCreateExtensionPage`, or whatever else you or @rehandalal want is equally fine.

recipe-server/client/control_new/components/extensions/ExtensionForm.js
Osmose: We still want to reset the fields if the extension is undefined or null, and `isEqual` handles them fine. But yeah, the `setState` stuff won't work if they're undefined. `?.` would actually be useful here (if we weren't using Immutable, that is).

Osmose: Also, upon further reflection, I'm not sure `isEqual` even works on Immutable.js objects anyway.

recipe-server/client/control_new/components/common/FormActions.js
Osmose: I didn't plan it, but I figured that we'd be adding action button containers to enough forms that abstracting it would be useful. If we decide in the future to hide secondary actions in a collapsible container or something, we don't have to update the container since it's purely semantic.

I can remove it if you think it's overkill without the planned extra logic.

recipe-server/client/control_new/routes.js
Osmose: I can't get the nested routing to work with a trailing slash added to `/extension`. I think only leaf nodes can safely have a trailing slash. We might have to just ship without trailing slashes anywhere for the first release.

Osmose:
I think I got all the feedback that didn't need input from others. Maybe.

recipe-server/client/control_new/components/extensions/CreateExtensionPage.js
Osmose: Oh, there's also the option of using the `Wrapped` prefix for our default imports, but then the name wouldn't matched the file, and I suspect no one wants to go sprinkling the word "Wrapped" in front of every redux-connected component's filename.

recipe-server/client/control_new/components/common/BaseAntForm.js
andymikulski: Yep, you're correct - this note was strictly on marking the function as `async`.

recipe-server/client/control_new/components/extensions/CreateExtensionPage.js
andymikulski: Fair points. I'd go with `Unwrapped`/`Unconnected` over the underscore, but I'm not leaning towards one or the other. Up to you.

recipe-server/client/control_new/components/extensions/EditExtensionPage.js
andymikulski: To elaborate on this: I understand _why_ we're parsing the param here, but I feel like the point of using selectors is to avoid doing these sorts of operations inside of `connect`. Maybe I'm wrong.

recipe-server/client/control_new/components/extensions/ExtensionForm.js
andymikulski: Good point re: first comment. re: `isEqual`, I think we can use a built-in Immutable function, e.g. `newProps.extension.equals(this.props.extension)`

recipe-server/client/control_new/routes.js
andymikulski: Interesting. It may be worth submitting a patch for `redux-little-router`, but that's a greater discussion to have with the group.

recipe-server/client/control_new/components/common/FormActions.js
andymikulski: Makes sense. We can leave it in.

recipe-server/client/control_new/components/common/FormActions.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { PropTypes as pt } from 'react';
```
andymikulski: Need to use `prop-types`

recipe-server/client/control_new/components/extensions/ExtensionForm.js
```diff
@@ -40,7 +37,7 @@ export class _ExtensionForm extends BaseAntForm {
 
   componentWillReceiveProps(newProps) {
     // Reset form if the extension changed
-    if (!isEqual(newProps.extension, this.props.extension)) {
+    if (!is(newProps.extension, this.props.extension)) {
```
andymikulski: Oh cool, I had seen this in Immutable's testing but didn't realize it could be pulled into actual code. Disregard my `.equals` comment on previous review.

recipe-server/client/control_new/components/extensions/EditExtensionPage.js
Osmose: Where would the new selector live? Under `state/extensions`? Or still in `state/router`?

recipe-server/client/control_new/components/extensions/ExtensionForm.js
Osmose: I ended up using [`Immutable.is`](https://facebook.github.io/immutable-js/docs/#/is) to avoid trying to handle when `newProps.extension` is undefined.

recipe-server/client/control_new/components/extensions/EditExtensionPage.js
andymikulski: I'd say `state/extensions`, since it's handling extension-specific data.

recipe-server/client/control_new/components/extensions/CreateExtensionPage.js
Osmose: Turns out connected components have a `WrappedComponent` attribute with the unconnected version of the component, so we don't need this naming convention to get around that anymore.

PR #853, opened 2017-06-29
Recipe listing

rehandalal:
Still need to get a couple little things fixed up.


recipe-server/client/control_new/components/recipes/Listing.js
```diff
@@ -0,0 +1,241 @@
...
+  getRecipeListingCount,
+  getRecipeListingFlattenedAction,
+} from 'control_new/state/recipes/selectors';
+import {
+  getCurrentURL,
+  getQueryParameter,
+} from 'control_new/state/router/selectors';
+
+
+@connect(
```
andymikulski: I actually prefer `connect`s to be above components, but we actually don't do this anywhere else. We should stick to one convention or the other - my vote is for this one.

recipe-server/client/control_new/components/recipes/Listing.js
```diff
@@ -0,0 +1,241 @@
...
+  state => ({
+    columns: getRecipeListingColumns(state),
+    getCurrentURL: queryParams => getCurrentURL(state, queryParams),
+    searchText: getQueryParameter(state, 'searchText'),
+  }),
+  dispatch => (
+    bindActionCreators(
```
andymikulski: This formatting is weird, can we do something like:
```
bindActionCreators({
  push,
  saveWhatever,
}, dispatch)

File: recipe-server/client/control_new/components/recipes/Listing.js [hunk 3]
```diff
@@ -0,0 +1,241 @@
...
+      {
+        push,
+        saveRecipeListingColumns,
+      }, dispatch
+    )
+  )
+)
+@autobind
+class ActionBar extends React.Component {
```
andymikulski: IMO components should be split up - there are like three in this file.

recipe-server/client/control_new/components/recipes/Listing.js
```diff
@@ -0,0 +1,241 @@
...
+  static propTypes = {
+    columns: PropTypes.object,
+    getCurrentURL: PropTypes.func.isRequired,
+    push: PropTypes.func.isRequired,
+    saveRecipeListingColumns: PropTypes.func.isRequired,
+    searchText: PropTypes.string,
+  };
+
+  handleChangeSearch(value) {
+    this.props.push(
```
andymikulski: This might be cleaner to `const {push, getCurrentURL} = this.props;`

recipe-server/client/control_new/components/tables/ColumnSelector.js
```diff
@@ -0,0 +1,26 @@
+import { Button, Checkbox, Dropdown } from 'antd';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+
+export default function ColumnSelector({ columns, onChange, options }) {
```
andymikulski: This seems generic enough that we could abstract/rename it to something else, like `CheckboxMenu` or something. We'd just need to pass in a `label` prop or something to display the `Columns` text.

recipe-server/client/control_new/less/main.less
```diff
@@ -4,12 +4,34 @@
...
+.righted {
+  text-align: right;
+}
+
+.boolean-icon {
+  font-size: 16px;
+  font-weight: bold;
+
+  &.true {
+    color: #0c0;
```
andymikulski: stylelint will complain the `c` is not capitalized.

recipe-server/client/control_new/state/recipes/reducers.js
```diff
@@ -60,8 +62,26 @@ function items(state = new Map(), action) {
...
 
 
+function listing(state = new Map(), action) {
+  switch (action.type) {
+    case RECIPES_PAGE_RECEIVE:
+      return state
+        .set('count', action.recipes.count)
+        .set('pageNumber', action.pageNumber)
+        .set('results', fromJS(action.recipes.results.map(result => result.id)));
+
```
andymikulski: I'd expect eslint to complain about missing `break;`s, but maybe not.

andymikulski:
I noticed changing column sort (e.g. `Last Updated`) fires off like a million Redux actions in the console. Is it possible to batch all the `[RECIPE|REVISION]_RECEIVED` actions?

recipe-server/client/control_new/components/data/QueryRecipes.js
```diff
@@ -0,0 +1,40 @@
...
+class QueryRecipes extends React.Component {
+  static propTypes = {
+    fetchFilteredRecipesPage: PropTypes.func,
+    filters: PropTypes.object,
+    pageNumber: PropTypes.number,
+  }
+
+  componentWillMount() {
+    const { filters, pageNumber } = this.props;
+    this.props.fetchFilteredRecipesPage(pageNumber, filters);
```
rehandalal: there's a conflict with the action creator existing in a higher scope (due to being imported with the same name). i'm open to suggestions to avoid this because this is annoying.

recipe-server/client/control_new/components/data/QueryRecipes.js
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  componentWillReceiveProps(nextProps) {
+    const { filters, pageNumber } = this.props;
+    if (pageNumber !== nextProps.pageNumber || !isEqual(filters, nextProps.filters)) {
```
rehandalal: it is not.

recipe-server/client/control_new/components/recipes/Listing.js
rehandalal: oops. yes that's what this should be. don't know what happened here.

recipe-server/client/control_new/components/recipes/Listing.js
rehandalal: Same scope issue as before.

recipe-server/client/control_new/components/data/QueryRecipes.js
mythmon: The plucking works. It shadows the imported value, but it doesn't cause any problems. It is kind of a footgun though. Something we could do would be to change the import to `import * as recipeActions from 'control_new/state/recipes/actions'`, and change `connect` to reference is as `recipeActions.fetchFilteredRecipesPage`. It is a bit more verbose, and we lose some static analysis, but it avoids the shadowing problem. I think shadowing is a bigger concern.

recipe-server/client/control_new/components/common/LoadingOverlay.js
```diff
@@ -0,0 +1,29 @@
+import { Spin } from 'antd';
+import PropTypes from 'prop-types';
+import React from 'react';
+import { connect } from 'react-redux';
+
+import { areAnyRequestsInProgress } from 'control_new/state/requests/selectors';
```
andymikulski: nit: This naming feels a little verbose, maybe something a bit more succinct like `hasActiveRequests` would be better?

recipe-server/client/control_new/components/common/LoadingOverlay.js
```diff
@@ -0,0 +1,29 @@
...
+
+
+@connect(
+  state => ({
+    loading: areAnyRequestsInProgress(state) || false,
```
andymikulski: I don't think you need the `|| false`, since `loading` is tested for truthiness later and doesn't necessarily need to be a bool.

Beyond that, I'd sort of expect `areAnyRequests` to return false so we don't need to `|| false`, but I haven't looked at that code yet.

recipe-server/client/control_new/components/recipes/Listing.js
```diff
@@ -0,0 +1,185 @@
...
+} from 'control_new/state/router/selectors';
+
+
+@connect(
+  state => ({
+    columns: getRecipeListingColumns(state),
+    count: getRecipeListingCount(state),
+    getCurrentURL: queryParams => getCurrentURL(state, queryParams),
+    ordering: getQueryParameter(state, 'ordering', '-last_updated'),
+    pageNumber: parseInt(getQueryParameter(state, 'page', 1), 10),
```
andymikulski: Since I bugged Osmose about this, I'll point it out here too: this should be set into its own selector to avoid playing with the data inside of `connect`. Maybe it's worth it to build a stretchier `getQueryParameter` that does parsing for you, since we seem to run into the need of parsing ints here somewhat frequently.

recipe-server/client/control_new/components/recipes/Listing.js
```diff
@@ -0,0 +1,185 @@
...
+      push: pushAction,
+    }, dispatch)
+  )
+)
+@autobind
+export default class Listing extends React.Component {
+  static propTypes = {
+    columns: PropTypes.object,
+    count: PropTypes.number,
+    fetchFilteredRecipesPage: PropTypes.func.isRequired,
```
andymikulski: I understand this is alphabetized, but I prefer to keep the `isRequired`s towards the top of the propTypes, so it's easier to grok when looking over a component.

Perhaps we could hoist requireds to the top, then alphabetize non/requireds separately, e.g.:

```
...
fetchFilteredRecipesPage: PropTypes.func.isRequired,
getCurrentURL: PropTypes.func.isRequired,
columns: PropTypes.object,
count: PropTypes.number,
...
```

recipe-server/client/control_new/components/recipes/Listing.js
```diff
@@ -0,0 +1,185 @@
...
+    },
+
+    action() {
+      return (
+        <Table.Column
+          title="Action"
+          dataIndex="action"
+          key="action"
+          render={Listing.renderLinkedText}
+        />
```
andymikulski: No `sorter`?

recipe-server/client/control_new/components/recipes/Listing.js
```diff
@@ -0,0 +1,185 @@
...
+    },
+
+    lastUpdated({ ordering }) {
+      return (
+        <Table.Column
+          title="Last Updated"
+          key="last_updated"
+          dataIndex="last_updated"
+          render={(text, record) => {
+            const lastUpdated = moment(record.last_updated);
```
andymikulski: I think it's better practice to provide `moment` with the date format instead of having it guess. I'm not sure how `last_updated` is formatted though so maybe this is a non-issue.

recipe-server/client/control_new/components/recipes/Listing.js
```diff
@@ -0,0 +1,185 @@
...
+    const { columns, count, ordering, pageNumber, recipes, status } = this.props;
+
+    return (
+      <div>
+        <QueryRecipes
+          pageNumber={pageNumber}
+          filters={this.getFilters()}
+        />
+        <LoadingOverlay>
+          <ListingActionBar />
```
andymikulski: UX nit: Currently, this spinner disables the `New Recipe` button while loading:
screen shot 2017-06-30 at 1 12 38 pm

which is minor, but if a user is not interested in the table data, we're holding them up from doing what they want. I'd suggest just moving `ListingActionBar` out from under the `LoadingOverlay` for this reason.

--

Actually, better idea that requires more work: I think it'd look nice if we moved the `New Recipe` button up and in line with the breadcrumbs. That way, the only components 'attached' to the table are those that directly affect it.

recipe-server/client/control_new/components/recipes/ListingActionBar.js
```diff
@@ -0,0 +1,80 @@
...
+            onChange={this.props.saveRecipeListingColumns}
+            options={[
+              { label: 'Name', value: 'name' },
+              { label: 'Action', value: 'action' },
+              { label: 'Enabled', value: 'enabled' },
+              { label: 'Last Updated', value: 'lastUpdated' },
+            ]}
+          />
+        </Col>
+        <Col span={8} className="righted">
```
andymikulski: We should probably use Ant.Grid's `push` attribute instead of a `righted` class, but I tried setting it up and it came out real wonky, so maybe not.

recipe-server/client/control_new/components/tables/DataList.js
```diff
@@ -0,0 +1,81 @@
...
+    return false;
+  }
+
+  handleChangeSortFilters(pagination, filters, sorter) {
+    const filterParams = mapObject(filters, values => values.join(',') || undefined);
+
+    let ordering;
+    if (!isEmpty(sorter)) {
+      const prefix = sorter.order === 'ascend' ? '' : '-';
+      ordering = prefix + sorter.field;
```
andymikulski: I'm surprised this wasn't an eslint issue, I'd expect you to have to use a template string, e.g. 
```
ordering = `${prefix}${sorter.field}`;

File: recipe-server/client/control_new/less/main.less [hunk 19]
```diff
@@ -28,6 +31,27 @@ body {
...
+
+.righted {
+  text-align: right;
+}
+
+.boolean-icon {
+  font-size: 16px;
+  font-weight: bold;
+
+  &.true {
```
andymikulski: nit: Should this be more representative of state, e.g. `is-true`?

recipe-server/client/control_new/state/recipes/actions.js
```diff
@@ -48,29 +51,46 @@ export function fetchRecipe(pk) {
 }
 
 
-export function fetchAllRecipes() {
+export function fetchRecipesPage(pageNumber = 1) {
   return async dispatch => {
-    const requestId = 'fetch-all-recipes';
-    const recipes = await dispatch(makeApiRequest(requestId, 'v2/recipe/'));
+    const requestId = `fetch-recipes-page-${pageNumber}`;
+    const recipes = await dispatch(makeApiRequest(requestId, 'v2/recipe/', {
```
andymikulski: This probably needs to be wrapped in a `try/catch` in case the API errors out.

recipe-server/client/control_new/state/recipes/actions.js
```diff
@@ -48,29 +51,46 @@ export function fetchRecipe(pk) {
...
 }
 
 
-export function fetchFilteredRecipes(filters) {
+export function fetchFilteredRecipesPage(pageNumber = 1, filters = {}) {
   return async dispatch => {
     const filterIds = Object.keys(filters).map(key => `${key}-${filters[key]}`);
-    const requestId = `fetch-filtered-recipes-${filterIds.join('-')}`;
+    const requestId = `fetch-filtered-recipes-page-${pageNumber}-${filterIds.join('-')}`;
     const recipes = await dispatch(makeApiRequest(requestId, 'v2/recipe/', {
```
andymikulski: Try/catch thing again. Do we have an error notification/toast system set up? Maybe that needs to be another PR all together.

recipe-server/client/control_new/state/requests/selectors.js
```diff
@@ -12,3 +13,17 @@ export function isRequestInProgress(state, id) {
   const request = getRequest(state, id);
   return request.get('inProgress');
 }
+
+
+export function areAnyRequestsInProgress(state) {
+  if (state.app.requests.size === 0) {
```
andymikulski: Can we pull out `state.app.requests` for brevity?

recipe-server/client/control_new/state/requests/selectors.js
```diff
@@ -12,3 +13,17 @@ export function isRequestInProgress(state, id) {
...
+    return false;
+  } else if (state.app.requests.size === 1) {
+    return state.app.requests.first().get('inProgress');
```
andymikulski: This can become `...get('inProgress', false);` to prevent the `|| false` business later on.

recipe-server/client/control_new/state/requests/selectors.js
```diff
@@ -12,3 +13,17 @@ export function isRequestInProgress(state, id) {
...
+  }
+
+  return state.app.requests.reduce((reduced, value) => {
+    const boolValue = reduced instanceof Map ? reduced.get('inProgress') : reduced;
```
andymikulski: This seems weird - shouldn't anything that comes through here be an Immutable Map?

recipe-server/client/control_new/state/router/selectors.js
```diff
@@ -7,3 +7,24 @@
...
+
+  let queryString = '';
+  Object.keys(query).forEach(key => {
+    if (query[key]) {
+      queryString += queryString ? '&' : '?';
+      queryString += `${key}=${query[key]}`;
+    }
+  });
+
+  return `${state.router.pathname}${queryString}`;
```
andymikulski: redux-little-router has a `router.search` property which I'm pretty sure makes a majority of this function redundant.

recipe-server/client/control_new/components/recipes/Listing.js
```diff
@@ -0,0 +1,185 @@
...
+
+          <DataList
+            columns={columns}
+            columnRenderers={Listing.columnRenderers}
+            dataSource={recipes.toJS()}
+            ordering={ordering}
+            onRowClick={this.handleRowClick}
```
andymikulski: Since we use `onRowClick`, we should set `cursor: pointer;` on rows to denote to the user they can be clicked.

recipe-server/client/control_new/components/recipes/Listing.js
andymikulski: Noticed this also affects the pagination buttons, which I think is undesirable. If a user is on `Page 1` and clicks `Page 2`, but then wants to go back to `Page 1` quickly, they have to wait for the request to finish loading.

recipe-server/client/control_new/components/recipes/Listing.js
rehandalal: Nope we're sorting server side.

recipe-server/client/control_new/state/recipes/actions.js
rehandalal: Yeah, I'll file an issue for this.

recipe-server/client/control_new/components/recipes/ListingActionBar.js
rehandalal: I couldn't get that to work right.

recipe-server/client/control_new/state/recipes/actions.js
rehandalal: Yeah I'm gonna deal with notifications in another PR

rehandalal:
anyone got an idea why circle ci is failing on this?

Osmose:
> anyone got an idea why circle ci is failing on this?

The mock recipe server scripts loop through actions/recipes/etc. returned by the API and rewrite fields they known to have URLs to correct the domain name used. They assume that the APIs return lists, but this PR adds pagination, so they return objects now instead. They'll need to be updated to deal with that, and potentially to fetch all the items necessary instead of a single page of items.

This might be a sign that APIs that the system add-on hits are now being paginated, and we need to either update the client to deal with that, or make pagination optional and opt-in. Have we tested this PR against the system add-on yet?

rehandalal:
derp! we should not be paginating the v1 API. the whole point of branching off into a v2 API was to not break v1 and the system add-on.

recipe-server/client/control_new/components/recipes/Listing.js
rehandalal: We use ISO-8601 date strings so we don't need to pass the format in.

PR #857, opened 2017-06-30
Update eslint packages

mythmon:


recipe-server/package.json
```diff
@@ -70,14 +70,14 @@
     "babili-webpack-plugin": "0.1.1",
     "css-loader": "0.24.0",
     "enzyme": "2.4.1",
-    "eslint": "3.19.0",
-    "eslint-config-airbnb": "10.0.1",
-    "eslint-import-resolver-webpack": "0.7.0",
+    "eslint": "^3.19.0",
```
rehandalal: shouldn't these be pinned?

mythmon:
The lints that were leftover were because I forgot to commit the .eslint file I added to the old control interface. I also fixed the test failures I introduced. I disabled a few rules in the old interface to avoid making too many changes.

PR #858, opened 2017-07-03
Recipe detail page

rehandalal:
Building on @mythmon's branch.


recipe-server/client/control_new/components/recipes/Details.js
```diff
@@ -0,0 +1,48 @@
...
+import { Map } from 'immutable';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+
+export default function Details({ recipe }) {
+  return (
+    <div>
+      <Card key="recipe-details" title="Recipe">
+        <dl className="details">
```
mythmon: This probably also needs to show the simple filter data, if it exists. I'm not sure if you want to do that in this PR or not though.

recipe-server/client/control_new/components/recipes/DetailsActionBar.js
```diff
@@ -0,0 +1,63 @@
...
+      revisionId,
+    };
+  },
+)
+export default class DetailsActionBar extends React.Component {
+  static propTypes = {
+    isLatestApproved: PropTypes.bool,
+    recipe: PropTypes.object,
+    recipeId: PropTypes.number,
+    revisionId: PropTypes.string,
```
mythmon: The new lint rules in #857 is going to complain that these aren't required and also don't have defaults. I suggest making them required, since the selector always passed them in.

recipe-server/client/control_new/components/recipes/HistoryTimeline.js
```diff
@@ -0,0 +1,60 @@
...
+
+            let url = `/recipes/${recipeId}`;
+            if (!isLatestRevision(revision.get('id'))) {
+              url += `/rev/${revision.get('id')}`;
+            }
+
+            return (
+              <Timeline.Item color="grey" key={revision.get('id')}>
+                <Link href={url}>
+                  <Tag color={revision.get('id') === selectedRevisionId ? 'blue' : null}>
```
mythmon: Making this blue to indicate that it is selected is pretty subtle. It is easy to get it confused with the approval tags. I think we should either use a different UI treatment to show the selected revision. In @Osmose's version, the `Timeline.Item` used a different icon for the selected revision, I think that worked well.

recipe-server/client/control_new/components/recipes/RevisionApprovalTag.js
```diff
@@ -0,0 +1,73 @@
...
+    isRejected: isRejectedRevision(state, revision.get('id')),
+    isPendingApproval: isRevisionPendingApproval(state, revision.get('id')),
+  }),
+)
+export default class RevisionApprovalTag extends React.Component {
+  static propTypes = {
+    isApproved: PropTypes.bool.isRequired,
+    isLive: PropTypes.bool.isRequired,
+    isRejected: PropTypes.bool.isRequired,
+    isPendingApproval: PropTypes.bool.isRequired,
```
mythmon: I'm not sure about all these `is...` variables. They seem to be mutually exclusive, because of how you're treating them below. Maybe these should all be grouped together into something like `recipeState` that returns a string or a Symbol?

recipe-server/client/control_new/components/recipes/Details.js
rehandalal: Do any of the recipes have simple filter data?

mythmon: I'm not sure. I think we should either add the data here as new UI, or migrate everything off.

Osmose:
Nitpick: Should the clone/edit buttons be inside the header of the recipe card? It'd help get rid of that gap at the top left.

If we're planning on adding buttons for hitting the review interface from this page and for deleting the recipe, that list might get a bit crowded, too.

rehandalal:
@mythmon assigned you for the python bits.

recipe-server/client/control_new/state/serviceInfo/selectors.js
```diff
@@ -2,8 +2,8 @@ export function getCurrentUser(state, defaultsTo = null) {
   return state.app.serviceInfo.get('user', defaultsTo);
 }
 
-export function isPeerApprovalEnforced(state, defaultsTo = true) {
-  return state.app.serviceInfo.get('peer_approval_enforced', defaultsTo);
+export function isPeerApprovalEnforced(state) {
+  return state.app.serviceInfo.get('peer_approval_enforced') !== false;
```
rehandalal: ensures that `peer_approval_enforced` is always `true` unless explicitly set to `false`.

recipe-server/client/control_new/components/tables/DataList.js
```diff
@@ -46,7 +46,12 @@ export default class DataList extends React.Component {
 
   handleChangeSortFilters(pagination, filters, sorter) {
     const { getCurrentURL, push } = this.props;
-    const filterParams = mapObject(filters, values => values.join(',') || undefined);
+    const filterParams = mapObject(filters, values => {
```
andymikulski: This could go back to the one-liner with a small change:
`const filterParams = mapObject(filters, values => values && values.join(',') || undefined);`

recipe-server/client/control_new/less/ant-override.less
```diff
@@ -4,6 +4,9 @@
 @white: #FFF;
 @black: #000;
 
+// Colors
+@primary-color: #359946;
```
andymikulski: I think this can just be removed entirely.

recipe-server/client/control_new/less/main.less
```diff
@@ -23,6 +22,40 @@ body {
   }
 }
 
+dl.details {
```
andymikulski: Can this be put into `recipe-details`?

recipe-server/client/control_new/less/main.less
```diff
@@ -23,6 +22,40 @@ body {
...
+  display: grid;
```
andymikulski: Great, now I have to learn CSS grid!

recipe-server/client/control_new/less/main.less
```diff
@@ -113,9 +155,8 @@ body {
 
 /* Buttons at the bottom of a form */
 .form-actions {
-  background: #FBFBFB;
-  border: 1px solid #D9D9D9;
-  padding: 15px;
+  border-top: 1px solid #eee;
```
andymikulski: Linter should be complaining about `#eee` being lowercase. Double check that the npm script is running against `less` files.

recipe-server/client/control_new/utils/apiFetch.js
```diff
@@ -61,5 +61,10 @@ export default async function apiFetch(url, options = {}) {
 
     throw new ApiError(message, data);
   }
-  return response.json();
+
+  if (response.status !== 204) {
+    return response.json();
```
andymikulski: This should probably be in a `try/catch` block, since `.json()` likes to throw.

recipe-server/client/control_new/components/recipes/DetailsActionBar.js
```diff
@@ -0,0 +1,150 @@
...
+          isApprovable &&
+            <Button icon="question-circle" type="primary" onClick={this.handleRequestClick}>
+              Request Approval
+            </Button>
+        }
+
+        {
+          isPendingApproval &&
+            <Link href={`/recipe/${recipeId}/approval_history`}>
+              <Button icon="message" type="primary">Approval Request</Button>
```
mythmon: I don't think we should  use an animated icon here. To me that implies that we are waiting for some network request or other processing, not that the recipe is in a given state.

recipe-server/client/control_new/components/common/CurrentUserDetails.js
```diff
@@ -11,18 +12,14 @@ import {
 
 @connect(
   state => ({
-    user: getCurrentUser(state),
-    logoutUrl: getLogoutUrl(state),
+    user: getCurrentUser(state, Map()),
```
rehandalal: d'oh! good spot.

recipe-server/client/control_new/components/data/QueryRecipeHistory.js
```diff
@@ -0,0 +1,35 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
+
+import * as recipeActions from 'control_new/state/recipes/actions';
+
+@connect(
+  null,
+  dispatch => (bindActionCreators({
```
rehandalal: I have a "cleanup" branch where I've done this.

recipe-server/client/control_new/components/recipes/ApprovalForm.js
```diff
@@ -0,0 +1,52 @@
+import { Button, Form, Input } from 'antd';
+import autobind from 'autobind-decorator';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import FormActions from 'control_new/components/common/FormActions';
+import { createForm, FormItem } from 'control_new/components/common/forms';
+
+@createForm({
+  validateFields(values) {
```
rehandalal: nope, that's left over from when I was trying to figure out how to capture which button was clicked.

recipe-server/client/control_new/components/recipes/ApprovalForm.js
```diff
@@ -0,0 +1,52 @@
...
+
+  handleRejectClick(event) {
+    this.props.onSubmit(event, { approved: false });
+  }
+
+  render() {
+    const { onSubmit } = this.props;
+
+    return (
+      <Form onSubmit={onSubmit}>
```
rehandalal: oh! yeah, good call.

recipe-server/client/control_new/components/recipes/ApprovalRequest.js
```diff
@@ -0,0 +1,169 @@
...
+          </Col>
+          <Col span={8}>
+            <Card title="Approval Request" extra={extra}>
+              <div className="request-metadata">
+                <dl className="details narrow">
+                  <dt>Requested by</dt>
+                  <dd>{approvalRequest.getIn(['creator', 'email'])}</dd>
+
+                  <dt>Request Date</dt>
+                  <dd>{moment(approvalRequest.get('created')).format('MMMM Do YYYY, h:mm a')}</dd>
```
rehandalal: i dont think there will be a need to copy/paste. if there is we can deal with that when it's brought up.

recipe-server/client/control_new/components/recipes/DetailPage.js
```diff
@@ -0,0 +1,70 @@
...
+        <Row gutter={24}>
+          <Col span={16}>
+            <DetailsActionBar />
+            <RevisionNotice revision={revision} />
+            <LoadingOverlay>
+              <Details recipe={revision.get('recipe', new Map())} />
+            </LoadingOverlay>
+          </Col>
+          <Col span={8} className="recipe-history">
+            <Card title="History">
```
rehandalal: I tried this right now and it's not a quick drop in change. I don't think we should block on this. We can change this up later.

recipe-server/client/control_new/components/recipes/DetailPage.js
```diff
@@ -0,0 +1,70 @@
...
+    const { history, recipeId, revision, revisionId } = this.props;
+    return (
+      <div className="page-recipe-details">
+        <QueryRecipe pk={recipeId} />
+        <QueryRecipeHistory pk={recipeId} />
+        <Row gutter={24}>
+          <Col span={16}>
+            <DetailsActionBar />
+            <RevisionNotice revision={revision} />
+            <LoadingOverlay>
```
rehandalal: This doesn't work because it doesn't show the loading spinner after the recipe has loaded into state. Even though it is "loading" every time the page loads. Leads to weird unexpected flashes.

recipe-server/client/control_new/components/recipes/Details.js
```diff
@@ -0,0 +1,45 @@
+import { Card } from 'antd';
+import { Map } from 'immutable';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+
+export default function Details({ recipe }) {
+  return (
+    <div className="recipe-details">
+      <Card key="recipe-details" title="Recipe">
```
rehandalal: I couldn't for the life of me figure out how to do this. And this doesn't seem to work either.

rehandalal: Oh I see it's in the docs. This was definitely a super-recent addition. We probably need to upgrade ant. But I definitely don't want to do that in this PR.

recipe-server/client/control_new/components/recipes/DetailsActionBar.js
```diff
@@ -0,0 +1,150 @@
...
+
+    return (
+      <div className="details-action-bar clearfix">
+        <Link href={cloneUrl}>
+          <Button icon="swap" type="primary">Clone</Button>
+        </Link>
+
+        {
+          isLatest &&
+            <Link href={`/recipe/${recipeId}/edit`}>
```
rehandalal: I think this a larger discussion.

recipe-server/client/control_new/components/recipes/DetailsActionBar.js
rehandalal: I agree with @mythmon. That doesn't look like a button to me.

recipe-server/client/control_new/components/recipes/HistoryTimeline.js
```diff
@@ -0,0 +1,66 @@
...
+      selectedRevisionId,
+    } = this.props;
+
+    return (
+      <Timeline>
+        {
+          history.map((revision, index) => {
+            const recipeId = getRecipeIdForRevision(revision.get('id'));
+
+            const icon = <Icon type="circle-left" style={{ fontSize: '16px' }} />;
```
rehandalal: I think this is a later thing that needs some further design thought.

recipe-server/client/control_new/components/recipes/RevisionApprovalTag.js
```diff
@@ -0,0 +1,83 @@
...
+  render() {
+    const { revision, status } = this.props;
+
+    const email = revision.getIn(['approval_request', 'creator', 'email']);
+
+    let color;
+    let label;
+    let popoverContent;
+
+    if (status === REVISION_LIVE) {
```
rehandalal: nope, good call!

recipe-server/client/control_new/components/recipes/RevisionApprovalTag.js
```diff
@@ -0,0 +1,83 @@
...
+      color = 'yellow';
+      label = 'Pending Approval';
+    } else {
+      return null;
+    }
+
+    if ([REVISION_LIVE, REVISION_DISABLED, REVISION_APPROVED, REVISION_REJECTED].includes(status)) {
+      popoverContent = (
+        <div>
+          <div><strong>&quot;{revision.getIn(['approval_request', 'comment'])}&quot;</strong></div>
```
rehandalal: Sure I didn't have any styling thoughts for these but this is definitely something we can look at in the future.

rehandalal: oh I see what you mean, I'll get some class styles in for these.

recipe-server/client/control_new/components/recipes/DetailsActionBar.js
andymikulski: Fair point re: animated icon, though I still think the treatment of the `pending approval` state needs a bit more prominence. As it stands, the only visual cue we're giving to the user that the recipe is pending approval is the `Pending Approval` text in the Timeline, and a 12x12 icon in a button.

I think we can do more to visually signify "hey this recipe is pending approval and this button can help you wrap it up." Maybe use an [Alert](https://ant.design/components/alert/) near the top of the content, and drop a button in there with some text?

rehandalal: I think maybe the `RevisionNotice` should say it's pending review.

recipe-server/client/control_new/utils/apiFetch.js
```diff
@@ -62,9 +62,9 @@ export default async function apiFetch(url, options = {}) {
     throw new ApiError(message, data);
   }
 
-  if (response.status !== 204) {
+  try {
     return response.json();
```
andymikulski: Just tested this locally and we have to change how the `return` works a bit:

```
let json = null;
try {
    json = await response.json();
} catch (error) {}
return json;
```
It seems the `try` wasn't `await`ing for the json call to wrap up before moving on, so when a JSON error comes up after closing a request, the page throws an error and the user isn't redirected. Using the above approach fixed the issue for me.

mythmon: I think a slightly cleaner way to put that would be 

```js
try {
  return await response.json();
} catch(e) {
  return null;
}
```

This has the `return await` that we've been talking about in linting lately, but in this case it is actually needed (for the reason @andymikulski pointed out), and the linter understands this and will let it go.

rehandalal: @mythmon yeah that's what I was doing but then I realized I think this should go back to checking status code for 204 or actually error out if there is an issue with the JSON. 204's are the only time I expect this to error out. Anything else is probably an actual issue.

PR #859, opened 2017-07-05
New control interface: Editing recipes

Osmose:
Here's some livestream footage from while I was writing this patch:

There are a few known bugs with this patch:

- If a branch in a preference experiment has a validation error, the message overflows because it's absolutely positioned and doesn't cause the container to expand. This is deep in Ant's HTML/CSS and I'm not entirely sure how we should solve this. Should we gather branch errors in one spot?
- Switching between preference value types has interesting results. And if you just switch the type but don't update the values, I think the form still posts them as the old datatype.

I don't think they're blockers, though. Here's other relevant notes:

- I ran into this before with the `ant-redesign` branch, and briefly talked to @rehandalal about it, but the v2 recipe API uses the `action_id` field for writing what action is being used, but recipe form doesn't know the IDs of actions; it'd have to fetch all of them and then do a name lookup if it wanted the right ID. 

  I couldn't find anything in the new interface that was yet relying on nested actions being returned by the recipe API, so I refactored the API such that `recipe.action` is a string with the action's unique name. I imagine things like the recipe listing might have to do some extra work to fetch action data, so let me know if that's a problem.
- I left some comments explaining more, but Ant's form handling is not great at nesting more than one level down. The preference branches input deals with this by not using nested field names outside of displaying errors.
- I removed the lint rule that prefers functional components because Ant also had issues with them (it uses refs to track components wrapped by Form.Item). We've also had issues in the past testing functional components, so from now on I'm just going to default to class-based components where possible.


rehandalal:
The recipe details page relies on `recipe.action` so this is going to be a problem. Was there an issue with populating the options from the state other than needing to load the actions into the state?

Osmose:
> The recipe details page relies on recipe.action so this is going to be a problem. Was there an issue with populating the options from the state other than needing to load the actions into the state?

Not really.

I guess it comes down to where we want the related bits of data to be combined; currently the server returns denormalized data, actions normalize it and store it in the state, and selectors denormalize it again. My instinct would be to have two steps instead of three; server returns normalized data, and selectors denormalize it. The query components can handle triggering a fetch for the action as well if necessary, although I dunno how we'd handle components that are loading the recipe, since currently I believe we do stuff like `if (recipe)`, which doesn't say anything about whether the action we also might need has been loaded yet. Maybe the selector returns null until both are present?

@rehandalal What do you think? I may be forgetting that we've already decided that the API should return denormalized data.

rehandalal:
You'll probably want to rebase or merge in the branch `redesign`.

Osmose:
@rehandalal and I talked offline and decided to remove the backend changes and just load the actions from the API when the app starts up. This should be ready for another pass.

recipe-server/client/control_new/state/recipes/selectors.js
```diff
@@ -25,6 +26,14 @@ export function getRecipe(state, id, defaultsTo = null) {
   return defaultsTo;
 }
 
+export function getCurrentRecipePk(state) {
```
rehandalal: I ditched all the `currentRecipe` stuff which you can also see in the recipe details PR.

recipe-server/client/control_new/components/common/NavigationMenu.js
```diff
@@ -18,17 +18,17 @@ export function NavigationMenu({ router }) {
       <Item key="/"><Link href="/">Home</Link></Item>
 
       <SubMenu title="Recipes" key="Recipes">
-        <Item key="/recipes">
-          <Link href="/recipes">View All</Link>
+        <Item key="/recipe">
```
Osmose: I thought we had settled during the workweek on singular nouns since that's what the API already uses for the URLs backing these views. That's the main reason I've been changing them around, I think we should be consistent across our URLs one way or the other.

rehandalal: /me nods. I'm fine with going all singular.

recipe-server/normandy/control/templates/control/new.html
```diff
@@ -4,7 +4,7 @@
 <!DOCTYPE html>
 {% get_current_language as LANGUAGE_CODE %}
 
-<html lang="{{ LANGUAGE_CODE|default:"en-us" }}">
+<html lang="{{ LANGUAGE_CODE|default:"en-us" }}" data-csrf="{{csrf_token}}">
```
andymikulski: This should be safe for removal, right? As far as I can tell, the new `service_info` API endpoint should/would return the csrf token if needed, beyond that it looks like `apiFetch` pulls the actual cookie value itself.

recipe-server/client/control_new/state/actions/selectors.js
```diff
@@ -1,5 +1,13 @@
-/* eslint import/prefer-default-export: "off" */
-
 export function getAction(state, id, defaultsTo = null) {
-  return state.app.actions.items.get(id, defaultsTo);
+  // Convert string IDs to integers, since the state is keyed by ints.
```
andymikulski: I actually think this conversion business should remain; `getAction` is not always used within a `connect` or with router params. I found these cases after searching:

`getAction(state, revision.getIn(['recipe', 'action_id']));`
`getAction(state, recipe.get('action_id'));`
```@connect(
  (state, props) => ({
    selectedAction: getAction(
      state,
      props.form.getFieldValue('action_id'),
      new Map(),
    ),
  }),
)```

In order to keep these sections concise, it makes sense to me to leave the int parsing within `getAction`.

andymikulski:
Closing in favor of PR #875

recipe-server/normandy/control/templates/control/new.html
Osmose: I put that in here to make this PR function before the CSRF fix landed. If this is landing into a branch that already has the CSRF fix then we can drop it.

PR #860, opened 2017-07-05
#749 - Track study addon data

### LINKED ISSUE #749: Store info about installed studies
Users should be able to view a list of current and past studies that they have participated in. To show this, we need to store that data, and keep it up-to-date.

The install function in #748 should, when called, add a new entry to our internal list of studies the user has participated, storing:

- The unique name for the experiment
- The install date
- The study end date (starts off null)
- The add-on ID
- The add-on version
- The add-on description
- The active status of the study (starts off true)

In order to detect when the user removes an add-on, we need to monitor two things:

- If we detect the add-on in question is uninstalled (using the [AddonManager.addAddonListener](https://developer.mozilla.org/en-US/Add-ons/Add-on_Manager/AddonManager#addAddonListener()) method), we should mark the study as inactive and record the time the study ended. We'll need to register these listeners on startup, similar to what we do for preference experiments.
- If, on startup (before we fetch recipes) we notice that the addon is no longer installed, we should mark the study as inactive and record the time the study ended.

@kmaglione Is IndexedDB, via your promise-based helper, a reasonable place to store this kind of data? Would we be able to ship in 54 (post-release, potentially) if we used IndexedDB.jsm?

If not, then we should consider consolidating the JSON files we're storing in the profile folder (currently two of them) rather than adding yet another one.

andymikulski:
Fixes part of #749 

- Adds `StudyStorage`, a singleton for managing a study data IndexedDB.
- Adds relevant tests

I initially tried to abstract out some of the IDB business (to avoid duplicate code, specifically around objectStores), but ultimately it was simpler to just drop all of that and rely on the [example provided on the firefox-dev mailing list](https://mail.mozilla.org/pipermail/firefox-dev/2017-March/005220.html).


recipe-client-addon/lib/StudyStorage.jsm
```diff
@@ -0,0 +1,83 @@
...
+  namespace: "shield-study-addons",
+  version: 1,
+  storage: "persistent",
+  schema: {
+    name: { autoIncrement: false, unique: false },
+    version: { autoIncrement: false, unique: false },
+    description: { autoIncrement: false, unique: false },
+    installDate: { autoIncrement: false, unique: false },
+    studyEndDate: { autoIncrement: false, unique: false },
+    addonId: { autoIncrement: false, unique: true },
```
Osmose: We might run an experiment with the same add-on multiple times, which would lead to duplicate `addonId`s. I think `name` is a better candidate for the unique key, as we enforce unique experiment names on the server.

recipe-client-addon/lib/StudyStorage.jsm
```diff
@@ -0,0 +1,83 @@
...
+      installDate = Date.now(), // The install date
+      studyEndDate = null, // The study end date (starts off null)
+      addonId = study.id, // The add-on ID
+      active = true, // The active status of the study (starts off true)
+    } = study;
+
+
+    const db = await this.getDatabase();
+
+    return await db.objectStore(this.dbName, "readwrite")
```
Osmose: Async functions can just return a Promise and it will automatically get awaited correctly.

recipe-client-addon/lib/StudyStorage.jsm
```diff
@@ -0,0 +1,83 @@
...
+    studyEndDate: { autoIncrement: false, unique: false },
+    addonId: { autoIncrement: false, unique: true },
+    active: { autoIncrement: false, unique: false },
+  },
+};
+
+this.StudyStorage = {
+  dbName: "shield-studies",
+
+  async reset(clearData) {
```
Osmose: Clearing all the data and closing the database connection should be two separate functions rather than a single one, to avoid mixing them up; closing the connection should be common, while clearing the data should not.

recipe-client-addon/lib/StudyStorage.jsm
```diff
@@ -0,0 +1,83 @@
...
+  async saveStudyData(study) {
+    const {
+      name, // The unique name for the experiment
+      version, // The add-on version
+      description, // The add-on description
+      installDate = Date.now(), // The install date
+      studyEndDate = null, // The study end date (starts off null)
+      addonId = study.id, // The add-on ID
+      active = true, // The active status of the study (starts off true)
+    } = study;
```
Osmose: A schema for the incoming data, [similar to what we did for Preference Experiments](https://github.com/mozilla/normandy/blob/master/recipe-client-addon/lib/PreferenceExperiments.jsm#L21), would be helpful.

recipe-client-addon/lib/StudyStorage.jsm
```diff
@@ -0,0 +1,83 @@
...
+    return this.database;
+  },
+
+  async getStudyData(studyId) {
+    const db = await this.getDatabase();
+
+    return await db.objectStore(this.dbName, "readwrite").get(studyId);
+  },
+
+  async saveStudyData(study) {
```
Osmose: Nit: `Data` is redundant for naming purposes here; we're just saving the study.

Osmose: There's no way to update existing study data without overwriting the rest of the data (particularly the install date). Splitting this into an `update` and `create` method might be a good idea.

recipe-client-addon/lib/StudyStorage.jsm
```diff
@@ -0,0 +1,83 @@
...
+    const {
+      name, // The unique name for the experiment
+      version, // The add-on version
+      description, // The add-on description
+      installDate = Date.now(), // The install date
```
Osmose: Technically this isn't when the add-on was installed. `studyStartDate` is more accurate and matches `studyEndDate`.

recipe-client-addon/lib/StudyStorage.jsm
```diff
@@ -0,0 +1,83 @@
...
+    if (!this.database) {
+      this.database = await IndexedDB.open(this.dbName, DB_OPTIONS, db => {
+        db.createObjectStore(this.dbName, { keyPath: "addonId" });
+      });
+    }
+
+    return this.database;
+  },
+
+  async getStudyData(studyId) {
```
Osmose: I think this throws if the study ID cannot be found. We should add a `has` method to allow users to check first if a study exists or not.

recipe-client-addon/test/browser/browser_StudyStorage.js
```diff
@@ -0,0 +1,92 @@
...
+
+  await StudyStorage.getDatabase();
+  ok(StudyStorage.database, "StudyStorage creates an IDB instance.");
+
+  await StudyStorage.reset();
+  ok(!StudyStorage.database, "StudyStorage correctly removes its IDB instance.");
+});
+
+add_task(async function checkGetDatabase() {
+  await StudyStorage.reset(true);
```
Osmose: Nit: We can't really guarantee test execution order, so to avoid side-effects, we should clean up the study storage after the test is done rather than at the start of the test.

A `withStudyStorage` wrapper that automatically does this might be a good idea.

recipe-client-addon/test/browser/browser_StudyStorage.js
```diff
@@ -0,0 +1,92 @@
...
+  ok(!StudyStorage.database, "StudyStorage starts without an IDB instance.");
+
+  const db1 = await StudyStorage.getDatabase();
+  ok(db1, "StudyStorage creates an IDB instance.");
+
+  const db2 = await StudyStorage.getDatabase();
+  is(db1, db2, "StudyStorage creates only one IDB instance at a time.");
```
Osmose: This is a rad test that I wouldn't have thought of.

recipe-client-addon/test/browser/browser_StudyStorage.js
```diff
@@ -0,0 +1,92 @@
...
+  try {
+    await StudyStorage.saveStudyData({
+      name: "hi",
+      version: 2,
+      description: "fake",
+      id: 12345
+    });
+  } catch (err) {
+    success = false;
+    throw err;
```
Osmose: Since you're throwing here, the assertion below won't run anyway, so catching and saving the success status isn't necessary.

gijsk:
Apologies for the delay on my side here, the email for this got a bit lost, but looks like you already got some feedback. Let me know if you want me to have a look before addressing Osmose's notes.

recipe-client-addon/lib/StudyStorage.jsm
```diff
@@ -0,0 +1,83 @@
...
+      if (clearData) {
+        await this.database.objectStore(this.dbName, "readwrite").clear();
+      }
+
+      await this.database.close();
+      this.database = null;
+    }
+  },
+
+  async getDatabase() {
```
andymikulski: How should we handle testing this function? We have a couple of tests that test `getDatabase` and check against `StudyStorage.database`. Should those be exported as their own object, something like `StudyStorageInternals` so they can be imported into the test suite, but still 'hidden' from the end devs?

Osmose: That's a good point. Could we mock the IndexedDB module and assert that `open` only gets called once instead?

recipe-client-addon/lib/StudyStorage.jsm
```diff
@@ -0,0 +1,108 @@
...
+        studyEndDate,
+        addonId,
+        active,
+      });
+  },
+
+  async updateStudy(study) {
+    const db = await getDatabase();
+
+    return db.objectStore(dbName, "readwrite").put(study);
```
gijsk: What happens if the consumer passes in an incomplete object? In particular, if I just want to save a study end date, can I pass just `{name, studyEndDate}`, or will that wipe other values? It would be nice to have this documented and tested.

recipe-client-addon/lib/StudyStorage.jsm
```diff
@@ -0,0 +1,108 @@
...
+  },
+
+  async saveStudy(study) {
+    const {
+      name, // The unique name for the study
+      version, // The add-on version
+      description, // The add-on description
+      studyStartDate = Date.now(), // The study start date
+      studyEndDate = null, // The study end date (starts off null)
+      addonId = study.id, // The add-on ID
```
gijsk: In the docs at the top, this is just `addonId`, not `id`. This is generally confusing because you're basically doing:

```js
const {
  bar = foo.baz
} = foo;
```

... which I guess makes sense in that now `bar` will be either `foo.bar` or, if that wasn't defined, `foo.baz` - was that intentional? In any case, this isn't documented, and you've omitted this funky thing in the `updateStudy` implementation, but you're passing `id` (and not `addonId`) to `updateStudy` in the tests... This will need to be consistent and documented.

recipe-client-addon/test/browser/browser_StudyStorage.js
```diff
@@ -0,0 +1,70 @@
...
+    version: 1,
+    description: "fake",
+    id: "12345"
+  });
+
+  await storage.updateStudy({
+    name: "hey",
+    version: 123,
+    description: "test",
+    id: "12345"
```
gijsk: As noted above, this won't do anything. Also note that you're sometimes passing a string and sometimes a number for this property.

recipe-client-addon/lib/StudyStorage.jsm
```diff
@@ -0,0 +1,108 @@
...
+    }
+  },
+
+  async getStudy(studyName) {
+    const db = await getDatabase();
+
+    return db.objectStore(dbName, "readonly").get(studyName);
+  },
+
+  async saveStudy(study) {
```
gijsk: Should this and/or `updateStudy` do any schema checks, or are we just letting indexeddb take care of this?

recipe-client-addon/test/browser/browser_StudyStorage.js
```diff
@@ -0,0 +1,70 @@
...
+function withStudyStorage(testFn) {
+  return async () => {
+    await testFn(StudyStorage);
+
+    await StudyStorage.clearAllData();
+    await StudyStorage.closeConnection();
+  };
+}
+
+// -------
```
gijsk: Nit: missing comment

recipe-client-addon/lib/StudyStorage.jsm
```diff
@@ -0,0 +1,108 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * @typedef {Object} Study
+ * @property {string} name
+ *   Unique name of the study
+ * @property {number} version
```
andymikulski: @Osmose Can you confirm what the `version` type will be? I imagined using ints since we don't support major/minor revisions to recipes, just full revisions.

Osmose: This is documented as the add-on version number, which would be a string as @gijsk suggests. I wasn't thinking of storing the revision number of the recipe as well, but that might be useful info to have as well (although how we'd expose it, I don't know). If we did, it'd be a different field. We store revision numbers on the backends as strings too, though.

Let's make this a string, and consider whether we want to send and store the recipe revision number as well in a separate issue.

recipe-client-addon/lib/StudyStorage.jsm
andymikulski: I thought IDB checked the schema, but I was incorrect. I've added a simple schema-checking function and added it to `saveStudy` and `updateStudy`.

recipe-client-addon/lib/StudyStorage.jsm
andymikulski: > was that intentional?

It was! However, I've gone ahead and changed all `study.id` references to use the `addonId` naming instead.

recipe-client-addon/lib/StudyStorage.jsm
andymikulski: In this format, the entire object is overwritten. I've changed `updateStudy` to handle partial updates accordingly.

andymikulski:
I've put in some features such as schema checking, as well as returning a promise for `getDatabase` which calls `IDB.open` only once internally (a race condition @gijsk had pointed out).

This is ready for another review.

Osmose:
@andymikulski and I are swapping some work, I'm going to throw some more commits at this instead of just reviewing it.

recipe-client-addon/lib/StudyStorage.jsm
Osmose: @gijsk Preferably we'd use something standard for schema checks; there's a copy of [ajv](https://github.com/epoberezkin/ajv) in `testing-common`, but I imagine it would be frowned upon to use it in non-testing code?

gijsk: Yes, using it from testing wouldn't really work - we'd have to package it separately.

I'm not really up-to-scratch with schema checks (I mostly meant that we'd need checks *somewhere*, given that previously you could pass in just about anything), but it seems like ajv is 28k of JS once it's minified? Personally, that feels like overkill for what is needed here, but I'm not going to object to either going with ajv (or some other off-the-shelf solution) or hand-rolling some basic checks.

recipe-client-addon/lib/StudyStorage.jsm
```diff
@@ -0,0 +1,210 @@
...
+  },
+
+  /**
+   * If a database exists and is connected, disconnect and remove the reference
+   * to that database.
+   *
+   * @async
+   */
+  async closeConnection() {
+    if (hasDatabaseRequest()) {
```
gijsk: Thinking about it, this races if called multiple times, too. :-)

To prevent this, store `gDatabase` in a local (can just do this directly rather than calling `getDatabase()`, given we're directly nulling it out anyway), then null out the global, then wait for the local `db` ref, then await `db.close()`. This results in the global being nulled out synchronously from the first call, meaning any subsequent calls won't try to close it again.

recipe-client-addon/lib/StudyStorage.jsm
```diff
@@ -0,0 +1,210 @@
...
+   * @async
+   * @param  {string} studyName Unique name of study to retrieve.
+   * @return {Promise<Study>}
+   */
+  async getStudy(studyName) {
+    const db = await getDatabase();
+
+    const foundStudy = await db.objectStore(DB_NAME, "readonly").get(studyName);
+
+    return foundStudy || null;
```
gijsk: What does the promise resolve to if the study is not present? The extra `|| null` feels odd - does it resolve to `false` ? Or `undefined` ? A comment might be helpful if it's really something other than `null` (which would kind of surprise me, but hey).

recipe-client-addon/lib/StudyStorage.jsm
```diff
@@ -0,0 +1,210 @@
...
+   * Given a Study, validates its schema and saves the data into the DB's
+   * objectStore instance.
+   *
+   * @async
+   * @param  {Study}  study [description]
+   * @return {Promise}
+   */
+  async saveStudy(study) {
+    const {
+      addonId, // The add-on ID
```
gijsk: Nit-pickety me is confused about the order being different here compared to the `newStudy` assignment further down. It should probably be consistent. :-)

recipe-client-addon/lib/StudyStorage.jsm
```diff
@@ -0,0 +1,210 @@
...
+      name,
+      version,
+      description,
+      studyStartDate,
+      studyEndDate,
+      addonId,
+      active,
+    };
+
+    validateSchema(newStudy);
```
gijsk: I wonder if the default value assignments could be part of the validation step.

recipe-client-addon/lib/StudyStorage.jsm
```diff
@@ -0,0 +1,210 @@
...
+      addonId, // The add-on ID
+      name, // The unique name for the study
+      version, // The add-on version
+      description, // The add-on description
+      studyStartDate = Date.now(), // The study start date
+      studyEndDate = null, // The study end date (starts off null)
+      active = true, // The active status of the study (starts off true)
+    } = study;
+
+    const db = await getDatabase();
```
gijsk: Would it make sense to wait for the db ref after the validation step, or is there some reason that's not possible? Then the validation results would return sync.

recipe-client-addon/lib/StudyStorage.jsm
Osmose: I remember looking for a schema library in mozilla-central at least one time before this, plus we probably eventually want to shift to a single storage solution across the entire add-on, instead of the split between IndexedDB and JSONFile that we'll have once this lands. I think we'll be doing enough schema stuff that pulling it in is worth it, although ideally mozilla-central would eventually ship it built-in and we wouldn't have to pull it in ourselves.

Osmose:
I'm having trouble getting this PR to show the correct commits, so I'm going to make a new one instead.

Issue #852, opened 2017-06-29
Ability to log out


andymikulski:
Currently, users have to `/control/logout` in order to log out of the site. This brings up a couple questions:

- Are we still going to use this server-provided logout page? Or are we going to do something like create a React page that hits an API to log out?
- ~redux-little-router `Link` components only allow linking _within_ the React app - it doesn't allow external links. This is only a problem because the `NavigationMenu` is built on `Link`s, and we'd need to find a solve if we intend to use the server's logout page.~ This info is outdated and now irrelevant, as we have changed how the `NavigationMenu` is built.

rehandalal:
We decided to keep using the server-provided logout page.

PR #862, opened 2017-07-06
Store service info and show logout button

### LINKED ISSUE #765: Send backend settings to frontend via API instead of data attributes

rehandalal:
Fixes #765 and #852.

r?


recipe-server/client/control_new/components/common/CurrentUser.js
```diff
@@ -0,0 +1,38 @@
...
+
+import { getCurrentUser } from 'control_new/state/serviceInfo/selectors';
+
+
+@connect(
+  state => ({
+    user: getCurrentUser(state),
+  }),
+)
+export default class CurrentUser extends React.Component {
```
andymikulski: This name feels a little ambiguous, but I can't think of anything better. `CurrentUserPanel` or something? I was expecting this to be a query component when I first saw the name.

PR #863, opened 2017-07-06
Document code owners for automatic review requests

mythmon:
This is a new feature that Github just added. It lets us enforce the implicit rules we have about who must review what.

See https://github.com/blog/2392-introducing-code-owners for more details.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,7 +4,7 @@
 # Later rules override earlier rules
 
 # As a fallback, ping the team
-* @mozilla/normandy-comitters
+* @mozilla/normandy-owners
 
 # add-on work needs to be reviewed by both a Firefox Peer and a comitter
-* @mozilla/normandy-comitters @mozilla/normandy-firefox-peers
+recipe-client-addon/* @mozilla/normandy-owners @mozilla/normandy-firefox-peers
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,10 @@
+# This file creates automatic review requests for PRs based on the files touched.
+# See this blog post for more info: https://github.com/blog/2392-introducing-code-owners
+
+# Later rules override earlier rules
+
+# As a fallback, ping the team
+* @mozilla/normandy-comitters
+
+# add-on work needs to be reviewed by both a Firefox Peer and a comitter
+* @mozilla/normandy-comitters @mozilla/normandy-firefox-peers


.github/CODEOWNERS
```diff
@@ -0,0 +1,10 @@
+# This file creates automatic review requests for PRs based on the files touched.
+# See this blog post for more info: https://github.com/blog/2392-introducing-code-owners
+
+# Later rules override earlier rules
+
+# As a fallback, ping the team
+* @mozilla/normandy-comitters
```
Osmose: I'd suggest making a `normandy-owners` team or something with just core committers (our immediate team, probably) rather than using `normandy-committers`, which includes QA and ops and will notify them when normally we don't have them doing reviews.

.github/CODEOWNERS
```diff
@@ -0,0 +1,10 @@
...
+
+# add-on work needs to be reviewed by both a Firefox Peer and a comitter
+* @mozilla/normandy-comitters @mozilla/normandy-firefox-peers
```
Osmose: This is missing the marker limiting it to the add-on directory.

gijsk:
This doesn't seem to be working quite right? I just got pinged on https://github.com/mozilla/normandy/pull/864 which seems to make no changes to the add-on...

mythmon:
That's odd. #864 was filed at 2:46PM, before this PR was merged (4:26pm PST). Also, #864 was on the redesign branch, which this feature shouldn't affect. I also don't see any evidence of an automatic code-owner review request on #864, but maybe Github did something weird.

I think that actually happened was that I initially asked for review from the normandy-comitters team, which you are a part of. @Osmose pointed out that is probably not ideal since it casts too wide a net. I later changed it to request review from the normandy-owners team.

I'd like to keep this feature on for now, but if you get any more spurious notifications, we can roll it back.

gijsk:
> I think that actually happened was that I initially asked for review from the normandy-comitters team, which you are a part of. @Osmose pointed out that is probably not ideal since it casts too wide a net. I later changed it to request review from the normandy-owners team.

This sounds plausible. I'll let you know if I get anything else. I'm continuing to get emails for #864, I guess that's just github deciding "once you're in, you're in"?

mythmon:
> once you're in, you're in

Sort of. Github often subscribes one to things very easily. If you open #864 in a browser, there should be a control on the right side of the page to change your level of notification. It looks like this: 

screenshot-2017-7-7 document code owners for automatic review requests by mythmon pull request 863 mozilla normandy

PR #864, opened 2017-07-06
Use OIDC logout url if it is provided.

mythmon:


recipe-server/client/control_new/components/common/CurrentUserDetails.js
```diff
@@ -3,21 +3,23 @@ import PropTypes from 'prop-types';
...
 )
 export default class CurrentUserDetails extends React.Component {
   static propTypes = {
     user: PropTypes.object.isRequired,
+    logoutUrl: PropTypes.string.isRequired,
   }
 
   render() {
-    const { user } = this.props;
+    const { user, logoutUrl } = this.props;
```
rehandalal: nit: I've been trying to keep these alphabetized

recipe-server/client/control_new/components/common/CurrentUserDetails.js
```diff
@@ -3,21 +3,23 @@ import PropTypes from 'prop-types';
...
   state => ({
     user: getCurrentUser(state),
-  }),
+    logoutUrl: getLogoutUrl(state),
+  })
 )
 export default class CurrentUserDetails extends React.Component {
   static propTypes = {
     user: PropTypes.object.isRequired,
+    logoutUrl: PropTypes.string.isRequired,
```
rehandalal: nit: I've been trying to keep these alphabetized

recipe-server/normandy/base/api/views.py
```diff
@@ -64,6 +64,7 @@ def get(self, request):
         return Response(ServiceInfoSerializer({
             'user': user,
             'peer_approval_enforced': settings.PEER_APPROVAL_ENFORCED,
+            'logout_url': settings.OIDC_LOGOUT_URL or '/control/logout',
```
rehandalal: We should probably use `reverse` to get the fallback URL so we don't have to keep this updated.

recipe-server/client/control_new/components/common/CurrentUserDetails.js
mythmon: This sounds like something that should be a lint so it always gets enforced.

PR #867, opened 2017-07-07
Extensions listing

### LINKED ISSUE #828: Redesigned extensions listing page
We need to redo the extensions listing page using Ant.

rehandalal:
Fixes #828.

r?


Osmose:
Tagging in @andymikulski to review this instead of me.

rehandalal:
@Osmose there were a bunch of python changes which is why I tagged you. Maybe just do a once over on those.

recipe-server/client/control_new/components/extensions/Listing.js
```diff
@@ -0,0 +1,139 @@
...
+
+  getFilters() {
+    const { ordering } = this.props;
+
+    const filters = {
+      ordering,
+    };
+
+    Object.keys(filters).forEach(key => {
+      if ([undefined, null].includes(filters[key])) {
```
andymikulski: This seems strange - would a simple `if (!filters[key]) { ... }` work? Would we ever expect the value of `filters[key]` to be falsy, like`0`?

recipe-server/client/control_new/components/extensions/ListingActionBar.js
```diff
@@ -0,0 +1,65 @@
...
+    push(getCurrentURL({ searchText: value || undefined }));
+  }
+
+  render() {
+    const { columns, saveExtensionListingColumns } = this.props;
+    return (
+      <Row gutter={16} className="list-action-bar">
+        <Col span={16}>
+          <CheckboxMenu
+            checkboxes={columns.toJS()}
```
andymikulski: I wonder if we should always pass Immutable objects to nested components, and then handle the `toJS` within the component where it needs it (e.g. in its `render`).

recipe-server/client/control_new/state/action-types.js
```diff
@@ -4,6 +4,8 @@ export const ACTION_RECEIVE = 'ACTION_RECEIVE';
 export const APPROVAL_REQUEST_DELETE = 'APPROVAL_REQUEST_DELETE';
 export const APPROVAL_REQUEST_RECEIVE = 'APPROVAL_REQUEST_RECEIVE';
 export const EXTENSION_RECEIVE = 'EXTENSION_RECEIVE';
+export const EXTENSIONS_LISTING_COLUMNS_CHANGE = 'EXTENSIONS_LISTING_COLUMNS_CHANGE';
+export const EXTENSIONS_PAGE_RECEIVE = 'EXTENSIONS_PAGE_RECEIVE';
```
andymikulski: This is not alphabetical.

recipe-server/client/control_new/state/extensions/actions.js
```diff
@@ -74,3 +86,15 @@ export function updateExtension(pk, extensionData) {
     });
   };
 }
+
+export function saveExtensionListingColumns(columns) {
+  return async dispatch => {
+    await localForage.setItem('extension_listing_columns', columns);
```
andymikulski: Technically, this `await` could be removed, since the UI doesn't need to be hung up on waiting for localForage.

Also, I noticed there's only a `save` and no `load` function for these columns - is that for another PR?

recipe-server/client/control_new/components/data/QueryExtensions.js
```diff
@@ -3,27 +3,38 @@ import React from 'react';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 
-import { fetchAllExtensions } from 'control_new/state/extensions/actions';
+import * as extensionActions from 'control_new/state/extensions/actions';
```
andymikulski: I notice this method is used in a few spots - maybe it requires a larger team discussion.

recipe-server/client/control_new/components/extensions/Listing.js
rehandalal: yes it could be a `0` or `''`. so a truthy check doesn't work here.

recipe-server/client/control_new/state/action-types.js
rehandalal: isn't it?

recipe-server/client/control_new/state/extensions/actions.js
rehandalal: /me nods. i need to create a query component that pulls from localForage. I'll add that in.

recipe-server/client/control_new/components/data/QueryExtensions.js
rehandalal: yeah it probably requires a larger discussion. and since it doesn't solve anything right now, i think we can defer. but it probably makes sense to file an issue about upgrading webpack and addressing this then.

recipe-server/client/control_new/state/action-types.js
rehandalal: I'll make them all singular to avoid confusion.

rehandalal: Actually i think maybe I should do that in another PR that does the same for recipes once this lands.

andymikulski: Ah, you're right - `EXTENSION_` vs `EXTENSIONS_` tricked me.

rehandalal: Fixed in #874

recipe-server/client/control_new/components/data/QueryExtensionListingColumns.js
```diff
@@ -0,0 +1,27 @@
...
+import React from 'react';
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
+
+import { loadExtensionListingColumns } from 'control_new/state/extensions/actions';
+
+
+@connect(
+  null,
+  dispatch => (bindActionCreators({
```
andymikulski: We sort of do this differently in certain components - in some we just pass an object of functions (`{ loadExtensionListingColumns }`) and other times we use `bindActionCreators`.

@rehandalal Is there any reason we should use one syntax over the other? If they're the same, I'd suggest just going with the object for brevity.

PR #868, opened 2017-07-07
Get CSRF token from cookie

### LINKED ISSUE #865: apiFetch in control_new should pull the CSRF token from the user's cookies
Django stores the CSRF token in the user's cookies; we should be pulling it from there instead of redundantly adding it to the HTML for the page. I'm not entirely sure, but this might also help with stale CSRF tokens, as a failed request to the service can return an updated cookie with a fresh token.

rehandalal:
Fixes #865.

f?

### DIFF
--- a/recipe-server/client/control_new/utils/apiFetch.js
+++ b/recipe-server/client/control_new/utils/apiFetch.js
@@ -18,7 +18,7 @@ export default async function apiFetch(url, options = {}) {
   if (!(options.body && options.body instanceof FormData)) {
     headers.append('Content-Type', 'application/json');
   }
-  headers.append('X-CSRFToken', Cookie.get('csrftoken'));
+  headers.append('X-CSRFToken', Cookie.get('csrftoken-20170707'));
 
   const settings = {
     headers,
--- a/recipe-server/client/control_new/utils/apiFetch.js
+++ b/recipe-server/client/control_new/utils/apiFetch.js
@@ -1,3 +1,5 @@
+import * as Cookie from 'js-cookie';
+
 const API_ROOT = '/api/';
 
 export class ApiError extends Error {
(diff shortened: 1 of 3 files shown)


recipe-server/normandy/settings.py
```diff
@@ -280,7 +280,7 @@ def RAVEN_CONFIG(self):
     SECURE_PROXY_SSL_HEADER = values.TupleValue()
     SECURE_HSTS_SECONDS = values.IntegerValue(3600)
     SECURE_HSTS_INCLUDE_SUBDOMAINS = values.BooleanValue(True)
-    CSRF_COOKIE_HTTPONLY = values.BooleanValue(True)
+    CSRF_COOKIE_HTTPONLY = values.BooleanValue(False)
```
rehandalal: Changing this to `False` since that's the only way to use this in javascript on the front end. 

This should be fine as per the [Django Docs](https://docs.djangoproject.com/en/1.11/ref/settings/#csrf-cookie-httponly):

> Designating the CSRF cookie as HttpOnly doesn’t offer any practical protection because CSRF is only to protect against cross-domain attacks. If an attacker can read the cookie via JavaScript, they’re already on the same domain as far as the browser knows, so they can do anything they like anyway. (XSS is a much bigger hole than CSRF.)

Osmose: Django still warns with a system check if this is false unless you ignore the warning, that's what's breaking the contract tests for this PR.

recipe-server/client/control_new/utils/apiFetch.js
```diff
@@ -16,7 +18,7 @@ export default async function apiFetch(url, options = {}) {
   if (!(options.body && options.body instanceof FormData)) {
     headers.append('Content-Type', 'application/json');
   }
-  headers.append('X-CSRFToken', document.getElementsByTagName('html')[0].dataset.csrf);
+  headers.append('X-CSRFToken', Cookie.get('csrftoken'));
```
Osmose: We should rename this to something like `csrftoken2` or `csrftoken-cachebust-2017-7-7` so that users don't have to clear their cookies manually when we deploy this.

recipe-server/normandy/settings.py
mythmon: To ignore system checks we use `SILENCED_SYSTEM_CHECKS`. You can see an example [here](https://github.com/rehandalal/normandy/blob/42a3404db0fa3e9e635245309b15b246f20a53df/recipe-server/normandy/settings.py#L379).

PR #869, opened 2017-07-07
Fix #861 octothrope comments

### LINKED ISSUE #861: allow comments of some sort in jexl
Per previous discussions, allowing 'comments' would help a lot.

I evaluated several paths:
1 upgrading `jexl` to support comments of this form '//' and '/* */'
2 pre-stripping WHOLE LINE comments ONLY `// this is a comment` or `# comment`

I think the *dirtiest quickest way* is to do 2 above, maybe using `#` as the character.  

Claims and support: 
- Doing whole line strip is pretty safe.  
- Whole line comments are better for readability and style.  

Rejecting the 'fix jexl' option.  In jexl, the lexer / parser are all hand-written and idiosyncratic.  A fully complete and robust 'inline' comments lexer / parser has a lots of edge cases  "This is my /* string */", and such.  It might all just work though :) 

Example Lexing:
```
> L2.tokenize("a == '/* comment \n blah */' /* a comment (b == 2) /* */")
[ { type: 'identifier', value: 'a', raw: 'a ' },
  { type: 'binaryOp', value: '==', raw: '== ' },
  { type: 'literal',
    value: '/* comment \n blah */',
    raw: '\'/* comment \n blah */\' ' },
  { type: 'openComment', value: '/*', raw: '/* ' },
  { type: 'identifier', value: 'a', raw: 'a ' },
  { type: 'identifier', value: 'comment', raw: 'comment ' },
  { type: 'openParen', value: '(', raw: '(' },
  { type: 'identifier', value: 'b', raw: 'b ' },
  { type: 'binaryOp', value: '==', raw: '== ' },
  { type: 'literal', value: 2, raw: '2' },
  { type: 'closeParen', value: ')', raw: ') ' },
  { type: 'openComment', value: '/*', raw: '/* ' },
  { type: 'closeComment', value: '*/', raw: '*/' } ]
```

(This supports 'example filters', and a bunch of other pieces for recipe creation smoothing)

gregglind:
So I did a LOT MORE WORK ON THIS.

What does this do?

1.  In the `Lexer` catches OCTOTHORPE comments using a REGEX, and DROPS those tokens.  Thus the parser never has to see them, and never has to do anything.
2.  If we can write a regex for C-comments, we could do the same thing.  PS, those comments are much harder to regex, and I gave up with trying to get the right level of slash escaping to work :) , but it's easy to extend.
3.  PARSING the C-Comments is a LOT MORE CONFUSING for me.  I tried writing some grammar above, but I got totally lost in `states.js`.
4.  PS, I don't know how you are handling multi-line.  I wrote a test for that in the jexl, and it fell-apart.

### DIFF
--- a/mozjexl/lib/Lexer.js
+++ b/mozjexl/lib/Lexer.js
@@ -14,7 +14,9 @@ var numericRegex = /^-?(?:(?:[0-9]*\.[0-9]+)|[0-9]+)$/,
 		'\\s+',
 		// Booleans
 		'\\btrue\\b',
-		'\\bfalse\\b'
+		'\\bfalse\\b',
+		// Octothorpe comments #861
+		'#.*[\\n\\r]?'
 	],
 	postOpRegexElems = [
 		// Identifiers
@@ -71,6 +73,8 @@ Lexer.prototype.getTokens = function(elements) {
 			if (tokens.length)
 				tokens[tokens.length - 1].raw += elements[i];
 		}
+		else if (this._isComment(elements[i]))
+			continue;
 		else if (elements[i] === '-' && this._isNegative(tokens))
 			negate = true;
 		else {
@@ -217,6 +221,16 @@ Lexer.prototype._isNegative = function(tokens) {
 var _whitespaceRegex = /^\s*$/;
 Lexer.prototype._isWhitespace = function(str) {
 	return _whitespaceRegex.test(str);
+};
+
+/**
+ * A utility function to determine if a string is an octothorpe comment
+ * @param {string} str A string to be tested
+ * @returns {boolean} true if the string is a comment, false otherwise.
+ * @private
+ */
+Lexer.prototype._isComment = function(str) {
+	return str.startsWith('#');
 };
 
 /**
--- a/mozjexl/lib/Lexer.js
+++ b/mozjexl/lib/Lexer.js
@@ -15,8 +15,8 @@ var numericRegex = /^-?(?:(?:[0-9]*\.[0-9]+)|[0-9]+)$/,
 		// Booleans
 		'\\btrue\\b',
 		'\\bfalse\\b',
-		// Octothorpe comments #861
-		'#.*[\\n\\r]?'
+		// comments
+		'//.*[\\n\\r]?'
 	],
 	postOpRegexElems = [
 		// Identifiers
@@ -224,13 +224,13 @@ Lexer.prototype._isWhitespace = function(str) {
 };
 
 /**
- * A utility function to determine if a string is an octothorpe comment
+ * A utility function to determine if a string is a comment
  * @param {string} str A string to be tested
  * @returns {boolean} true if the string is a comment, false otherwise.
  * @private
  */
 Lexer.prototype._isComment = function(str) {
-	return str.startsWith('#');
+	return str.startsWith('//');
 };
 
 /**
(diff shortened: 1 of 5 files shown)


mozjexl/lib/Lexer.js
```diff
@@ -67,10 +69,13 @@ Lexer.prototype.getTokens = function(elements) {
 	var tokens = [],
 		negate = false;
 	for (var i = 0; i < elements.length; i++) {
+
```
mythmon: Super nit: this newline is probably not needed. That being said, all of mozjexl needs a style pass, so I wouldn't stress much about it.

Osmose:
Can we consider using `//` instead of `#`? To match JavaScript, which JEXL is at least trying to do.

mozjexl/lib/Lexer.js
gregglind: That is a foo foo.  Annoying there is no eslint in this package!

gregglind:
1. Maybe y'all need to think on this, but is OCTOTHORPE enough / right?
2. I will augment tests to do some grosser things, including
    - multline
    - comments inside strings (which should not be stupid)

Thanks!

gregglind:
@Osmose, noted, will try `//`.  I think the way you handle whitespace might be causing my multiline foo foo, and I will try the things to show that, or think on that.

gregglind:
@Osmose @mythmon 
1.  `//` is used for floor division, so '#' it is.
2.  Test cases updated to work with multiline, nested and '#' in values.
3.  Same basic design:  lexer just 'throws comments away' so parser never has to deal with them.  It is possible to put these nodes on the AST, but a lot more work.  

So, what do other people do?
- dropping comments at the Lexer is common.
- python ran into issues once they started type-hinting.  http://www.psf.upfronthosting.co.za/issue24119
- esprima drops them at the Lexer:  http://esprima.org/demo/parse.html#

gregglind:
@Osmose if you REALLY want jshint / highlight, then we could DROP floorDiv and '|' from our spec.  Pipe forward seems very un-natural anyway, fwiw.

mozjexl/test/Lexer.js
```diff
@@ -125,6 +125,42 @@ describe('Lexer', function() {
 			fn.should.throw();
 		});
 	});
+	describe('Comments', function() {
+		// per 861, currently support BOTH # and // comments
```
mythmon: I don't think this comment matches the code.

gregglind: Fixed by removal, pushed.

mythmon:
If we want to switch to `//` comments, instead of `#`, lets do that now. Having multiple historical syntaxes can't bring any good.

gregglind:
So, floor division seems useful in general.  I am in favor of a "fork the
syntax" bug :). It should be much more js like.  To me the warts are floor
division, forward pipe, and comments.  I can see why they made those
choices though.

I am fine with code base listing first.

mythmon:
@gregglind I pushed to your branch to change to `//` style comments. If you're happy with that, I say we merge this.

gregglind:
Claim: if we want floor div back, we do it as a pipe function?  Is that the
solve?

mythmon:
If we switch from pipe syntax to JS syntax for function calls, then I think we should make floordiv a function that can be called, instead of an operator. That way, we stick close to JS syntax.

gregglind:
+1.  Approved.

PR #871, opened 2017-07-08
Support loading arbitrary npm dependencies in the add-on via webpack.

Osmose:
In working on #749, I needed a way to load up [ajv](https://github.com/epoberezkin/ajv), but didn't want to manually import it like we did with mozjexl. Webpack to the rescue!

What this does is add a webpack config to the add-on that includes one configuration per-library that we want to include. Each config builds the library into a single file and outputs it to a `vendor` subdirectory. It also includes a plugin that appends `this.EXPORTED_SYMBOLS = ["libname"];` to the end of the output file so that it can be imported via `Cu.import`.

The rest is plumbing for automatically including these files within the add-on.

I think this is more preferable to running webpack on the entire add-on for a few reasons:
- We can continue to use imports that look familiar to Firefox devs.
- The code that we land in mozilla-central is still amenable to large refactors that happen from time-to-time.
- No need to mess around with source maps for the entire add-on.

This _should_ work with most browser-compatible libraries that we care about, although some browser globals might cause issues, like `URL` or `fetch`. If we run into that, I'd consider just adding an `importGlobalProperties` call to the top of the output file as necessary. Could be an optional argument to `libraryConfig`.

This is a WIP because I still need to write docs and also have a question:

How should the workflow for triggering builds go? Should we keep using the scripts in `bin` and modify them to run webpack every time? Should we switch to `npm` script commands and call the scripts within those instead? Should we just expect devs to run webpack when necessary (since you only need a rebuild when dependencies change)?

I'd appreciate feedback from anyone, but want it particularly from @mythmon. :D

### DIFF
--- a/recipe-client-addon/package.json
+++ b/recipe-client-addon/package.json
@@ -7,9 +7,8 @@
   "repository": "https://github.com/mozilla/normandy-addon",
   "license": "MPL-2.0",
   "scripts": {
-    "watch-vendor": "webpack --config ./webpack.config.js --watch",
-    "build-vendor": "webpack --config ./webpack.config.js",
-    "build": "webpack --config ./webpack.config.js && ./bin/make-xpi.sh"
+    "watch": "webpack --config ./webpack.config.js --watch",
+    "build": "webpack --config ./webpack.config.js"
   },
   "devDependencies": {
     "babel-eslint": "7.2.3",
--- a/recipe-client-addon/package.json
+++ b/recipe-client-addon/package.json
@@ -5,11 +5,10 @@
   "main": "index.js",
   "author": "Mike Cooper <mcooper@mozilla.com>",
   "repository": "https://github.com/mozilla/normandy-addon",
-  "scripts": {},
   "license": "MPL-2.0",
   "scripts": {
     "watch-vendor": "webpack --config ./webpack.config.js --watch",
-    "build-bendor": "webpack --config ./webpack.config.js",
+    "build-vendor": "webpack --config ./webpack.config.js",
     "build": "webpack --config ./webpack.config.js && ./bin/make-xpi.sh"
   },
   "devDependencies": {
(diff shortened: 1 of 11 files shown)


recipe-client-addon/package.json
```diff
@@ -5,15 +5,20 @@
   "main": "index.js",
   "author": "Mike Cooper <mcooper@mozilla.com>",
   "repository": "https://github.com/mozilla/normandy-addon",
+  "scripts": {},
+  "license": "MPL-2.0",
   "scripts": {
+    "watch-vendor": "webpack --config ./webpack.config.js --watch",
+    "build-bendor": "webpack --config ./webpack.config.js",
```
mythmon: typo: bendor

recipe-client-addon/package.json
```diff
@@ -5,15 +5,20 @@
   "main": "index.js",
   "author": "Mike Cooper <mcooper@mozilla.com>",
   "repository": "https://github.com/mozilla/normandy-addon",
+  "scripts": {},
```
mythmon: Duplicate key?

mythmon:
I think switching over to using npm scripts instead of running things in `bin/` directly makes sense. That way we can make sure that requirements are always satisfied. As long as we update the docs and CI, I have no problems moving away from running scripts directly.

Osmose:
> This looks like a good idea. I'm a little unsure about one webpack config per vendor library (even if they are autogenerated). But I also couldn't figure out a way to make ExportedSymbols work another way.

The hard part is the `output.library` option. If that supported multiple entry points better, we could get the rest of the info from the compilation object and loop through each asset.

mythmon:
> The hard part is the output.library option. If that supported multiple entry points better, we could get the rest of the info from the compilation object and loop through each asset.

Oh, I know how to deal with that. If you set it to `"[name]"`, the same variable expansion that you use for `filename` will expand to the right thing. So if you  know how to make the plugin work with multiple entrypoints, I think we're all set.

Osmose:
Circle and TaskCluster seem to have stopped making new builds for this PR. I'm going to make a new one to see if that fixes anything.

Issue #707, opened 2017-04-24
Refactor the recipe form


rehandalal:
We need to update the recipe form component to work with the new state tree.

Osmose:
Andy has graciously agreed to handle the remaining bits of this that aren't in #859, namely the create new recipe page.

rehandalal:
oops! may have jumped the gun on closing this.

PR #872, opened 2017-07-10
New Create Recipe Form

andymikulski:
Completes issue #707 
Depends on PR ~#859~ #875 

- Adds `CreateRecipePage` component, route
- Adds `disabled` props to form fields/components
- Adds `isLoading` prop to `FormActions`, which displays a spinner when the form is submitting/working.
- Adds `isProcessing` states to `Create` and `Update` recipe components

Commit 6da8f069597bab0b1fc8dfcdbd5a8de325c704ba is ready for review, but this whole PR depends on @Osmose 's PR ( ~#859~ #875 )

### DIFF
--- a/recipe-server/client/control_new/components/common/LoadingOverlay.js
+++ b/recipe-server/client/control_new/components/common/LoadingOverlay.js
@@ -14,16 +14,30 @@ import { areAnyRequestsInProgress } from 'control_new/state/requests/selectors';
 export default class LoadingOverlay extends React.Component {
   static propTypes = {
     children: PropTypes.any,
+    condition: PropTypes.bool,
     loading: PropTypes.bool.isRequired,
+    useCondition: PropTypes.bool,
   };
 
   static defaultProps = {
     children: null,
+    condition: false,
+    useCondition: false,
   };
 
   render() {
-    const { children, loading } = this.props;
-    const Wrapper = loading ? Spin : 'div';
+    const {
+      children,
+      condition,
+      loading,
+      useCondition,
+    } = this.props;
+
+    // We can use a conditional expression instead of the global request state.
+    const isCurrentlyLoading = useCondition ? condition : loading;
+
+    const Wrapper = isCurrentlyLoading ? Spin : 'div';
+
     return (
       <Wrapper>
         {children}
--- a/recipe-server/client/control_new/components/common/LoadingOverlay.js
+++ b/recipe-server/client/control_new/components/common/LoadingOverlay.js
@@ -6,6 +6,34 @@ import { connect } from 'react-redux';
 import { areAnyRequestsInProgress } from 'control_new/state/requests/selectors';
 
 
+export class SimpleLoadingOverlay extends React.Component {
+  static propTypes = {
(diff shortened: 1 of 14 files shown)


rehandalal:
this could probably use some rebase love.

andymikulski:
I've got this rebased, however I want to change how the form handles being disabled while submitting forms etc, so I need to add another commit before review.

andymikulski:
This is ready for review. I've changed how I initially approached disabling the form fields during submission, so now:

- `RecipeForm` will automatically disable form fields via `areAnyRequestsInProgress`
- `LoadingOverlay` now has an optional `condition` prop
  - I ran into an issue where the `LoadingOverlay` would pop up when saving an existing recipe, thereby removing all the field changes (due to elements being swapped in/out of the dom and the fact we don't track form changes anywhere).
  - The `condition` prop gives a little finer control over when to display the `LoadingOverlay`. With this approach, we can use `LoadingOverlay` for global requests, or use it as a 'dumb' component, as I have in this PR.
- `disable` props have been added to form controls. I was trying to find a way around adding `disabled={disabled}` all over the place, but ultimately couldn't find a better approach.
- With the `LoadingOverlay`, `RecipeForm`, and `disabled` changes, we get the desired result of a disabled form while submitting info to the server, _without_ losing changes due to dom flips.
   - However, we lose the actual spinner + blur, so it doesn't feel as cohesive.

Ultimately, we _should_ do some sort of form state tracking, however I think to get this page landed this approach should suffice.

@rehandalal Ready for review!

recipe-server/client/control_new/components/common/LoadingOverlay.js
```diff
@@ -14,16 +14,30 @@ import { areAnyRequestsInProgress } from 'control_new/state/requests/selectors';
...
   static propTypes = {
     children: PropTypes.any,
+    condition: PropTypes.bool,
     loading: PropTypes.bool.isRequired,
+    useCondition: PropTypes.bool,
   };
 
   static defaultProps = {
     children: null,
+    condition: false,
```
rehandalal: can't we change this to `null` and then we don't need `useCondition`.

we would just do this:
```javascript
const isCurrentlyLoading = condition === null ? loading : condition;
```

rehandalal: oops, meant to start a review but clicked the wrong button.

andymikulski: I imagined a scenario where `condition` could be `undefined`, `null`, or an actual value. I realize now, though, that I propTyped it as `bool` so it shouldn't have to worry about nulls/etc anyway. Good catch.

andymikulski: Fixed this in 8cc8d3aa75eacb192cf2b26a0ed5d71d19d5321c

rehandalal:
I think your last commit added all the clone page stuff in. This is fine, but I just want to make sure you got all of it. If not either add all of it in or fix the commit.

PR #875, opened 2017-07-11
New control interface: Editing recipes (pt 2)

andymikulski:
This is a continuation of PR #859 - I've taken over for @Osmose to finish landing code review/feedback items.

This should only require a quick pass - other feedback has been addressed in #859 already, relevant commits should be the last two (5e4312fc72a75ecba06170b9578ca3f0e3a1978b and c0e4743ae76d8d440adfd3cc5ed41925f51384b6)


rehandalal:
Does this also invalidate #872 ?

andymikulski:
@rehandalal #872 just needs a rebase onto this branch when ready. The reason for opening the new PR was because I don't have write access to Osmose's fork, so I just pushed up my own branch with the feedbacked code for review instead.

rehandalal:
oh okay. it seemed like you had all the same commits here as in #872? is the "Add CreateRecipePage" commit different in the two PRs?

andymikulski:
@rehandalal You're correct, the `Add CreateRecipePage` commit should _not_ have been in this PR. Not sure how I pulled that off, but this PR should have all the proper commits now.

I've also rebased #872 onto this branch to keep that one up to speed.

andymikulski:
TC failure appears unrelated.

recipe-server/client/control_new/components/recipes/RecipeForm.js
```diff
@@ -0,0 +1,128 @@
...
+          name="extra_filter_expression"
+          label="Filter Expression"
+          initialValue={recipe.get('extra_filter_expression')}
+        >
+          <Input type="textarea" rows="4" />
+        </FormItem>
+        <FormItem
+          name="action_id"
+          label="Action"
+          initialValue={recipe.getIn(['action', 'id']).toString(10)}
```
rehandalal: do we have to do this `toString`?

recipe-server/client/control_new/state/actions/selectors.js
```diff
@@ -1,5 +1,13 @@
-/* eslint import/prefer-default-export: "off" */
-
 export function getAction(state, id, defaultsTo = null) {
-  return state.app.actions.items.get(id, defaultsTo);
+  // Convert string IDs to integers, since the state is keyed by ints.
```
rehandalal: Is there ever a time we are not passing an integer in here?

recipe-server/client/control_new/components/common/forms.js
```diff
@@ -0,0 +1,242 @@
+import { Form, message } from 'antd';
+import autobind from 'autobind-decorator';
+import get from 'lodash.get';
+import pt from 'prop-types';
```
rehandalal: nit: we've been using the full form `PropTypes` everywhere else. we should stay consistent.

recipe-server/client/control_new/components/data/QueryRecipe.js
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  componentWillMount() {
+    const { pk } = this.props;
+    this.props.fetchRecipe(pk);
+  }
+
+  componentWillReceiveProps(nextProps) {
+    const { pk } = this.props;
+    if (nextProps.pk && pk !== nextProps.pk) {
```
rehandalal: i don't think we need the first `nextProps.pk` check.

recipe-server/client/control_new/components/recipes/EditRecipePage.js
```diff
@@ -0,0 +1,81 @@
...
+
+  render() {
+    const { recipe, recipeId } = this.props;
+    const Wrapper = recipe ? 'div' : Spin;
+
+    return (
+      <Wrapper>
+        <h2>Edit Recipe</h2>
+        <QueryRecipe pk={recipeId} />
+        {recipe &&
```
rehandalal: you don't need to do this check since `getRecipe` above has a default value.

recipe-server/client/control_new/components/recipes/EditRecipePage.js
```diff
@@ -0,0 +1,81 @@
...
+      );
+      if (error.data) {
+        this.setState({ formErrors: error.data });
+      }
+    }
+  }
+
+  render() {
+    const { recipe, recipeId } = this.props;
+    const Wrapper = recipe ? 'div' : Spin;
```
rehandalal: We have a `LoadingOverlay` common component that you can use instead.

recipe-server/client/control_new/components/recipes/EditRecipePage.js
```diff
@@ -0,0 +1,81 @@
...
+
+    return (
+      <Wrapper>
```
rehandalal: I think this should just be a `<div>` and we should wrap the form in `LoadingOverlay`. That's what I did on the extensions form.

recipe-server/client/control_new/state/actions/selectors.js
andymikulski: It's called in `RecipeForm`'s connect, which passes `props.form.getFieldValue('action_id')` as the `id` param. I suspect that since we're pulling from Ant's `form` for the value, we could _possibly_ get a numeric value.

I think it makes sense to leave the type parsing in here, but I'm open to counterpoints.

recipe-server/client/control_new/components/recipes/RecipeForm.js
andymikulski: Nope! Looking into it, the `ActionSelect` component already `toString`s the value, so we don't need to do so here. I've updated the `ActionSelect`'s propTypes to resolve the console warning.

recipe-server/client/control_new/components/extensions/ExtensionForm.js
```diff
@@ -4,7 +4,7 @@ import { is, Map } from 'immutable';
 import pt from 'prop-types';
```
rehandalal: nit: this guy still needs changing

recipe-server/client/control_new/components/recipes/ConsoleLogFields.js
```diff
@@ -0,0 +1,30 @@
+import { Input } from 'antd';
+import { Map } from 'immutable';
+import pt from 'prop-types';
```
rehandalal: same here

recipe-server/client/control_new/components/recipes/EditRecipePage.js
```diff
@@ -0,0 +1,82 @@
+import { message } from 'antd';
+import autobind from 'autobind-decorator';
+import { Map } from 'immutable';
+import pt from 'prop-types';
```
rehandalal: and here.

recipe-server/client/control_new/components/recipes/PreferenceExperimentFields.js
```diff
@@ -0,0 +1,334 @@
+/* eslint-disable react/jsx-boolean-value */
+import { Alert, Button, Icon, Input, InputNumber, Radio, Select } from 'antd';
+import autobind from 'autobind-decorator';
+import { List, Map } from 'immutable';
+import pt from 'prop-types';
```
rehandalal: and here.

recipe-server/client/control_new/components/recipes/ShowHeartbeatFields.js
```diff
@@ -0,0 +1,112 @@
+import { Checkbox, Input, InputNumber, Select } from 'antd';
+import { Map } from 'immutable';
+import pt from 'prop-types';
```
rehandalal: same deal.

andymikulski:
@rehandalal Sorry about that, didn't realize there were some `pt`s left. I've gone through and converted the rest.

rehandalal:
nice job! thanks for picking this up and sticking with it!!

PR #877, opened 2017-07-11
Support loading arbitrary npm dependencies in the add-on via webpack.

Osmose:
See #871 for previous discussion.


Osmose:
TC had a lint error but otherwise everything passed (I just pushed a fix for the lint issue). This should be ready for review!

Osmose:
Github hates freedom and refuses to trigger our builds. We've already created at least one new PR in an attempt to get the builds triggering again, and the only thing that was broken in the last successful build was linting (which I just added an ignore comment for since the vendor built files aren't meant to be handled by humans). So I'm going to just merge this.

PR #878, opened 2017-07-11
#749: Add StudyStorage.jsm for storing data about studies

Osmose:
This is a continuation of #860. It also relies on #877; commits from 03dc6f9 onward are the interesting ones.

The biggest change between the last PR and this is the use of ajv for schema validation of data. I chose it over other schema validation or rolling our own because we already use JSONSchema elsewhere within Normandy, and I'd rather be consistent in our schemas. One mild downside of this is that types that we could actually store natively in IndexedDB, like dates, are instead stored as strings.

### DIFF
--- a/recipe-client-addon/lib/StudyStorage.jsm
+++ b/recipe-client-addon/lib/StudyStorage.jsm
@@ -1,0 +1,83 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+"use strict";
+
+const {utils: Cu} = Components;
+Cu.import("resource://gre/modules/XPCOMUtils.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "IndexedDB", "resource://gre/modules/IndexedDB.jsm");
+
+this.EXPORTED_SYMBOLS = ["StudyStorage"];
+
+const DB_OPTIONS = {
+  namespace: "shield-study-addons",
+  version: 1,
+  storage: "persistent",
+  schema: {
+    name: { autoIncrement: false, unique: false },
+    version: { autoIncrement: false, unique: false },
+    description: { autoIncrement: false, unique: false },
+    installDate: { autoIncrement: false, unique: false },
+    studyEndDate: { autoIncrement: false, unique: false },
+    addonId: { autoIncrement: false, unique: true },
+    active: { autoIncrement: false, unique: false },
+  },
+};
+
+this.StudyStorage = {
+  dbName: "shield-studies",
+
+  async reset(clearData) {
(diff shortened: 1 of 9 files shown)


Osmose:
Rebased to only include the commits that matter, now.

recipe-client-addon/lib/StudyStorage.jsm
```diff
@@ -0,0 +1,170 @@
...
+    const db = await getDatabase();
+    return getStore(db).add(study);
+  },
+
+  async update(studyName, data) {
+    const db = await getDatabase();
+    const savedStudy = await getStore(db).get(studyName);
+    if (!savedStudy) {
+      throw new Error(`Cannot update study ${studyName}: could not find study.`);
+    } else if (!validateUpdate(data)) {
```
gijsk: Nit: no else after return/throw.

recipe-client-addon/lib/StudyStorage.jsm
```diff
@@ -0,0 +1,170 @@
...
+const DB_OPTIONS = {
+  version: 1,
+  storage: "persistent",
+};
+const DB_SCHEMA = {
+  type: "object",
+  additionalProperties: false,
+  properties: {
+    name: {
+      type: "string",
```
gijsk: Based on a *very* quick read of docs / the interwebz (I tried to find a live demo site but couldn't), I *think* this doesn't require `name` or any of the other string props to be non-empty, so passing `{name: "", addonId: "", addonVersion: "", description: ""}` would be valid. Should we fix that (seems like `minLength: 1` on those props might be enough, though we could try to do regex validation of the versions) ?

Osmose:
Feedback'd.

mythmon:
Can we leave this PR open and unmerged for a bit? I've got a reply from GitHub support about our CI issue, and having an open PR that shows the issue should help.

mythmon:
Closing and re-opening to help debug CI not running on PRs.

recipe-client-addon/lib/StudyStorage.jsm
```diff
@@ -0,0 +1,184 @@
...
+  async clear() {
+    const db = await getDatabase();
+    await getStore(db).clear();
+  },
+
+  async close() {
+    if (databasePromise) {
+      const db = await databasePromise;
+      await db.close();
+      databasePromise = null;
```
gijsk: Seconded! I noted this in https://github.com/mozilla/normandy/pull/860#discussion_r126279145 and then forgot about it when reviewing this version of the PR. Oops...

Osmose:
Feedback'd. Any chance we can create a new PR to demonstrate this issue instead of blocking this? It's starting to block other work.

mythmon:
I think that we've got everything we can out of this for now. Lets go ahead and merge it.

mythmon:
Using admin powers to merge this PR since CI isn't running.

PR #884, opened 2017-07-12
Add eslint-config-airbnb-base to lints project for mozjexl.

mythmon:

PR #879, opened 2017-07-11
Lint mozjexl

mythmon:
Re-creating #870, since it seems to have gotten stuck.

PR #885, opened 2017-07-12
Revert "Lint mozjexl"

mythmon:
Reverts mozilla/normandy#879

These tests aren't working with the current lint setup (see #884), and are making my life harder. Reverting them until next week when we have the better lint setup from the redesign branch.


mythmon:
Taskcluster failure is expected: mozilla-central has a known bug that is causing our tasks to fail. If you look at the logs you'll see that all of our tests passed, and the failure comes from the teardown.

PR #887, opened 2017-07-14
#873: New Clone Recipe page

### LINKED ISSUE #873: New Clone Recipe UI
We need to recreate the `clone` page for recipes and extensions using Ant components and the new state tree.

Routes should be:
- `/recipes/:recipeId/clone` - clone the latest revision
- `/recipes/:recipeId/rev/:revisionId/clone` - clone a specific revision

andymikulski:
Fixes #873 
~Depends on PR #872~

- Adds `/clone` route to recipes
- Adds `CloneRecipePage` component
- Adds gear dropdown menu on `EditRecipePage` with link to clone page

This is probably going to require a rebase and some work from #858 , since the ticket requires the `revision` routes.

### DIFF
--- a/recipe-server/client/control_new/components/common/FormActions.js
+++ b/recipe-server/client/control_new/components/common/FormActions.js
@@ -1,3 +1,4 @@
+import { Spin } from 'antd';
 import PropTypes from 'prop-types';
 import React from 'react';
 
@@ -16,18 +17,23 @@ import React from 'react';
  *   </FormActions.Secondary>
  * </FormActions>
  */
-export default function FormActions({ children }) {
+export default function FormActions({ children, isLoading }) {
+  const Wrapper = isLoading ? Spin : 'span';
   return (
     <div className="form-actions">
-      {children}
+      <Wrapper>
+        {children}
+      </Wrapper>
     </div>
   );
 }
 FormActions.propTypes = {
   children: PropTypes.node,
+  isLoading: PropTypes.bool,
 };
 FormActions.defaultProps = {
   children: null,
+  isLoading: false,
 };
 
 /**
--- a/recipe-server/client/control_new/components/recipes/CloneRecipePage.js
+++ b/recipe-server/client/control_new/components/recipes/CloneRecipePage.js
@@ -6,7 +6,7 @@ import React from 'react';
 import { connect } from 'react-redux';
 import { Link, push as pushAction } from 'redux-little-router';
 
-import LoadingOverlay from 'control_new/components/common/LoadingOverlay';
+import { SimpleLoadingOverlay } from 'control_new/components/common/LoadingOverlay';
 import RecipeForm from 'control_new/components/recipes/RecipeForm';
 import QueryRecipe from 'control_new/components/data/QueryRecipe';
 
@@ -64,11 +64,9 @@ export default class CloneRecipePage extends React.Component {
         'Recipe cannot be saved. Please correct any errors listed in the form below.',
       );
 
-      if (error) {
-        this.setState({
-          formErrors: error.data || error,
-        });
-      }
+      this.setState({
+        formErrors: error.data || error,
+      });
     }
   }
 
@@ -83,7 +81,7 @@ export default class CloneRecipePage extends React.Component {
     return (
       <div className="clone-page">
         <QueryRecipe pk={recipeId} />
-        <LoadingOverlay useCondition condition={!recipeName}>
+        <SimpleLoadingOverlay isVisible={!recipeName}>
           <h2>New Recipe</h2>
           { recipeName &&
             <h3>
@@ -96,7 +94,7 @@ export default class CloneRecipePage extends React.Component {
             onSubmit={this.handleSubmit}
             errors={this.state.formErrors}
           />
-        </LoadingOverlay>
+        </SimpleLoadingOverlay>
       </div>
     );
   }
--- a/recipe-server/client/control_new/components/recipes/CloneRecipePage.js
+++ b/recipe-server/client/control_new/components/recipes/CloneRecipePage.js
@@ -1,18 +1,18 @@
-import { message, Button, Icon, Dropdown, Menu } from 'antd';
+import { message } from 'antd';
 import autobind from 'autobind-decorator';
 import { Map } from 'immutable';
 import PropTypes from 'prop-types';
 import React from 'react';
 import { connect } from 'react-redux';
+import { Link, push as pushAction } from 'redux-little-router';
 
-import { SimpleLoadingOverlay } from 'control_new/components/common/LoadingOverlay';
+import LoadingOverlay from 'control_new/components/common/LoadingOverlay';
 import RecipeForm from 'control_new/components/recipes/RecipeForm';
 import QueryRecipe from 'control_new/components/data/QueryRecipe';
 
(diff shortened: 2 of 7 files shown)


rehandalal:
You can probably go ahead and rebase this.

andymikulski:
This is ready. It's missing the routes to clone specific revisions, but revision stuff hasn't landed in `redesign` yet anyway. I think that'll end up being another PR.

Note @rehandalal I added a commit to this changing `SimpleLoadingOverlay`'s prop from `condition` to `visible`. To the world outside of the component, `condition` is pretty ambiguous, so I thought it made sense to switch it. BUT, this brings up the question that I don't think has been answered yet:

For React prop names, do we use `is`/`has` prefixes? Should we use
`<Whatever visible={false} />` or `<Whatever isVisible={false} />`?

recipe-server/client/control_new/components/recipes/CloneRecipePage.js
```diff
@@ -0,0 +1,103 @@
...
+    recipeId: PropTypes.number.isRequired,
+    recipe: PropTypes.instanceOf(Map),
+  };
+
+  static defaultProps = {
+    recipe: null,
+  };
+
+  state = {
+    formErrors: undefined,
```
mythmon: I think this needs to be inside a constructor, like below, but I'm not actually sure. Is this ok?

```js
constructor(props) {
  super(props);
  this.state = { formErrors: undefined };
}
```

recipe-server/client/control_new/components/recipes/CloneRecipePage.js
```diff
@@ -0,0 +1,103 @@
...
+        formErrors: undefined,
+      });
+
+      push(`/recipe/${newId}`);
+    } catch (error) {
+      message.error(
+        'Recipe cannot be saved. Please correct any errors listed in the form below.',
+      );
+
+      if (error) {
```
mythmon: When would `error` ever be falsey?

recipe-server/client/control_new/components/recipes/CloneRecipePage.js
```diff
@@ -0,0 +1,103 @@
...
+
+    const recipeName = recipe.get('name');
+
+    // Remove the 'name' field value.
+    const displayedRecipe = recipe.set('name');
+
+    return (
+      <div className="clone-page">
+        <QueryRecipe pk={recipeId} />
+        <LoadingOverlay useCondition condition={!recipeName}>
```
mythmon: Should this use `loading` instead of `useCondition` and `condition`?

recipe-server/client/control_new/components/recipes/CloneRecipePage.js
```diff
@@ -0,0 +1,103 @@
...
+    const recipeId = getRouterParamAsInt(state, 'recipeId');
+
+    return {
+      recipeId,
+      recipe: getRecipe(state, recipeId, new Map()),
+    };
+  },
+  {
+    push: pushAction,
+    createRecipe: createAction,
```
andymikulski: Yep! If `connect` is given an object of actions, it loops through and `bindActionCreators`s 'em.

recipe-server/client/control_new/components/recipes/CloneRecipePage.js
andymikulski: Nope, we can actually get away with just setting it like this. It's equivalent to `this.state = {...};`

mythmon:
TC failure is lint errors from #890. This will keep failing until master and the redesign branch sync up.

recipe-server/client/control_new/components/recipes/CloneRecipePage.js
```diff
@@ -0,0 +1,101 @@
...
+import React from 'react';
+import { connect } from 'react-redux';
+import { Link, push as pushAction } from 'redux-little-router';
+
+import { SimpleLoadingOverlay } from 'control_new/components/common/LoadingOverlay';
+import RecipeForm from 'control_new/components/recipes/RecipeForm';
+import QueryRecipe from 'control_new/components/data/QueryRecipe';
+
+import { createRecipe as createAction } from 'control_new/state/recipes/actions';
+import { getRecipe } from 'control_new/state/recipes/selectors';
```
rehandalal: You'll need to use the new `getRecipeForRevision` selector similar to:
https://github.com/mozilla/normandy/pull/858/files#diff-816a70a0a0eec30c1392dd3fabceb9eeR14

recipe-server/client/control_new/components/common/FormActions.js
```diff
@@ -1,3 +1,4 @@
+import { Spin } from 'antd';
```
rehandalal: is there a reason we're not using `SimpleLoadingOverlay` here?

recipe-server/client/control_new/components/recipes/CloneRecipePage.js
```diff
@@ -0,0 +1,101 @@
...
+      this.setState({
+        formErrors: error.data || error,
+      });
+    }
+  }
+
+  render() {
+    const { recipe, recipeId } = this.props;
+
+    const recipeName = recipe.get('name');
```
rehandalal: you should have this fallback to `''` like:
```javascript
const recipeName = recipe.get('name', '');
```

recipe-server/client/control_new/components/recipes/CloneRecipePage.js
```diff
@@ -0,0 +1,101 @@
...
+
+    // Remove the 'name' field value.
+    const displayedRecipe = recipe.set('name');
+
+    return (
+      <div className="clone-page">
+        <QueryRecipe pk={recipeId} />
+        <SimpleLoadingOverlay isVisible={!recipeName}>
+          <h2>New Recipe</h2>
+          { recipeName &&
```
rehandalal: You don't need this check with the fallback I described above.

recipe-server/client/control_new/components/recipes/EditRecipePage.js
```diff
@@ -39,6 +39,14 @@ export default class EditRecipePage extends React.Component {
     recipe: null,
   };
 
+  static ActionMenu = (
```
rehandalal: I don't think the form should include any controls that are non-form related. That's the whole point of the details page. I think we should ditch this.

recipe-server/client/control_new/components/recipes/CloneRecipePage.js
andymikulski: IMO I'd prefer to keep it without the fallback. If we default to an empty string, that means that there will be some time where `Cloning recipe values from ""` is on the page - even if it's under a `Spin`. With this approach, we don't display the string until we have the complete thing.

recipe-server/client/control_new/components/common/FormActions.js
andymikulski: Nope, good catch!

recipe-server/client/control_new/components/recipes/CloneRecipePage.js
rehandalal: yeah, that's valid.

andymikulski:
This is going to take a significant rewrite to accommodate the rebase + revisions. I will open another PR.

PR #888, opened 2017-07-14
Fix Less linting

andymikulski:
- Fixes npm script that was still searching sass files.
- Fixes handful of lint errors


mythmon:
TC failure is #890 and #757.

PR #890, opened 2017-07-14
Bug 1380009 - Enable eslint no-return-await.

mythmon:
This is a backport from mozilla-central [bug 1380009](https://bugzilla.mozilla.org/show_bug.cgi?id=1380009).


andymikulski:
IMO, I prefer to use `return await`, but if it has been bugzilla'd...

mythmon:
> IMO, I prefer to use return await, but if it has been bugzilla'd...

Yeah, I agree that `return await` seems clearer. We could override the general mozilla-central eslint rules about this. In this case I don't think it is a good idea though, since this is sort of related to actual execution and not just style.

PR #891, opened 2017-07-14
Fixes #743: Add add-on information to filter expression context.

### LINKED ISSUE #743: Filtering by add-ons
We want to be able to target recipes based on the add-ons that a user has installed. Telemetry may contain this information, but to make filers more readable, and to allow for future expansion, we want to add another API to query the AddonManager directly. Here's my strawman:

```js
// normandy.addons is a map; keys are add-on IDs, and the values are objects
normandy.addons['shield-recipe-client@mozilla.org'] // Truthyness means the add-on is installed

// The value contains version information
normandy.addons['shield-recipe-client@mozilla.org'].version == '1.0.0'
```

One thing that is important to test here is if `[]` syntax for indexing into an object gracefully handles missing keys. I _think_ JEXL actually throws if a key doesn't exist (whereas `.` syntax does not, it just returns `undefined` in all cases). It may be time for us to fork JEXL to change this behavior.

@rayborn @gregglind @MattGrimes Thoughts on the syntax above?

mythmon:


recipe-client-addon/lib/Addons.jsm
```diff
@@ -0,0 +1,108 @@
...
+ */
+
+this.Addons = {
+  /**
+   * Get information about an installed add-on by ID.
+   *
+   * @param {string} addonId
+   * @returns {SafeAddon?} Add-on with given ID, or null if not found.
+   */
+  get: async function get(addonId) {
```
gijsk: I think in es6 you can just:

```js
async get(addonId) {
  ...;
}
```

recipe-client-addon/lib/Addons.jsm
```diff
@@ -0,0 +1,108 @@
...
+ *   "extension", "theme", etc.
+ * @property {string} version
+ */
+
+this.Addons = {
+  /**
+   * Get information about an installed add-on by ID.
+   *
+   * @param {string} addonId
+   * @returns {SafeAddon?} Add-on with given ID, or null if not found.
```
gijsk: Note that `AddonManager` can throw if you e.g. pass an empty string as an id. Not sure that needs to be in this docstring or not, just flagging this up.

recipe-client-addon/lib/Addons.jsm
```diff
@@ -0,0 +1,108 @@
...
+  },
+
+  /**
+   * Get information about all installed add-ons.
+   * @async
+   * @returns {Array<SafeAddon>}
+   */
+  getAll: async function get(addonId) {
+    const addons = await AddonManager.getAllAddons();
+    return addons.map(Addons.serializeForSandbox);
```
gijsk: Maybe `.bind(Addons)` so we don't leave the footgun of later trying to use `this` inside `serializeForSandbox` and that then breaking?

recipe-client-addon/lib/NormandyDriver.jsm
```diff
@@ -157,61 +158,9 @@ this.NormandyDriver = function(sandboxManager) {
...
-      uninstall: sandboxManager.wrapAsync(async function uninstall(addonId) {
-        const addon = await AddonManager.getAddonByID(addonId);
-        if (addon === null) {
-          throw new Error(`No addon with ID [${addonId}] found.`);
-        }
-        addon.uninstall();
-        return null;
-      }),
+      get: sandboxManager.wrapAsync(Addons.get, {cloneInto: true}),
+      install: sandboxManager.wrapAsync(Addons.get),
```
Osmose: This should be passing `Addons.install` instead of `Addons.get`.

mythmon:
Feedback'd and rebased on master. I opted to leave the tests as is for now, since it helps lend reliability to this change.

Osmose:
@mythmon The latest code looks fine to me, but when I try to run the tests locally mochitest hangs. Updating to the latest mozilla-central and rebuilding doesn't help. Does this work locally for you? Any tips on how to diagnose?

mythmon:
The tests work fine for me on mozilla-central changeset e0b0865639ce.

recipe-client-addon/lib/Addons.jsm
```diff
@@ -47,9 +48,9 @@ this.Addons = {
    * @async
    * @returns {Array<SafeAddon>}
    */
-  getAll: async function get(addonId) {
+  async getAll(addonId) {
     const addons = await AddonManager.getAllAddons();
-    return addons.map(Addons.serializeForSandbox);
+    return addons.map(Addons.serializeForSandbox.bind(this));
```
gijsk: This looks odd - `Addons` and `this` are equivalent, it looks like, but it would be more obvious if we bound to the same thing we accessed the method on (so either `Addons.serializeForSandbox.bind(Addons)` or `this.serializeForSandbox.bind(this)`).

Osmose:
Tests pass locally for me, this should be fine (and blame me and my admin powers if it is not, of course).

PR #892, opened 2017-07-17
JEXL: explore AST traversal for JEXL

gregglind:
Claim:
1.  we can do it.
2.  nested (dot) ids are a pain

usage:
cd lib
node idsintree.js [jexl]

Next:
1.  do we WANT to do this?
2.  expose what exactly?

### DIFF
--- a/mozjexl/lib/idsintree.js
+++ b/mozjexl/lib/idsintree.js
@@ -1,0 +1,37 @@
+
+expression = process.argv[2] || 'normandy.clientId=="abced" && channel=="nightly"';
+console.log(`
+
+  EXPRESSION TO PARSE
+
+  ${expression}
+
+`)
+
+var Evaluator = require('./evaluator/Evaluator'),
+  Lexer = require('./Lexer'),
+  Parser = require('./parser/Parser'),
+  defaultGrammar = require('./grammar').elements;
+
+P = new Parser(defaultGrammar);
+P.addTokens(new Lexer(defaultGrammar).tokenize(expression))
+
+function findIdsInTree (jexlParseTree) {
+  const out = [];
+  const tree = jexlParseTree;
+  function recurse (tree) {
+    if (!tree) {
+      return
+    }
+    console.log(tree);
+    if (tree.type === "Identifier") {
+      out.push(tree.value);
+    }
+    recurse(tree.left);
+    recurse(tree.right);
+  };
+  recurse(tree);
+  return out;
+}
+out = findIdsInTree(P._tree);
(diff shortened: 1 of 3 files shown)


Osmose:
FWIW [pyjexl](https://github.com/Osmose/pyjexl) uses AST traversal to implement validation of things like variable names used in the expression.

mythmon:
TC failure is #890. CircleCI failure is because we have mozjexl configured to run with Node 4 because it's gulp build tasks aren't compatible with newer versions of Node. I guess that means we can't use things like template strings (backticks).

mozjexl/lib/idsintree.js
```diff
@@ -0,0 +1,37 @@
...
+
+`)
+
+var Evaluator = require('./evaluator/Evaluator'),
+  Lexer = require('./Lexer'),
+  Parser = require('./parser/Parser'),
+  defaultGrammar = require('./grammar').elements;
+
+P = new Parser(defaultGrammar);
+P.addTokens(new Lexer(defaultGrammar).tokenize(expression))
```
mythmon: I think mozjexl could expose a nicer API than this. We could probably expose a tree walking API directly.

mythmon:
@gregglind this is an interesting proof-of-concept. Where are you going with this? What kind of things do you want to eventually build out of a AST traverser?

gregglind:
(Also this really shouldn't be merged anyway... it has an example runner script.

Claims:
- simplest: mention what fields the user is filtering on, add those as a recall / descriptor in recipes
- we can validate names and properties to lint field typos
- we can replace / break up '&&' clauses into fields automagically
- we can check arity of arguments for filter functions.

mythmon:
@gregglind Do you want to do anything else with this PR? Do you  have a goal in mind for when this will be complete?

gregglind:
So, I used this to do dashboarding for recipe-history.

Not sure if there is an ask, but I am definitely ready to talk through the
in-flight work.  I made my own tools, and I have some sense of real
questions I have asked.

I am also quite fine with moving / promoting a lot of things we do in jexl
(including sampling) into "1st class" things that don't require traversal.

GL

Project work at:
http://www.github.com/gregglind/x-shield-inflight-dashboard

mythmon:
Sorry, I don't really see how any of that relates to this PR. Do you want to do anything else with this PR? Do you have a list of things to do before this PR will be complete?

gregglind:
I don't think there is a PR here.  It's notice that there is a valuable
work to be done with the knowledge from this.

PR #893, opened 2017-07-17
Fix #886 - Prefetch signatures in recipe health checks.

### LINKED ISSUE #886: DoesNotExist: Signature matching query does not exist.
The current theory is that this is a race condition of some sorts.

https://sentry.prod.mozaws.net/operations/normandy-stage/issues/624465/

```
DoesNotExist: Signature matching query does not exist.
(8 additional frame(s) were not displayed)
...
  File "django/core/checks/registry.py", line 81, in run_checks
    new_errors = check(app_configs=app_configs)
  File "normandy/recipes/checks.py", line 70, in recipe_signatures_use_good_certificates
    urls = set(r.signature.x5u for r in signed_recipes)
  File "normandy/recipes/checks.py", line 70, in <genexpr>
    urls = set(r.signature.x5u for r in signed_recipes)
  File "django/db/models/fields/related_descriptors.py", line 179, in __get__
    rel_obj = qs.get()
  File "django/db/models/query.py", line 385, in get
    self.model._meta.object_name

DoesNotExist: Signature matching query does not exist.
```

mythmon:
I expect this will avoid the race condition of checking a recipe while its signature is deleted that we see on Sentry sometimes.

### DIFF
--- a/recipe-server/normandy/recipes/checks.py
+++ b/recipe-server/normandy/recipes/checks.py
@@ -39,7 +39,8 @@ def recipe_signatures_are_correct(app_configs, **kwargs):
     errors = []
     try:
         Recipe = apps.get_model('recipes', 'Recipe')
-        signed_recipes = list(Recipe.objects.exclude(signature=None))
+        # pre-fetch signatures, to avoid race condition with deleted signatures
+        signed_recipes = list(Recipe.objects.exclude(signature=None).select_related('signature'))
     except (ProgrammingError, OperationalError, ImproperlyConfigured) as e:
         errors.append(Info(f'Could not retrieve recipes: {e}', id=INFO_COULD_NOT_RETRIEVE_RECIPES))
     else:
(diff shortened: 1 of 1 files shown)


Osmose:
The original patch failed because it removed the `list()` cast around the fetch, which meant that the `try`, which was detecting if the database wasn't available at all, wasn't catching the connection error, since the recipes were now being fetched lazily. Adding the `list()` cast back in fixes this.

Since this PR went up, we disabled the checks in question due to #900, so nothing is running these checks anyway, including the tests. You can test this issue locally by un-commenting the lines in `normandy.recipes.checks.register` that disabled the checks, and running `./manage.py check` with Normandy pointing at an uninitialized database.

mythmon:
Osmose's explanation makes sense, and I can confirm that his fix works for me.

TC Failure is #757, failure while cloning. I think we can bypass TC on this PR, since it is a server-side check.

Someone still needs to review this, and it can't be me, since I filed the PR.

PR #894, opened 2017-07-18
Clean up redesign branch

rehandalal:
r?


recipe-server/client/control_new/routes.js
```diff
@@ -68,20 +68,25 @@ export const {
...
+  };
 
-  if (router.route) {
-    content = <router.result.component />;
-  }
+  render() {
+    const { router } = this.props;
 
-  return <App>{content}</App>;
+    return (
```
andymikulski: IMO, I kinda liked this the old way - feels cleaner to do `<App>{content}</App>`. What if:
```
const content = router.route ? <router.result.component /> : <MissingPage />;
return <App>{content}</App>;
```
?

recipe-server/client/control_new/utils/api.js
```diff
@@ -0,0 +1,75 @@
+import * as Cookie from 'js-cookie';
+
+
+export default class APIClient {
```
andymikulski: Can you touch on why you decided to turn this into a class? I debated internally for a few minutes and I think this is a smart move, but I'd be interested to hear your thoughts on it.

andymikulski: Can you elaborate on why you made this a class? After internal debate, it makes sense to do this, but I'd be interested to hear your thoughts on it.

rehandalal: Sure. To me there was nothing inherently wrong with the way we were doing it. I do however think this is: a) cleaner to read. b) more flexible should we for some reason need to call an API with a different root. c) in the future we could create shortcut methods like `.post`, `.get`, `.patch` etc. and perhaps then consider making some changes to the requests action creator. 

There's not a lot of immediate benefit to this change but to my mind there are only potential upsides with no downsides.

Issue #889, opened 2017-07-14
Add ability to fetch and run recipes early on firstrun


Osmose:
We'd like to add a feature to Normandy that triggers fetching and executing recipes as early on a user's first run of Firefox as we can reasonably get.

- Once all recipes have finished executing, we should write the time since launch to telemetry so that we can analyze how reliably we can modify things like preferences at firstrun. TBD on how to structure this (a custom one-time packet? on the main ping? data review?).
  - Thinking more, we probably want to include individual recipe IDs and their finish times so we can measure a specific preference change, for example.
- We should add a field to filter expressions to allow filtering recipes to only execute on firstrun (`normandy.isFirstRun`?)

rhelmer:
I think that the earliest we can reasonably start running something like fetching code is after `sessionstore-windows-restored` (see https://developer.mozilla.org/en-US/docs/Observer_Notifications) - we'll need an observer like:

```
Cu.import("resource://gre/modules/XPCOMUtils.jsm");
Cu.import("resource://gre/modules/Services.jsm");

let observer = {
  observe: function(subject, topic, data)
  {
    Services.obs.removeObserver("sessionstore-windows-restored", this);
    // fetch and execute recipes
  },
  QueryInterface: XPCOMUtils.generateQI([
    Components.interfaces.nsIObserver,
    Components.interfaces.nsISupportsWeakReference
  ])
};
Services.obs.addObserver("sessionstore-windows-restored", observer, true);
```
Since there's already an update timer being used I suspect we can just trigger it to run when the observer notification fires.

rhelmer:
OK. So it looks like `RecipeRunner.jsm` registers a timer with the ID `recipe-client-addon-run`, which fetches and executes recipes.

What we should be able to do here is add a `"sessionstore-windows-restored"` observer early in the startup of the system add-on, which calls `notify()` on the timer.

rhelmer:
On further reflection I think it's simpler and more readable to just run once and set the timer up after `"sessionstore-windows-restored"` is observed, rather than let the timer be set up and notify it explicitly etc.

Osmose:
#895 fixed this and it got uplifted and is currently in release.

PR #895, opened 2017-07-18
fetch and run recipes early in startup

rhelmer:
This is just the first bit for issue #889 - do a first fetch+run of recipes and then creates the update timer after `"sessionstore-windows-restored"` has been observed.

### DIFF
--- a/recipe-client-addon/lib/ClientEnvironment.jsm
+++ b/recipe-client-addon/lib/ClientEnvironment.jsm
@@ -203,6 +203,10 @@ this.ClientEnvironment = {
       return Utils.keyBy(addons, "id");
     });
 
+    XPCOMUtils.defineLazyGetter(environment, "isFirstRun", () => {
+      return Preferences.get("extensions.shield-recipe-client.first_run");
+    });
+
     return environment;
   },
 };
--- a/recipe-client-addon/lib/RecipeRunner.jsm
+++ b/recipe-client-addon/lib/RecipeRunner.jsm
@@ -42,6 +42,9 @@ const log = LogManager.getLogger("recipe-runner");
 const prefs = Services.prefs.getBranch("extensions.shield-recipe-client.");
 const TIMER_NAME = "recipe-client-addon-run";
 const RUN_INTERVAL_PREF = "run_interval_seconds";
+const FIRST_RUN_PREF = "first_run";
+const UI_AVAILABLE_NOTIFICATION = "sessionstore-windows-restored";
+const SHIELD_INIT_NOTIFICATION = "shield-init-complete";
 
 this.RecipeRunner = {
   init() {
@@ -54,6 +57,28 @@ this.RecipeRunner = {
       this.run();
     }
 
+    if (prefs.getBoolPref(FIRST_RUN_PREF)) {
+      // Run once immediately after the UI is available. Do this before adding the
+      // timer so we can't end up racing it.
+      const observer = {
+        observe: (subject, topic, data) => {
+          Services.obs.removeObserver(observer, UI_AVAILABLE_NOTIFICATION);
+
+          this.run();
+          this.registerTimer();
+          prefs.setBoolPref(FIRST_RUN_PREF, false);
+
+          Services.obs.notifyObservers(null, SHIELD_INIT_NOTIFICATION);
+        },
+      };
+      Services.obs.addObserver(observer, UI_AVAILABLE_NOTIFICATION);
+      CleanupManager.addCleanupHandler(() => Services.obs.removeObserver(observer, UI_AVAILABLE_NOTIFICATION));
+    } else {
+      this.registerTimer();
+    }
+  },
+
+  registerTimer() {
     this.updateRunInterval();
     CleanupManager.addCleanupHandler(() => timerManager.unregisterTimer(TIMER_NAME));
 
--- a/recipe-client-addon/lib/ShieldRecipeClient.jsm
+++ b/recipe-client-addon/lib/ShieldRecipeClient.jsm
@@ -39,6 +39,7 @@ const DEFAULT_PREFS = {
   "logging.level": Log.Level.Warn,
   user_id: "",
   run_interval_seconds: 86400, // 24 hours
+  first_run: true,
 };
 const PREF_DEV_MODE = "extensions.shield-recipe-client.dev_mode";
 const PREF_LOGGING_LEVEL = PREF_BRANCH + "logging.level";
--- a/recipe-client-addon/test/browser/browser_ClientEnvironment.js
+++ b/recipe-client-addon/test/browser/browser_ClientEnvironment.js
@@ -134,3 +134,16 @@ add_task(withDriver(Assert, async function testAddonsInContext(driver) {
 
   await driver.addons.uninstall(addonId);
 }));
+
+add_task(async function isFirstRun() {
+  let environment = ClientEnvironment.getEnvironment();
+
+  // isFirstRun is set to false after the recipe client runs
+  ok(!environment.isFirstRun, "isFirstRun has a default value");
+
+  // isFirstRun is read from a preference
+  await SpecialPowers.pushPrefEnv({set: [["extensions.shield-recipe-client.first_run", true]]});
+  environment = ClientEnvironment.getEnvironment();
+  ok(environment.isFirstRun, "isFirstRun is read from preferences");
+});
+
--- a/recipe-client-addon/test/browser/browser_RecipeRunner.js
+++ b/recipe-client-addon/test/browser/browser_RecipeRunner.js
@@ -330,7 +330,13 @@ add_task(async function testStartup() {
   const updateRunIntervalStub = sinon.stub(RecipeRunner, "updateRunInterval");
 
   // in dev mode
-  await SpecialPowers.pushPrefEnv({set: [["extensions.shield-recipe-client.dev_mode", true]]});
+  await SpecialPowers.pushPrefEnv({
+    set: [
+      ["extensions.shield-recipe-client.dev_mode", true],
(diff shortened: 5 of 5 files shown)


rhelmer:
@Osmose - does this need to *only* run shortly after startup on the *very* first run of a new profile?

Osmose:
> @Osmose - does this need to only run shortly after startup on the very first run of a new profile?

Hmmmm... maybe we can always run shortly after startup? @mythmon @gijsk Thoughts?

gijsk:
I don't really have well-developed thoughts here. The cost is slowing down startup a bit, the benefits aren't really super-clear to me. I expect that in something like 90% of cases we wouldn't need to refetch that quickly (ie the refetch will be a no-op), so not sure it would really make sense to continually refetch like that without further conditions (e.g. recent update (so version and therefore eligibility changes for studies), more than X amount of time since last update, etc.). But that's just a gut sense of things, so I could be wrong.

Osmose:
> I don't really have well-developed thoughts here. The cost is slowing down startup a bit, the benefits aren't really super-clear to me. I expect that in something like 90% of cases we wouldn't need to refetch that quickly (ie the refetch will be a no-op), so not sure it would really make sense to continually refetch like that without further conditions (e.g. recent update (so version and therefore eligibility changes for studies), more than X amount of time since last update, etc.). But that's just a gut sense of things, so I could be wrong.

I think I agree with this; there's high value in fetching immediately on first run, but not really so much after that.

mythmon:
I'd say that except for the special case of first 15 minute attrition, we don't need fast startup. This makes sense for first run, but I don't think we want to do it every time.

rhelmer:
Makes sense, thanks. Should we then set a pref on first-run, so it can be reset later if needed?

mythmon:
Storing it in a preference sounds like a good route. Any other storage mechanism would probably get us in trouble for doing extra IO during.

recipe-client-addon/lib/RecipeRunner.jsm
```diff
@@ -50,12 +51,26 @@ this.RecipeRunner = {
...
+        // Watch for the run interval to change, and re-register the timer with the new value
+        prefs.addObserver(RUN_INTERVAL_PREF, this);
+        CleanupManager.addCleanupHandler(() => prefs.removeObserver(RUN_INTERVAL_PREF, this));
+      },
+      QueryInterface: XPCOMUtils.generateQI([
+        Components.interfaces.nsIObserver,
+        Components.interfaces.nsISupportsWeakReference,
+      ]),
+    };
+    Services.obs.addObserver(observer, UI_AVAILABLE_NOTIFICATION);
```
gijsk: The default 3rd arg is a strong reference. If we use a strong reference, you can just pass a function and remove the QI goop. If you want a weak reference, pass `true` as the third arg here and in the test.

gijsk: Separately, what happens if we update this add-on out of band and the notification has already happened? Would we just wait forever and never call `.run()`?

If we only want this to happen on first run, as said in the comments, I guess we want a pref of sorts.

recipe-client-addon/test/browser/browser_RecipeRunner.js
```diff
@@ -212,25 +212,37 @@ add_task(async function testStartup() {
...
+      runStub.restore();
+      addCleanupHandlerStub.restore();
+      updateRunIntervalStub.restore();
+    },
+    QueryInterface: XPCOMUtils.generateQI([
+      Components.interfaces.nsIObserver,
+      Components.interfaces.nsISupportsWeakReference,
+    ]),
+  };
+  Services.obs.addObserver(observer, UI_AVAILABLE_NOTIFICATION);
```
gijsk: I have to admit I don't understand this change. Surely there's no guarantee that the test runs after this observer notification has been sent? Where is it being sent instead?

recipe-client-addon/lib/RecipeRunner.jsm
gijsk: Finally, even if this is weak, should we explicitly remove it with the cleanup manager? And if we do that, is it worth having the weak observer?

rhelmer: I'll just use a strong reference here and add to cleanup manager.

Regarding updating out-of-band, you're correct and I should've resolved the question about first-run before bothering you to review :) I'll add a pref as discussed.

recipe-client-addon/test/browser/browser_RecipeRunner.js
rhelmer: Not... sure what I was thinking w/ the test change.

recipe-client-addon/lib/ShieldRecipeClient.jsm
```diff
@@ -62,7 +63,7 @@ this.ShieldRecipeClient = {
       () => Preferences.ignore(PREF_LOGGING_LEVEL, LogManager.configure),
     );
     log = LogManager.getLogger("bootstrap");
-
+/* FIXME this is preventing first run.
```
rhelmer: @mythmon mentioned that there is a PR to remove this bit, if we land and I rebase this can go away.

recipe-client-addon/test/browser/browser_RecipeRunner.js
```diff
@@ -225,6 +226,7 @@ add_task(async function testStartup() {
 
   // not in dev mode
   await SpecialPowers.pushPrefEnv({set: [["extensions.shield-recipe-client.dev_mode", false]]});
+  await SpecialPowers.pushPrefEnv({set: [["extensions.shield-recipe-client.first_run", false]]});
   RecipeRunner.init();
   ok(!runStub.called, "RecipeRunner.run is not called immediately when not in dev mode");
   ok(addCleanupHandlerStub.called, "A cleanup function is registered when not in dev mode");
```
rhelmer: I'm having trouble coming up with a good testing strategy here - for the moment this just makes the current tests pass.

The problem is that shield completes way too early in startup, so listening for a notification or for the pref to change will be far too late to really prove anything the existing test doesn't.

I suppose we could have the shield client execute a recipe that spins the event loop enough times that the test can see a notification? Or perhaps execute a recipe that flips a pref or something else that the test could verify? Any thoughts? :)

gijsk: I think this is impossible to test well with browser mochitests. You may stand a chance using marionette, but equally I don't think that needs to block landing.

recipe-client-addon/lib/ShieldRecipeClient.jsm
rhelmer: I split this bit out to PR #913

rhelmer:
@gijsk @mythmon mind taking another look at this one now? It depends on PR #913 but I think it's ready other than that.

recipe-client-addon/test/browser/browser_ClientEnvironment.js
```diff
@@ -134,3 +134,16 @@ add_task(withDriver(Assert, async function testAddonsInContext(driver) {
 
   await driver.addons.uninstall(addonId);
 }));
+
+add_task(async function isFirstRun() {
+  let environment = ClientEnvironment.getEnvironment();
+
+  // isFirstRun defaults to false
+  ok(environment.isFirstRun, "isFirstRun has a default value");
```
mythmon: I think the comment and the code are mismatched here.

recipe-client-addon/test/browser/browser_RecipeRunner.js
```diff
@@ -331,6 +331,7 @@ add_task(async function testStartup() {
 
   // in dev mode
   await SpecialPowers.pushPrefEnv({set: [["extensions.shield-recipe-client.dev_mode", true]]});
+  await SpecialPowers.pushPrefEnv({set: [["extensions.shield-recipe-client.first_run", false]]});
```
mythmon: You should combine these two calls to `pushPrefEnv`:

```js
await SpecialPower.pushPrefEnv({
  set: [
    ["extensions.shield-recipe-client.dev_mode", true],
    ["extensions.shield-recipe-client.first_run", false],
  ]
});
```

recipe-client-addon/test/browser/browser_RecipeRunner.js
```diff
@@ -342,6 +343,7 @@ add_task(async function testStartup() {
 
   // not in dev mode
   await SpecialPowers.pushPrefEnv({set: [["extensions.shield-recipe-client.dev_mode", false]]});
+  await SpecialPowers.pushPrefEnv({set: [["extensions.shield-recipe-client.first_run", false]]});
```
mythmon: These should also be combined.

recipe-client-addon/lib/ClientEnvironment.jsm
```diff
@@ -203,6 +203,10 @@ this.ClientEnvironment = {
       return Utils.keyBy(addons, "id");
     });
 
+    XPCOMUtils.defineLazyGetter(environment, "isFirstRun", () => {
```
gijsk: Are you deliberately not using `XPCOMUtils.defineLazyPreferenceGetter()` to avoid the observer, or...?

mythmon:
All of the CI failures are due to an error like "ReferenceError: SpecialPower is not defined". Looks like you forgot the `s` on the end of <code>SpecialPower<strong>s</strong></code> somewhere.

recipe-client-addon/lib/ClientEnvironment.jsm
rhelmer: Just following the style of the rest of the file, really. I'm not sure that the observer would be helpful since I think these getters are only used when evaluating a recipe and not something shield needs to worry about changing out from under it.

PR #898, opened 2017-07-19
#873: Clone Page

### LINKED ISSUE #873: New Clone Recipe UI

andymikulski:
Fixes #873 

- Adds `/clone` routes to revision + latest recipe routes
- Adds `getRecipeFromURL` selector which examines URL params for recipe and revision IDs and returns the complete recipe.

Github is showing a commit that should already be landed (`Add CreateRecipePage`), but on further inspection it seems that the commit message doesnt align with what the commit does. I'm wondering if this is left over from an `amend` or something.

### DIFF
--- a/recipe-server/client/control_new/components/recipes/CloneRecipePage.js
+++ b/recipe-server/client/control_new/components/recipes/CloneRecipePage.js
@@ -6,7 +6,7 @@ import React from 'react';
 import { connect } from 'react-redux';
 import { Link, push as pushAction } from 'redux-little-router';
 
-import LoadingOverlay from 'control_new/components/common/LoadingOverlay';
+import { SimpleLoadingOverlay } from 'control_new/components/common/LoadingOverlay';
 import RecipeForm from 'control_new/components/recipes/RecipeForm';
 import QueryRecipe from 'control_new/components/data/QueryRecipe';
 
@@ -64,11 +64,9 @@ export default class CloneRecipePage extends React.Component {
         'Recipe cannot be saved. Please correct any errors listed in the form below.',
       );
 
-      if (error) {
-        this.setState({
-          formErrors: error.data || error,
-        });
-      }
+      this.setState({
+        formErrors: error.data || error,
+      });
     }
   }
 
@@ -83,7 +81,7 @@ export default class CloneRecipePage extends React.Component {
     return (
       <div className="clone-page">
         <QueryRecipe pk={recipeId} />
-        <LoadingOverlay useCondition condition={!recipeName}>
+        <SimpleLoadingOverlay isVisible={!recipeName}>
           <h2>New Recipe</h2>
           { recipeName &&
             <h3>
@@ -96,7 +94,7 @@ export default class CloneRecipePage extends React.Component {
             onSubmit={this.handleSubmit}
             errors={this.state.formErrors}
           />
-        </LoadingOverlay>
+        </SimpleLoadingOverlay>
       </div>
     );
   }
--- a/recipe-server/client/control_new/components/recipes/CloneRecipePage.js
+++ b/recipe-server/client/control_new/components/recipes/CloneRecipePage.js
@@ -1,38 +1,37 @@
-import { message, Button, Icon, Dropdown, Menu } from 'antd';
+import { message } from 'antd';
 import autobind from 'autobind-decorator';
 import { Map } from 'immutable';
 import PropTypes from 'prop-types';
 import React from 'react';
 import { connect } from 'react-redux';
+import { Link, push as pushAction } from 'redux-little-router';
 
-import { SimpleLoadingOverlay } from 'control_new/components/common/LoadingOverlay';
+import LoadingOverlay from 'control_new/components/common/LoadingOverlay';
 import RecipeForm from 'control_new/components/recipes/RecipeForm';
 import QueryRecipe from 'control_new/components/data/QueryRecipe';
 
-import { updateRecipe } from 'control_new/state/recipes/actions';
+import { createRecipe as createAction } from 'control_new/state/recipes/actions';
 import { getRecipe } from 'control_new/state/recipes/selectors';
-import { getRecipeForRevision } from 'control_new/state/revisions/selectors';
-import { getUrlParamAsInt } from 'control_new/state/router/selectors';
-
+import { getRouterParamAsInt } from 'control_new/state/router/selectors';
 
 @connect(
   state => {
-    const recipeId = getUrlParamAsInt(state, 'recipeId');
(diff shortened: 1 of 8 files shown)


recipe-server/client/control_new/components/recipes/CloneRecipePage.js
```diff
@@ -0,0 +1,117 @@
...
+import { getRecipeFromURL } from 'control_new/state/recipes/selectors';
+
+@connect(
+  state => {
+    const {
+      isLatestRevision,
+      recipe,
+      recipeId,
+      revisionId,
+    } = getRecipeFromURL(state);
```
rehandalal: So I went back and looked at what I was doing on the recipe details page and on further thought I think this logic is actually clean enough that it doesn't need another selector or a "container":
```javascript
    const recipeId = getUrlParamAsInt(state, 'recipeId');
    const latestRevisionId = getLatestRevisionIdForRecipe(state, recipeId, '');
    const revisionId = getUrlParam(state, 'revisionId', latestRevisionId);
    const revision = getRevision(state, revisionId, new Map());
    const isLatestRevision = isLatestRevision(state, revisionId);
```

rehandalal: I think this makes it transparent what the logic is and I don't think it's actually complicated enough that it needs to be pulled out into it's own function.

recipe-server/client/control_new/components/recipes/CloneRecipePage.js
```diff
@@ -0,0 +1,117 @@
...
+  {
+    push: pushAction,
+    createRecipe: createAction,
+  },
+)
+@autobind
+export default class CloneRecipePage extends React.Component {
+  static propTypes = {
+    createRecipe: PropTypes.func.isRequired,
+    isLatestRevision: PropTypes.bool,
```
rehandalal: this can be required as you will always get a boolean value from the selector.

recipe-server/client/control_new/components/recipes/CloneRecipePage.js
```diff
@@ -0,0 +1,117 @@
...
+    recipeId: PropTypes.number.isRequired,
+    recipe: PropTypes.instanceOf(Map),
```
rehandalal: this will also always have an instance of `Map`.

recipe-server/client/control_new/components/recipes/CloneRecipePage.js
```diff
@@ -0,0 +1,117 @@
...
+    revisionId: PropTypes.string.isRequired,
+  };
+
+  static defaultProps = {
```
rehandalal: you can ditch these and make everything required.

recipe-server/client/control_new/components/recipes/CloneRecipePage.js
```diff
@@ -0,0 +1,117 @@
...
+
+    // Only display revision hash if we're _not_ on the latest version.
+    const revisionAddendum = isLatestRevision ? '' : `(revision ${revisionId.slice(0, 7)}...)`;
+    const originalRecipeText = `"${recipeName}" ${revisionAddendum}`;
+
+    return (
+      <div className="clone-page">
+        <QueryRecipe pk={recipeId} />
+        <SimpleLoadingOverlay isVisible={!recipeName}>
+          <h2>New Recipe</h2>
```
rehandalal: This should be:
```jsx
<h2>Clone Recipe</h2>
```

recipe-server/client/control_new/components/recipes/CloneRecipePage.js
```diff
@@ -0,0 +1,117 @@
...
+      });
+    }
+  }
+
+  render() {
+    const { recipe, recipeId, isLatestRevision, revisionId } = this.props;
+    const recipeName = recipe.get('name');
+
+    // Remove the 'name' field value.
+    const displayedRecipe = recipe.set('name');
```
rehandalal: I don't think we need to strip the name away. I think it's easy enough to delete and annoying to have to copy over if you want to modify slightly.

recipe-server/client/control_new/state/recipes/selectors.js
```diff
@@ -90,3 +91,30 @@ export function getRecipeApprovalHistory(state, id) {
   const history = getRecipeHistory(state, id);
   return history.filter(revision => revision.get('approval_request'));
 }
+
+export function getRecipeFromURL(state) {
```
rehandalal: we should be able to ditch this in favor of:
https://github.com/mozilla/normandy/pull/898/files#diff-cf8531aa9cc5d72ba402b1ed599a8dd7R23

recipe-server/client/control_new/components/recipes/CloneRecipePage.js
andymikulski: This was a UX thing that had been brought up before, I think maybe even in the latest all hands review.

When you clone a recipe, it's confusing because we set the name to `[name of recipe] (clone)` (or `clone of [name of recipe]`? Can't remember). I'd argue removing the name entirely solves that issue, plus if the user wanted to use the last recipe name, we provide it to them under the "New Recipe" header.

rehandalal: ah! i don't remember them bringing that up, but that's cool. we can keep this in and if they complain about it change accordingly.

recipe-server/client/control_new/components/recipes/CloneRecipePage.js
andymikulski: I think it's better to signify that it's a `new` recipe versus a `cloned` recipe. 'Cloning' here is essentially using the values from one recipe to autofill the New Recipe form. Users can then mess with the values, the original history is not brought over, etc. There's no connection to the original recipe and the new recipe beyond any values that were left untouched.

To the end user, the only difference between hitting `new recipe` and `clone` is that on the clone page it has `Cloning from "some recipe"` and the fields are pre-populated. At the end of the day, you're still just creating a new recipe.

andymikulski: Ahh, but then there's the disconnect between the actual button saying `Clone` and the page the user lands on saying `New Recipe`.... Curses!

andymikulski:
@rehandalal All set. I had to change `getRevision` to `getRecipeForRevision` but then it worked great. Good work!

This is ready for a final pass and merge.

PR #902, opened 2017-07-21
about:studies

Osmose:
Adds the about:studies page. Seems to work. WIP because it still needs tests.

@gijsk When I run the test that I added with this PR (`browser_about_studies.js`) with this PR I get the following failure:

```
18 INFO TEST-UNEXPECTED-FAIL | browser/extensions/shield-recipe-client/test/browser/browser_about_studies.js | A promise chain failed to handle a rejection: Component returned failure code: 0xc1f30001 (NS_ERROR_NOT_INITIALIZED) [nsIMessageSender.sendAsyncMessage] - stack: null
Rejection date: Thu Jul 20 2017 14:56:59 GMT-0700 (PDT) - false == true - JS frame :: resource://testing-common/PromiseTestUtils.jsm :: assertNoUncaughtRejections :: line 265
Stack trace:
    resource://testing-common/PromiseTestUtils.jsm:assertNoUncaughtRejections:265
    chrome://mochikit/content/browser-test.js:Tester.prototype.nextTest<:429
    testScope/test_finish/<@chrome://mochikit/content/browser-test.js:1075:11
    run@chrome://mochikit/content/browser-test.js:1012:9
```

I don't really know how to track this down. Any ideas?


Osmose:
I added some functional tests for `about:studies` and am no longer getting the `NS_ERROR_NOT_INITIALIZED` error. Hooray!

Instead, though, I'm seeing the following logs during a mochitest run:

```
448 INFO Console message: [JavaScript Error: "NS_ERROR_FACTORY_EXISTS: Component returned failure code: 0xc1f30100 (NS_ERROR_FACTORY_EXISTS) [nsIComponentRegistrar.registerFactory]" {file: "resource://shield-recipe-client/content/AboutPages.jsm" line: 74}]
```

and

```
GECKO(4353) | JavaScript error: resource://shield-recipe-client/content/AboutPages.jsm, line 87: TypeError: Cm is undefined
```

Perhaps something about how mochitests are run is re-running the about page initialization without cleaning it up?

Osmose:
Barring the logged issues I mentioned above, I think this is ready for review.

recipe-client-addon/content/AboutPages.jsm
```diff
@@ -0,0 +1,208 @@
...
+    CleanupManager.addCleanupHandler(() => {
+      // Stop loading processs scripts and notify existing scripts to clean up.
+      Services.ppmm.removeDelayedProcessScript(PROCESS_SCRIPT);
+      Services.ppmm.broadcastAsyncMessage("Shield:ShuttingDown");
+      Services.mm.removeDelayedFrameScript(FRAME_SCRIPT);
+      Services.mm.broadcastAsyncMessage("Shield:ShuttingDown");
+
+      // Clean up about pages
+      this.aboutStudies.unregisterParentListeners();
+      this.aboutStudies.unregister();
```
gijsk: This is called twice, I think? Once here and once from the process script. I guess we should only add this to the cleanup manager if we're in the parent process, or something - and the content process script, is that only loaded in the content process, or also in the parent? Maybe that needs an exception, too, or we delegate the about page stuff to one of the two...

gijsk: Ah, hm, and we can probably avoid doing this on actual app shutdown, but I don't know if that's complicated. Might help with the errors you were seeing if they're at shutdown rather than earlier (or just wallpaper them... checking would be a good idea).

recipe-client-addon/lib/ShieldRecipeClient.jsm
```diff
@@ -39,9 +41,9 @@ const DEFAULT_PREFS = {
   "logging.level": Log.Level.Warn,
   user_id: "",
   run_interval_seconds: 86400, // 24 hours
+  shieldLearnMoreUrl: "https://support.mozilla.org/1/firefox/%VERSION%/%OS%/%LOCALE%/shield",
 };
 const PREF_DEV_MODE = "extensions.shield-recipe-client.dev_mode";
-const PREF_SELF_SUPPORT_ENABLED = "browser.selfsupport.enabled";
```
gijsk: woop!

recipe-client-addon/jar.mn
```diff
@@ -3,8 +3,9 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 [features/shield-recipe-client@mozilla.org] chrome.jar:
-% resource shield-recipe-client %content/
-  content/lib/ (./lib/*)
-  content/data/ (./data/*)
-  content/skin/  (skin/*)
-  content/vendor/ (./vendor/*)
+% resource shield-recipe-client %
```
gijsk: Note that this technically exposes all the resources shipped with the add-on to web content.  We're changing that in https://bugzilla.mozilla.org/show_bug.cgi?id=863246 but it might be something to be more cautious about here? Could use 2 resource contexts (shield-recipe-client-public or something) to mitigate this.

gijsk: Plus after that bug the public one will need `contentaccessible=yes` setting, but that's a separate issue...

recipe-client-addon/test/browser/browser_about_studies.js
```diff
@@ -0,0 +1,147 @@
+"use strict";
+
+Cu.import("resource://shield-recipe-client/content/AboutPages.jsm", this);
+
+function withAboutStudies(testFunc) {
+  return async function inner(...args) {
+    return BrowserTestUtils.withNewTab(
+      {gBrowser, url: "about:studies"},
```
gijsk: protip: you can just pass the url string instead of the object, and it'll just use the topmost browser window's gBrowser.

recipe-client-addon/test/browser/browser_about_studies.js
```diff
@@ -0,0 +1,147 @@
...
+  }),
+  withAboutStudies,
+  async function testLearnMore(browser) {
+    browser.contentDocument.getElementById("shield-studies-learn-more").click();
+    await BrowserTestUtils.waitForLocationChange(gBrowser);
+
+    const location = browser.contentWindow.location.href;
+    is(
+      location,
+      AboutPages.aboutStudies.getShieldLearnMoreHref(),
```
gijsk: Shouldn't we compare this to the constant ref as well, and check it fully matches?

recipe-client-addon/content/about-studies/about-studies.js
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  render() {
+    const { page, selected } = this.props;
+    return (
+      r("li", {
+        className: classnames("category", {selected}),
+        onClick: this.handleClick,
+      },
+        page.icon && r("img", {className: "category-icon", src: page.icon}),
```
mythmon: `page.icon` is marked as required in the proptypes, so this shouldn't ever not render, right?

recipe-client-addon/bootstrap.js
```diff
@@ -24,6 +24,7 @@ this.shutdown = function(data, reason) {
   // modules are not unloaded accidentally during tests.
   const log = LogManager.getLogger("bootstrap");
   const modules = [
+    "content/AboutPages.jsm",
```
Osmose: I figured it was a good signal that at least some of the code in this module runs in the content process and is restricted by that. Also I copied the pattern from Pocket, which puts it there too.

recipe-client-addon/content/AboutPages.jsm
Osmose: This cleanup handler is only called in the parent process, so `unregister` only gets called once per-process (the process script handles unregistering in the content processes). I was having trouble tracking what was running where, so I made the process script only perform any work if it's run in the content process, and parent-only logic happens in this method.

Maybe a comment here explaining that this is the split point for stuff that happens in the content process vs the parent?

> Ah, hm, and we can probably avoid doing this on actual app shutdown, but I don't know if that's complicated. Might help with the errors you were seeing if they're at shutdown rather than earlier (or just wallpaper them... checking would be a good idea).

Is this expensive enough to be worth avoiding during shutdown? I was trying to avoid adding more conditional logic given how confusing the parent/content split already is.

recipe-client-addon/content/about-studies/shield-studies.js
```diff
@@ -0,0 +1,151 @@
...
+    this.setState({
+      learnMoreHref: value,
+    });
+  }
+
+  render() {
+    return (
+      r("div", {},
+        r(InfoBox, {},
+          r("span", {}, "What's this? Firefox may install and run studies from time to time."),
```
Osmose: I checked with flod, and because users only get to this page by typing it into the address bar, and because we won't be running studies outside of English locales for a while anyway, he's okay shipping this without anything hiding the page for non-en-* locales. I didn't get an "official" signoff though.

The plan is to ship l10n support ASAP after we ship the initial version.

recipe-client-addon/test/browser/browser_about_studies.js
Osmose: What do you mean by the constant ref?

recipe-client-addon/content/about-studies/.eslintrc.js
```diff
@@ -0,0 +1,15 @@
...
+    FxButton: false,
+    InfoBox: false,
+    PropTypes: false,
+    r: false,
+    React: false,
+    ReactDOM: false,
+    remoteValues: false,
+    sendPageEvent: false,
+    ShieldStudies: false,
+    stringToDate: false,
```
mythmon: This, and the bunch of `<script>` tags make me nervous. (BuddyUp flashbacks, fwiw). Is it possible to use Cu.import or some other module system in the JS files?

recipe-client-addon/content/AboutPages.jsm
```diff
@@ -0,0 +1,208 @@
...
+XPCOMUtils.defineLazyModuleGetter(
+  this, "CleanupManager", "resource://shield-recipe-client/lib/CleanupManager.jsm",
+);
+XPCOMUtils.defineLazyModuleGetter(
+  this, "StudyStorage", "resource://shield-recipe-client/lib/StudyStorage.jsm",
+);
+
+this.EXPORTED_SYMBOLS = ["AboutPages"];
+
+const SHIELD_LEARN_MORE_URL_PREF = "extensions.shield-recipe-client.shieldLearnMoreUrl";
```
Osmose: I think we should forgo making this lazy unless it's a big deal so that, if the pref changes, the next page load has the updated URL vs it being cached until next reboot, or some complex pref watching.

recipe-client-addon/content/about-studies/shield-studies.js
gijsk: If flod's happy, I'm happy. :-)

recipe-client-addon/content/AboutPages.jsm
gijsk: Well, we could just avoid the cleanupmanager calls on shutdown, unless that leaks the world? Could be follow-up fodder if it does.

recipe-client-addon/content/AboutPages.jsm
gijsk: `XPCOMUtils.defineLazyPreferenceGetter` will take care of all of this, but if you use formatURLPref this problem goes away anyway.

recipe-client-addon/test/browser/browser_about_studies.js
gijsk: Sorry, I meant the string href that you set the pref to, regex'd or whatever so it deals with the `%OS` replacement.

recipe-client-addon/content/AboutPages.jsm
Osmose: I'm not sure if it does or not. I'd prefer doing this as a followup to think harder about the difference between add-on removal and shutdown.

recipe-client-addon/test/browser/browser_about_studies.js
Osmose: I didn't really want to bother replicating the functionality of `formatURLPref` just for the test, especially when there's the helper function that just does that anyway, but if you think it's valuable I'm happy to do it anyway.

recipe-client-addon/content/about-studies/about-studies.js
Osmose: `icon` is totally not marked as required.

recipe-client-addon/test/browser/browser_about_studies.js
gijsk: Eh, I guess it doesn't matter too much either way. I just figured, right now we're essentially verifying that code that called `getShieldLearnMoreHref()` produces the same result as `getShieldLearnMoreHref()`... :-)

recipe-client-addon/content/about-studies/.eslintrc.js
Osmose: Nope.

Osmose:
Feedback'd, this is ready for another pass.

recipe-client-addon/content/about-studies/common.js
```diff
@@ -0,0 +1,149 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+"use strict";
+/* eslint-disable no-unused-vars */
```
gijsk: I think technically, this wants to just say it's exporting whatever is being used by the other files? Then local unused vars inside the functions will still get flagged up.

recipe-client-addon/content/about-studies/about-studies.html
```diff
@@ -0,0 +1,23 @@
...
+    <link rel="stylesheet" href="chrome://global/skin/global.css">
+    <link rel="stylesheet" href="chrome://global/skin/in-content/common.css">
+    <link rel="stylesheet" href="resource://shield-recipe-client-content/about-studies/about-studies.css">
+  </head>
+  <body>
+    <div id="app"></div>
+    <script src="resource://shield-recipe-client/vendor/React.js"></script>
+    <script src="resource://shield-recipe-client/vendor/ReactDOM.js"></script>
+    <script src="resource://shield-recipe-client/vendor/PropTypes.js"></script>
+    <script src="resource://shield-recipe-client/vendor/classnames.js"></script>
```
gijsk: I expect the react stuff here will want to be in the -content thing as well?

recipe-client-addon/content/shield-content-frame.js
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  onShutdown() {
+    removeMessageListener("Shield:SendStudyList", this);
+    removeMessageListener("Shield:ShuttingDown", this);
+    removeEventListener("Shield", this);
+  }
+
+  navigateToDataPreferences() {
+    if (Services.prefs.getBoolPref(USE_OLD_PREF_ORGANIZATION_PREF)) {
```
gijsk: Will this ever run against 54? If so, please add a fallback value of `true`, as it doesn't seem like the pref exists there.

recipe-client-addon/content/shield-content-frame.js
```diff
@@ -0,0 +1,130 @@
...
+      content.location = "about:preferences#advanced";
+      addEventListener("load", () => {
+        const tabbox = content.document.getElementById("advancedPrefs");
+        const tab = content.document.getElementById("dataChoicesTab");
+        tabbox.selectedTab = tab;
```
gijsk: So... the tests show this works, right, even with e10s enabled? Meanwhile, though, I am very confused, because about:preferences opens in the parent process, but this page opens in the child process. While the same script text will be loaded, because they run in different processes, I'm confused how this load listener would fire for the load that's happening in the parent process, and how the `content.document` reference would point to that thing. Can you doublecheck `about:studies` is really loading in the content process? Checking with the browser console for `gBrowser.selectedBrowser.isRemoteBrowser` with e10s turned on and about:studies loaded in the current tab should work (ie return `true`).

gijsk: Ah, in fact it seems like you're using `Ci.nsIAboutModule.URI_MUST_RUN_IN_CHILD`, but the correct flag is `Ci.nsIAboutModule.URI_MUST_LOAD_IN_CHILD`, so your page is still running in the parent, I expect.

The easiest way to fix this is to use a message to the parent process and do the load from there. In the parent process, you can use the `openAdvancedPreferences` helper and pass it the right 'advanced' tab directly. See https://dxr.mozilla.org/mozilla-central/rev/7d2e89fb92331d7e4296391213c1e63db628e046/browser/base/content/browser-data-submission-info-bar.js#69-75 for code you can almost copy and paste verbatim (though you'll want to use a different `origin` object, maybe)?

Note that this will open the prefs in a new tab, which the test will need to cope with, but as a bonus, it will do the right thing if the user already has about:preferences open in another tab.

gijsk:
Sorry this is such a minefield. We should write docs on this stuff. I'll try to find some time to do so.

recipe-client-addon/content/about-studies/common.js
Osmose: The [no-unused-vars](http://eslint.org/docs/rules/no-unused-vars) docs say you can use an `exported` comment block for this, but for reasons unknown to me this doesn't work with the eslint setup we have. I spent like an hour yesterday trying to get past it and failing, and this comment is a sign of my failure. :P

I can file a followup issue for `exported` comment blocks not working, would that be a good compromise?

recipe-client-addon/content/shield-content-frame.js
Osmose: I don't think we will.

recipe-client-addon/content/about-studies/common.js
gijsk: Yep, that wfm! :-)

Osmose:
Rebased and updated with feedback.

recipe-client-addon/test/browser/browser_about_studies.js
```diff
@@ -64,22 +64,27 @@ compose_task(
   }),
   withAboutStudies,
   async function testUpdatePreferencesOldOrganization(browser) {
-    browser.contentDocument.getElementById("shield-studies-update-preferences").click();
-    await BrowserTestUtils.waitForLocationChange(gBrowser);
-    await BrowserTestUtils.waitForEvent(browser.contentWindow, "load");
+    // We have to use gBrowser instead of browser in most spots since we're
+    // dealing with a new tab outside of the about:studies tab.
+    const tab = await BrowserTestUtils.switchTab(gBrowser, () => {
+      browser.contentDocument.getElementById("shield-studies-update-preferences").click();
```
gijsk: This will use a CPOW (cross process wrapper) now that this is really running in the child process. We're aiming to get rid of those.

Instead, you can use:
```js
ContentTask.spawn(browser, null, () => {
  content.document.getElementById("shield-studies-update-preferences").click();
});
```
which will run that function in the content process (as a new frame script, essentially).

recipe-client-addon/jar.mn
```diff
@@ -7,6 +7,7 @@
   lib/ (./lib/*)
   data/ (./data/*)
   skin/  (skin/*)
-  vendor/ (./vendor/*)
 % resource shield-recipe-client-content %content/
   content/ (./content/*)
+% resource shield-recipe-client-vendor %vendor/
```
gijsk: I wonder, does it work as-is if you add `contentaccessible=yes` flags to the end? If so, might as well do that here in preparation for bug 863246.

Osmose: It doesn't. `./mach build` fails with an error about `contentaccessible` not being supported for `resource` files, although my build is a few days old. It also breaks if I put it in the chrome manifest we use the the XPI.

Osmose:
@gijsk Updated with gratuitous use of `ContentTask`, did you want to check that out, since it was a bit larger than what you initially thought?

recipe-client-addon/test/browser/browser_about_studies.js
```diff
@@ -1,5 +1,6 @@
 "use strict";
 
+Cu.import("resource://testing-common/ContentTask.jsm", this);
```
gijsk: You don't need to manually import this. It's automatically available on test scopes, like BrowserTestUtils and friends.

recipe-client-addon/test/browser/browser_about_studies.js
```diff
@@ -116,57 +121,66 @@ compose_task(
...
-        study1Row.querySelector(".remove-button"),
-        "Active studies show a remove button"
-      );
-      is(
-        study1Row.querySelector(".study-icon").textContent.toLowerCase(),
-        "a",
-        "Study icons use the first letter of the study name."
-      );
+      await ContentTask.spawn(browser, [study1, study2, study3], async ([cStudy1, cStudy2, cStudy3]) => {
+        Components.utils.import("resource://testing-common/ContentTaskUtils.jsm");
```
gijsk: You shouldn't need to import this either, I think.

PR #903, opened 2017-07-21
Minify webpack dependencies and run React in prod mode

Osmose:
- Adds Babili to minify the add-on dependencies. This saves almost a megabyte in the XPI with the dependencies added in #902.
- Sets NODE_ENV to production so that React (added in #902) runs in production mode, which skips some unnecessary checks. Long-term it'd be nice to disable this when the `dev_mode` pref is set, but I don't have the time to investigate that right now.

I'm only flagging mythmon since this only touches the webpack config for the add-on and not the actual code. I informally ran the idea of minifying by @Mossop to make sure this won't be a problem for the mozilla-central check-in.

### DIFF
--- a/recipe-client-addon/package.json
+++ b/recipe-client-addon/package.json
@@ -12,16 +12,16 @@
   },
   "devDependencies": {
     "babel-eslint": "7.2.3",
-    "babili-webpack-plugin": "^0.1.2",
+    "babili-webpack-plugin": "0.1.2",
     "eslint": "3.19.0",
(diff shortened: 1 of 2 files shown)


recipe-client-addon/package.json
```diff
@@ -12,6 +12,7 @@
   },
   "devDependencies": {
     "babel-eslint": "7.2.3",
+    "babili-webpack-plugin": "^0.1.2",
```
mythmon: Pin this.

PR #904, opened 2017-07-21
Bug 1383221: Add tests that weren't being run to manifests.

Osmose:
ActionSandboxManager.jsm imports ShellService.jsm, which is not supported by
xpcshell, so it had to be ported to a mochitest.

### DIFF
--- a/recipe-client-addon/test/browser/browser_ActionSandboxManager.js
+++ b/recipe-client-addon/test/browser/browser_ActionSandboxManager.js
@@ -85,6 +85,9 @@ add_task(async function testError() {
 });
 
 add_task(async function testDriver() {
+  // The value returned by runAsyncCallback is cloned without the cloneFunctions
+  // option, so we can't inspect the driver itself since its methods will not be
+  // present. Instead, we inspect the properties on it available to the sandbox.
   const script = `
     registerAsyncCallback("testCallback", async function(normandy) {
       return Object.keys(normandy);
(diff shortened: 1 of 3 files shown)


recipe-client-addon/test/browser/browser_ActionSandboxManager.js
```diff
@@ -79,28 +79,23 @@ add_task(async function testError() {
...
     const referenceDriver = new NormandyDriver(manager);
-    equal(
-      sandboxDriver.constructor.name,
-      "NormandyDriver",
-      "runAsyncCallback passes a driver as the first parameter",
-    );
-    for (const prop in referenceDriver) {
-      ok(prop in sandboxDriver, "runAsyncCallback passes a driver as the first parameter");
+    for (const prop of Object.keys(referenceDriver)) {
+      ok(sandboxDriverKeys.includes(prop), `runAsyncCallback's driver has the "${prop}" property.`);
```
gijsk: Sorry, can you briefly explain why this needed to change?

Osmose: I should've called this out, my bad. The old test wasn't working, and I was confused about it, but figured it out while writing this comment. :D 

Values returned from an async callback in the sandbox are cloned back into the privileged context, but the driver has several functions, and when we clone the result, we don't pass the option to clone functions, so they get removed from the driver. Returning a list of keys that are available within the sandbox gets around this, and to some extent is a more accurate test anyway (since it explicitly tests what the sandbox can see, not what the value returned by the sandbox has).

gijsk: Ah, that makes sense. Adding a comment here as to why this is comparing keys rather than just checking the objects might be a good idea.

Osmose: Done!

PR #907, opened 2017-07-23
Ditch devtools

rehandalal:
I copied the devtools stuff over from the old control panel code but:
- if anyone is actually using the devtools
- if this integration even works. (I couldn't get it working).

So I figured we should ditch it for now and add it back in if necessary.

This also makes the logger only run in development environments.

small r?


recipe-server/client/control_new/index.js
```diff
@@ -20,33 +19,40 @@ const reducers = combineReducers({
...
+  thunk,
+];
 
-const store = createStore(reducers, reducers(undefined, { type: 'initial' }), compose(
-  applyMiddleware(
-    routerMiddleware,
-    thunk,
+if (DEVELOPMENT) {
+  middleware.push(
     createLogger({
```
andymikulski: I comment out the `createLogger` business pretty consistently while working on things; it really hits performance on a page that has a lot of redux actions (listings, history timeline). If you don't have a particular use for it, I'd vote to take out `redux-logger` all together.

PR #909, opened 2017-07-24
Disable Django system check for HTTPOnly CSRF cookies.

mythmon:
We are purposely disabling HTTPOnly on the CSRF cookies because we are reading this cookie in JS for the new React forms.

We found while trying to deploy this branch to dev.


mythmon:
TC failure is an unrelated lint error that is fixed on master.

PR #912, opened 2017-07-24
Disable querystring auth for AWS

rehandalal:
r?


recipe-server/normandy/settings.py
```diff
@@ -333,6 +333,7 @@ def RAVEN_CONFIG(self):
     AWS_ACCESS_KEY_ID = values.Value()
     AWS_SECRET_ACCESS_KEY = values.Value()
     AWS_STORAGE_BUCKET_NAME = values.Value()
+    AWS_QUERYSTRING_AUTH = values.BooleanValue(False)
```
mythmon: I think this should probably be a part of `Core`, where settings we don't expect to change live. Either that, or this new configurable setting needs documentation.

PR #914, opened 2017-07-25
New UI Tests

andymikulski:
- Tests for `common` React components
- Tests for `data` React components
- Tests for `page` React components
- A handful of tests for `extension` React components
  - The `CreateExtensionPage` tests are failing, but I'm not totally sure why.
- Updates enzyme from 2.4.1 to 2.9.1
- Lint and filename corrections
  - Renamed `QueryRecipes` to `QueryMultipleRecipes`
  - Renamed `extension/Listing.js` to `extension/ExtensionListing.js`
  - Renamed `recipe/Listing.js` to `recipe/RecipeListing.js`

This is about halfway through all the files that need to be tested, though it's pretty big already so I reckon it's ready for some feedback.


rehandalal:
Should probably rebase when #916 lands.

andymikulski:
I'm actually starting to hit `out of memory` errors when running `karma`.

Considering test build times are like 5 minutes or something crazy, I suggest we review the tests in this PR as-is, open another PR to investigate fixing/replacing karma entirely, and then come back to wrap the component tests after karma is settled.

@rehandalal @mythmon @Osmose Thoughts?

andymikulski:
Alright, I've added 90+ tests for the React components, and utils used on the front end. The tests are pretty much just "does this throw an error when we use it?", but for the sake of landing the `redesign` branch sooner than later I think this should suffice for now.

r?

mythmon:
TC failure is #924, and is fixed on master. We can probably ignore it here.

CircleCI failure is "process out of memory", which does not bode well. We may need to move up plans to replace karma, or switch to doing dev builds in karma.

rehandalal:
@mythmon could you do a pass on this?

PR #916, opened 2017-07-25
Fix getCurrentUser selector

### LINKED ISSUE #915: 'state is undefined' in users/selectors.js
In the user selector here we expect a state object:

https://github.com/mozilla/normandy/blob/redesign/recipe-server/client/control_new/state/app/users/selectors.js#L3

But when we call into it here, we look up a field on the state object (which could be undefined) and pass that in:

https://github.com/mozilla/normandy/blob/redesign/recipe-server/client/control_new/state/app/serviceInfo/selectors.js#L5

Looks like we should just be passing the state?

rehandalal:
Fixes #915 

r?


rehandalal:
/me nods. 

I wasn't sure if I'd be stepping on @andymikulski's toes but I just looked at his PR and he hasn't gotten there yet.

rehandalal:
@mythmon tests included.

PR #918, opened 2017-07-25
Upgrade to Webpack 3, and split vendor libs into a separate bundle.

### LINKED ISSUE #896: Upgrade webpack.
We are on a pretty old version of webpack. We should try to upgrade to the latest version (3.3.0 at the time of filing this issue).

Somewhat related: it would nice to find a way to bring the webpack build time down.

mythmon:
I didn't notice this have much an an influence on our build or test times, but it is still a good change, I think.

Fixes #896.


recipe-server/karma.conf.js
```diff
@@ -28,6 +29,10 @@ module.exports = function (config) {
     // The first config is for the control interface. It is only coincidence
     // that this config works for the action code/tests as well.
     webpack: Object.assign(WEBPACK_CONFIG[0], {
+      plugins: WEBPACK_CONFIG[0].plugins.filter(
+        // karma-webpack is not compatible with CommonsChunkPlugin, so exclude it
+        p => !(p instanceof webpack.optimize.CommonsChunkPlugin)
```
andymikulski: This is a bummer, since we lose the benefits when running tests. It's better to have working tests, though, I reckon.

mythmon: Yeah, I agree, since this is where we wanted it the most. We can watch this issue to see if it ever gets fixed: https://github.com/webpack-contrib/karma-webpack/issues/22. There may also be a workaround, but I didn't want to dig into it right now.

andymikulski:
After working through some experimenting and digging through the webpack docs, I've managed to drop my local build time from 25s -> 10-12s.

This commit:
- Adds `allChunks` props to `ExtractTextPlugin`s
- Replaces the vendor `CommonsChunk` with `DllReferencePlugin`
- Adds `build:app`, `build:vendor` npm scripts (`build` just calls them both)
- Change sourcemapping from `cheap-module` to `cheap-eval`
- Adds `vendors.webpack.config.js` for building the `vendors.bundle.js` file
- Adds UglifyJS (back) to the project, for compressing the vendor bundle (5mb -> 3mb)

With this change, my build runs considerably faster. Also, considering we're still running two builds (`control` and `control_new`), we can expect further speed increases once we phase out the current UI.

**However**, it seems that `karma` does not take advantage of this pre-made vendor bundling, so `npm run test` still takes forever. I have an idea on how to fix it, but I can push that up in another commit.

andymikulski:
Oh, wait, looks like I actually broke the site if you try to visit it. @mythmon Can you review what I've added here while I fix it?

recipe-server/package.json
```diff
@@ -109,14 +111,14 @@
...
     "redux-logger": "3.0.6",
     "redux-mock-store": "1.1.4",
-    "sass-loader": "4.0.1",
+    "sass-loader": "6.0.6",
     "style-loader": "0.13.1",
     "stylelint": "7.6.0",
     "stylelint-config-standard": "15.0.0",
     "stylelint-order": "0.2.2",
-    "webpack": "1.13.2",
+    "uglifyjs-webpack-plugin": "^0.4.6",
```
rehandalal: should be pinned.

rehandalal: also since we're already using babili for minification (and switched to it from uglify) i think we should continue using that.

andymikulski: I went with UglifyJS because I was aiming more for just condensing the vendor code than try to do any further transformations. I will see if we can use babili on non-babel code and swap it out accordingly.

andymikulski: Took me a bit, but I got babili to compile the vendors file.

For what it's worth, babili is slower _and_ its end file size is larger - ~4.4mb after ~2 min. UglifyJS gives us ~3mb after ~50s. I understand the point of using babili across the board, but IMO it makes more sense to use the process that gives us the better compression.

andymikulski:
I'm still running into an issue with the `render_bundle` command for the vendor js file. @mythmon Would you mind taking a look and seeing what I'm missing? I thought I had everything set in the last commit, but I guess not.

recipe-server/package.json
rehandalal: If we get better compression with uglify we should definitely go with it. I just thought @Osmose ended up also seeing better compression with babili. But maybe that was because he was specifically compressing ES6 content.

mythmon:
The issue with `render_bundle` was that to pull from a non-default webpack config, you have to say that's what you're doing. Pretty simple fix.

I also had to revert your change to the webpack devtool, because our CSP policy rejects `eval`.

mythmon:
Oh, I also rebased on redesign to pick up some bug fixes so I could be sure the new UI was working.

PR #919, opened 2017-07-26
Refactor StudyStorage.jsm to AddonStudies.jsm, add opt-out-study action

Osmose:
This is based off of #902; 4bbf9df and onward are the commits you should care about. ([Here's a convenient link to the commit range](https://github.com/mozilla/normandy/pull/919/files/85b47b33ac9c0cd02aa8ce88bfbae7ec71482e9b..5f048a2a06a2f0d0ce516970802ba035eb24b48f).)

When adding the remove buttons to about:studies, I realized that there was common logic around stopping studies that we wanted to share between the add-on and the action. So I refactored `StudyStorage.jsm` to have an API closer to `PreferenceExperiments.jsm`, with start and stop methods. It now handles the basic logic for installing and uninstalling add-on based studies.

This also adds an action implementation to the service that uses this API. I didn't bother with tests for the control interface changes, since it's being replaced. Once the new control interfaces merges we'll need to land admin UI for opt-out studies to it, or modify it if it already exists to support the argument changes.


Osmose:
Oh, I meant to document it, but you'll need to set `extensions.install.requireBuiltInCerts` to false to test this out locally, as the add-on manager won't install add-ons hosted on an HTTPS server if the cert isn't signed by a built-in root.

gijsk:
Github has now marked some of my comments as outdated, but it's not really clear to me why...

Anyway, re:  https://github.com/mozilla/normandy/pull/919/commits/4bbf9df5cd70339c87ef7a6789ddc7a4d4edee3a, I'm struggling with the commit message. I kind of assume that cloning isn't important as long as we *do* use the error message *before* the sandbox gets nuked, but that seems to be the opposite of what the commit message says. Maybe I'm misunderstanding what that change is about?

recipe-client-addon/bootstrap.js
```diff
@@ -25,6 +25,8 @@ this.shutdown = function(data, reason) {
   const log = LogManager.getLogger("bootstrap");
   let modules = [
     "lib/ActionSandboxManager.jsm",
+    "lib/Addons.jsm",
+    "lib/AddonStudies.jsm",
```
Osmose: Someday. SOMEDAY. I will write a lint rule for this.

recipe-client-addon/content/about-studies/shield-studies.js
```diff
@@ -78,12 +78,7 @@ class StudyList extends React.Component {
   }
 
   receiveRemoteValue(name, value) {
-    // Convert date strings to dates.
-    const studies = value.map(study => {
-      study.studyStartDate = stringToDate(study.studyStartDate);
```
Osmose: IndexedDB can store Date objects; we were stringifying them before because the JSONSchema validation didn't handle date objects (which really indicated that it wasn't quite the right fit for our use case, but I didn't want to have two different schema formats in our codebase).

Since we're no longer accepting the entire object to be stored in IndexedDB, we don't really need full schema validation, and thus don't need to stringify the dates.

recipe-client-addon/lib/AddonStudies.jsm
```diff
@@ -0,0 +1,260 @@
...
+XPCOMUtils.defineLazyModuleGetter(this, "Addons", "resource://shield-recipe-client/lib/Addons.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "LogManager", "resource://shield-recipe-client/lib/LogManager.jsm");
+
+Cu.importGlobalProperties(["fetch"]); /* globals fetch */
+
+this.EXPORTED_SYMBOLS = ["AddonStudies"];
+
+const KEY_TEMPDIR = "TmpD";
+const DB_NAME = "shield";
+const STORE_NAME = "addon-studies";
```
Osmose: Yes, as it seemed like a bad idea to reuse the database name as the store name. We don't need to migrate as we haven't shipped the old one yet.

recipe-client-addon/lib/NormandyDriver.jsm
```diff
@@ -182,11 +182,11 @@ this.NormandyDriver = function(sandboxManager) {
 
     // Study storage API
     studies: {
-      create: sandboxManager.wrapAsync(StudyStorage.create, {cloneArguments: true}),
-      update: sandboxManager.wrapAsync(StudyStorage.update, {cloneArguments: true}),
-      get: sandboxManager.wrapAsync(StudyStorage.get, {cloneInto: true}),
-      getAll: sandboxManager.wrapAsync(StudyStorage.getAll, {cloneInto: true}),
-      has: sandboxManager.wrapAsync(StudyStorage.has),
+      start: sandboxManager.wrapAsync(AddonStudies.start, {cloneArguments: true, cloneInto: true}),
```
Osmose: Ability to install add-ons was one of the things we brought up in the original security review of the sandbox, but it wasn't implemented or concrete at the time, so it may not have gotten enough scrutiny. I'll double-check.

Although, if the answer is no, we have a larger problem, as the entire point of opt-out studies is to install these add-ons. What would be the alternative, shipping the sandbox-level code in the add-on itself and just downloading the argument payload from Shield? Wouldn't that be susceptible to the same issues? Although there's a case to be made that logic errors in the sandbox code are more likely than errors in a static payload.

Osmose:
> Anyway, re: 4bbf9df, I'm struggling with the commit message. I kind of assume that cloning isn't important as long as we do use the error message before the sandbox gets nuked, but that seems to be the opposite of what the commit message says. Maybe I'm misunderstanding what that change is about?

Nah I wrote it backwards. Probably some long-term neurological disorder.

recipe-client-addon/lib/NormandyDriver.jsm
gijsk: Yeah... maybe I'm overly paranoid. But I'm assuming the installs don't warn in any way, so I wonder if we could restrict them in some way to mitigate risk (e.g. require that it's signed with the relevant key in the code here, irrespective of pref values, and/or require that it's downloaded from a specific domain, or something).

Osmose:
Rebased and updated. The security aspects of this PR are being discussed in an email thread with the relevant folks.

recipe-client-addon/lib/AddonStudies.jsm
```diff
@@ -216,19 +221,18 @@ this.AddonStudies = {
     }
 
     // Create temporary file to store add-on.
-    const file = FileUtils.getDir(KEY_TEMPDIR, []);
-    const random = Math.random().toString(36).replace(/0./, "").substr(-3);
-    const filename = `tmp-${random}.xpi`;
-    file.append(filename);
-    file.createUnique(Ci.nsIFile.NORMAL_FILE_TYPE, FileUtils.PERMS_FILE);
+    const path = FileUtils.getFile(KEY_TEMPDIR, ["study.xpi"]).path;
```
gijsk: Nit: `OS.File.join(OS.Constants.Path.tmpDir, "study.xpi")`

Osmose:
pauljt and ulfr have agreed that the workflow we have in place for reviewing and signing both recipes and add-ons is sufficient, and we're free to ship this. :D

@mythmon just need an approval on this and we'll be good to go.

PR #921, opened 2017-07-26
Fix attribute name in getUser selector fixes #920

### LINKED ISSUE #920: Error in getUser Selector
Here we attempt to access state.app.users:

https://github.com/mozilla/normandy/blob/redesign/recipe-server/client/control_new/state/app/users/selectors.js#L4

But it looks like it should be:

state.app.user
Object { items: Object }

from this reference:

https://github.com/mozilla/normandy/blob/redesign/recipe-server/client/control_new/state/app/reducer.js#L21

jaredkerim:


rehandalal:
gah! this should actually get fixed here:
https://github.com/mozilla/normandy/blob/redesign/recipe-server/client/control_new/state/app/reducer.js#L21

it should be
`users` not `user`.

sorry!!

rehandalal:
TC failure seems unrelated.

PR #924, opened 2017-07-26
Cachebust addon taskcluster base image to pick up Rust 1.19.

mythmon:
This fixes some Taskcluster breakage we saw.


Osmose:
@mythmon Looks like TC couldn't find the tag, does this need a re-run?

mythmon:
I guess I pushed the tag to dockerhub a little too late, and TC won the race. This is fixed now, and ready to land.

PR #926, opened 2017-07-26
New UI: Split recipe loading overlays

andymikulski:
- Swaps out `LoadingOverlay`s from `recipes/DetailPage` and `HistoryTimeline`
  - Previously, one request would lock up both sections of the Recipe Details page (the actual details, and the revision history)
  - With this approach, if the recipe loads before the history, it unlocks accordingly
- Adds `card-no-hover` CSS class to make up for the `noHovering` prop of `Card`s not working

### DIFF
--- a/recipe-server/client/control_new/components/common/LoadingOverlay.js
+++ b/recipe-server/client/control_new/components/common/LoadingOverlay.js
@@ -41,15 +41,15 @@ export class SimpleLoadingOverlay extends React.PureComponent {
 
 
 @connect(
-  (state, { requests }) => {
+  (state, { requestIds }) => {
     let isLoading;
 
     // If we're given one or more request IDs, check if at least one is in progress.
     // If nothing is given, simply check if _any_ request is in progress.
-    if (requests) {
-      let requestArray = requests;
+    if (requestIds) {
+      let requestArray = requestIds;
       if (!(requestArray instanceof Array)) {
-        requestArray = [requests];
+        requestArray = [requestIds];
       }
 
       isLoading = !!requestArray.find(reqId => isRequestInProgress(state, reqId));
@@ -65,11 +65,11 @@ export class SimpleLoadingOverlay extends React.PureComponent {
 export default class LoadingOverlay extends React.PureComponent {
   static propTypes = {
     isLoading: PropTypes.bool.isRequired,
-    requests: PropTypes.oneOf([PropTypes.arrayOf(PropTypes.string), PropTypes.string]),
+    requestIds: PropTypes.oneOf([PropTypes.arrayOf(PropTypes.string), PropTypes.string]),
   };
 
   static defaultProps = {
-    requests: null,
+    requestIds: null,
   };
 
   render() {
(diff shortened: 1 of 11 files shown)


andymikulski:
I've added a `requests` prop to the `LoadingOverlay` component, which accepts one or more request ID's and shows the overlay accordingly. I've swapped out the `Simple` overlays with `request` props elsewhere, too.

I really like the idea of being able to target the request we're waiting on, though I'm unsure of how maintainable using string ID's will be. For instance, the recipe listing page needs to do a bit of data handling before it knows what the actual request ID is (cause filters) - if that code changes in one spot and not the other, our LoadingOverlay will malfunction.

I don't know if this is the PR to address it, but we may want to brainstorm a more scalable solution to fetching request ID's.

recipe-server/client/control_new/components/common/LoadingOverlay.js
```diff
@@ -35,13 +38,35 @@ export class SimpleLoadingOverlay extends React.Component {
...
+    return {
+      isLoading,
+    };
+  },
 )
-export default class LoadingOverlay extends React.Component {
+export default class LoadingOverlay extends React.PureComponent {
   static propTypes = {
     isLoading: PropTypes.bool.isRequired,
+    requests: PropTypes.oneOf([PropTypes.arrayOf(PropTypes.string), PropTypes.string]),
```
rehandalal: I think this should be `requestIds`

recipe-server/client/control_new/less/ant-override.less
```diff
@@ -20,3 +20,10 @@
 .ant-menu-submenu-title {
   background: rgba(0, 0, 0, 0.025);
 }
+
+// ant Card's `noHovering` doesn't seem to do anything, so this class makes up for it.
```
rehandalal: I believe this is just until we upgrade Ant. I'd rather not include this and this will get dealt with once we upgrade Ant (which should happen shortly after this branch merges).

recipe-server/client/control_new/components/recipes/RecipeDetails.js
```diff
@@ -14,7 +14,7 @@ export default class RecipeDetails extends React.Component {
 
     return (
       <div className="recipe-details">
-        <Card key="recipe-details" title="Recipe">
+        <Card className="card-no-hover" key="recipe-details" title="Recipe">
```
rehandalal: i would maybe just add the `noHovering` in so it just works when we upgrade Ant.

rehandalal:
I agree we need a better solution for getting request ids but we can brainstorm that later.

recipe-server/client/control_new/less/ant-override.less
andymikulski: IMO, we should leave this and make removing it a piece of the Ant upgrade, then.

At the moment, this patch does what we need it to, given the version we're using. If a later patch is applied to update Ant, I don't see why the duties of that patch wouldn't be to also phase out things such as this. That way, the end user gets a consistent UI, both before and after the code changes.

rehandalal: sure.

andymikulski:
I've swapped `requests` for `requestIds`, and also changed the `card-no-hover` class name to `noHovering` to closer mimic the ultimate implementation. This should be ready for a final pass!

andymikulski:
TC failure appears unrelated.

mythmon:
TC failures are #806, and can be fixed by rebasing on master.

PR #929, opened 2017-07-27
Namespace listing and detail components

rehandalal:
quick r?


andymikulski:
I shouldn't have landed this, there are issues of undefined `Listing`s due to use of static members. I'm about to open another small PR which will include the fix.

PR #934, opened 2017-07-27
Add signatures to actions and verify them in the add-on

mythmon:
This adds signing to actions, including all the APIs and checks that exist for Recipes. This comes dangerously close to wanting some deduplication between the systems, but I think for now leaving them separate will be cleaner.

Nothing is checking this yet. I'm planning on adding on to this PR to include the signature checking bits in the Shield extension. The server side of this is ready to review now.


recipe-server/normandy/recipes/checks.py
```diff
@@ -55,28 +57,60 @@ def recipe_signatures_are_correct(app_configs, **kwargs):
...
 
 
-def recipe_signatures_use_good_certificates(app_configs, **kwargs):
+def action_signatures_are_correct(app_configs, **kwargs):
+    errors = []
+    try:
+        Action = apps.get_model('recipes', 'Action')
+        signed_actions = list(Action.objects.exclude(signature=None))
+    except (ProgrammingError, OperationalError, ImproperlyConfigured):
+        errors.append(Info('Could not retrieve actions', id=INFO_COULD_NOT_RETRIEVE_ACTIONS))
```
Osmose: Maybe include the error message as well?

recipe-server/normandy/recipes/checks.py
```diff
@@ -55,28 +57,60 @@ def recipe_signatures_are_correct(app_configs, **kwargs):
...
+                action_count = matching_actions.count()
+                count = recipe_count + action_count
+                msg = (f'{count} objects are signed with a bad cert. {exc.detail}. '
                        f'Certificate url is {url}')
+                if recipe_count:
+                    ids = ', '.join(str(r.id) for r in matching_recipes)
+                    msg += f'\n\tRecipe ids: {ids}'
+                if action_count:
+                    names = ', '.join(str(a.name) for a in matching_actions)
+                    msg += f'\n\tAction names: {names}'
```
Osmose: This is fine as it is, but a minor improvement might be to rely on their `__str__` representations to mark which is which, and just join the lists together at the start.

recipe-server/normandy/recipes/management/commands/update_action_signatures.py
```diff
@@ -0,0 +1,41 @@
...
+        if force:
+            actions_to_update = Action.objects.all()
+        else:
+            actions_to_update = self.get_outdated_actions()
+
+        count = actions_to_update.count()
+        if count == 0:
+            self.stdout.write('No out of date actions to sign')
+        else:
+            self.stdout.write('Signing {} actions:'.format(count))
```
Osmose: Rejoice! We have `f` strings now!

recipe-server/normandy/recipes/management/commands/update_action_signatures.py
```diff
@@ -0,0 +1,41 @@
...
+            for action in actions_to_update:
+                self.stdout.write(' * ' + action.name)
+            actions_to_update.update_signatures()
+
+    def get_outdated_actions(self):
+        outdated_age = timedelta(seconds=settings.AUTOGRAPH_SIGNATURE_MAX_AGE)
```
Osmose: It'd be neat if django-configurations had a timedelta value type. Oh well.

recipe-server/normandy/recipes/models.py
```diff
@@ -201,7 +201,7 @@ def update_signature(self):
             return
 
         logger.info(
-            f'Requesting signatures for recipes with ids [{self.id}] from Autograph',
+            f'Requesting signatures for recipe with id {self.id} from Autograph',
```
Osmose: Should signature be plural too?

recipe-server/normandy/recipes/models.py
```diff
@@ -477,15 +477,48 @@ def close(self):
         recipe.save()
 
 
+class ActionQuerySet(models.QuerySet):
+    def update_signatures(self):
```
Osmose: Should we wrap this in a transaction? That way we don't have partially-updated actions if something goes wrong.

recipe-server/normandy/recipes/models.py
```diff
@@ -517,19 +550,55 @@ def recipes_used_by(self):
     def __str__(self):
         return self.name
 
+    def canonical_json(self):
+        # Avoid circular import
+        from normandy.recipes.api.v1.serializers import ActionSerializer
+        data = ActionSerializer(self).data
+        return CanonicalJSONRenderer().render(data)
```
Osmose: We can use `normandy.base.utils.canonical_json_dumps` here instead of the renderer.

recipe-server/normandy/recipes/models.py
```diff
@@ -517,19 +550,55 @@ def recipes_used_by(self):
...
+
+        logger.info(
+            f'Requesting signatures for action named {self.name} from Autograph',
+            extra={'code': INFO_REQUESTING_ACTION_SIGNATURES, 'action_names': [self.name]}
+        )
+
+        signature_data = autographer.sign_data([self.canonical_json()])[0]
+        signature = Signature(**signature_data)
+        signature.save()
+        self.signature = signature
```
Osmose: I'm still mildly grumpy that this method saves the signature without saving this object it assigns the signature to, but I also still don't know a great alternative. Man.

recipe-server/normandy/recipes/models.py
```diff
@@ -517,19 +550,55 @@ def recipes_used_by(self):
...
-        super().save(*args, **kwargs)
+        if self.is_dirty(check_relationship=True):
+            dirty_fields = self.get_dirty_fields(check_relationship=True)
+            dirty_field_names = list(dirty_fields.keys())
+
+            if (len(dirty_field_names) > 1 and 'signature' in dirty_field_names
+                    and self.signature is not None):
+                # Setting the signature while also changing something else is probably
+                # going to make the signature immediately invalid. Don't allow it.
+                raise ValidationError('Signatures must change alone')
```
Osmose: There seems to be enough duplicated logic between the recipe signature implementation and the action one that we could share the logic with a model mixin, and maybe a shared queryset class/mixin. What do you think?

recipe-server/normandy/recipes/tests/test_commands.py
```diff
@@ -158,3 +176,35 @@ def test_it_resigns_signed_recipes_with_force(self, mocked_autograph):
         call_command('update_recipe_signatures', '--force')
         r.refresh_from_db()
         assert r.signature.signature is not 'old signature'
+
+
+@pytest.mark.django_db
+class TestUpdateActionSignatures(object):
```
Osmose: We could also use a test that it does not sign signatures that haven't expired yet.

recipe-server/normandy/recipes/tests/test_models.py
```diff
@@ -50,6 +52,40 @@ def test_recipes_used_by_empty(self):
...
+        action.save()
+        assert action.signature is not None
+        assert action.signature.signature == 'fake signature'
+
+    def test_canonical_json(self):
+        action = ActionFactory(name='test-action', implementation='console.log(true)')
+        # Yes, this is ugly, but it needs to compare an exact byte
+        # sequence, since this is used for hashing and signing
+        expected = (
+            f'{{'
```
Osmose: I don't see any formatting happening, we can probably just make these all normal strings.

recipe-server/normandy/recipes/api/fields.py
```diff
@@ -42,3 +42,11 @@ def get_url(self, obj, view_name, request, format):
             url = reverse(view_name, kwargs=kwargs, request=request, format=format)
             assert url[0] == '/'
             return settings.CDN_URL + url[1:]
+
+    def to_representation(self, value):
+        # HyperlinkedIdentityField demands a request when creating
+        # urls, but one is not always available. Instead of erroring
+        # out hard, fake it.
```
Osmose: When is this needed? If it's only for tests, can we use a setting or something to avoid doing this in production?

recipe-server/normandy/recipes/checks.py
```diff
@@ -55,28 +57,60 @@ def recipe_signatures_are_correct(app_configs, **kwargs):
     return errors
 
 
-def recipe_signatures_use_good_certificates(app_configs, **kwargs):
+def action_signatures_are_correct(app_configs, **kwargs):
```
Osmose: Do we write tests for our system checks? Should we?

recipe-server/normandy/recipes/api/v1/serializers.py
```diff
@@ -34,6 +58,19 @@ class Meta:
         ]
 
 
+class SignedActionSerializer(serializers.ModelSerializer):
+    signature = SignatureSerializer()
+    action = serializers.SerializerMethodField()
```
mythmon: No, because then it would look for a member called `action` on the Action instance, when instead we want the Action itself.

recipe-server/normandy/recipes/models.py
mythmon: I resisted doing this because it is only the second time we've done this, and it was already pretty hard to keep everything straight. I tried to limit the scope of the change to get it out faster and avoid breaking existing code.

recipe-server/normandy/recipes/api/fields.py
mythmon: When calling `Action#canonical_json()`, such as for signing, it isn't part of a request/response cycle, so there is no request.

recipe-server/normandy/recipes/models.py
mythmon: I tried this, and I ended up causing really weird import errors I couldn't track down, so I left it as is.

mythmon:
@Osmose This is ready for another round of review, and all the add-on bits are here too now.

@gijsk you'll probably only be interested in [the last commit](https://github.com/mozilla/normandy/pull/934/commits/c9726109d31dff7ea39500169554a7206e66160f).

recipe-server/normandy/base/utils.py
```diff
@@ -82,5 +82,7 @@ def sri_hash(data):
     :param data:
         Bytes-like object containing the data to hash.
     """
-    data_hash = b64encode(sha384(data).digest())
-    return 'sha384-' + data_hash.decode()
+    data_hash = b64encode(sha384(data).digest()).decode()
+    if url_safe:
+        data_hash = data_hash.replace('+', '-').replace('/', '_')
```
Osmose: We should use [`urlsafe_b64encode`](https://docs.python.org/3.6/library/base64.html#base64.urlsafe_b64encode) instead of rolling this ourselves.

I wonder if SRI supports url-safe variants of hashes? If so, maybe we should just always use the URL safe version?

recipe-server/normandy/recipes/models.py
```diff
@@ -560,7 +560,7 @@ def get_absolute_url(self):
         return reverse('action-detail', args=[self.name])
 
     def compute_implementation_hash(self):
-        return hashlib.sha1(self.implementation.encode()).hexdigest()
+        return sri_hash(self.implementation.encode(), url_safe=True)
```
Osmose: `sri_hash` is specifically for the purposes of subresource integrity, but it's being used here as a generic hash util. If we want a helper util for this, it should be separate from `sri_hash`, and maybe `sri_hash` should depend on it instead.

recipe-server/normandy/recipes/urls.py
```diff
@@ -28,7 +28,7 @@
     url(r'^api/v2/', include(v2_router.urls)),
     url(r'^api/v1/', include(v1_router.urls)),
     url(
-        r'^api/v1/action/(?P<name>[_\-\w]+)/implementation/(?P<impl_hash>[0-9a-f]{40})/$',
+        r'^api/v1/action/(?P<name>[_\-\w]+)/implementation/(?P<impl_hash>.*)/$',
```
Osmose: Would this match slashes as well? We probably want to keep this limited to base64 symbols.

recipe-server/normandy/recipes/api/fields.py
Osmose: I think I'm a little confused then; I would think that `reverse`, when not given a `request` parameter, returns a non-absolute URL. If this is used in `canonical_json` during signing, wouldn't the data being signed have a domain-less URL and thus not match up with the data we server the user, which has a domain as part of the URL?

mythmon:
I can't seem to reply inline to the comments about `reverse` and `canonical_json`, so I'll reply here.

I've pondered this for a bit, and I'm not actually sure why it works. In prod, we have `CDN_URL` set, and so we actually ignore the `request` argument completely, and use `CDN_URL` to generate the hostname.

When `CDN_URL` is not set, this should behave as you suggest, but it doesn't. If it did, the add-on's `mock_api` would have badly signed data, or wrong URLs, which it doesn't.

recipe-server/normandy/recipes/models.py
mythmon: Actually I intended to use `sri_hash` specifically because it matched the spec for subresource integrity. That matches well with what I wanted (integrity of a subresource) and is the most easily available spec I know of that includes the algorithm of the hash in the string. I think that it's use is appropriate here.

recipe-server/normandy/base/utils.py
mythmon: I'm pretty sure that SRI doesn't support url safe base64. At least, [the spec](https://www.w3.org/TR/SRI/) doesn't talk about it.

recipe-client-addon/lib/NormandyApi.jsm
```diff
@@ -145,20 +149,65 @@ this.NormandyApi = {
...
+      const dataToHash = new TextEncoder().encode(responseText);
+      hasher.update(dataToHash, dataToHash.length);
+      const useBase64 = true;
+      const hash = hasher.finish(useBase64).replace(/\+/g, '-').replace(/\//g, '_');
+      if (hash !== expectedHash) {
+        throw new Error(
+          `Failed to fetch action implementation for ${action.name}: ` +
+          `Integrity hash does not match content. Expected ${expectedHash} got ${hash}.`
+        );
+      }
```
Osmose: Yes. I understand this. It is totally a thing I know how it works.

recipe-client-addon/lib/NormandyApi.jsm
```diff
@@ -145,20 +149,65 @@ this.NormandyApi = {
...
-    return res.json();
+  async fetchActions(filters = {}) {
+    return this.fetchSignedObjects('action', filters);
   },
 
   async fetchImplementation(action) {
-    const response = await fetch(action.implementation_url);
-    if (response.ok) {
-      return response.text();
+    const implementationUrl = this.absolutify(action.implementation_url);
```
Osmose: Do the URLs on the server not have domains, then? I think we decided a while ago to include domains in all our URL fields, or at least I prefer that (so clients don't have to do absolutify logic like this).

recipe-client-addon/lib/NormandyApi.jsm
```diff
@@ -82,19 +82,23 @@ this.NormandyApi = {
     return this.absolutify(url);
   },
 
-  async fetchRecipes(filters = {enabled: true}) {
-    const signedRecipesUrl = await this.getApiUrl("recipe-signed");
-    const recipesResponse = await this.get(signedRecipesUrl, filters);
-    const rawText = await recipesResponse.text();
-    const recipesWithSigs = JSON.parse(rawText);
+  async fetchSignedObjects(type, filters) {
```
Osmose: I think this is not the best level of reuse; having a function that verified objects with signatures, and maybe a shorthand for fetching and parsing JSON from the API, would be a little cleaner (less state to keep in your head when trying to understand the function, etc.). 

I'm haven't yet found a great articulation for that feeling though, so I wouldn't call it a blocker.

recipe-server/normandy/recipes/models.py
Osmose: Ah, got it. Can you leave a comment explaining that an SRI-style hash is intentional, then?

recipe-client-addon/lib/NormandyApi.jsm
```diff
@@ -82,19 +82,23 @@ this.NormandyApi = {
...
+    const signedObjectsUrl = await this.getApiUrl(`${type}-signed`);
+    const objectsResponse = await this.get(signedObjectsUrl, filters);
+    const rawText = await objectsResponse.text();
+    const objectsWithSigs = JSON.parse(rawText);
```
gijsk: The .parse() can throw if this isn't valid JSON. Should we cope and throw a specific error, or just let the original throw propagate?

recipe-client-addon/lib/NormandyApi.jsm
```diff
@@ -82,19 +82,23 @@ this.NormandyApi = {
...
 
-    const verifiedRecipes = [];
+    for (const objectWithSig of objectsWithSigs) {
+      const {signature, x5u} = objectWithSig.signature;
+      const object = objectWithSig[type];
 
-    for (const {recipe, signature: {signature, x5u}} of recipesWithSigs) {
-      const serialized = CanonicalJSON.stringify(recipe);
+      const serialized = CanonicalJSON.stringify(object);
       if (!rawText.includes(serialized)) {
```
gijsk: Can you remind me why this isn't a plain equality check but a substring one? Can we add a comment about this? In general, this feels a bit odd because we're trying to verify the hashed thing matches the canonical JSON, and we pass in the serialized string for this purpose, not the raw text.

recipe-client-addon/lib/NormandyApi.jsm
```diff
@@ -145,20 +149,65 @@ this.NormandyApi = {
...
+    if (!response.ok) {
+      throw new Error(
+        `Failed to fetch action implementation for ${action.name}: ${response.status}`
+      );
+    }
+    const responseText = await response.text();
+
+    // Try to verify integrity of the implementation text. If no
+    // integrity value is found, skip. If the integrity value doesn't
+    // match thte content or uses an unknown algorithm, fail.
```
gijsk: ththe!

recipe-client-addon/lib/NormandyApi.jsm
mythmon: Our public APIs tend to, but for various reasons our dev servers don't always. We have this utility already, so it probably works fine to keep it.

recipe-client-addon/lib/NormandyApi.jsm
mythmon: We could leave this check out entirely, and it would be fine because it still wouldn't match the signatures. This is mostly a benefit to give better error messages to developers, especially in tests and QA.

The reason it is a substring check is because `rawText` includes both the object that is signed and the signature for that object. It isn't required that the entire respons be in canonical json format, only the object that is signed, which is what `serialized` is a representation of.

I'll add a comment.

mythmon:
Added two commits to handle the feedback, and fix CI problems. Then I rebased on master to pick up the changes from #953.

mythmon:
Tagging @rehandalal  for review instead of @Osmose, since @Osmose is supposed to be on PTO now.

mythmon:
@jvehent since this is a security fix, can you look over this, or have someone else?

recipe-client-addon/lib/NormandyApi.jsm
```diff
@@ -145,20 +153,67 @@ this.NormandyApi = {
...
+    // Try to verify integrity of the implementation text.  If the
+    // integrity value doesn't match the content or uses an unknown
+    // algorithm, fail.
+
+    // Get the last non-empty portion of the url path, and split it
+    // into two to get the aglorithm and hash.
+    const parts = implementationUrl.pathname.split("/");
+    const lastNonEmpty = parts.filter(p => p !== "").slice(-1)[0];
+    const [algorithm, ...hashParts] = lastNonEmpty.split("-");
+    const expectedHash = hashParts.join("-");
```
jvehent: I wonder if there's really added value in storing the hash in the URL. It requires extra parsing, and we could just store the hash in the action JSON document, which, I think, would be much cleaner.

recipe-client-addon/lib/NormandyApi.jsm
```diff
@@ -145,20 +153,67 @@ this.NormandyApi = {
...
+
+    if (algorithm !== "sha384") {
```
jvehent: If you're only implement `sha384`, why store the name of the alrgorithm at all?
You could also take then `len()` of the hash to determine which algorithm was used. SHA384 hashes are 96 bytes long.

recipe-client-addon/test/unit/mock_api/api/v1/action/index.json
```diff
@@ -15,12 +15,44 @@
...
+      },
+      "required": [
+        "name",
+        "description",
+        "addonUrl"
+      ],
+      "title": "Enroll a user in an opt-out SHIELD study",
+      "type": "object"
+    },
+    "implementation_url": "https://localhost:8443/api/v1/action/opt-out-study/implementation/sha384-HM_avYcD00o27ufwU1V7PIBtiuMAXML6MMwlYrDEqDX-XzGVuOfL52RCM680JExN/",
```
jvehent: As mentioned above, I think it would be cleaner to split this into
```json
{
"implementation_url": "https://localhost:8443/api/v1/action/opt-out-study/implementation/foo",
"implementation_hash": "HM_avYcD00o27ufwU1V7PIBtiuMAXML6MMwlYrDEqDX-XzGVuOfL52RCM680JExN",
}
```
which, in passing, doesn't really need to be base64url, since it's not in a url.

recipe-server/normandy/recipes/management/commands/update_recipe_signatures.py
```diff
@@ -41,7 +41,7 @@ def handle(self, *args, force=False, **options):
         if count == 0:
             self.stdout.write('No disabled recipes to unsign')
         else:
-            self.stdout.write('Unsigning {} disabled recipes:'.format(count))
+            self.stdout.write(f'Unsigning {count} disabled recipes:')
```
jvehent: "unsigning"?
I know it's not strictly related to this patch, but could you explain what this is meant to provide?

recipe-server/normandy/base/utils.py
```diff
@@ -70,7 +70,7 @@ def filter_m2m(qs, field, values):
     return qs
 
 
-def sri_hash(data):
+def sri_hash(data, url_safe=False):
```
jvehent: Would calling this a "SRI Hash" cause confusion later on? This is simply a hash, and we're not exactly in the context of subresource integrity here.

recipe-client-addon/lib/NormandyApi.jsm
mythmon: As a part of this change, I changed from sha1 to sha384 hashes. Changing the algorithm was rather painful. I'd like to make this extensible so it is easier to change should we need to do it again.

recipe-server/contract-tests/test_api.py
```diff
@@ -117,6 +117,27 @@ def test_recipe_signatures(conf, requests_session):
         signing.verify_x5u(url)
 
 
+def test_action_signatures(conf, requests_session):
```
mythmon: Contract tests are designed to validate existing servers, not as unit tests. As such there isn't anyway to use them to force a invalid hash or signature.

recipe-client-addon/test/unit/mock_api/api/v1/action/index.json
mythmon: The implementation url must have some unique identifier in it, for caching purposes. Hashes make excellent choices for this. We could also duplicate the hash and put it in another field, but I'm not sure the duplication makes sense.

recipe-server/normandy/recipes/management/commands/update_recipe_signatures.py
mythmon: When a recipe is disabled (turned off for Firefox users), we delete the signature for it, so that it can't be executed easily from the API anymore.

recipe-server/normandy/base/utils.py
mythmon: I'd argue that we are in the context of subresource integrity here. This usage isn't covered by the SRI spec, but we are using the hashes in the same way for the same purpose: verifying that we can trust remote servers to serve the correct data. Although it is true that this is simply a hash, it's purpose is something specific that gives more meaning than just a hash.

recipe-client-addon/lib/NormandyApi.jsm
```diff
@@ -145,20 +153,67 @@ this.NormandyApi = {
...
+    }
+
+    // verify integrity hash
+    const hasher = Cc["@mozilla.org/security/hash;1"].createInstance(Ci.nsICryptoHash);
+    hasher.init(hasher.SHA384);
+    const dataToHash = new TextEncoder().encode(responseText);
+    hasher.update(dataToHash, dataToHash.length);
+    const useBase64 = true;
+    const hash = hasher.finish(useBase64).replace(/\+/g, "-").replace(/\//g, "_");
+    if (hash !== expectedHash) {
```
mythmon: Using proper crypto practices here might be a good idea, though I hesitate to bring in another library for this. I also expect that that particular  library won't work well in Firefox.

I don't think the savings from not reencoding in base64 are useful, because we would have to decode from base64 in order to do binary comparison. Either way we are doing a conversion.

PR #935, opened 2017-07-28
Remove extra nav menu items

rehandalal:
Since these don't actually do anything, we should get rid of them.

small r?


andymikulski:
@rehandalal  What's the level of effort to add `action name` to the existing filters on the back end? It'd be great to land that functionality in so we can tweak these links instead of removing them all together.

rehandalal:
More effort than I can give this right now. I don't think it's pressing and/or should block this branch landing. So we should remove these links for now and can re-add them in later.

PR #936, opened 2017-07-28
call recipe runner async on first run, so recipes have a chance to fi…

rhelmer:
…lter


Osmose:
I cherry-picked the test fixes from #938 to this branch, and dropped an extra fix to handle the switch to async here. I think that should get the tests passing, although this probably needs a re-review as a result.

mythmon:
This looks good to me, but this is more than a backport now, so I'm tagging @gijsk for review as well.

Issue #937, opened 2017-07-28
Abort opt-out studies / preference studies if the user has opted-out


Osmose:
#741 adds a global opt-out preference, and #746 will add functions for drivers to be able to read preferences. I'm filing this to make sure we don't forget to actually use these in the actions and abort participation in experiments if the user has signaled that they do not want to participate.

PR #938, opened 2017-07-28
Add global opt-out study preference to add-on

Osmose:
This adds a new preference `app.shield.optoutstudies.enabled` that will track whether a user has opted-out of participating in opt-out Shield studies. A checkbox is injected into about:preferences (only the new organization) that controls the pref. See [bug 1377192](https://bugzilla.mozilla.org/show_bug.cgi?id=1377192) for the spec.

This also fixes our preferences so that they're set as default values immediately on startup instead of as user-set values. Existing users will still have user-set values, and we should follow up soon with a migration of some kind to get them on default values if possible. But switching now will help existing users, and also affect the new preference being added by this PR, so it's worth doing.

I cleared shipping this as en-US only with flod a few weeks ago, as long as we do not display the checkbox to users in non en-* locales. Accordingly, we will not run any studies outside of those locales until we hook up translation, so that users can properly opt-out of feature studies.

I filed #937 separately to actually use this preference in the `opt-out-study` and `preference-experiment` actions.


recipe-client-addon/lib/ShieldPreferences.jsm
```diff
@@ -0,0 +1,133 @@
...
+const FHR_UPLOAD_ENABLED_PREF = "datareporting.healthreport.uploadEnabled";
+const OPT_OUT_STUDIES_ENABLED_PREF = "app.shield.optoutstudies.enabled";
+
+/**
+ * Handles Shield-specific preferences, including their UI.
+ */
+this.ShieldPreferences = {
+  async init() {
+    // Disabled outside of en-* locales temporarily (bug 1377192).
+    // Disabled when MOZ_DATA_REPORTING is false since the FHR UI is also hidden
```
gijsk: Should we default the pref to false if MOZ_DATA_REPORTING is false?

recipe-client-addon/lib/ShieldPreferences.jsm
```diff
@@ -0,0 +1,133 @@
...
+      checkbox.setAttribute("preference", OPT_OUT_STUDIES_ENABLED_PREF);
+      checkbox.setAttribute("checked", Services.prefs.getBoolPref(OPT_OUT_STUDIES_ENABLED_PREF));
+      checkbox.setAttribute("disabled", !Services.prefs.getBoolPref(FHR_UPLOAD_ENABLED_PREF));
+      hContainer.appendChild(checkbox);
+
+      const viewStudies = this.refs.viewStudies = document.createElementNS(XUL_NS, "label");
+      viewStudies.setAttribute("id", "viewShieldStudies");
+      viewStudies.textContent = "View Firefox Studies";
+      viewStudies.classList.add("learnMore", "text-link");
+      viewStudies.addEventListener("click", this);
```
gijsk: Don't do this, just set the href attribute? See e.g. this: https://dxr.mozilla.org/mozilla-central/source/browser/components/preferences/in-content-new/privacy.xul#499-501

recipe-client-addon/lib/ShieldPreferences.jsm
```diff
@@ -0,0 +1,133 @@
...
+      const parent = this.document.getElementById("submitHealthReportBox").closest("vbox");
+      parent.appendChild(this.refs.container);
+
+      const preferences = this.document.getElementById("privacyPreferences");
+      preferences.appendChild(this.refs.optOutPref);
+      preferences.appendChild(this.refs.fhrPref);
+    }
+
+    remove() {
+      this.refs.container.remove();
```
gijsk: So I'm nervous here because `fhrPref` has a change listener which points to this object, it's a JSM so it stays alive ~forever, and because the object keeps DOM references it will keep the window alive.

Additionally... I don't see anything calling `.remove()` here? So I guess we're hoping the cycle collector works well here? Should we be more defensive about ensuring this doesn't leak?

recipe-client-addon/lib/ShieldPreferences.jsm
```diff
@@ -0,0 +1,133 @@
...
+      case "advanced-pane-loaded":
+        if (!Services.prefs.getBoolPref("browser.preferences.useOldOrganization", false)) {
+          const checkbox = new this.OptOutStudyCheckbox(subject.document);
+          checkbox.inject();
+        }
+        break;
+      // If the FHR pref changes, set the opt-out-study pref to the value it is changing to.
+      case NS_PREFBRANCH_PREFCHANGE_TOPIC_ID:
+        if (data === FHR_UPLOAD_ENABLED_PREF) {
+          const fhrUploadEnabled = Services.prefs.getBoolPref(FHR_UPLOAD_ENABLED_PREF);
```
gijsk: Note that we're not running this on startup, so if this runs against an existing profile where the pref has been flipped, or it was flipped while Firefox was shut down, we wouldn't ever change the shield pref.

recipe-client-addon/lib/ShieldPreferences.jsm
```diff
@@ -0,0 +1,133 @@
...
+
+      const hContainer = this.refs.hContainer = document.createElementNS(XUL_NS, "hbox");
+      hContainer.setAttribute("align", "center");
+      container.appendChild(hContainer);
+
+      const checkbox = this.refs.checkbox = document.createElementNS(XUL_NS, "checkbox");
+      checkbox.setAttribute("id", "optOutStudiesEnabled");
+      checkbox.setAttribute("label", "Allow Firefox to install and run studies");
+      checkbox.setAttribute("preference", OPT_OUT_STUDIES_ENABLED_PREF);
+      checkbox.setAttribute("checked", Services.prefs.getBoolPref(OPT_OUT_STUDIES_ENABLED_PREF));
```
gijsk: I am surprised you have to init this here first. If we insert the preference element before the checkbox, does that help?

recipe-client-addon/lib/ShieldPreferences.jsm
Osmose: I tried that and got this error:

```
Security Error: Content at moz-nullprincipal:{14b21f4f-1fc8-8046-b8b9-d2b3376239c0} may not load or link to about:studies.
Error: Cannot open a about: link using the text-link binding.
```

Is `about:preferences` running with a null principal, even though it's in the parent process?

recipe-client-addon/lib/ShieldPreferences.jsm
Osmose: `remove` is a leftover I forgot to remove.

> So I'm nervous here because fhrPref has a change listener which points to this object, it's a JSM so it stays alive ~forever, and because the object keeps DOM references it will keep the window alive.

The references are to the OptOutStudyCheckbox object, not the JSM itself. But maybe that's not a significant difference?



> So I guess we're hoping the cycle collector works well here? Should we be more defensive about ensuring this doesn't leak?

Yeah. Would adding an event listener to the `unload` event for the `document` element be a good way to detect when it's time to clean up?

recipe-client-addon/lib/ShieldPreferences.jsm
gijsk: No, it's not, but the XBL binding for `label.text-link[href]` does a manual security check against a null principal ( http://searchfox.org/mozilla-central/rev/09c065976fd4f18d4ad764d7cb4bbc684bf56714/toolkit/content/widgets/text.xml#315-329 ), which is unfortunate. You can add `useoriginprincipal` as well, and then it should work. https://bugzilla.mozilla.org/show_bug.cgi?id=315166 added the original restriction because in some cases, the URLs are controlled by people other than us. That isn't the case here, so we should be OK with the added attribute.

recipe-client-addon/lib/ShieldPreferences.jsm
gijsk: Yep, `unload` would be the usual thing. It's possible I'm being overly paranoid, though... if there are leaks, in theory they should show up on a debug build when running an automated test that would exercise this code. If nothing holds a reference to the `OptOutStudyCheckbox` object besides the page itself (if anything), it's possible that the cycle collector deals with this for us.

Osmose: I talked to @Mossop briefly about this and he agrees that the cycle collector handles this, and debug build tests should catch this, so we're probably fine. I'd prefer to not add extra cleanup handling if possible, shall we leave this as-is (besides removing the `remove` method, of course)?

Osmose:
This is ready for another pass.

@msujaws I can't assign you as a reviewer since you're not in the Mozilla org or on the team I invited you to, but could you r? this PR?

recipe-client-addon/lib/ShieldPreferences.jsm
gijsk: Yep, that wfm!

recipe-client-addon/bootstrap.js
```diff
@@ -4,17 +4,53 @@
...
+Cu.import("resource://gre/modules/Services.jsm");
 Cu.import("resource://gre/modules/XPCOMUtils.jsm");
 
 XPCOMUtils.defineLazyModuleGetter(this, "LogManager",
   "resource://shield-recipe-client/lib/LogManager.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "ShieldRecipeClient",
   "resource://shield-recipe-client/lib/ShieldRecipeClient.jsm");
 
+const DEFAULT_PREFS = {
+  "extensions.shield-recipe-client.api_url": ["char", "https://normandy.cdn.mozilla.net/api/v1"],
```
msujaws: We shouldn't need to include the type of the pref here. In this.startup() we can switch off of `typeof value`.

recipe-client-addon/lib/ShieldPreferences.jsm
```diff
@@ -0,0 +1,133 @@
...
+const XUL_NS = "http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul";
+const NS_PREFBRANCH_PREFCHANGE_TOPIC_ID = "nsPref:changed"; // from modules/libpref/nsIPrefBranch.idl
+const FHR_UPLOAD_ENABLED_PREF = "datareporting.healthreport.uploadEnabled";
+const OPT_OUT_STUDIES_ENABLED_PREF = "app.shield.optoutstudies.enabled";
+
+/**
+ * Handles Shield-specific preferences, including their UI.
+ */
+this.ShieldPreferences = {
+  async init() {
```
msujaws: This function doesn't need to be async. The caller users await on it, but since this function doesn't have any awaits internal to it and the caller doesn't treat the promise that this function returns as anything special, we can just make this a plain function.

recipe-client-addon/lib/ShieldPreferences.jsm
```diff
@@ -0,0 +1,133 @@
...
+      optOutPref.setAttribute("id", OPT_OUT_STUDIES_ENABLED_PREF);
+      optOutPref.setAttribute("name", OPT_OUT_STUDIES_ENABLED_PREF);
+      optOutPref.setAttribute("type", "bool");
+
+      // Weirdly, FHR doesn't have a <preference> element on the page, so we create it.
+      const fhrPref = this.refs.fhrPref = document.createElementNS(XUL_NS, "preference");
+      fhrPref.setAttribute("id", FHR_UPLOAD_ENABLED_PREF);
+      fhrPref.setAttribute("name", FHR_UPLOAD_ENABLED_PREF);
+      fhrPref.setAttribute("type", "bool");
+      fhrPref.addEventListener("change", this);
```
msujaws: I'm just a little worried that this object will live longer than it needs, especially since this.refs holds references to all of these elements. Can you inline the event handler here and change OptOutStudyCheckbox to not use this.refs?

recipe-client-addon/lib/ShieldRecipeClient.jsm
```diff
@@ -80,20 +68,16 @@ this.ShieldRecipeClient = {
       log.error("Failed to initialize preference experiments:", err);
     }
 
+    try {
+      await ShieldPreferences.init();
```
msujaws: `await` isn't needed here

Osmose:
Feedback'd.

@gijsk Should I dismiss your rejection since @msujaws approved?

msujaws:
I don't know how to see Gijs' comments except for the one about startup and pre-existing profiles. You should still make changes or take into account that feedback.

Osmose:
> I don't know how to see Gijs' comments except for the one about startup and pre-existing profiles. You should still make changes or take into account that feedback.

You can click "Show outdated" next to the grey boxes around that comment to see his old review comments. I already applied the feedback from his comments before asking you to review, but Github is still storing his rejection from the previous review. I just didn't know if he wanted to do another pass himself or just dismiss his old rejection so that it doesn't disable the UI for merging.

recipe-client-addon/lib/ShieldPreferences.jsm
```diff
@@ -0,0 +1,119 @@
...
+    optOutPref.setAttribute("name", OPT_OUT_STUDIES_ENABLED_PREF);
+    optOutPref.setAttribute("type", "bool");
+
+    // Weirdly, FHR doesn't have a <preference> element on the page, so we create it.
+    const fhrPref = doc.createElementNS(XUL_NS, "preference");
+    fhrPref.setAttribute("id", FHR_UPLOAD_ENABLED_PREF);
+    fhrPref.setAttribute("name", FHR_UPLOAD_ENABLED_PREF);
+    fhrPref.setAttribute("type", "bool");
+    fhrPref.addEventListener("change", () => {
+      checkbox.disabled = !Services.prefs.getBoolPref(FHR_UPLOAD_ENABLED_PREF);
```
gijsk: Note that this will still keep a ref for `this` because the callback is closured on `this` - I'm not sure if by "inline", @msujaws meant using an attribute, which I think would also work, ie:

```js
fhrPref.setAttribute("onchange", `document.getElementById("optOutStudiesEnabled").disabled = Services.prefs.getBoolPref("${FHR_UPLOAD_ENABLED_PREF}")`);
```

But as noted earlier, I'm probably overly paranoid. If this works on a debug build when exercised in a test, and doesn't show up as leaking, I think we're good.

msujaws: Yes, by "inline" I indeed meant an attribute. Thank you for making it clearer :)

msujaws: I should add that I'm paranoid and skeptical of this code too. I would make the change even if our testing doesn't show it because it's a common code smell and what doesn't leak today may leak in the future if someone makes an innocent-looking change.

Osmose: The class is now just a method, though, and `this` is now the `ShieldPreferences` object, and will get unloaded when the add-on shuts down (and should live otherwise). Or am I missing something else that will leak?

Also, would not using an arrow function work too, to avoid saving the reference to `this`?

msujaws: Well, because you don't reference `this` I don't think that will be a problem. But since you don't use `this`, you might as well just use a plain function since your code doesn't use the semantics of arrow functions. But this does hold a reference to `checkbox`. That's why Gijs' snippet uses `document.getElementById`.

gijsk: Oh, OK, I hadn't looked at more context, sorry. So, yeah, this is about lifetimes of things. You're right that the JSM should stay loaded pretty much ~forever (though it would get unloaded if the add-on is disabled). The worry I had originally was that the JSM would keep references to the document, thus meaning the tab's document (and with that likely the entire parent tree ie browser window) would live as long as the jsm would live, because the jsm had permanent-ish references to the document (or ones that were cyclical via event handlers). That's now no longer the case, so I think this is fine and shouldn't leak. The document has references to the JSM, but that's fine because it shouldn't prevent the document from being GC'd, because although the document can reach the JSM, the JSM can't reach the document.

Osmose: Aha, got it, I think. So then would this be a good final version?

```js
fhrPref.addEventListener("change", function() {
  document.getElementById("optOutStudiesEnabled").disabled = Services.prefs.getBoolPref(FHR_UPLOAD_ENABLED_PREF);
});
```

msujaws:
fhrPref.addEventListener("change", function(event) {
  event.target.ownerDocument.getElementById("optOutStudiesEnabled").disabled
= Services.prefs.getBoolPref(FHR_UPLOAD_ENABLED_PREF);
});

recipe-client-addon/lib/ShieldPreferences.jsm
msujaws: `fhrPref.addEventListener("change", function(event) {
    event.target.ownerDocument.getElementById("optOutStudiesEnabled").disabled = Services.prefs.getBoolPref(FHR_UPLOAD_ENABLED_PREF);
});`

recipe-client-addon/lib/ShieldPreferences.jsm
mythmon: I believe that we set the normandy pref during start up of the add-on, in bootstrap.js.

PR #939, opened 2017-07-29
Fix #746: Add driver functions for reading preferences.

### LINKED ISSUE #746: Add driver functions for reading preferences
The opt-out study action will need to check certain preferences to determine if it should install the study add-on or not. We'll need to add an API to the driver for reading preferences.

I think mirroring [Preferences.jsm](https://dxr.mozilla.org/mozilla-central/source/toolkit/modules/Preferences.jsm) on the driver is our best bet:

```js
normandy.preferences.get('foo.bar.baz', 'default_value');
normandy.preferences.has('foo.bar.baz')
normandy.preferences.isSet('foo.bar.baz');
```

@crankycoder How does the API above look to you?

Osmose:
I opted to leave out checked whether a preference has a user-set value because I couldn't think of a reason we'd need that in an action.

I also opted to use a type-specific API similar to `nsIPrefBranch` to enforce type expectations (especially given recent issues around mis-typed preference values).


rehandalal:
needs conflict resolution

PR #942, opened 2017-07-30
Fix #940: Mark studies as ended if their add-ons are uninstalled.

### LINKED ISSUE #940: Monitor in-progress study add-ons for removal
Splitting this off from #749 since it already has a sizeable PR. There's two major cases where we need to detect if an in-progress study add-on has been removed:

- If we detect the add-on in question is uninstalled (using the [AddonManager.addAddonListener](https://developer.mozilla.org/en-US/Add-ons/Add-on_Manager/AddonManager#addAddonListener()) method), we should mark the study as inactive and record the time the study ended. We'll need to register these listeners on startup, similar to what we do for preference experiments.
- If, on startup (before we fetch recipes) we notice that the addon is no longer installed, we should mark the study as inactive and record the time the study ended.

Osmose:
This is based off #919; 3b45b72 is the new commit to review.

@rhelmer The tests pass on this PR, but when I test it out manually with an XPI, my uninstall listener in `AddonStudies.jsm` is not called when I uninstall a study add-on via `about:addons`. Is there a reason why a listener for add-on uninstalls would not be called?

The TC run for this PR should have a task for building the XPI; check the artifacts there for a copy of the XPI if you want to test it out for yourself. You can use the following line in the browser console to create a new study entry in `about:studies`:

```js
// recipeId must be unique for each new study. 
Cu.import("resource://shield-recipe-client/lib/AddonStudies.jsm", {}).AddonStudies.start({recipeId: 9000, name: "My Study", description: "Don't worry about it", addonUrl: "https://github.com/Osmose/jsonfile-example/releases/download/v1/jsonfile-example.xpi"})
```

### DIFF
--- a/recipe-client-addon/lib/AddonStudies.jsm
+++ b/recipe-client-addon/lib/AddonStudies.jsm
@@ -88,6 +88,18 @@ function getStore(db) {
   return db.objectStore(STORE_NAME, "readwrite");
 }
 
+/**
+ * Mark a study object as having ended. Modifies the study in-place.
+ * @param {IDBDatabase} db
+ * @param {Study} study
+ */
+async function markAsEnded(db, study) {
+  study.active = false;
+  study.studyEndDate = new Date();
+  await getStore(db).put(study);
+  Services.obs.notifyObservers(study, STUDY_ENDED_TOPIC);
+}
+
 this.AddonStudies = {
   /**
    * Test wrapper that temporarily replaces the stored studies with the given
@@ -132,9 +144,7 @@ this.AddonStudies = {
     for (const study of activeStudies) {
       const addon = await AddonManager.getAddonByID(study.addonId);
       if (!addon) {
-        this.markAsEnded(study);
-        await getStore(db).put(study);
-        Services.obs.notifyObservers(study, STUDY_ENDED_TOPIC);
+        await markAsEnded(db, study);
       }
     }
     await this.close();
@@ -157,10 +167,8 @@ this.AddonStudies = {
       // Use a dedicated DB connection instead of the shared one so that we can
       // close it without fear of affecting other users of the shared connection.
       const db = await openDatabase();
-      await this.markAsEnded(matchingStudy);
-      await getStore(db).put(matchingStudy);
+      await markAsEnded(db, matchingStudy);
       await db.close();
-      Services.obs.notifyObservers(matchingStudy, STUDY_ENDED_TOPIC);
     }
   },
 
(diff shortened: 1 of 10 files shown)


rhelmer:
@Osmose re: the addon listener - so the way undo works in `about:addons` is that it does not actually commit the change until you navigate away, at which point it calls `uninstall()` on the add-on (and any other pending changes).

An easy way to test if this is what you're seeing would be to reload `about:addons` after clicking "Remove".

Osmose:
Ah, that makes sense. If I refresh the page, then it disables the study as expected. Thanks for the help!

This should be ready for review (or, at least, the last commit of this PR should be ready for review :P).

Osmose:
Moving the review to @mythmon given @rehandalal's PTO

recipe-client-addon/lib/AddonStudies.jsm
```diff
@@ -112,6 +124,46 @@ this.AddonStudies = {
...
+    // If the add-on an active study has been removed since we last ran, stop
+    // the study.
+    const activeStudies = (await this.getAll()).filter(study => study.active);
+    const db = await getDatabase();
+    for (const study of activeStudies) {
+      const addon = await AddonManager.getAddonByID(study.addonId);
+      if (!addon) {
+        this.markAsEnded(study);
+        await getStore(db).put(study);
+        Services.obs.notifyObservers(study, STUDY_ENDED_TOPIC);
```
gijsk: Can we just do this in `markAsEnded` directly? Updating the study based on an optional (default to this.db) passed DB connection would also reduce duplication and reduce the risk that we later add some other callsite for markAsEnded that doesn't do this.

recipe-client-addon/test/browser/browser_AddonStudies.js
```diff
@@ -282,3 +271,70 @@ compose_task(
...
+  }
+);
+
+compose_task(
+  AddonStudies.withStudies([
+    studyFactory({active: true, addonId: "installed@example.com", studyEndDate: null}),
+  ]),
+  withInstalledWebExtension({id: "installed@example.com"}),
+  async function testInit([study], [id, addonFile]) {
+    // Wait (or throw if we time out ) for the shield-study-ended event to
```
gijsk: You can just use `TestUtils.topicObserved` instead of rolling your own. I don't think the timeout is a good idea because it will likely intermittently fail on debug build runs, but if you want to keep using it, use `Promise.race()` with a different resolution value and then use that value once the result of `Promise.race` resolves.

recipe-client-addon/lib/AddonStudies.jsm
Osmose: I hesitated doing that because I don't want to send the notification out until the save has actually happened. I didn't save inside the method because the add-on watcher needs to use its own database instance to save.

But I could pass in the database to use as a parameter to solve that. And it's pretty much an internal-only API anyway, I should just remove it from the exported object.

Osmose:
Feedback'd.

mythmon:
```
654 INFO TEST-UNEXPECTED-FAIL | browser/extensions/shield-recipe-client/test/browser/browser_ClientEnvironment.js | isFirstRun has a default value - 
Stack trace:
    chrome://mochitests/content/browser/browser/extensions/shield-recipe-client/test/browser/browser_ClientEnvironment.js:isFirstRun:142
    Tester_execTest@chrome://mochikit/content/browser-test.js:789:9
    Tester.prototype.nextTest</<@chrome://mochikit/content/browser-test.js:689:7
    SimpleTest.waitForFocus/waitForFocusInner/focusedOrLoaded/<@chrome://mochikit/content/tests/SimpleTest/SimpleTest.js:795:59
```

Osmose:
Updated; I renamed `compose` and `compose_task` to use the term `decorate` instead, which is a more accurate name for what they were doing.

PR #943, opened 2017-07-30
Uglify test

Osmose:
I was looking at the test slowness in #914; even locally I was starting to get out-of-memory errors. I merged in #918 and it fixed the problem, but prod builds were still taking 80 seconds or so. 

I suspected Babili was the issue given that prod builds were the ones mainly affected. Since we switched to it, the Uglify harmony branch (`uglify-es` on npm) has landed support for async/await, which was the main reason we switched away. The webpack plugin has a beta version `1.0.0-beta.2` that uses `uglify-es`; when I swapped it out for Babili in the main build and ran it locally, the build time went down to 24 seconds. Yay!

This PR is mostly to see how Circle runs the tests with all these changes piled together.


Osmose:
Whoo! So #918 fixes the test problem, and `uglify-es` can minify our action code.

PR #945, opened 2017-08-01
Fix #937: Honor the opt-out study preference.

### LINKED ISSUE #937: Abort opt-out studies / preference studies if the user has opted-out

Osmose:
Both preference experiments and opt-out studies make changes without prompting the user, and thus should both honor the opt-out pref.

This depends on #919, #938, _and_ #939. Luckily, the actual change is tiny: 1e4cdff. The rest is just a mediocre merge of the the three other branches and can be ignored.

@MattGrimes Can you confirm whether the opt-out pref we're adding to about:preferences should affect preference experiments in addition to opt-out studies? Or only opt-out studies?

### DIFF
--- a/recipe-server/client/actions/opt-out-study/index.js
+++ b/recipe-server/client/actions/opt-out-study/index.js
@@ -1,5 +1,7 @@
 import { Action, registerAction, registerAsyncCallback } from '../utils';
 
+const SHIELD_OPT_OUT_PREF = 'app.shield.optoutstudies.enabled';
+
 let seenRecipeIds = [];
 
 /**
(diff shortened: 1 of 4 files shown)


gregglind:
Whew, this feels above my pay grade to decide.  My hunch is that it should affect prefs as well, except rollouts (when they come).

Osmose:
> My hunch is that it should affect prefs as well, except rollouts (when they come).

@MattGrimes agrees with that, via Slack. So we're good there. Thanks!

mythmon:
Now that #919, #938, and #939 have been merged. I've rebased this PR on to master. I think it is ready to merge once it's been reviewed.

mythmon:
Stealing this review to get this merged asap.

PR #946, opened 2017-08-01
Fix #820: Add option for pausing enrollment in an opt-out study. 

### LINKED ISSUE #820: Ability to pause and rollback enrollment for studies
There are three states that a study can be in:

- Enrolling: Actively enrolling users
- Paused: Do not enroll new users, but continue to run on existing users
- Rollback: End the study for all users that are still active, do not enroll new users

One proposal for handling these three states is to add an option to the arguments for study recipes that controls whether to enroll new users, and adding pre- and post-execution handlers that uninstall any studies that are active but didn't match the user, similar to what we do for preference experiments. 

To pause enrollment, you check the box, and to rollback a study, you disable the entire recipe.

@rehandalal @MattGrimes Thoughts on this UI?

Osmose:
Adds a checkbox to the opt-out study form fields that allows users to pause enrollment without stopping the study completely. Long-term I think we need a better design around handling the specifics of enroll-unenroll-style studies, but this will do for the short term, I think.

### DIFF
--- a/recipe-server/client/actions/opt-out-study/package.json
+++ b/recipe-server/client/actions/opt-out-study/package.json
@@ -30,6 +30,11 @@
                     "type": "string",
                     "format": "uri",
                     "minLength": 1
+                },
+                "isEnrollmentPaused": {
+                    "description": "Whether the study is currently enrolling new users",
+                    "type": "boolean",
+                    "default": false
                 }
             }
         }
--- a/recipe-server/client/actions/opt-out-study/package.json
+++ b/recipe-server/client/actions/opt-out-study/package.json
@@ -32,7 +32,7 @@
                     "minLength": 1
                 },
                 "isEnrollmentPaused": {
-                    "description": "Whether the study is currently enrolling new users",
+                    "description": "If true, new users will not be enrolled in the study.",
                     "type": "boolean",
                     "default": false
                 }
--- a/recipe-server/client/actions/opt-out-study/index.js
+++ b/recipe-server/client/actions/opt-out-study/index.js
@@ -16,7 +16,7 @@ export function resetAction() {
 export default class OptOutStudyAction extends Action {
   async execute() {
     const recipeId = this.recipe.id;
-    const { name, description, addonUrl } = this.recipe.arguments;
+    const { name, description, addonUrl, isEnrollmentPaused } = this.recipe.arguments;
     const studies = this.normandy.studies;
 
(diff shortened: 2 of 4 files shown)


recipe-server/client/actions/opt-out-study/package.json
```diff
@@ -30,6 +30,11 @@
                     "type": "string",
                     "format": "uri",
                     "minLength": 1
+                },
+                "isEnrollmentPaused": {
+                    "description": "Whether the study is currently enrolling new users",
```
mythmon: I think this has the opposite meaning as the name and the code. The description alone makes me think that a value of `true` would enroll new users and `false` would not.

recipe-server/client/actions/opt-out-study/package.json
```diff
@@ -30,6 +30,11 @@
                     "type": "string",
                     "format": "uri",
                     "minLength": 1
+                },
+                "isEnrollmentPaused": {
```
mythmon: I was going to suggest that this setting needs to be documented, but I don't see any docs at all for this action [on readthedocs](https://normandy.readthedocs.io/en/latest/user/actions/index.html). Should we add those docs in a separate PR?

Osmose: Yeah, a separate docs PR is next on my list.

mythmon:
```
The following tests failed:
635 INFO TEST-UNEXPECTED-FAIL | browser/extensions/shield-recipe-client/test/browser/browser_ActionSandboxManager.js | test left unexpected property on window: Preferences - 
636 INFO TEST-UNEXPECTED-FAIL | browser/extensions/shield-recipe-client/test/browser/browser_ClientEnvironment.js | isFirstRun has a default value - 
Stack trace:
    chrome://mochitests/content/browser/browser/extensions/shield-recipe-client/test/browser/browser_ClientEnvironment.js:isFirstRun:142
    Tester_execTest@chrome://mochikit/content/browser-test.js:789:9
    Tester.prototype.nextTest</<@chrome://mochikit/content/browser-test.js:689:7
    SimpleTest.waitForFocus/waitForFocusInner/focusedOrLoaded/<@chrome://mochikit/content/tests/SimpleTest/SimpleTest.js:795:59
```

Osmose:
The test failures are independent of this PR, and the fixes are in #936. I can rebase once that lands.

Osmose:
Rebased, test failures have been dealt with.

PR #948, opened 2017-08-02
Bind AddonStudies.jsm methods in the driver

Osmose:
I noticed an issue during manual testing of add-on studies; `driver.studies.start` fails when called in a sandbox because `this` is unbound when it's called.

This fleshes out the study API tests a bit and fixes the binding issue. I expect this to fail on TC with the errors fixed by #936, but if you run this locally, you can see the test added in the first commit fail, and then be fixed by the second commit.

This is small enough to not need two reviewers.

### DIFF
--- a/recipe-client-addon/lib/NormandyDriver.jsm
+++ b/recipe-client-addon/lib/NormandyDriver.jsm
@@ -182,11 +182,14 @@ this.NormandyDriver = function(sandboxManager) {
 
     // Study storage API
     studies: {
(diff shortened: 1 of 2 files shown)


Osmose:
> Looks good, but for my sanity, can you explain in a comment on the github PR why the original test didn't fail? Naively, that looks like it was calling studies.get from inside the sandbox, which I would expect to fail?

It's `AddonStudies.start` that had the binding bug, since it calls `this.downloadAddonToTemporaryFile`. `AddonStudies.get` doesn't use `this` at all.

PR #955, opened 2017-08-04
Merge the redesign feature branch into master

rehandalal:
The time has come to slay this beast... 

All commits except 9e2455b3 onwards have been reviewed.

r?


rehandalal:
The only two commits that are un-reviewed are both merge commits with no conflicts. I think this should be good to land if/when tests pass.

PR #961, opened 2017-08-04
Add missing commas

rehandalal:
r?


rehandalal:
@mythmon I think the TC failure is unrelated

mythmon:
TC failure is #757. I'll merge this with admin powers.

PR #966, opened 2017-08-07
control-new: Show an error when arguments fields aren't available.

mythmon:
Specifically, opt-out studies aren't implemented in the new control interface yet. This has been confusing to QA, and would be confusing to other users too.

### DIFF
--- a/recipe-server/client/control_new/components/recipes/RecipeForm.js
+++ b/recipe-server/client/control_new/components/recipes/RecipeForm.js
@@ -171,7 +171,7 @@ class ArgumentEditorMissingError extends React.PureComponent {
     name: PropTypes.string.isRequired,
   }
 
-  defaultPropTypes = {
+  static defaultProps = {
     githubUrl: null,
   }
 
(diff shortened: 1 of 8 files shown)


recipe-server/client/control_new/components/recipes/RecipeForm.js
```diff
@@ -155,3 +159,48 @@ class ActionSelect extends React.Component {
...
+
+  defaultPropTypes = {
+    githubUrl: null,
+  }
+
+  render() {
+    const { githubUrl, name } = this.props;
+    let fileIssueUrl;
+    if (githubUrl) {
+      const u = new URL(githubUrl);
```
andymikulski: We should use more descriptive variable names. I'm surprised this wasn't hit by linting, I thought there was a 3 character minimum or something.

recipe-server/client/control_new/components/recipes/RecipeForm.js
```diff
@@ -155,3 +159,48 @@ class ActionSelect extends React.Component {
...
+      u.pathname += '/issues/new';
+      u.searchParams.set('title', `Argument fields missing for action "${name}"`);
+      fileIssueUrl = u.toString();
+    }
+
+    console.log('QueryServiceInfo', QueryServiceInfo, 'Alert', Alert);
```
andymikulski: `console.log`

recipe-server/normandy/settings.py
```diff
@@ -351,6 +351,8 @@ def RAVEN_CONFIG(self):
     AWS_SECRET_ACCESS_KEY = values.Value()
     AWS_STORAGE_BUCKET_NAME = values.Value()
 
+    GITHUB_URL = values.Value('https://github.com/mozilla/normandy')
```
andymikulski: Will this ever be used outside of the `ArgumentEditorMissingError` component? I'm leaning towards just setting this as the default prop versus adding it to the service API call. That way, the link href won't ever be null, too.

mythmon: This setting is also used in our `/__version__` endpoint. I think it's fine to have it be null for a moment. This is an uncommon case, and we should be fetching service info on every page any way, since it powers other parts of the UI.

PR #967, opened 2017-08-07
#695: Add EnvAlert component

### LINKED ISSUE #695: Add status bar to control interface showing which environment is currently being browsed
If you have multiple tabs open to control interfaces in different environments (local vs stage vs prod), it's hard to tell which environment you're looking at. Other projects I've worked on have had an indicator of some kind that is shown when browsing a staging or local development instance. The indicator usually states which environment you're currently viewing, and is hidden on the production instance.

Because this isn't visible on prod, we don't have/need a specific design for this. A bar at the top of the page would be fine. It might even be nice to link to the [Environments](http://normandy.readthedocs.io/en/latest/qa/environments.html) documentation for that environment where possible.

A follow up feature that I think we should avoid doing for the first iteration of this is to show the status of the `__heartbeat__` checks in the status bar to help notify testers if something fails during their testing. This would help catch issues like our recent signature problems that the heartbeat check caught.

andymikulski:
Fixes #695 

- Adds component to top of page indicating whether the user is viewing a development or staging environment
- Adds link to the docs
- Adds appropriate tests

screen shot 2017-08-07 at 2 48 12 pm

### DIFF
--- a/recipe-server/client/control_new/components/common/EnvAlert.js
+++ b/recipe-server/client/control_new/components/common/EnvAlert.js
@@ -1,11 +1,20 @@
 import { Alert } from 'antd';
+import PropTypes from 'prop-types';
 import React from 'react';
+import { connect } from 'react-redux';
 
 /**
  * Simple component which tells the user whether they are viewing a dev or staging
  * environment. On production, nothing is shown.
  */
+@connect(() => ({
+  currentUrl: window.location.href,
+}))
 export default class EnvAlert extends React.PureComponent {
+  static propTypes = {
+    currentUrl: PropTypes.string.isRequired,
+  };
+
   static productionFragments = ['normandy-admin.prod.'];
   static stageFragments = ['normandy-admin.stage.'];
 
@@ -39,7 +48,7 @@ export default class EnvAlert extends React.PureComponent {
   }
 
   render() {
-    const currentUrl = window.location.href;
+    const { currentUrl } = this.props;
 
     // Never show this component when on production.
     if (EnvAlert.checkProduction(currentUrl)) {
--- a/recipe-server/client/control_new/components/common/EnvAlert.js
+++ b/recipe-server/client/control_new/components/common/EnvAlert.js
@@ -6,8 +6,8 @@ import React from 'react';
  * environment. On production, nothing is shown.
  */
 export default class EnvAlert extends React.PureComponent {
-  static productionUrls = ['normandy-admin.prod.'];
-  static stageUrls = ['normandy-admin.stage.'];
+  static productionFragments = ['normandy-admin.prod.'];
+  static stageFragments = ['normandy-admin.stage.'];
 
   static learnMoreLink = (
     <a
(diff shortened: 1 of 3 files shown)


recipe-server/client/control_new/components/common/EnvAlert.js
```diff
@@ -0,0 +1,66 @@
...
+
+  /**
+   * Find a string fragment in an array of strings.
+   *
+   * @param  {String}        needle   String fragment to find.
+   * @param  {Array<String>} haystack String collection to search.
+   * @return {Boolean}                Was the fragment found?
+   */
+  static findPartialString(needle, haystack) {
+    return !!haystack.find(straw => straw.indexOf(needle) > -1);
```
mythmon: I think this doesn't work right. The `needle` will be the the full url, `https://normandy-admin.stage.mozaws.net/control/...`, which won't be a substring of any of the `haystack`.

recipe-server/client/control_new/tests/components/common/test_EnvAlert.js
```diff
@@ -0,0 +1,42 @@
...
+      expect(result).toBe(false);
+
+      // abcdef is not present
+      result = EnvAlert.findPartialString('abcdef', testStrings);
+      expect(result).toBe(false);
+    });
+  });
+
+  it('should have accurate production URLs', () => {
+    expect(EnvAlert.productionUrls).toEqual(['normandy-admin.prod.']);
```
mythmon: I don' t think this is a valuable test.

recipe-server/client/control_new/components/common/EnvAlert.js
```diff
@@ -0,0 +1,66 @@
...
+  static checkProduction(url) {
+    return EnvAlert.findPartialString(url, EnvAlert.productionUrls);
+  }
+
+  static checkStaging(url) {
+    return EnvAlert.findPartialString(url, EnvAlert.stageUrls);
+  }
+
+  render() {
+    const currentUrl = window.location.href;
```
mythmon: We could take this from the state (via `redux-little-router`), which would make testing this easier.

recipe-server/client/control_new/tests/components/common/test_EnvAlert.js
```diff
@@ -0,0 +1,42 @@
...
+  it('should work', () => {
+    const wrapper = () => shallow(<EnvAlert />);
+
+    expect(wrapper).not.toThrow();
+  });
+
+  describe('findPartialString', () => {
+    const testStrings = ['abc', 'def'];
+
+    it('should find a piece of a string amongst an array of strings', () => {
```
mythmon: This test should probably use representative urls, instead of filler data. That way it's clearer if it's testing the right thing.

recipe-server/client/control_new/components/common/EnvAlert.js
andymikulski: We can't actually, as `lit'l` only deals with 'internal' app routes. That is, the router doesn't know anything about the `location.origin`, the full URL, etc. The `router` state object only gives us app-relevant data (`you're on this Component on this route at the path /whatever/cool`).

Do you have any suggestions on stubbing out the `window` object? In the past, I've used something like [window-or-global](https://github.com/purposeindustries/window-or-global/blob/master/lib/index.js) for a similar case, but not sure if it would work here. I'm open to suggestions.

recipe-server/client/control_new/tests/components/common/test_EnvAlert.js
andymikulski: I agree, looking at it now. The idea was that I'd like some sort of red flag to be raised when the URLs change (if ever), to prevent some of the headaches of using hardcoded URLs.

An idea I was kicking around was defining the staging/prod URLs in the webpack config (using `webpack.DefinePlugin`), then replacing the hardcoded arrays with the webpack-global variable (e.g. `static prodUrls = PRODUCTION_URLS;`). How do you feel about that?

recipe-server/client/control_new/components/common/EnvAlert.js
mythmon: Ah, you're right.

I don't know a great way to stub window. Last time I tried to do something like this I only got sad. I think the best thing I found was to have a function like "getUrl", and then we stubbed that method out instead of stubbing window.

Osmose: Could also make the current URL a prop.

andymikulski:
@Osmose 's prop idea worked out really well. I just `connect`ed the component and then passed in the URL as a prop for testing.

Ready for another review!

PR #970, opened 2017-08-09
#927: Add more unique IDs for selenium

### LINKED ISSUE #927: Create unique ids for certain DOM elements to make automated UI testing more sustainable
I've added TODO comments in my locators.py file, which indicate which DOM elements I would like to have unique ids to make automated testing easier and more sustainable. 

https://github.com/18epedersen/Normandy-e2e-tests/blob/dev/pages/locators.py

andymikulski:
Fixes #927 , adds more Selenium ID's.

### DIFF
--- a/recipe-server/client/control_new/components/recipes/PreferenceExperimentFields.js
+++ b/recipe-server/client/control_new/components/recipes/PreferenceExperimentFields.js
@@ -221,10 +221,10 @@ export class BooleanPreferenceField extends React.Component {
   render() {
     return (
       <Radio.Group {...this.props}>
-        <Radio.Button value={true} id="pref-true">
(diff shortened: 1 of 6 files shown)


recipe-server/client/control_new/components/recipes/PreferenceExperimentFields.js
```diff
@@ -215,10 +221,10 @@ export class BooleanPreferenceField extends React.Component {
   render() {
     return (
       <Radio.Group {...this.props}>
-        <Radio.Button value={true}>
+        <Radio.Button value={true} id="pref-true">
```
mythmon: Won't this component get rendered multiple times on one page, making using an id here incorrect?

Issue #541, opened 2017-02-22
Add a favicon


Osmose:
Firefox shows a 404 when it attempts to fetch a favicon for the site. IIRC @designakt made a SHIELD logo that we could use.

Alternatively, a blank image like what http://stackoverflow.com/questions/1321878/how-to-prevent-favicon-ico-requests recommends would be fine too. The 404 is mildly confusing because it looks like an error, so we might as well make it go away _somehow_.

mastizada:
Favicon will be always requested even in api view (json text only). Suggestion is stackoverflow is to add html favicon tag with empty base64 image. I think better way will be to add static favicon.ico file and redirect it using webserver configuration or using django url as I did it previously in Pontoon for robots.txt

Osmose:
That's a good point about viewing API views directly (with a browser).

I think your redirect idea is good (`django.contrib.staticfiles.storage.staticfiles_storage.url` should be able to get you the URL for the resource). You'll also want to make sure the redirect view has cache headers for 30 seconds (you should probably add a new setting for this), as we get enough traffic on Normandy that no caching would melt our app servers.

I'm apparently unable to assign the issue to you unless you're a collaborator on the repo, but you should feel free to submit a PR for this issue. Thanks for the help!

andymikulski:
As of #955 we have a favicon asset, set using a simple `link` tag:

```
<link rel="icon" href="{{STATIC_URL}}img/favicon.ico">
```

This meets the front end needs, however I'm 99% sure that doesn't solve the API view business. @mastizada , if you're interested in taking over the django side, you're more than welcome to! Otherwise, this issue can probably be closed.

mastizada:
Hi, for some reason it was in my spam box. Will try to send PR tomorrow, thanks for informing.

mastizada:
@andymckay @mythmon We should test it after deployment and close that issue.

mythmon:
We actually close issues when the code merges. We can re-open if something goes wrong after deployment.

PR #972, opened 2017-08-09
fix issue #541 - favicon url

mastizada:


recipe-server/normandy/base/views.py
```diff
@@ -1,5 +1,9 @@
-from django.http import HttpResponse
+from django.http import HttpResponse, HttpResponseRedirect
 
 
 def index(request):
     return HttpResponse('<h1>Normandy</h1>')
+
+
+def favicon(request):
+    return HttpResponseRedirect('/static/img/favicon.ico')
```
mythmon: Hardcoding this url means that it won't work on our production servers, which serve static content with a CDN. You're better off building this url with [django.contrib.staticfiles.templatetags.staticfiles.static](https://stackoverflow.com/questions/11721818/django-get-the-static-files-url-in-view/17738606#17738606), which will work in all environments.

mastizada: @mythmon Templates use `STATIC_URL` variable for url, not the `static` tag. Shoul I use `settings.STATIC_URL` instead of `static` tag in view too?

mythmon: I think both would work in this case, but I think `static` is a little cleaner.

mythmon:
Using my admin powers to merge this since it doesn't affect TC builds, and I don't know how to get TC to run for just this PR.

PR #983, opened 2017-08-14
Fixes #981 - Add shield identicon component.

### LINKED ISSUE #981: Recipe Identicons (IdentiSHIELDs?)
**Description**
Automatically create a visually unique SHIELD/emblem for recipes.

**Goal**
- Provide identifier that is easily rememberable and communicable between end users.

**Level of Effort**
Medium. @mythmon has already done a bulk of the work with regards to creating the prototype shields, which is awesome. The remaining work is implementing them into the site; the shields introduce a new aesthetic element in several places, which would require minor design QA/reworking.

mythmon:
Fixes #981 

This is likely way over engineered, but it was fun.

@andymikulski Can you push your color and emoticon changes to this branch, and any other changes you want to make to the design integration?


andymikulski:
I can push my commit in a moment. @mythmon Where did `utils/Genome.js` come from? Can that be made into a dependency?

mythmon:
I wrote `Genome.js` specifically for this. We could make it a separate npm package, but I don't think there is a lot of benefit to that.

recipe-server/client/control_new/components/common/ShieldIdenticon.js
```diff
@@ -146,88 +145,120 @@ class Cover extends React.Component {
...
+
+function pickPair(genome, base) {
+  const baseLuminance = hexLuma(base);
+  let chosenColor;
+  let pairLuminance;
+
+  do {
+    chosenColor = genome.choice(colors, 'counter color');
+    pairLuminance = hexLuma(chosenColor);
+  } while (Math.abs(baseLuminance - pairLuminance) < 75);
```
mythmon: This works well, but I worry about it running a variable number of times, each time using up a bunch of entropy from the `Genome`. Could we instead filter `colors` to remove colors that aren't different enough in luminescence, and then run `genome.choice` only once?

recipe-server/client/control_new/components/common/ShieldIdenticon.js
```diff
@@ -3,52 +3,51 @@ import PropTypes from 'prop-types';
...
+  '🔔', '📢', '⏳', '⏰', '🔓', '🔒', '🔏', '🔐', '🔑', '🔎', '💡', '🔦', '🔆', '🔅', '🔌', '🔋',
+  '🔍', '🛁', '🛀', '🚿', '🚽', '🔧', '🔨', '🚪', '💣', '🔫', '🔪', '💊', '💉', '💰', '💸',
+  '📨', '📬', '📌', '📎', '📕', '📓', '📚', '📖', '🔬', '🔭', '🎨', '🎬', '🎤', '🎵', '🎹', '🎻',
+  '🎺', '🎷', '🎸', '👾', '🎮', '🃏', '🎲', '🎯', '🏈', '🏀', '⚽', '🎾', '🎱', '🏉', '🎳', '⛳',
+  '🚴', '🏁', '🏇', '🏆', '🎿', '🏂', '🏄', '🎣', '🍵', '🍶', '🍼', '🍺', '🍻', '🍸', '🍹', '🍷',
+  '🍴', '🍕', '🍔', '🍟', '🍗', '🍤', '🍞', '🍩', '🍮', '🍦', '🍨', '🍧', '🎂', '🍰', '🍪', '🍫',
+  '🍬', '🍭', '🍯', '🍎', '🍏', '🍊', '🍋', '🍒', '🍇', '🍉', '🍓', '🍑', '🍌', '🍐', '🍍', '🍆',
+  '🍅', '🌽', '🏠', '🏡', '⛵', '🚤', '🚣', '🚀', '🚁', '🚂', '🚎', '🚌', '🚍', '🚙', '🚘', '🚗',
+  '🚕', '🚖', '🚛', '🚚', '🚨', '🚓', '🚔', '🚒', '🚑', '🚐', '🚲', '🚜', '💈', '🚦', '🚧', '🏮',
+  '🎰', '🗿', '🎪', '🎭', '📍', '🚩', '💯',
```
mythmon: A small trick, if we wanted this to be more compact and easier to edit:

```js
const emojis = `
  😄😃😀😊😉😍😘😚
  😗😙😜😝😛😳😁😔`
  .split(/\s*/));
```

On the other hand, we should probably avoid changing this much, since it would make the same seed generate different colors after the change. We probably don't want to do this. Overall we probably can't change the identishields much after we deploy them.

recipe-server/client/control_new/components/common/ShieldIdenticon.js
andymikulski: Good thinking! Makes sense to me.

recipe-server/client/control_new/components/common/ShieldIdenticon.js
andymikulski: >Overall we probably can't change the identishields much after we deploy them.

Yeah, I've been thinking about that. It'd be cool if we could save the shield config to the recipe, versus relying on a seed value to correctly regenerate the icon.

By saving the shield and not relying on the seed, we could alter the generator in the future and not alter existing recipes. We'd also be able to add a `shuffle` button to the `create recipe page`, if the user decided they would like a different randomized shield icon before saving.

The low-hanging idea I have is to save the resultant SVG to the recipe as plaintext HTML/data URI. I'm not in love with that, though. Maybe we could have it serialize as a basic JS object: `{ stripes: { count: 2, padding: 0.3, direction: 'horizontal'  }, ... }`?

The curveball for me is how the `Genome.wrapComponent` business ties into that approach, any thoughts @mythmon ?

andymikulski: Also, turns out `'😗😙😜😝😛😳😁😔'.split(/\s*/);` doesn't work. I tried the solution in [this SO post](https://stackoverflow.com/a/37535876) which worked: `[...'😗😙😜😝😛😳😁😔']`

andymikulski: ^ Just kidding, since we're using template strings I also have to filter out `\n`s and empty strings:

```
[...`😗😙
😜😝😛`].filter(emo => emo.indexOf(/\n/) > -1 || emo.trim());
```

mythmon: Huh. What problem did `.split(/\s*/)` have for you? It worked fine in my testing. It's not too important though.

andymikulski: The array would look like `['😗', '😙', '\n', ' ', ' ', '😜', '😝', '😛']`

mythmon: > The low-hanging idea I have is to save the resultant SVG to the recipe as plaintext HTML/data URI. I'm not in love with that, though. Maybe we could have it serialize as a basic JS object: { stripes: { count: 2, padding: 0.3, direction: 'horizontal' }, ... }?
>
> The curveball for me is how the Genome.wrapComponent business ties into that approach, any thoughts @mythmon ?

If we want to go this way we could probably ditch `wrapComponent`, and just use Genome to generate a set of props to pass to the component. The question that raises for me is where do we save those serialized props? We can't save them locally, because then different users would have different versions of the shield (or might, if we update things).

Storing stuff like those props in the API doesn't make a lot of sense to me. It is sort of extra noise that I don't think matters to the server. I think I'd like to keep it specific to the front end. Either that, or we should have a generic metadata API for recipes, but those are kind of a pain, and I'm not sure I want to go down that road right now.

Maybe if we ever change this, we could define recipe "generations". These shield identicons are part of generation 1, and if we want to add more features to the identicons, only generation 2 recipes get to use them. We could do it based on recipe creation date or something.

Overall, I think we can leave it as is. The identicons are a minor feature for now, and if we need to improve them in the future, we can deal with that then.

mythmon: About the `.split` thing: I did my tests with simple alphabet characters, so I didn't run into the problem with emoji often being astral plane characters.

mythmon:
CircleCI failure is some lint issues.

andymikulski:
I think this can probably be closed in favor of #988 , this code would probably be good reference for @marcrowo though.

PR #985, opened 2017-08-15
#984: Add Opt-Out action UI

### LINKED ISSUE #984: Opt-out study UI

andymikulski:
Fixes #984 

- Adds `opt-out-study` action UI
- Adds `SwitchBox` form component (basically Ant's `Switch` with a label wrapper)

screen shot 2017-08-15 at 10 05 13 am

### DIFF
--- a/recipe-server/client/control_new/components/recipes/OptOutStudyFields.js
+++ b/recipe-server/client/control_new/components/recipes/OptOutStudyFields.js
@@ -1,0 +1,88 @@
+import { Col, Row, Input, Select } from 'antd';
+import { Map, List } from 'immutable';
+import PropTypes from 'prop-types';
+import React from 'react';
+import { connect } from 'react-redux';
+
+import SwitchBox from 'control_new/components/forms/SwitchBox';
+import FormItem from 'control_new/components/forms/FormItem';
+import { connectFormProps } from 'control_new/utils/forms';
+
+import { getExtensionListing } from 'control_new/state/app/extensions/selectors';
+
+@connectFormProps
+@connect(
+  state => ({
+    extensions: getExtensionListing(state),
+  }),
+)
+export default class OptOutStudyFields extends React.Component {
+  static propTypes = {
+    disabled: PropTypes.bool,
+    extensions: PropTypes.instanceOf(List).isRequired,
+    recipeArguments: PropTypes.instanceOf(Map).isRequired,
+  };
+
+  static defaultProps = {
+    disabled: false,
+    extensions: new List(),
+    recipeArguments: new Map(),
+  };
+
+  render() {
+    const {
+      recipeArguments,
+      disabled,
+    } = this.props;
+    return (
+      <Row>
+        <p className="action-info">Enroll the user in a study which they can remove themselves from.</p>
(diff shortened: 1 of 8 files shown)


recipe-server/client/control_new/components/recipes/OptOutStudyFields.js
```diff
@@ -0,0 +1,88 @@
...
+            </FormItem>
+
+            <FormItem
+              label="Prevent New Enrollment"
+              name="arguments.isEnrollmentPaused"
+              initialValue={recipeArguments.get('isEnrollmentPaused', false)}
+            >
+              <SwitchBox>
+                Prevents new users from joining this study&nbsp;cohort. <br /> Existing
+                users will remain in&nbsp;the&nbsp;study.
```
mythmon: What's up with the `&nbsp;`s here? Would it be better to put these two items in `<p>` tags?

andymikulski: They're non-breaking spaces, to prevent widows:

```
Long text long text existing users will remain in the
study.
```
Here, I'm ensuring that `cohort`/`study` don't become widows. Technically, this could be accomplished via CSS, but I've never gotten the `widows`/`orphans` properties to work correctly.

mythmon: Huh, I've never seen that before. That's a good idea.

recipe-server/client/control_new/tests/components/recipes/test_SwitchBox.js
```diff
@@ -0,0 +1,18 @@
...
+  const props = {
+    children: null,
+    onChange: () => {},
+    value: null,
+  };
+
+  it('should work', () => {
+    const wrapper = () => shallow(<SwitchBox {...props} />);
+
+    expect(wrapper).not.toThrow();
```
mythmon: We could probably write a more in depth version of this test. Maybe checking that clicking acts as expected, and that the children get rendered?

andymikulski:
@mythmon I've added tests for the `SwitchBox` component, however I'm having trouble doing much more than `shallow` testing the `OptOutStudyFields` component. It seems there's an issue with `connectFormProps` not exposing the `WrappedComponent` we use for testing `connect`ed components. I reckon @Osmose can shed more light on this, but I'm thinking that in-depth form tests may have to be its own PR all together.

recipe-server/client/control_new/tests/components/recipes/test_SwitchBox.js
```diff
@@ -10,9 +11,46 @@ describe('<SwitchBox>', () => {
...
+  });
+
+  describe('the label', () => {
+    it('should not render if there is no label', () => {
+      const wrapper = factory();
+      expect(wrapper.find('.label').length).toBe(0);
+    });
+
+    it('should render the passed in children', () => {
+      const label = 'Root beer\'s flavor comes from sassafras and/or sarsaparilla!';
```
mythmon: Nit: I'm pretty sure that in  this case the linter will let you use double quotes to avoid the `\'` escape. This doesn't need to block this PR though.

PR #992, opened 2017-08-17
Fix #988: Add server side identicons

### LINKED ISSUE #988: Add server side shield identicons
This is specs for @marcrowo's benefit.

`/api/v2/recipe/` should have a new field, `identicon_seed`. It should validate that the seed is of the format `generation:seed`, like `v1:asdfasdf`, where the first part is always `v1` (for now), and the second part is any string. If no generation is supplied in the  API, it may add a `v1:` prefix.

There should be a new API, `/api/v2/identicon/${generation}:${seed}.svg` that returns a SVG shield identicon using stable-random generation techniques, like in #983. This should always return the exact same output for the same input, and there should be a test for this.

marcrowo:
Fixes #988, as specified in that issue.


recipe-server/normandy/base/genome.py
```diff
@@ -0,0 +1,138 @@
...
+        self.log = []
+
+    @staticmethod
+    def generate(seed=random.random()):
+        """Asynchronously returns a Genome based on the given seed.
+
+        seed -- Any JSON serializable value. Will be used to seed the
+                genome. If not provided, will be randomly generated.
+        """
+        hash = sha256(json.dumps(seed).encode('utf-8'))
```
mythmon: I think the Python version of this should just call `str()` on seed, instead of using JSON, since Python is better at converting things to strings by default.

recipe-server/normandy/base/genome.py
```diff
@@ -0,0 +1,138 @@
...
+        if self.entropy < needed:
+            raise Exception('Not enough entropy left in genome.')
+        self.entropy -= needed
+        quotient, remainder = divmod(self.chromosome, mod)
+        self.log.append({
+            'reason': reason,
+            'mod': mod,
+            'remainder': str(remainder),
+            'before': str(self.chromosome),
+            'after': str(quotient),
```
mythmon: Converting these to strings isn't needed in Python. It was needed in JS because the big int library didn't print out nicely.

recipe-server/normandy/recipes/api/v2/views.py
```diff
@@ -203,3 +205,98 @@ def close(self, request, pk=None):
...
+        base_luminance = self.hex_luma(base)
+        chosen_color = None
+        pair_luminance = None
+
+        while True:
+            chosen_color = genome.color('counter color')
+            pair_luminance = self.hex_luma(chosen_color)
+
+            if abs(base_luminance - pair_luminance) < 75:
+                break
```
mythmon: From my [comment on #983](https://github.com/mozilla/normandy/pull/983#discussion_r133260738):

> This works well, but I worry about it running a variable number of times, each time using up a bunch of entropy from the Genome. Could we instead filter colors to remove colors that aren't different enough in luminescence, and then run genome.choice only once?

marcrowo: Note: This is also the wrong comparison. It should be `abs(...) > 75`.

recipe-server/normandy/recipes/api/v2/views.py
```diff
@@ -292,4 +292,4 @@ def get(self, request, *, generation, seed):
             }
             context['transform'] = f'scale(100) rotate({rotations[direction]} 0.5,0.5)'
 
-        return render(request, 'identicon.svg', context)
+            return render(request, 'identicon.svg', context, content_type='image/svg+xml')
```
mythmon: I don't think this shouldn't have been indented

mythmon: I don't think this should be indented.

recipe-server/normandy/recipes/api/v2/views.py
```diff
@@ -203,3 +205,91 @@ def close(self, request, pk=None):
...
+        pattern_color = self.pick_pair(genome, field_color)
+
+        context = {
+            'treatment': treatment,
+            'emoji': emoji,
+            'field_color': field_color,
+            'pattern_color': pattern_color,
+        }
+
+        # Add treatment-specific information to context
```
Osmose: Instead of a big if statement based off the treatment names, we could make the treatment values functions that take the `genome` as an argument. Then we can define them statically outside of this function and call them instead of having all of their logic here.

recipe-server/normandy/recipes/urls.py
```diff
@@ -32,4 +32,9 @@
         api_v1_views.ActionImplementationView.as_view(),
         name='action-implementation'
     ),
+    url(
+        r'^api/v2/identicon/(?P<generation>v[0-9]):(?P<seed>.{1,64}).svg',
```
Osmose: @mythmon Thoughts on an API endpoint with `.svg` at the end? I'm on the fence but don't have a great alternative.

recipe-server/normandy/recipes/tests/api/v2/test_api.py
```diff
@@ -722,3 +743,22 @@ def test_cancel_approval(self, api_client, mocked_autograph):
...
+        assert res.status_code == 200
+
+    def test_it_returns_the_same_output(self, api_client):
+        res1 = api_client.get('/api/v2/identicon/v1:foobar.svg')
+        res2 = api_client.get('/api/v2/identicon/v1:foobar.svg')
+        assert res1.content == res2.content
+
+    def test_it_returns_known_output(self, api_client):
+        res = api_client.get('/api/v2/identicon/v1:foobar.svg')
+        with open(Path(settings.BASE_DIR).joinpath(
```
Osmose: Instead of a hanging indent, create the path object on a previous line.

recipe-server/normandy/recipes/tests/api/v2/test_api.py
```diff
@@ -722,3 +743,22 @@ def test_cancel_approval(self, api_client, mocked_autograph):
...
+                'normandy', 'recipes', 'tests',
+                'api', 'v2', 'foobar.svg'), 'rb') as svg_file:
+            assert(svg_file.read() == res.content)
```
Osmose: `assert` is a statement, not a function.

recipe-server/normandy/recipes/templates/identicon.svg
```diff
@@ -0,0 +1,46 @@
...
+                <path d="M 0.5 0.0
+                         C 0.6 0.1, 0.7 0.15, 0.9 0.15
+                         C 0.9 0.5, 0.9 0.8, 0.5 1.0
+                         C 0.1 0.8, 0.1 0.5, 0.1 0.15
+                         C 0.3 0.15, 0.4 0.1, 0.5 0.0"/>
+            </clipPath>
+        </defs>
+        <g clip-path="url(#shield-border)">
+            <rect fill="{{  field_color }}" width="100" height="100"/>
+            {% if treatment == 'TwoColor' %}
```
Osmose: Above I was assuming that the treatment was generating this XML and the template didn't need to know which treatment was being used, and I think that's a cleaner design anyway.

recipe-server/normandy/recipes/migrations/0047_reciperevision_identicon_seed.py
```diff
@@ -0,0 +1,39 @@
...
+        ('recipes', '0046_reset_signatures'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='reciperevision',
+            name='identicon_seed',
+            field=normandy.recipes.fields.IdenticonSeedField(blank=True, null=True, max_length=64),
+        ),
+        migrations.RunPython(generate_identicon_seeds, generate_identicon_seeds),
```
Osmose: The backwards migration should just be an empty function rather than just re-generating the seeds again.

recipe-server/normandy/recipes/models.py
```diff
@@ -330,6 +336,7 @@ def data(self):
             'channels': list(self.channels.all()),
             'countries': list(self.countries.all()),
             'locales': list(self.locales.all()),
+            'identicon_seed': self.identicon_seed,
```
Osmose: @mythmon This means we'll need another signature-busting migration, right?

recipe-server/normandy/recipes/fields.py
```diff
@@ -34,3 +35,21 @@ def deconstruct(self):
         del kwargs['max_length']
         del kwargs['editable']
         return name, path, args, kwargs
+
+
+class IdenticonSeedField(models.CharField):
```
Osmose: This field class is rad, I like it! :D

recipe-server/normandy/recipes/api/v2/views.py
```diff
@@ -203,3 +205,91 @@ def close(self, request, pk=None):
         approval_request = self.get_object()
         approval_request.close()
         return Response(status=status.HTTP_204_NO_CONTENT)
+
+
+class IdenticonView(views.APIView):
+    @staticmethod
+    def hex_luma(hex):
```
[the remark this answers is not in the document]
marcrowo: @mythmon I think you mentioned something about `@staticmethod` when we discussed this previously? I don't remember exactly what we discussed however.

recipe-server/normandy/base/genome.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+class Genome:
+    """A seedable source that provides arbitrarily sized chunks of randomness."""
+    def __init__(self, chromosome, entropy):
+        self.chromosome = chromosome
+        self.entropy = entropy
+        self.log = []
+        self.colors = [
+            '#002b36',  # base03
```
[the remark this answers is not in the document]
marcrowo: These colors and labeling come from the [Solarized colorscheme](http://ethanschoonover.com/solarized):

solarized palette

recipe-server/normandy/base/genome.py
```diff
@@ -0,0 +1,136 @@
...
+            '#073642',  # base02
+            '#586e75',  # base01
+            '#657b83',  # base00
+            '#839496',  # base0
+            '#93a1a1',  # base1
+            '#eee8d5',  # base2
+            '#fdf6e3',  # base3
+            # '#b58900',  # yellow
```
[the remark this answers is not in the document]
marcrowo: I believe that this is here because we are using a different color than the default yellow from Solarized, but I agree it should be removed.

recipe-server/normandy/recipes/api/v2/views.py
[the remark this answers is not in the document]
mythmon: The standard advice is to not use `@staticmethod`, but instead use a non-class function or `@classmethod`. Since you are not using anything from the class in this method, a non-class function would probably be appropriate.

recipe-server/normandy/base/genome.py
```diff
@@ -0,0 +1,136 @@
+import math
+import random
+import string
+
+from hashlib import sha256
+
+
+class Genome:
```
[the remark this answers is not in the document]
marcrowo: I think this is a question for @mythmon.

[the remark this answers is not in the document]
mythmon: As long as the same guarantees are provided, I'm ok with this. Since we can seed Python's Random class, instead of JS's single global object, we can exert more control over than we could in JS. As long as the same calls in the same order always produce the same results (and we have tests for this), I'm ok with subclassing `Random`.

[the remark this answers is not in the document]
Osmose: Oh, and to clarify, the example above isn't how I think we should end up using the Random class (since it's not subclassing it like I think we should), just on how the objects created by the class work.

recipe-server/normandy/recipes/api/v2/views.py
```diff
@@ -203,3 +205,91 @@ def close(self, request, pk=None):
...
+        # hex color code to RGB
+        hex = hex[1:]  # remove '#'
+        rgb = [int(h, 16) for h in [hex[i: i + 2] for i in range(0, len(hex), 2)]]
+
+        # Determine color luminance based on perceived values.
+        return (0.2126 * rgb[0]) + (0.7152 * rgb[1]) + (0.0722 * rgb[2])
+
+    def pick_pair(self, genome, base):
+        base_luminance = self.hex_luma(base)
+        good_pairs = list(filter(lambda c: abs(base_luminance - self.hex_luma(c)) > 75,
```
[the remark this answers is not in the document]
marcrowo: I think that the `contrasts_well` approach is a good idea. I have added a `Color` class which encompasses various helper functions applied to colors and this class includes `contrasts_well(self, pair_color)`.

recipe-server/normandy/base/genome.py
```diff
@@ -0,0 +1,136 @@
...
+        """Return an integer in the range [0, mod), and deduct the correct
+        amount of entropy from the pool. Vulnerable to modulus-bias when
+        `mod` is not a power of two.
+        This is the fundamental way to consume a Genome. All the other
+        methods below that use entropy defer to this method.
+        """
+        needed = math.log2(mod)
+        if self.entropy < needed:
+            raise Exception('Not enough entropy left in genome.')
+        self.entropy -= needed
```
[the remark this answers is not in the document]
marcrowo: This seems reasonable. AFAIK `divmod` will only throw if attempting to divide by 0.

recipe-server/normandy/recipes/templates/identicon.svg
marcrowo: As discussed in person, we decided that since there are only two treatments, it's better to leave the template as is.

recipe-server/normandy/recipes/urls.py
mythmon: I'm ok with it. We aren't doing it here, but DRF can be configured to vary based on the extension of the API endpoint called. So in theory, `/api/v2/identicon/v1:foo.svg` could return the image, and `/api/v2/identicon/v1:foo.json` could return a JSON version of the data used to generate the svg. I think for non-json endpoints, this is a good idea.

recipe-server/normandy/recipes/models.py
mythmon: No, because `identicon_seed` isn't a part of the signed part in `canonical_json`.

recipe-server/normandy/recipes/api/v2/shield_identicon.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def pick_pair(genome, base_color):
+    good_pairs = [
+        color for color in genome.colors if base_color.contrasts_well(color)]
+    assert len(good_pairs), 'no colors satisfy the luminance requirement'
+    return genome.choice(good_pairs)
+
+
+def single_color(genome, context):
```
Osmose: Passing in an object that gets modified is a pattern we typically avoid; instead, consider returning an object, and then updating the context with it using `context.update(extra)` in `generate_svg`.

Modifying an argument as a way of returning data isn't recommended because it makes it harder to track where mutations are happening to your data. In this small case it's pretty easy to understand, but broadly applied, the pattern makes code harder to comprehend, so we just avoid using it at all.

recipe-server/normandy/recipes/api/v2/shield_identicon.py
```diff
@@ -0,0 +1,81 @@
...
+        'top',
+        'bottom',
+        'left',
+        'right',
+        'topLeft',
+        'topRight',
+        'bottomLeft',
+        'bottomRight',
+    ]
+    area = genome.choice(areas)
```
Osmose: You're not using the names anywhere, could you just use `choice` on a list of the values from the `rotations` dict? You could even just generate it with the `range` builtin:

```python
range(0, 360, 45)  # == an iterator containing the angles you want
```

recipe-server/normandy/recipes/api/v2/shield_identicon.py
```diff
@@ -0,0 +1,81 @@
...
+    }
+    context['transform'] = f'scale(100) rotate({rotations[area]} 0.5,0.5)'
+    return context
+
+
+def stripes(genome, context):
+    context['treatment'] = 'Stripes'
+    count = genome.randint(1, 4)
+    padding = genome.uniform(0.1, 0.4)
+    directions = ['vertical', 'horizontal', 'diagonal1', 'diagonal2']
```
Osmose: Same comment here with the directions as above with the areas.

recipe-server/normandy/recipes/api/v2/shield_identicon.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def generate_svg(genome):
+    treatments = [
+        {'function': single_color, 'weight': 1},
+        {'function': two_color, 'weight': 4},
+        {'function': stripes, 'weight': 6},
+    ]
+    update_context_with_treatment = genome.weighted_choice(treatments)['function']
+    emoji = genome.emoji()
```
Osmose: Nit: No need to store this in a variable, you can just call it directly in the dict below.

recipe-server/normandy/base/color.py
```diff
@@ -0,0 +1,31 @@
+"""This module implements the Color class for Shield identicons."""
+
+
+class Color:
```
Osmose: This class is well-contained and useful. Nice work!

recipe-server/normandy/recipes/api/v2/views.py
```diff
@@ -23,6 +24,8 @@
     RecipeRevisionSerializer,
     RecipeSerializer,
 )
+import normandy.recipes.api.v2.shield_identicon as shield_identicon
```
Osmose: This is equivalent to a `from` import, since we're not renaming the module: `from normandy.recipes.api.v2 import shield_identicon`. We should use that.

Also, these imports should follow our import style guidelines (that no one probably told you about that we should have): http://mozweb.readthedocs.io/en/latest/reference/python-style.html#import-statements

recipe-server/normandy/recipes/api/v2/views.py
```diff
@@ -222,3 +225,15 @@ def close(self, request, pk=None):
         approval_request = self.get_object()
         approval_request.close()
         return Response(status=status.HTTP_204_NO_CONTENT)
+
+
+class IdenticonView(views.APIView):
```
Osmose: This is so much easier to follow, thanks!

recipe-server/normandy/base/tests/test_genome.py
```diff
@@ -0,0 +1,61 @@
...
+        ]
+        assert emoji_values == ['😅', '🐯', '😈']
+
+    def test_color(self, genome):
+        color_values = [
+            genome.color().get_rgb_color(),
+            genome.color().get_rgb_color(),
+            genome.color().get_rgb_color(),
+        ]
+        print(color_values)
```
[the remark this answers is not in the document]
marcrowo: 🙈

recipe-server/normandy/base/genome.py
```diff
@@ -0,0 +1,88 @@
...
+            Color((0x6c, 0x71, 0xc4)),
+            Color((0x26, 0x8b, 0xd2)),
+            Color((0x2a, 0xa1, 0x98)),
+            Color((0x85, 0x99, 0x00)),
+        ]
+
+    def weighted_choice(self, options):
+        """
+        Choose a random object from an array by weight.
+        `options` should be objects with at least a `weight` key.
```
mythmon: Nit: `object` isn't the correct term here. `dict` would be better.

recipe-server/normandy/base/genome.py
```diff
@@ -0,0 +1,88 @@
...
+        sum_weights = sum(o['weight'] for o in options)
+
+        choice = self.randrange(sum_weights)
+        for option in options:
+            choice -= option['weight']
+            if choice <= 0:
+                return option
+        raise Exception('No choices chosen.')
+
+    def letter(self):
```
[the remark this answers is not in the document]
marcrowo: I believe this is correct. I think that this is from a previous version of identicons which used letters instead of emoji.

recipe-server/normandy/recipes/api/v2/shield_identicon.py
```diff
@@ -0,0 +1,169 @@
...
+    def luminance(self):
+        """Determine color luminance based on perceived values."""
+        return ((0.2126 * self.red) +
+                (0.7152 * self.green) +
+                (0.0722 * self.blue))
+
+    def contrasts_well(self, pair_color):
+        """Determine whether self and pair_color contrast well."""
+        base_luminance = self.luminance
+        return abs(base_luminance - pair_color.luminance) > 75
```
Osmose: Nit: Might as well just use `self.luminance` her directly.

recipe-server/normandy/recipes/api/v2/shield_identicon.py
```diff
@@ -0,0 +1,169 @@
...
+        choice = self.randrange(sum_weights)
+        for option in options:
+            choice -= option['weight']
+            if choice <= 0:
+                return option
+        raise Exception('No choices chosen.')
+
+    def emoji(self):
+        """Generates a random emoji."""
+        emojis = ["😄", "😃", "😀", "😊", "😉", "😍", "😘", "😚", "😗", "😙", "😜", "😝", "😛",
```
Osmose: Nit: I totally missed suggesting making this a class property like the colors list, my bad.

recipe-server/normandy/recipes/api/v2/shield_identicon.py
```diff
@@ -0,0 +1,169 @@
...
+                  "🚑", "🚐", "🚲", "🚜", "💈", "🚦", "🚧", "🏮", "🎰", "🗿", "🎪", "🎭", "📍",
+                  "🚩", "💯"]
+        return self.choice(emojis)
+
+    def color(self):
+        """Generates a random color."""
+        return self.choice(self.colors)
+
+
+def pick_pair(genome, base_color):
```
Osmose: Since genome is the first argument, might as well make this a method on the class?

marcrowo: I think that it might make more sense to make this a class method of the Color class instead.

PR #993, opened 2017-08-18
Make implementation URLs more permissive of different hash formats.

mythmon:
This is intended to be a point release branched off from `v57`. It is a temporary measure to make `v57` and v69 compatible enough. I intend to tag the resulting merge of this as `v57.1`.

Note that this is not merging into master, but a branch named `v57-branch`, which currently points to the same place as the `v57` tag. We could delete this branch after `v57.1` is tagged, if we wanted.

CC @relud


mythmon:
TC failure is #806, and this is a server side only change, so we are bypassing it.

PR #997, opened 2017-08-22
#982: Recently Viewed Navigation

### LINKED ISSUE #982: Feature Proposal: Recently Viewed Nav Links
>*Note: This example also includes @mythmon 's [IdentiSHIELDs feature](https://github.com/mozilla/normandy/issues/981)

**Description**
Update the site navigation with `Recently Viewed` sections for recipes/extensions, which contain links to the last ~5 items the user has viewed.

I like the idea of using `localStorage` to save the last viewed links, as well. That way, if a user closes the tab with their session and re-open the site, they can quickly pick up where they left off.

**Goal**
- Eliminates the need for the user to navigate to and reload the Listing page to switch back to a recipe they were just viewing.
- Deter user losing context if they accidentally close their browser/tab.

**Level of Effort**
Low.

andymikulski:
Fixes #982 

screen shot 2017-08-22 at 11 49 33 am

- Adds `recently viewed` list for Recipes
- Adds `session` redux store (+ actions, reducers)
- Adds `QuerySessionInfo`
  - Loads/saves session info to `localStorage`
- Adds tests

This PR currently only supports recipes, but is ready for feedback while I add extension support.


recipe-server/client/control_new/state/app/session/actions.js
```diff
@@ -0,0 +1,71 @@
...
+  return async (dispatch, getState) => {
+    const url = getState().router.pathname;
+
+    dispatch({
+      type: SESSION_INFO_HISTORY_VIEW,
+      item: new Map({ url, caption }),
+    });
+
+    // Automatically save the session when views are added.
+    return saveSession()(dispatch, getState);
```
mythmon: The normal way to deal with thunk action creators would be `dispatch(saveSession())`, which would handle passing `dispatch` and `getState`. It doesn't matter much here, but it provides a consistent API for action creators that I think we should follow.

recipe-server/client/control_new/state/app/session/actions.js
```diff
@@ -0,0 +1,71 @@
...
+  return async dispatch => {
+    const requestId = 'fetch-session-info';
+    dispatch({
+      type: REQUEST_SEND,
+      requestId,
+    });
+
+    // Read the stringified object and convert it (and its nested objects) into
+    // proper immutable objects.
+    let savedHistory = JSON.parse(localStorage.getItem(STORAGE_KEY) || '{}');
```
mythmon: Should this use `localForage` instead of `localStorage`? We already have it as a dependency, and you already have all the async machinery in these action creators.

recipe-server/client/control_new/tests/state/extensions/test_reducers.js
```diff
@@ -1,29 +1,103 @@
...
   });
 
-  it('should handle EXTENSION_RECEIVE', () => {
-    expect(extensionsReducer(undefined, {
-      type: EXTENSION_RECEIVE,
-      extension,
-    })).toEqual({
-      ...INITIAL_STATE,
-      items: INITIAL_STATE.items.set(extension.id, fromJS(extension)),
+  describe('SESSION_INFO_HISTORY_VIEW', () => {
```
mythmon: Should these tests be in their own file, instead of adding on to `extensions/test_reducers.js`?

recipe-server/client/control_new/tests/state/extensions/test_reducers.js
```diff
@@ -1,29 +1,103 @@
...
+
+      expect(is(result.history, new List([fakeItem3, fakeItem2, fakeItem]))).toBe(true);
+
+
+      result = sessionReducer(result, {
+        type: SESSION_INFO_HISTORY_VIEW,
+        item: fakeItem,
+      });
+
+      expect(is(result.history, new List([fakeItem, fakeItem3, fakeItem2]))).toBe(true);
```
mythmon: I like this test. 👍

recipe-server/client/control_new/tests/state/extensions/test_reducers.js
```diff
@@ -1,29 +1,103 @@
...
+      expect(is(result.history, new List([fakeItem2, fakeItem]))).toBe(true);
+
+      result = sessionReducer(result, {
+        type: SESSION_INFO_HISTORY_VIEW,
+        item: fakeItem3,
+      });
+
+      expect(is(result.history, new List([fakeItem3, fakeItem2, fakeItem]))).toBe(true);
+
+
```
mythmon: Not: Extra blank line

recipe-server/client/control_new/tests/state/session/test_reducers.js
```diff
@@ -0,0 +1,29 @@
...
+  ActionFactory,
+} from 'control_new/tests/state/actions';
+
+
+describe('Actions reducer', () => {
+  it('should return initial state by default', () => {
+    expect(actionsReducer(undefined, { type: 'INITIAL' })).toEqual(INITIAL_STATE);
+  });
+
+  it('should handle ACTION_RECEIVE', () => {
```
mythmon: I think something got mixed up with this test file too?

recipe-server/client/control_new/tests/state/session/test_selectors.js
```diff
@@ -0,0 +1,37 @@
...
+import { getAction } from 'control_new/state/app/actions/selectors';
+import {
+  INITIAL_STATE,
+} from 'control_new/tests/state';
+import {
+  ActionFactory,
+} from 'control_new/tests/state/actions';
+
+
+describe('getAction', () => {
```
mythmon: This file seems out of place too.

andymikulski:
Whoops, I didn't realize my test files were all disorganized, sorry about that! I'm surprised I didn't notice..

andymikulski:
>I also think that the combination of localstorage and all the request machinery is a bit odd.

My reasoning behind using `local[St|F]orage` in conjunction with the `Query` component is to keep a uniform approach of requesting/receiving data on the front end. For instance, with this approach we can use the localStorage request ID with `LoadingOverlay`s to hide/show loading spinners in relevant sections. Another application is if we (for some reason) decided to store the session info server-side; we simply need to swap out the `localStorage` calls in the Query component with an `apiFetch` call, and the site continues to chug along as normal.

Unfortunately, we _do_ have to manually dispatch the `REQUEST_` events to keep in line with those conventions, but IMO that's not super weird.

mythmon:
> we can use the localStorage request ID with LoadingOverlays to hide/show loading spinners in relevant sections

This is actually ties to my main issue: since localStorage is a synchronous API, there will never be a chance to show a loading spinner here. LocalStorage blocks until the data is available, so React doesn't get a chance to re-render. If this were localForage, which is an asynchronous API, it would be fine. I tested this by artificially making the localStorage fetch take several seconds (by fetching 10000 times form local storage). This just makes the page hang on a white screen for a while.

Having the flexibility to switch to another backend is nice, but would also be easy to add later. I don't think we should use request/response for synchronous actions.

andymikulski:
>localStorage is a synchronous AP

Fair point. Do you have objections to switching to `localForage` and keeping the `REQUEST_` business in place, or do you feel that mechanic should be removed entirely?

andymikulski:
Wow, I really messed that one up. Looks like I copy/pasted some files with the intent of using them as templates, then just overwrote the originals instead. It's all straightened out now, so this is ready for re-review.

andymikulski:
@mythmon This now has extension support, localForage, and proper tests. Sorry for the large PR!

andymikulski:
Tagging @rehandalal in to check out the front end stuff if he's interested.

recipe-server/client/control_new/tests/state/session/test_reducers.js
```diff
@@ -0,0 +1,102 @@
...
+      type: SESSION_INFO_RECEIVE,
+      history: [1, 2, 3],
+    })).toEqual({ history: new List([1, 2, 3]) });
+  });
+
+  describe('SESSION_INFO_HISTORY_VIEW', () => {
+    it('should add an item to the state history', () => {
+      const url = 'mozilla.com';
+      const caption = 'test';
+      const fakeItem = new Map({ type: 'test', url, caption });
```
rehandalal: I think we should be using a factory to generate this data.

recipe-server/client/control_new/tests/state/session/test_selectors.js
```diff
@@ -0,0 +1,56 @@
...
+import { getSessionHistory } from 'control_new/state/app/session/selectors';
+import {
+  INITIAL_STATE,
+} from 'control_new/tests/state/session';
+
+const historyItem = (id, category = 'recipe') =>
+  new Map({ url: id, caption: 'caption', category });
+
+let lastId = 0;
+const generateFakeItems = (count, category) => {
```
rehandalal: This should be a factory like:
https://github.com/mozilla/normandy/blob/master/recipe-server/client/control_new/tests/state/actions/index.js#L12

mythmon:
My feedback is all handled here. Thanks for making those changes. I agree with keeping the request machinery since it is using an async API now.

TC failue is #1015, and can be fixed by rebasing on master.

> Sorry for the large PR!

I think this is a fine size, and nothing to apologize for. It is nicely focused on one feature, and every change here was needed for that feature. It is also well tested, which makes it bigger, but not unnecessarily so. Nice work 🎉

andymikulski:
@rehandalal I've switched over to using a `SessionFactory` and `faker` etc. Check it out and let me know if I've set it up correctly, though I think I've got it.

Thanks!

recipe-server/client/control_new/tests/state/session/test_selectors.js
```diff
@@ -3,17 +3,16 @@ import { List, Map, is } from 'immutable';
...
   INITIAL_STATE,
+  SessionFactory,
 } from 'control_new/tests/state/session';
 
-const historyItem = (id, category = 'recipe') =>
-  new Map({ url: id, caption: 'caption', category });
-
-let lastId = 0;
-const generateFakeItems = (count, category) => {
+const itemFactory = (count, category) => {
```
rehandalal: This name becomes a bit confusing since it's not actually a "factory" per se.

Maybe something like `bulkGenerateSessions`.

recipe-server/client/control_new/tests/state/session/test_selectors.js
```diff
@@ -3,17 +3,16 @@ import { List, Map, is } from 'immutable';
...
   const items = [];
   for (let idx = 0; idx < count; idx++) {
-    items.push(historyItem(++lastId + idx, category));
+    const newItem = SessionFactory.build({ category });
+    items.push(new Map(newItem));
```
rehandalal: I would just do `fromJS` instead of `new Map`.

andymikulski:
I've addressed @rehandalal 's feedback, so I think this should be landable once CI finishes.

andymikulski:
CI failures appear unrelated.

PR #1000, opened 2017-08-23
Fix mock-server to work with signed actions.

Osmose:
Since actions are signed now, we can't modify the URLs in the actions since
that would invalidate the signature. Instead, we switch to using the Django
test client for making API responses, which allows us to temporarily modify the
CDN_URL setting to the URL prefix we need for a testcase.

This is urgent as it is blocking QA from finishing.

### DIFF
--- a/mock-recipe-server/conftest.py
+++ b/mock-recipe-server/conftest.py
@@ -12,4 +12,4 @@ def root_path(request):
     build_dir = request.config.getoption('--mock-server-dir')
     if not build_dir:
         pytest.fail('--mock-server-dir was not provided, but is required for this test')
-    return APIPath(build_dir, 'http://example.com/')
+    return APIPath(build_dir)
--- a/mock-recipe-server/docker-compose.yml
+++ b/mock-recipe-server/docker-compose.yml
@@ -8,7 +8,6 @@ services:
     links:
       - database
       - autograph
-      - proxy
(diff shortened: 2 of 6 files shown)


mythmon:
```
___________________________ test_testcase_difference ___________________________

root_path = <utils.APIPath object at 0x7f0f8b6fb518>

    def test_testcase_difference(root_path):
        """Ensure that different testcases output different data."""
        recipes = set()
    
        testcase_paths = (
            APIPath(path, 'http://example.com')
            for path in root_path.path.iterdir() if path.is_dir()
        )
        for testcase_path in testcase_paths:
>           recipe_path = testcase_path.add('api', 'v1', 'recipe')

/mock-recipe-server/test_mock_server.py:16: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <utils.APIPath object at 0x7f0f88dc8b38>, paths = ('api', 'v1', 'recipe')

    def add(self, *paths):
        """Add segments to the current URL."""
>       return APIPath(self.base_path, self.segments + list(paths))
E       TypeError: must be str, not list

/mock-recipe-server/utils.py:33: TypeError
```

Osmose:
Updated!

PR #1001, opened 2017-08-23
recipe-server: Do some light checks of uploaded study XPIs.

mythmon:
This checks that uploaded XPIs:

* are zip-type files
* appear to be signed (but does not check the signatures)
* if they are web extensions
  * their manifest is JSON
  * they have a webextension ID


recipe-server/normandy/studies/api/v2/fields.py
```diff
@@ -0,0 +1,54 @@
...
+
+        if self._is_webext(xpi_zip):
+            self._has_webext_id(xpi_zip)
+        self._is_signed(xpi_zip)
+
+        return super().to_internal_value(data)
+
+    def _is_webext(self, xpi_zip):
+        files = xpi_zip.namelist()
+        if 'manifest.json' not in files or 'bootstrap.json' in files:
```
Osmose: Did you mean `bootstrap.js`?

recipe-server/normandy/studies/tests/api/v2/test_api.py
```diff
@@ -51,3 +53,50 @@ def test_detail_sets_no_cookies(self, api_client):
...
+    def _upload_extension(self, api_client, path):
+        with open(path, 'rb') as f:
+            res = api_client.post('/api/v2/extension/', {
+                'name': 'test extension',
+                'xpi': f,
+            }, format='multipart')
+        return res
+
+    def test_upload_works_webext(self, api_client):
+        path = './normandy/studies/tests/data/webext-signed.xpi'
```
Osmose: An absolute file path (maybe using `settings.BASE_DIR`) might be a bit more reliable than a relative one.

recipe-server/normandy/studies/api/v2/fields.py
mythmon: Oops, `bootstrap.js` would be more correct. However, I think this whole check is misguided. I don't have any real reason why a file would have both `manifest.json` and `bootstrap.js` so I'm just going to take this out.

Osmose:
TC failure looks unrelated and weird. It actually looks similar to a failure I got on Linux ASAN after merging a fix that hasn't yet made it to this repo. 

You're not touching the add-on at all so I'm fine merging with this.

mythmon:
Test failure is #1015. I'll rebase on master once that is fixed.

PR #1002, opened 2017-08-23
Fix bug 1393257: Stop in-progress studies when opt-out pref changes.

Osmose:
It looks like the `subject` of the notification is being wrapped in some way I'm not familiar with, so I used the data field (which must be a string) instead to transfer the recipe ID for the study. Not entirely pleased with it but it does the trick.

### DIFF
--- a/recipe-client-addon/lib/AddonStudies.jsm
+++ b/recipe-client-addon/lib/AddonStudies.jsm
@@ -97,7 +97,7 @@ async function markAsEnded(db, study) {
   study.active = false;
   study.studyEndDate = new Date();
   await getStore(db).put(study);
-  Services.obs.notifyObservers(study, STUDY_ENDED_TOPIC);
+  Services.obs.notifyObservers(study, STUDY_ENDED_TOPIC, `${study.recipeId}`);
 }
 
 this.AddonStudies = {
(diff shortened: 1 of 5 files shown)


recipe-client-addon/lib/ShieldPreferences.jsm
```diff
@@ -56,11 +65,31 @@ this.ShieldPreferences = {
...
+        Services.prefs.setBoolPref(OPT_OUT_STUDIES_ENABLED_PREF, prefValue);
+        break;
+
+      // If the opt-out pref changes to be false, disable all current studies.
+      case OPT_OUT_STUDIES_ENABLED_PREF:
+        prefValue = Services.prefs.getBoolPref(OPT_OUT_STUDIES_ENABLED_PREF);
+        if (!prefValue) {
+          for (const study of await AddonStudies.getAll()) {
+            if (study.active) {
+              dump(`\nSTOPPING ${study.recipeId}\n`);
```
mythmon: Oops

recipe-client-addon/lib/AddonStudies.jsm
```diff
@@ -97,7 +97,7 @@ async function markAsEnded(db, study) {
   study.active = false;
   study.studyEndDate = new Date();
   await getStore(db).put(study);
-  Services.obs.notifyObservers(study, STUDY_ENDED_TOPIC);
+  Services.obs.notifyObservers(study, STUDY_ENDED_TOPIC, `${study.recipeId}`);
```
gijsk: Yeah, the subject is an nsISupports object. You could add a `wrappedJSObject` property that points to itself (`study.wrappedJSObject = study`) and that might work to extract it on the other side, but it's possible you'd need to also define a `QueryInterface` implementation - not sure off-hand.

Osmose:
Updated!

PR #1004, opened 2017-08-24
Fix #998: Automatic whitespace trimming for preference name and preference string values.

### LINKED ISSUE #998: Strip whitespace from certain recipe fields
Today we launched a pref flip study, but when the pref name was entered into the Normandy system it included a leading space.  When the recipe was sent to clients to enroll, it was not being set correctly and so it appeared as if the experiment had no enrollees even though the match statement was correct.  

We should strip out whitespace from certain fields like pref names and values when a recipe is entered and validated in the Normandy admin.

marcrowo:
~~Fixes #998 by adding a validation rule that checks for any whitespace in the preference name.~~

~~**WIP**:Adding this same validation to string Preference values.~~

This does not work because of Ant funkiness, I will need to write my own validation code and display error messages manually.


andymikulski:
Can we do something where we force fields to run `value.trim()` versus making it a validation error? I'm thinking there are a few spots we could `trim` - on input blur, on value change, technically could be done during validation, etc.

If users are copy-pasting values which leads to this whitespace issue, we should support that process entirely, not add a hurdle for users to overcome when trying to work quickly.

I'm open to suggestions from the rest of the team though.

marcrowo:
@mythmon mentioned adding a "fix this" button if there is a validation error (ie. a whitespace issue).

Osmose:
I'm with @andymikulski on this one; if we don't want to ever accept a preference with whitespace (and I assert that we don't), there's no user benefit to telling them but not fixing it. If we're really worried about confusing the user by changing their input (which I don't think we are) then we can pop a small message saying "whitespace removed" or whatever to let them know.

andymikulski:
That seems like a really weird way to solve the issue. If we have a button that 'fixes' everything, why don't we just automatically do that for the user?

Currently, we're building roadblocks into the system where there doesn't need to be any, and forcing the user to think/act more than is required. Is there an argument against always trimming strings?

recipe-server/client/control_new/components/forms/TrimWhitespaceFormItem.js
```diff
@@ -0,0 +1,34 @@
...
+import { message } from 'antd';
+import React from 'react';
+
+import FormItem from 'control_new/components/forms/FormItem';
+
+@autobind
+export default class TrimWhitespaceFormItem extends React.Component {
+  constructor(props) {
+    super(props);
+    this.shouldNotify = true;
```
andymikulski: I think this should be more React-y and use `state` instead of an instance variable, then we would not need the constructor, only `state = { shouldNotify: true, };`

recipe-server/client/control_new/components/forms/TrimWhitespaceFormItem.js
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  trimValue(event) {
+    const trimmed = event.target.value.trim();
+    if (trimmed !== event.target.value && this.shouldNotify) {
+      this.shouldNotify = false;
```
andymikulski: Use `this.setState` to stay React-y

recipe-server/client/control_new/components/forms/TrimWhitespaceFormItem.js
```diff
@@ -0,0 +1,34 @@
...
+      message.info(
+        'Whitespace was automatically trimmed from the Preference Name.',
```
andymikulski: I almost don't really think any of the `message` stuff is necessary. ♫ It's not unusual ♪ for form fields to auto-trim, especially in this case where we're using it on a field that is whitespace-sensitive.

Regardless, `from the Preference Name` should not be hardcoded here, since this component could be used for other fields.

recipe-server/client/control_new/components/recipes/PreferenceExperimentFields.js
```diff
@@ -51,13 +52,13 @@ export default class PreferenceExperimentFields extends React.Component {
           </FormItem>
         </Col>
         <Col sm={24} md={{ span: 12, offset: 1 }}>
-          <FormItem
+          <TrimWhitespaceFormItem
```
andymikulski: I think I'd prefer turning the whitespace trim into a prop on our already-custom `FormItem` over using a new component. To me, it's cleaner to use one component and write `<FormItem trimWhitespace ... />`. This could be a personal preference though - @mythmon @rehandalal @Osmose  any thoughts?

recipe-server/client/control_new/components/recipes/PreferenceExperimentFields.js
```diff
@@ -219,17 +220,48 @@ export class ExperimentBranches extends React.Component {
...
+    this.shouldNotify = true;
+  }
+
+  handleChange(event) {
+    const { onChange } = this.props;
+    const value = event.target.value;
+    const trimmed = value.trim();
+    if (trimmed !== value && this.shouldNotify) {
+      this.shouldNotify = false;
+      message.info(
```
andymikulski: As above, I think we can get away without notifying on trim.

recipe-server/client/control_new/components/forms/TrimWhitespaceFormItem.js
mythmon: I advised @marcrowo to use an instance variable since `shouldNotify` doesn't affect the rendering of the component, only event handling. Using state for this would produce spurious calls to `render`.

Though if we are removing the messages, it doesn't matter either way.

recipe-server/client/control_new/components/recipes/PreferenceExperimentFields.js
mythmon: +1 to a prop instead of a separate component.

recipe-server/client/control_new/components/forms/FormItem.js
```diff
@@ -1,3 +1,4 @@
+import autobind from 'autobind-decorator';
 import { Form } from 'antd';
```
andymikulski: nit: `autobind` should be after `antd` to stay alphabetical

recipe-server/client/control_new/components/forms/FormItem.js
```diff
@@ -50,21 +55,40 @@ export default class FormItem extends React.Component {
...
+    // but Input passes it via event.target.value.
+    let value = event;
+    if (event && event.target) {
+      value = event.target.value;
+    }
+    return value.trim();
+  }
+
   render() {
+    let { config } = this.props;
```
andymikulski: While technically you can `let` component props, I think it's better to consider them constants and not alter their values.

Looking over how the `config` stuff is set up, I think it could be condensed a bit down to just including the `getValueFromEvent` business in the `getFieldDecorator`, ex:

```
 field = form.getFieldDecorator(name, {
         initialValue,
         rules,
         ...config,
        getValueFromEvent: trimWhitespace  ? this.trimValue : config.getValueFromEvent || null,
})(children);
```

We could still include the console error/warning re: multiple `getValue`s, but I don't think it's necessary to alter/set up a new config object.

recipe-server/client/control_new/components/recipes/PreferenceExperimentFields.js
```diff
@@ -219,17 +220,39 @@ export class ExperimentBranches extends React.Component {
   }
 }
 
+@autobind
 export class StringPreferenceField extends React.Component {
+  static propTypes = {
+    onChange: PropTypes.func.isRequired,
+  };
+
+  constructor(props) {
```
andymikulski: Don't think this is needed anymore since the `shouldNotify` stuff was taken out

recipe-server/client/control_new/components/recipes/PreferenceExperimentFields.js
```diff
@@ -219,17 +220,39 @@ export class ExperimentBranches extends React.Component {
...
 }
 
 export class BooleanPreferenceField extends React.Component {
   static propTypes = {
-    value: PropTypes.any,
+    value: PropTypes.bool,
+  }
+
+  static defaultProps = {
+    value: null,
```
andymikulski: This should probably be `false`.

recipe-server/client/control_new/components/forms/FormItem.js
```diff
@@ -50,21 +55,40 @@ export default class FormItem extends React.Component {
     initialValue: null,
     name: null,
     rules: null,
+    trimWhitespace: false,
   };
 
+  trimValue(event) {
```
andymikulski: After seeing the `WrappedFormItem` stuff in the tests, I realized this function could be made `static` since it would make testing easier, and doesn't actually do anything with the component instance.

recipe-server/client/control_new/tests/components/forms/test_FormItem.js
```diff
@@ -1,7 +1,13 @@
 import React from 'react';
 import { shallow } from 'enzyme';
 
-import FormItem from 'control_new/components/forms/FormItem';
+import WrappedFormItem from 'control_new/components/forms/FormItem';
```
andymikulski: See comment re: `static trimValue`, which should let this stay how it was.

recipe-server/client/control_new/tests/components/forms/test_FormItem.js
```diff
@@ -16,8 +22,21 @@ describe('<FormItem>', () => {
...
 
     expect(wrapper).not.toThrow();
   });
+
+  it('should correctly trim whitespace', () => {
+    const formItemInstance = new FormItem();
+
+    const whitespaceBefore = createFakeEvent('   foobar');
+    const whitespaceAfter = createFakeEvent('foobar   ');
+    const noWhitespace = createFakeEvent('foobar');
```
andymikulski: Super nit: could use another test here against `foo bar` or similar, to ensure spaces aren't removed within strings, only ends.

recipe-server/client/control_new/utils/forms.js
```diff
@@ -139,6 +139,8 @@ export function createForm({ validateFields, ...formConfig }) {
  */
 export function connectFormProps(Component) {
   return class Wrapper extends React.Component {
+    static wrappedComponent = Component;
```
andymikulski: I don't think we should do this. `connect` already uses a similar variable name, plus using the `static` method should remove the need for this entirely.

mythmon: I don't agree. Although it isn't needed in this case, I think that any sort of wrapper component should provide access to the wrapped component, for ease of testing. I suggested @marcrowo use `wrappedComponent` specifically because it matched what `connect` uses.

mythmon:
TC failure is #1015, and can be fixed by rebasing on master.

mythmon:
I'm assuming that when Andy assigned Rehan, that was a mistake, and he meant to reassign this to Marc. Fixing that.

recipe-server/client/control_new/components/recipes/PreferenceExperimentFields.js
marcrowo: I think that this is `null` initially because by default the Preference Value is neither true nor false.

screen shot 2017-08-31 at 10 37 55 am

recipe-server/client/control_new/tests/components/forms/test_FormItem.js
marcrowo: I don't think that making `trimValue()` static will fix this.

The issue is that the `@connectFormProps` decorator wraps the `FormItem` component, and therefore we can't access anything from `FormItem` without using something like `wrappedComponent`.

PR #1006, opened 2017-08-24
#1005: Fix XPI select field not populating correctly

### LINKED ISSUE #1005: Extension select element doesn't populate correctly
The 'extension' select element on the Opt Out Study fields does not populate correctly. Essentially, there is no Query component on the page to actually _load_ the extensions, which means that whatever is currently loaded in store is displayed in the select.

This can be reproduced by opening the `/recipe/new` page, switching to `opt-out-study`, and viewing the XPI select field. The field will be empty, until you navigate to `/extension` and actually load the list.

This should be resolved by adding a Query component near the extension select element on the Opt Out fields.

andymikulski:
Fixes #1005 

Basically, the issue here is that there was no query component on the recipe form. Any values displayed in the select field were taken from whatever was already loaded in the store. This PR adds a `PaginatedSelect` base component which handles loading paginated data and inserting it into a `Select` field.

- Adds `PaginatedSelect` abstract component
  - Given a list and query component, displays a `show more...` button in the list until all pages are loaded and displayed in the select
- Adds `ExtensionSelect`, `RecipeSelect` components. Technically `RecipeSelect` is unnecessary, but I wanted to make sure the base component would work for both use cases.
- Adds `numPages` and `all` properties to Recipes and Extension `listing` stores
- Adds appropriate selectors

screen shot 2017-08-24 at 10 21 36 am

This PR requires tests, but the rest is up for feedback!


recipe-server/client/control_new/state/app/recipes/reducers.js
```diff
@@ -67,11 +68,23 @@ function items(state = new Map(), action) {
...
   switch (action.type) {
-    case RECIPE_PAGE_RECEIVE:
+    case RECIPE_PAGE_RECEIVE: {
+      let totalRecipes = state.get('all', new Set());
+
+      fromJS(action.recipes.results.forEach(result => {
+        totalRecipes = totalRecipes.add(fromJS(result));
+      }));
+
+      totalRecipes = totalRecipes.sort((a, b) => a.get('id') <= b.get('id'));
```
mythmon: This isn't correct. The comparator function passed to `.sort()` [should return -1, 0, or 1](https://facebook.github.io/immutable-js/docs/#/Set/sort), not a boolean. Assuming `id` is a number, `a.get('id') - b.get('id')` would would.

recipe-server/client/control_new/state/app/extensions/reducers.js
```diff
@@ -24,11 +24,22 @@ function items(state = new Map(), action) {
 
 function listing(state = new Map(), action) {
   switch (action.type) {
-    case EXTENSION_PAGE_RECEIVE:
+    case EXTENSION_PAGE_RECEIVE: {
+      let totalExtensions = state.get('all', new Set());
+
+      fromJS(action.extensions.results.forEach(result => {
```
mythmon: What does `fromJS` do here?

recipe-server/client/control_new/components/recipes/OptOutStudyFields.js
```diff
@@ -61,13 +53,15 @@ export default class OptOutStudyFields extends React.Component {
...
-              </Select>
+              <ExtensionSelect disabled={disabled} />
+            </FormItem>
+
+            <FormItem
+              label="Add-on Extension"
+              name="arguments.addonUrl"
+              initialValue={recipeArguments.get('addonUrl', '')}
+            >
+              <RecipeSelect disabled={disabled} />
```
mythmon: While useful for testing, this probably shouldn't be here in the final version.

andymikulski:
>As an alternative approach, would it make sense to load all extensions in this list?

This is the most desirable path, though I was just working with the APIs that are available. If @marcrowo wants to hop on this and create an endpoint that lists all extensions, that'd remove the need for 98% of this PR.

mythmon:
I don't think we should make an endpoint that lists all extensions. Or rather, if we do, we shouldn't *also* have the paginated API. We should pick one. The action creator that fetches extensions can make multiple queries to the API, and keep requesting pages until it has everything.

andymikulski:
Based on yesterday's IRC conversation, it sounds like there are two API solves for this:

1. Change the existing paginated endpoint to allow returning the entire list in one call
1. Add filters to the 'extension listing' endpoint to allow text search (like on the Recipes Listing)

If @marcrowo can handle one of those, this can be wrapped up. I'm a fan of the `filters` approach since it adds more features to other parts of the site for the same amount of work.

rehandalal:
+1 for filters

andymikulski:
Taking this over now that @marcrowo 's filter stuff has landed!

andymikulski:
@mythmon I've added `filter` support for fetching extensions and added the `ExtensionSelect` component, which shows the first page of extensions, unless a text search is provided.

This is ready for review!

recipe-server/client/control_new/components/extensions/ExtensionSelect.js
```diff
@@ -0,0 +1,72 @@
...
+        search,
+      });
+    }, 200);
+  }
+
+  render() {
+    const { search } = this.state;
+    let displayedList = this.props.extensions;
+
+
```
mythmon: Nit: extra line here.

recipe-server/client/control_new/components/extensions/ExtensionSelect.js
```diff
@@ -0,0 +1,72 @@
...
+    if (this.props.isLoadingSearch) {
+      displayedList = new List();
+    }
+
+    return (
+      <div>
+        <QueryMultipleExtensions filters={queryFilters} pageNumber={1} />
+        <Select
+          filterOption={false}
+          placeholder="Select Extension"
```
mythmon: Overall, It isn't obvious that this i s a search field. Can we change this to something like "Search for Extension", to make it clear that users should type? Maybe a search icon would be helpful?

recipe-server/client/control_new/less/main.less
```diff
@@ -188,3 +188,7 @@ fieldset {
     max-width: 300px;
   }
 }
+
+.select-show-more {
+  font-weight: 900;
+}
```
mythmon: I don't think this is used anymore?

recipe-server/client/control_new/state/app/extensions/reducers.js
```diff
@@ -24,11 +24,22 @@ function items(state = new Map(), action) {
...
   switch (action.type) {
-    case EXTENSION_PAGE_RECEIVE:
+    case EXTENSION_PAGE_RECEIVE: {
+      let totalExtensions = state.get('all', new Set());
+
+      action.extensions.results.forEach(result => {
+        totalExtensions = totalExtensions.add(fromJS(result));
+      });
+
+      totalExtensions = totalExtensions.sort((a, b) => a.get('id') <= b.get('id'));
```
mythmon: This sort method won't work. Immutable's `.sort()` method (and JS's as well), want a comparator that returns `0`, `1`, or `-1`.

recipe-server/client/control_new/state/app/extensions/selectors.js
```diff
@@ -7,11 +7,13 @@ export function getExtension(state, id, defaultsTo = null) {
   return state.app.extensions.items.get(id, defaultsTo);
 }
 
-
 export function getExtensionListingCount(state) {
   return state.app.extensions.listing.get('count');
 }
 
+export function getExtensionPageCount(state, defaultsTo = 0) {
```
mythmon: Are we still using this? Do you want to keep it anyways?

recipe-server/client/control_new/state/app/extensions/reducers.js
```diff
@@ -24,11 +24,22 @@ function items(state = new Map(), action) {
...
+
       return state
         .set('count', action.extensions.count)
+        .set('numPages', Math.ceil(action.extensions.count / 10))
         .set('pageNumber', action.pageNumber)
-        .set('results', fromJS(action.extensions.results.map(extension => extension.id)));
+        .set('results', fromJS(action.extensions.results.map(extension => extension.id)))
+        .set('all', totalExtensions);
```
mythmon: Do we still use `all` or `totalExtensions` anywhere? I don't see it.

recipe-server/client/control_new/state/app/recipes/selectors.js
```diff
@@ -42,6 +46,10 @@ export function getRecipeListingPageNumber(state) {
   return state.app.recipes.listing.get('pageNumber');
 }
 
+export function getRecipePageCount(state, defaultsTo = 0) {
```
mythmon: Do you want to keep these recipes changes?

recipe-server/client/control_new/components/extensions/ExtensionSelect.js
andymikulski: How do you feel about this treatment:

screen shot 2017-09-05 at 9 29 06 am

andymikulski:
I've taken out the unnecessary code from earlier changes and corrected feedback items. This should be ready for a final pass!

recipe-server/client/control_new/components/extensions/ExtensionSelect.js
mythmon: 👍

mythmon:
TC failure is #1003, which has been fixed on master. Merging without it.

PR #1007, opened 2017-08-25
Disallow duplicate extension filenames and fix action admin

Osmose:
[Bug 1393751](https://bugzilla.mozilla.org/show_bug.cgi?id=1393751) came up due to extension objects with the same filename. This removes that possibility in prod environments by adding a config from django-storages. I'd like to have marked the field as unique as well, but doing that for FileFields wasn't supported until Django 1.11.

Since we'll need to clean up the data on staging once this deploys, I decided to add Extensions to the Django admin to make the cleanup easier. And, since extension URLs are in the recipe arguments, there's no relational cascade to warn when you delete an extension that is currently being used, so I added a list of recipes that are currently using an extension to the admin. While doing that, I noticed that the similar list we had for the action admin was broken, and fixed that.

### DIFF
--- a/recipe-server/normandy/settings.py
+++ b/recipe-server/normandy/settings.py
@@ -383,6 +383,7 @@ class Production(Base):
     LOGGING_USE_JSON = values.Value(True)
     SECURE_HSTS_SECONDS = values.IntegerValue(31536000)  # 1 year
     DEFAULT_FILE_STORAGE = values.Value('storages.backends.s3boto3.S3Boto3Storage')
+    AWS_S3_FILE_OVERWRITE = False
 
 
 class ProductionReadOnly(Production):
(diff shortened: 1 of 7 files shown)


recipe-server/normandy/settings.py
```diff
@@ -383,6 +383,7 @@ class Production(Base):
     LOGGING_USE_JSON = values.Value(True)
     SECURE_HSTS_SECONDS = values.IntegerValue(31536000)  # 1 year
     DEFAULT_FILE_STORAGE = values.Value('storages.backends.s3boto3.S3Boto3Storage')
+    AWS_S3_FILE_OVERWRITE = False
```
rehandalal: 👍🏼

PR #1009, opened 2017-08-28
recipe-server: Add caching to API indexes.

mythmon:
I opted not to totally split the API tests into v1 and v2 like we do with the recipes, mostly to have a smaller patch. Since the differences so far are very small, I think having a class for each version is enough separation.


rehandalal:
not sure about the TC failures.

mythmon:
Rebased to pick up TC fixes from #1015.

PR #1010, opened 2017-08-28
Bug 1392738: Update how we open the new preferences UI.

Osmose:
This brings us more inline with other implementations within mozilla-central. The leak from [bug 1392738](https://bugzilla.mozilla.org/show_bug.cgi?id=1392738) doesn't appear to affect the test for the old preferences, which leads me to believe that this may also solve that otherwise-puzzling leak.

### DIFF
--- a/recipe-client-addon/content/AboutPages.jsm
+++ b/recipe-client-addon/content/AboutPages.jsm
@@ -142,6 +142,7 @@ XPCOMUtils.defineLazyGetter(this.AboutPages, "aboutStudies", () => {
     registerParentListeners() {
       Services.mm.addMessageListener("Shield:GetStudyList", this);
       Services.mm.addMessageListener("Shield:RemoveStudy", this);
+      Services.mm.addMessageListener("Shield:OpenDataPreferences", this);
     },
 
     /**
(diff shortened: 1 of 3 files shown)


mythmon:
Is the theory here that we aren't properly dealing with the old UI, and that is what the leaking tests are using? so handling both the old and new UIs correctly will fix the leak?

Osmose:
> Is the theory here that we aren't properly dealing with the old UI, and that is what the leaking tests are using? so handling both the old and new UIs correctly will fix the leak?

Flip old and new (it's the new UI test that is leaking), but yes.

recipe-client-addon/content/AboutPages.jsm
```diff
@@ -204,9 +205,13 @@ XPCOMUtils.defineLazyGetter(this.AboutPages, "aboutStudies", () => {
       });
     },
 
-    openOldDataPreferences() {
+    openDataPreferences() {
       const browserWindow = Services.wm.getMostRecentWindow("navigator:browser");
-      browserWindow.openAdvancedPreferences("dataChoicesTab", {origin: "aboutStudies"});
+      if (Services.prefs.getBoolPref(USE_OLD_PREF_ORGANIZATION_PREF)) {
```
gijsk: Add a default value (false), otherwise this could throw.

Osmose:
> The old prefs are being removed in https://bugzilla.mozilla.org/show_bug.cgi?id=1349689 but if this needs to hit 56 then we should take this with the default value as I'm suggesting here - otherwise we can just simplify this code.

Given that this is intermittent anyway, I think it can just ride the trains without uplift. I'll update this now that #1031 landed.

Osmose:
Updated

PR #1012, opened 2017-08-29
recipe-server: Add cache-control headers when redirecting to add slashes.

mythmon:
This will allow us to remove blanket 301 caching from Nginx, replacing it with the automatic caching we get when there are cache control headers.

### DIFF
--- a/recipe-server/normandy/base/middleware.py
+++ b/recipe-server/normandy/base/middleware.py
@@ -1,10 +1,13 @@
 import os
 import re
 
+from django import http
 from django.conf import settings
 from django.utils import timezone
+from django.utils.cache import patch_cache_control
 from django.utils.deprecation import MiddlewareMixin
 from django.contrib.auth.middleware import RemoteUserMiddleware
+from django.middleware.common import CommonMiddleware
 
 from mozilla_cloud_services_logger.django.middleware import (
     RequestSummaryLogger as OriginalRequestSummaryLogger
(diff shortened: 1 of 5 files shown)


recipe-server/normandy/base/middleware.py
```diff
@@ -65,3 +68,27 @@ def is_immutable_file(self, path, url):
...
+    Permanent redirect that also includes cache headers.
+
+    This primarily helps our infrastructure stay reliable and
+    performant, and softens "permanent" redirects to merely be long
+    lived, which helps with maintainability.
+    """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        DAY = 60 * 60 * 24
+        patch_cache_control(self, public=True, max_age=30 * DAY)
```
Osmose: We should probably make the max age a setting, maybe reuse an existing one?

recipe-server/normandy/base/middleware.py
```diff
@@ -65,3 +68,27 @@ def is_immutable_file(self, path, url):
...
+
+
+class NormandyCommonMiddleware(CommonMiddleware):
+    """
+    Overrides CommonMiddleware to customize it to Normandy's.
```
Osmose: Normandy's needs?

recipe-server/normandy/base/tests/test_urls.py
```diff
@@ -0,0 +1,7 @@
+def test_slash_redirects_have_cache_control_headers(client, settings):
```
Osmose: I wonder: Should we test the middleware directly? Asserting this for a single endpoint only feels a little fragile to future changes.

PR #1014, opened 2017-08-30
recipe-server: Reduce the number of DB queries on API listing pages.

mythmon:
tl;dr: O(1) is better than O(n).

Prior to this change, many API listing were making several DB queries per object. On prod this could result in several hundred individual queries on a page, making things pretty slow. Concretely, `/api/v1/recipe/` was doing `9n+1` queries, for a total of 900 queries when there are 100 recipes. After this patch it does a constant 4 queries, regardless of the number of recipes.

In my testing this has has about a 5x speed up querying `/api/v1/recipe/` in a test when there are 1000 recipes. With 100 recipes, the control interface loaded about 20% faster (4.5 to 3.5 seconds) and the `/api/v1/recipe/` endpoint in HTML format loaded about 30% faster (4.5 to 3 seconds). These measurements are all very low sample size with lots of noise, so I wouldn't look too much in to them.

The tests are pretty duplicated, and I'm open for ideas to DRY them up where reasonable.

### DIFF
--- a/recipe-server/normandy/recipes/api/v1/views.py
+++ b/recipe-server/normandy/recipes/api/v1/views.py
@@ -83,7 +83,17 @@ class RecipeFilters(django_filters.FilterSet):
 
 class RecipeViewSet(CachingViewsetMixin, UpdateOrCreateModelViewSet):
     """Viewset for viewing and uploading recipes."""
-    queryset = Recipe.objects.all()
+    queryset = (
+        Recipe.objects.all()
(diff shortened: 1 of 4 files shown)


Osmose:
RE: de-duplicating test logic, perhaps we make this a standalone test generator instead of adding it to the per-view tests? Seems like the URL and the factory are the main things changing between tests.

mythmon:
Fixed.

PR #1018, opened 2017-08-31
Fix #1013 - Log HTTP to HTTPS redirects, and add cache headers.

### LINKED ISSUE #1013: Log HTTP -> HTTPS redirects
In our infrastructure, HTTP -> HTTPS redirects are generally a bug, because the app servers should not be accessible over HTTP. This can cause strange bugs like self-imploding 301 caching. To help track this down, we should add logging any time we would redirect to HTTPS.

There are two ways to do this:

1. Modify the Django middleware that is doing this redirect.
2. Install a separate middleware that logs all HTTP requests that would be redirected, independent of the middleware doing the actual redirecting. This should exclude URLs in `settings.SECURE_REDIRECT_EXEMPT`, which we use to filter out endpoints that are hit directly via HTTP by automation.

Reaching into Django internals is usually a bad idea, so I suggest we go with option 2.

@relud Can you specify what kind of logging you'd like to get out of this?

mythmon:
This depends on #1012. The first unique commit is 9ee2382.

@Osmose and I had previously talked about adding an independent middleware, and not modifying `SecurityMiddleware`. I started implementing this, and in the end it was pretty much a complete reimplementation of `SecurityMiddleware.processRequest`. Instead of duplicating that code, I instead subclassed it. When the original middleware returns a 301, the new middleware logs it and adds cache headers.

### DIFF
--- a/recipe-server/normandy/base/middleware.py
+++ b/recipe-server/normandy/base/middleware.py
@@ -108,7 +108,7 @@ class NormandySecurityMiddleware(SecurityMiddleware):
         response = super().process_request(request)
         if response is not None:
             assert type(response) is http.HttpResponsePermanentRedirect
-            patch_cache_control(response, public=True, max_age=settings.PERMANENT_REDIRECT_CACHE_TIME)
+            patch_cache_control(response, public=True, max_age=settings.HTTPS_REDIRECT_CACHE_TIME)
 
             # Pull out just the HTTP headers from the rest of the request meta
             headers = {
--- a/recipe-server/normandy/base/middleware.py
+++ b/recipe-server/normandy/base/middleware.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import re
 
@@ -8,11 +9,18 @@ from django.utils.cache import patch_cache_control
 from django.utils.deprecation import MiddlewareMixin
 from django.contrib.auth.middleware import RemoteUserMiddleware
 from django.middleware.common import CommonMiddleware
+from django.middleware.security import SecurityMiddleware
 
 from mozilla_cloud_services_logger.django.middleware import (
     RequestSummaryLogger as OriginalRequestSummaryLogger
 )
 from whitenoise.middleware import WhiteNoiseMiddleware
+
+
+DEBUG_HTTP_TO_HTTPS_REDIRECT = 'normandy.base.middleware.D001'
+
+
+logger = logging.getLogger(__name__)
 
 
 def request_received_at_middleware(get_response):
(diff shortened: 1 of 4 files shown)


recipe-server/normandy/base/middleware.py
```diff
@@ -91,3 +99,32 @@ class NormandyCommonMiddleware(CommonMiddleware):
...
+            patch_cache_control(response, public=True, max_age=settings.PERMANENT_REDIRECT_CACHE_TIME)
+
+            # Pull out just the HTTP headers from the rest of the request meta
+            headers = {
+                key.lstrip('HTTP_'): value
+                for (key, value) in request.META.items()
+                if key.startswith('HTTP_') or key == 'CONTENT_TYPE' or key == 'CONTENT_LENGTH'
+            }
+
+            logger.debug(
```
Osmose: Could this collect any personally-identifiable or otherwise sensitive info? CSRF tokens? Auth0 information? I think generally we should avoid storing that kind of data in the logs when possible (although we may already be storing that info in the logs and I'm just not aware or wrong that it's a potential issue).

Perhaps we should consider a whitelist of headers to collect, and no request body data.

recipe-server/normandy/base/tests/test_middleware.py
```diff
@@ -21,3 +33,58 @@ def test_append_slash_redirects_with_cache_headers(self, rf, settings):
         assert res['Location'] == url + '/'
         cache_control = set(res['Cache-Control'].split(', '))
         assert cache_control == {'public', f'max-age={cache_time}'}
+
+
+class TestNormandySecurityMiddleware(object):
+
+    @pytest.fixture
+    def enable_ssl_redirect(self, settings):
```
Osmose: Why's this fixture on the class but the other one isn't?

recipe-server/normandy/base/middleware.py
mythmon: You're right, this can collect personally-identifiable information. This is here specifically because @relud asked for "absolutely everything you can get, especially body and headers".

Also, our logs are already considered sensitive information. Some information gets scrubbed out before it gets put in something like parquet.

@relud does that sound correct?

recipe-server/normandy/base/tests/test_middleware.py
mythmon: Even though both fixtures are only used in this class, this fixture applies only to `NormandySecurityMiddleware` testing, whereas the logger applies to the entire module under test.

recipe-server/normandy/base/middleware.py
relud: yes. also in prod you aren't supposed to be able to hit this page.

relud:
could i also get an option to control the cache time for 301 to https specifically, so that i can set it to 0?

mythmon:
Updated to fix test and add the setting @relud requested (`DJANGO_HTTPS_REDIRECT_CACHE_TIME`).

Osmose:
Taskcluster failure is #1030, feel free to ignore.

mythmon:
Rebased to fix TC failure from #1003. This should be ready to merge once CI finishes.

PR #1024, opened 2017-08-31
Bug 1371350: Delay almost all startup tasks until after sessionstore-windows-restored

Osmose:
This addresses the early JSM loads from [bug 1371350](https://bugzilla.mozilla.org/show_bug.cgi?id=1371350), and also should address a significant amount of the regression from [bug 1390325](https://bugzilla.mozilla.org/show_bug.cgi?id=1390325). My try tests (which were on a mildly-older version of this current PR) showed a confident improvement of 4-5% in the fileio measurements on xperf tests, mostly due to moving the IndexedDB reads (and potentially writes if we're initializing the database) to happen after first paint.

Along with moving init to happen after first paint, this moves the initialization of default preference values for experiments into `bootstrap.js` so that it happens as early as possible, since experiments may need prefs to be set to affect behavior before first paint. In order to avoid file reads during this early part of startup, the values to set are now stored in a preference subtree that is updated whenever experiments start/stop, and when the add-on shuts down. 

Comments are inline for a few unresolved issues, but otherwise this should be ready for review.

### DIFF
--- a/recipe-client-addon/bootstrap.js
+++ b/recipe-client-addon/bootstrap.js
@@ -20,7 +20,7 @@ this.EXPORTED_SYMBOLS = ["Bootstrap"];
 
 const REASON_APP_STARTUP = 1;
 const UI_AVAILABLE_NOTIFICATION = "sessionstore-windows-restored";
-const STARTUP_EXPERIMENT_PREFS_BRANCH = "extensions.shield-recipe-client.startupExperimentPrefs";
+const STARTUP_EXPERIMENT_PREFS_BRANCH = "extensions.shield-recipe-client.startupExperimentPrefs.";
 const PREF_LOGGING_LEVEL = "extensions.shield-recipe-client.logging.level";
 const BOOTSTRAP_LOGGER_NAME = "extensions.shield-recipe-client.bootstrap";
 const DEFAULT_PREFS = {
@@ -67,37 +67,36 @@ this.Bootstrap = {
     const defaultBranch = Services.prefs.getDefaultBranch("");
     const experimentBranch = Services.prefs.getBranch(STARTUP_EXPERIMENT_PREFS_BRANCH);
 
-    for (const childPrefName of experimentBranch.getChildList("")) {
-      const experimentPrefType = experimentBranch.getPrefType(childPrefName);
-      const realPrefName = childPrefName.slice(1); // Remove leading dot
-      const realPrefType = defaultBranch.getPrefType(realPrefName);
+    for (const prefName of experimentBranch.getChildList("")) {
+      const experimentPrefType = experimentBranch.getPrefType(prefName);
+      const realPrefType = defaultBranch.getPrefType(prefName);
 
       if (realPrefType !== Services.prefs.PREF_INVALID && realPrefType !== experimentPrefType) {
-        log.error(`Error setting startup pref ${realPrefName}; pref type does not match.`);
+        log.error(`Error setting startup pref ${prefName}; pref type does not match.`);
         continue;
       }
 
       switch (experimentPrefType) {
         case Services.prefs.PREF_STRING:
-          defaultBranch.setCharPref(realPrefName, experimentBranch.getCharPref(childPrefName));
+          defaultBranch.setCharPref(prefName, experimentBranch.getCharPref(prefName));
           break;
 
         case Services.prefs.PREF_INT:
-          defaultBranch.setIntPref(realPrefName, experimentBranch.getIntPref(childPrefName));
+          defaultBranch.setIntPref(prefName, experimentBranch.getIntPref(prefName));
           break;
 
         case Services.prefs.PREF_BOOL:
-          defaultBranch.setBoolPref(realPrefName, experimentBranch.getBoolPref(childPrefName));
+          defaultBranch.setBoolPref(prefName, experimentBranch.getBoolPref(prefName));
           break;
 
         case Services.prefs.PREF_INVALID:
           // This should never happen.
-          log.error(`Error setting startup pref ${childPrefName}; pref type is invalid (${experimentPrefType}).`);
+          log.error(`Error setting startup pref ${prefName}; pref type is invalid (${experimentPrefType}).`);
           break;
 
         default:
           // This should never happen either.
-          log.error(`Error getting startup pref ${childPrefName}; unknown value type ${experimentPrefType}.`);
+          log.error(`Error getting startup pref ${prefName}; unknown value type ${experimentPrefType}.`);
       }
     }
   },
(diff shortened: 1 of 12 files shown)


recipe-client-addon/bootstrap.js
```diff
@@ -29,69 +38,146 @@ const DEFAULT_PREFS = {
...
+  initExperimentPrefs() {
+    const defaultBranch = Services.prefs.getDefaultBranch("");
+    const experimentBranch = Services.prefs.getBranch(STARTUP_EXPERIMENT_PREFS_BRANCH);
+
+    for (const childPrefName of experimentBranch.getChildList("")) {
+      const realPrefName = childPrefName.slice(1); // Remove leading dot
+      const realPrefType = defaultBranch.getPrefType(childPrefName);
+      const experimentPrefType = experimentBranch.getPrefType(childPrefName);
+
+      // TODO: This never passes?
```
Osmose: For some reason, as far as I can tell, `realPrefType` is _always_ `0` (`PREF_INVALID`) here. I marked the test as skipped just to get the suite to pass, but even if the test sets the pref value and outputs the right type before calling this function, the function still thinks the real pref doesn't exist and doesn't detect the value conflict.

Any insight on why `realPrefType` has the wrong value would be appreciated.

recipe-client-addon/lib/CleanupManager.jsm
```diff
@@ -5,26 +5,49 @@
...
+  }
+
+  async cleanup() {
+    if (this.cleanupPromise === null) {
+      this.cleanupPromise = (async () => {
+        for (const handler of this.handlers) {
+          try {
+            await handler();
+          } catch (ex) {
+            // TODO: Log error in a way that persists after shutdown
```
Osmose: I can't figure out a good way to log an error during shutdown that won't just disappear since the app is closing. Any ideas?

recipe-client-addon/bootstrap.js
mythmon: `realPrefType` comes from querying the default branch. That wouldn't ever have any values for any of our prefs, right? That would explain why this is always `PREF_INVALID`.

Osmose: This is for experimental preferences, not Shield preferences. Experiment preferences would totally have default values.

mythmon: Ah, ok. I'm not sure it is safe to say that ll experiment preferences will have default branches, since some of the experiments are themselves add-ons. But this doesn't explain the `PREF_INVALID` thing.

recipe-client-addon/lib/CleanupManager.jsm
mythmon: I think most logs like this will show up a terminal if Firefox is run from there, so this might not be an issue. Alternatively you could do something complex like jamming log lines in a preference and log them at startup.

recipe-client-addon/bootstrap.js
gijsk: You're passing `childPrefName` instead of `realPrefName`, so instead of asking for, say, "browser.chromeURL", you're asking for ".browser.chromeURL", which doesn't exist.

recipe-client-addon/bootstrap.js
```diff
@@ -29,69 +38,146 @@ const DEFAULT_PREFS = {
...
+      ShieldRecipeClient.startup();
+    }
+  },
+
+  async shutdown(data, reason) {
+    // Wait for async write operations during shutdown before unloading modules.
+    await ShieldRecipeClient.shutdown(reason);
+
+    // In case the observer didn't run, clean it up.
+    Services.obs.removeObserver(this, UI_AVAILABLE_NOTIFICATION);
```
gijsk: This will throw if the observer has already been removed.

recipe-client-addon/bootstrap.js
```diff
@@ -29,69 +38,146 @@ const DEFAULT_PREFS = {
...
+      "lib/RecipeRunner.jsm",
+      "lib/Sampling.jsm",
+      "lib/SandboxManager.jsm",
+      "lib/ShieldPreferences.jsm",
+      "lib/ShieldRecipeClient.jsm",
+      "lib/Storage.jsm",
+      "lib/Uptake.jsm",
+      "lib/Utils.jsm",
+    ].map(m => `resource://shield-recipe-client/${m}`);
+    modules = modules.concat([
```
gijsk: I know you're just moving this, but I think this would be easier to read and faster if written as:

```js
modules.push(
   "resource://shield-recipe-client-content/AboutPages.jsm",
   "resource://shield-recipe-client-vendor/mozjexl.js"
);
```

recipe-client-addon/lib/CleanupManager.jsm
gijsk: Yeah, I wouldn't worry too much about this.

recipe-client-addon/bootstrap.js
```diff
@@ -29,69 +38,146 @@ const DEFAULT_PREFS = {
...
+
+      // TODO: This never passes?
+      if (realPrefType !== Services.prefs.PREF_INVALID && realPrefType !== experimentPrefType) {
+        log.error(`Error setting startup pref ${childPrefName}; pref type does not match.`);
+        continue;
+      }
+
+      switch (experimentPrefType) {
+        case Services.prefs.PREF_STRING:
+          defaultBranch.setCharPref(realPrefName, experimentBranch.getCharPref(childPrefName));
```
gijsk: So, this only sets on the default branch. Are we no longer supporting pref experiments on the user branch at all? Do we need to encode that bit into the experiment branch (with "user." or "default." prefixes or something) ?

Osmose: User prefs persist between restarts, so we don't need to set them on startup. They get set when the experiment starts.

Osmose:
Updated!

recipe-client-addon/bootstrap.js
```diff
@@ -29,69 +38,147 @@ const DEFAULT_PREFS = {
...
 
-  ShieldRecipeClient.startup();
-};
+  initExperimentPrefs() {
+    const defaultBranch = Services.prefs.getDefaultBranch("");
+    const experimentBranch = Services.prefs.getBranch(STARTUP_EXPERIMENT_PREFS_BRANCH);
+
+    for (const childPrefName of experimentBranch.getChildList("")) {
+      const experimentPrefType = experimentBranch.getPrefType(childPrefName);
+      const realPrefName = childPrefName.slice(1); // Remove leading dot
```
mythmon: Would it make sense to put a dot at the end of `STARTUP_EXPERIMENT_PREFS_BRANCH`, so that this isn't needed?

PR #1037, opened 2017-09-05
Fix #1036: Remove OSX-specific close button CSS for Heartbeat bar.

### LINKED ISSUE #1036: Heartbeat close button styling needs fixing for Photon
On Nightly, the close button for the Heartbeat bar looks too small and offset now that the Photon UI changes are landing:

screen shot 2017-09-05 at 2 24 16 pm

I believe the styling is meant to match the close button on tabs, and on hover the close button looks large and centered like the current close tab buttons on hover, so fixing this I think is mostly a matter of fixing the non-hover styling.

Osmose:
Due to changes in browser styles for Photon, the fix is no longer necessary, and
in fact makes the close button look wrong. Without the fix, the close button
looks and behaves the same as the tab bar close buttons.

Here's what the bar looks like on Nightly OSX now:

screen shot 2017-09-05 at 2 24 16 pm

And here's what it looks like with the workaround removed:

screen shot 2017-09-05 at 3 03 03 pm

I tested the bar on Windows and Linux (via mythmon) and the close button looks fine on those platforms.

This is a small enough PR that one review is enough IMO.

### DIFF
--- a/recipe-client-addon/skin/osx/Heartbeat.css
+++ b/recipe-client-addon/skin/osx/Heartbeat.css
@@ -9,16 +9,3 @@ notification.heartbeat {
   border-bottom: 1px solid #C1C1C1 !important;
   height: 40px;
 }
-
-/* In themes/osx/global/notification.css the close icon is inverted because notifications
-   on OSX are usually dark. Heartbeat is light, so override that behaviour. */
-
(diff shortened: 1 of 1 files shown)


Osmose:
I've only observed it on 57 and have no intent of uplifting it. In fact, I'm going to hold off merging this until we do a sync for the performance fixes so we don't have to do a messy uplift.

PR #1038, opened 2017-09-05
Fixes #981: Add identicons to frontend

### LINKED ISSUE #981: Recipe Identicons (IdentiSHIELDs?)

marcrowo:
Based on previous work done in #983.


recipe-server/client/control_new/components/common/ShieldIdenticon.js
```diff
@@ -0,0 +1,29 @@
...
+    size: 64,
+    className: null,
+  }
+
+  render() {
+    const { seed, size, className } = this.props;
+
+    return (
+      <img
+        className={className}
```
andymikulski: Instead of explicitly using the `className` prop, we can just do something like:

```
const { seed, size, ...rest } = this.props;
return (<img src="..." height={size} width={size} {...rest} />)
```

This pulls out `seed` and `size` from props and then groups any other props as `rest`. We pass this down the line so we don't need to worry about things like `className`, `onClick`, etc.

recipe-server/client/control_new/components/common/ShieldTestPage.js
```diff
@@ -0,0 +1,49 @@
+import { Card, Row, Col } from 'antd';
+import autobind from 'autobind-decorator';
+import React from 'react';
+
+import ShieldIdenticon from 'control_new/components/common/ShieldIdenticon';
+
+
+@autobind
+export default class ShieldTestPage extends React.Component {
```
andymikulski: I'm fine with dropping this page, I don't see a huge need for it after landing the shields. I'm open to opposing thoughts though.

recipe-server/client/control_new/components/recipes/RecipeDetailPage.js
```diff
@@ -48,8 +49,15 @@ export default class RecipeDetailPage extends React.Component {
         <QueryRecipe pk={recipeId} />
         <Row gutter={24}>
           <Col span={16}>
-            <DetailsActionBar />
             <RevisionNotice revision={revision} />
+            <Row type="flex" align="middle">
+              <Col span={4}>
+                <ShieldIdenticon seed={recipeId} letter={revision.getIn(['recipe', 'name'])} />
```
andymikulski: If the `seed` value is just the `recipeId`, we can probably change `seed` to just `id` or something similar. The term `seed` was more appropriate when it was more directly responsible for the shield generation.

recipe-server/client/control_new/components/common/ShieldIdenticon.js
```diff
@@ -0,0 +1,29 @@
...
+        src={`/api/v2/identicon/v1:${seed}.svg`}
```
andymikulski: Currently this has the `v1:` version of the shield endpoint. If we make changes to shields and have 'generations' of them, in the future we will need to determine which version shield to display.

We can avoid this by saving the entire `src` string with the recipe and simply doing something like `` on the front end.  We can do this a few ways, but I think changing what we store requires backend work, so I'm not positive what is the most flexible approach there.

With all that said, I don't think we anticipate changing these shields anytime soon, so maybe this is a non-issue for now.

recipe-server/client/control_new/components/recipes/RecipeListing.js
```diff
@@ -74,7 +75,18 @@ export default class RecipeListing extends React.Component {
...
-          render={RecipeListing.renderLinkedText}
+          render={(text, record) =>
+            (
+              <div className="recipe-listing-name">
+                <ShieldIdenticon
+                  className="shieldicon"
+                  seed={record.id}
+                  size={32}
+                />
+                {RecipeListing.renderLinkedText(text, record)}
```
andymikulski: I don't see any other instances of `renderLinkedText` being called. Perhaps instead of adding the `div.recipe-listing-name` here it can be moved to the `renderLinkedText` function to keep this JSX simple (`render={RecipeListing.renderLinkedText}`) (Note that a different name for the function may be more suitable too).

If `renderLinkedText` is used elsewhere though, disregard this note entirely.

recipe-server/client/control_new/routes.js
```diff
@@ -65,6 +66,10 @@ const routes = {
         crumb: 'Edit Extension',
       },
     },
+    '/shield-demo': {
```
andymikulski: Again, not positive we need this page.

recipe-server/client/control_new/components/common/ShieldIdenticon.js
andymikulski: Having thought on it for a few minutes, it makes more sense to me to save the resulting svg URL and then using that to recall the shield icon.

An idea I liked for the 'create recipe' flow was to add a random shield generator, the premise being that if the user wanted to cycle through a few shields before settling on one they could do so. If we wanted to incorporate something like that, we would need to allow/save `seed` values that weren't just recipe ID's. The current setup would only allow for one shield which is tied to the ID.

andymikulski: @mythmon @Osmose @rehandalal If y'all have any thoughts on this, I'd be interested to hear 'em!

recipe-server/client/control_new/components/common/ShieldIdenticon.js
mythmon: I advised @marcrowo not to do this, since we've had instances in the past of too many props being passed down to HTML components, causing errors. This was mostly with regards to redux-form. I think having a whitelist of props to pass is a better policy when dealing with HTML components, which are strict about which props they accept.

recipe-server/client/control_new/components/common/ShieldIdenticon.js
mythmon: Now that #992 has landed, we now [have a `identicton_seed` attribute on recipes](https://github.com/mozilla/normandy/pull/992/files#diff-4b47ef5bd0860f8070da44d56085d838R110). We should be using that instead of `id`.

recipe-server/client/control_new/components/recipes/RecipeListing.js
marcrowo: `renderLinkedText` is also used for the Action column.

recipe-server/client/control_new/components/recipes/RecipeDetailPage.js
marcrowo: Since this should be using the `identicon_seed` attribute from the recipe, I think that `seed` is appropriate.

andymikulski:
I've fixed some minor issues with the shields (they weren't centered in the viewport and were missing drop shadows), as well as added an `IdenticonField` which lets users cycle through randomly generated identicons on the Recipe Form in order to choose one they like.

This needs a handful of tests, but it's getting a little large, so I'm going to say this is **ready for feedback**!

IdenticonField on the edit recipe page
IdenticonField on the edit recipe page

On hover, users can see a larger view of the icon for inspection.
On hover, users can see a larger view of the icon for inspection.

recipe-server/client/control_new/components/common/ShieldIdenticon.js
```diff
@@ -1,29 +1,91 @@
...
   render() {
     const { seed, size, className } = this.props;
 
+    if (typeof seed === 'undefined') {
+      return null;
+    }
+
+    // In order to prevent the the page jumping while the image is loading, we'll
+    // squish and hide it until it has loaded.
+    const imgStyle = { height: 0, width: 0, position: 'absolute' };
```
mythmon: I don't understand this. Wouldn't this cause the page to page to jump around when the image loads? That's what I see.

recipe-server/client/control_new/components/forms/IdenticonField.js
```diff
@@ -0,0 +1,126 @@
...
+      return;
+    }
+
+    next = history[newIndex];
+
+    let newHistory = [...history];
+    if (!next) {
+      next = IdenticonField.generateSeed();
+      if (newHistory.indexOf(next) === -1) {
+        newHistory = [...history, next];
```
mythmon: `history` was already cloned into `newHistory` a few lines above. There isn't a reason to do it again here. Alternately, should this be using `Immutable.List`?

recipe-server/client/control_new/components/forms/IdenticonField.js
```diff
@@ -0,0 +1,126 @@
...
+
+  handleChange(direction) {
+    let next;
+    const {
+      index,
+      history,
+    } = this.state;
+    const newIndex = index + direction;
+
+
```
mythmon: nit: blank line

recipe-server/client/control_new/components/forms/IdenticonField.js
mythmon: `newHistory` is already a cloned object, because of `let newHistory = [...history]`. No point in doing i t again here.

Alternatively, would it make sense to use Immutable here?

recipe-server/client/control_new/components/forms/IdenticonField.js
```diff
@@ -0,0 +1,126 @@
...
+  state = {
+    history: [],
+    index: -1,
+  };
+
+  componentDidMount() {
+    const { value } = this.props;
+
+    if (value) {
+      this.setState({
```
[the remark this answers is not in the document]
andymikulski: I ended up changing this approach a little bit. The initial state is now set in the constructor, but still allows `willReceiveProps` to update accordingly (for instance, if the icon seed is not loaded on mount but then loads later on). This should avoid the re-render issue you mentioned and avoids disabling the eslint rule.

recipe-server/client/control_new/components/forms/IdenticonField.js
andymikulski: Yep, this should be Immutable - I've converted it over accordingly.

andymikulski:
Added tests and addressed feedback, this is ready for review!

andymikulski:
Having trouble with this after the rebase, going to need some looking over.

andymikulski:
Alright, this should be fixed, just want to make sure CI is happy about it.

mythmon:
Circle failures are related. I think the Python tests failed because the template was updated without updating one of the tests that rely on them.

andymikulski:
Yep, you are correct. I had changed most of it but missed some height/width attributes that needed updating. Fix has been pushed, just waiting on CI.

andymikulski:
Hm. Looks like the test is still failing, but now on another portion of the same file. Not totally sure what the issue is here, @mythmon any thoughts? The issue stems from the differences in `foobar.svg` and `identicon.svg`, though I've applied changes to both files, so I'm not sure if there is a third missing file I need to update or what.

mythmon:
There was some trailing whitespace in the file you changed, I think that's the difference now. If you strip the same whitespace out of the other file, I think that would fix it.

mythmon:
There was some trailing whitespace in the file you changed, I think that's the difference now. If you strip the same whitespace out of the other file, I think that would fix it.

PR #1041, opened 2017-09-06
PureComponents

andymikulski:
- Converts almost every React component we have into a `PureComponent`.
  - The exception being some of the wrapped form inputs.

With this approach, our pages will render far fewer times than necessary, which is a decent performance increase. For upcoming features such as graphs, this will be a great. I'd like to point out this was very easy to accomplish because we went with Immutable!

### DIFF
--- a/recipe-server/client/control_new/components/App.js
+++ b/recipe-server/client/control_new/components/App.js
@@ -14,7 +14,7 @@ import QueryServiceInfo from 'control_new/components/data/QueryServiceInfo';
 const { Content, Header, Sider } = Layout;
 
 
-export default class App extends React.Component {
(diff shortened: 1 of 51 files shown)


andymikulski:
@mythmon Mainly, the only components that gave me issue were form-related, which I expected. Essentially, form inputs wouldn't update, acting as if they were disabled.

I narrowed the issue down to the form wrappers we have in `utils/forms`, which wraps all of the inputs. Returning `PureComponent`s there would not trigger updates, I'm assuming since Ant is not using immutable data structures internally to track form stuff.

Outside of those wrapped components everything seemed fine, which makes sense since we use immutable everywhere else.

PR #1042, opened 2017-09-06
 #996: Process recipes in bulk

### LINKED ISSUE #996: Listing Pages: Long load times
>Note: This is something that has been discussed previously, but I feel needs an actual ticket to track these issues.

Currently, we reload the recipe/extension listing pages every time the user navigates to them. We designed the system this way to prevent stale data from appearing as the latest data. However, this leads us into some UX issues:

- Recipes landing page takes an average of **~3.4s** to load (tracking the `fetchFilteredRecipesPage` call)
  - For reference, 100ms is considered the threshold between whether a user can notice a delay or not
  - Also note, this is when viewing a listing page without any filters applied at all
- UI hangs during loading
  - This prevents clicking ANY interface element until the app is ready.

This UX is pretty subpar IMO - navigating between multiple recipes takes 5s+, even with recently viewed links in the nav (#982)

Things to investigate:
- Handling `_RECEIVED` actions in bulk
  - Currently, hitting the recipes list causes a bunch of `[RECIPE|REVISION]_RECEIVED` actions to fire. I suspect that the approach of iterating over each action plus the use of redux-logger to console.log reduced actions is causing a considerable chunk of the slowdown.
  - There are several redux things we can experiment with, IMO the easiest way is to simply add a `RECIPE_BULK_RECEIVED` or similar, which would update the redux stores in one fell swoop (versus looping over an array and `dispatch`ing a bunch).
- Caching API results
  - Since we want to ensure the data is fresh, caching seems like something that's off the table. However, if we could do something like assign a hash to an API response, we could use that for cachebusting.
  - Ex: user loads recipe listing, gets hash `abc123` back with the data. user does stuff, navigates back to recipe page, hits the _hash only_ endpoint to see that the hash is now `bcd345`, which means it should request the full lookup.
  - With this approach, we at least remove the need for waiting for the listing call to respond, and keep data fresh.

I'm open to ideas on how to approach this, but IMO looking at the redux side of things may help with a good chunk of the slow down.

andymikulski:
Fixes #996 - Long listing page load times

- Replaces firing multiple `dispatch`ed actions with one 'bulk' handler action
  - Specifically, instead of firing a bunch of `recipeReceived`s, we fire one action with type `RECIPE_PAGE_RECEIVE` and process all the recipes in one go in the reducer.
- This approach cuts the measured front end 'load time' by ~40-60% and greatly improves the UX around the two components
- This PR contains the changes for bulk processing the recipe listing and recipe histories.

### DIFF
--- a/recipe-server/client/control/state/app/recipes/actions.js
+++ b/recipe-server/client/control/state/app/recipes/actions.js
@@ -46,27 +46,6 @@ export function fetchRecipe(pk) {
   };
 }
 
-
-export function fetchRecipesPage(pageNumber = 1) {
-  return async dispatch => {
-    const requestId = `fetch-recipes-page-${pageNumber}`;
-    const recipes = await dispatch(makeApiRequest(requestId, 'v2/recipe/', {
-      data: { page: pageNumber },
-    }));
-
-    recipes.results.forEach(recipe => {
-      dispatch(recipeReceived(recipe));
-    });
-
-    dispatch({
-      type: RECIPE_PAGE_RECEIVE,
-      pageNumber,
-      recipes,
-    });
-  };
-}
-
-
 export function fetchFilteredRecipesPage(pageNumber = 1, filters = {}) {
   return async dispatch => {
     const filterIds = Object.keys(filters).map(key => `${key}-${filters[key]}`);
@@ -77,10 +56,6 @@ export function fetchFilteredRecipesPage(pageNumber = 1, filters = {}) {
         page: pageNumber,
       },
     }));
-
-    recipes.results.forEach(recipe => {
-      dispatch(recipeReceived(recipe));
-    });
 
     dispatch({
       type: RECIPE_PAGE_RECEIVE,
@@ -159,10 +134,6 @@ export function fetchRecipeHistory(pk) {
   return async dispatch => {
     const requestId = `fetch-recipe-history-${pk}`;
     const revisions = await dispatch(makeApiRequest(requestId, `v2/recipe/${pk}/history/`));
-
-    revisions.forEach(revision => {
-      dispatch(revisionReceived(revision));
-    });
 
     dispatch({
       type: RECIPE_HISTORY_RECEIVE,
--- a/recipe-server/client/control/state/app/recipes/reducers.js
+++ b/recipe-server/client/control/state/app/recipes/reducers.js
@@ -39,13 +39,13 @@ function history(state = new Map(), action) {
 }
 
 const formatRecipe = recipe =>
-  recipe
-    .set('action_id', recipe.getIn(['action', 'id'], null))
-    .set('latest_revision_id', recipe.getIn(['latest_revision', 'id'], null))
-    .set('approved_revision_id', recipe.getIn(['approved_revision', 'id'], null))
-    .remove('action')
-    .remove('latest_revision')
-    .remove('approved_revision');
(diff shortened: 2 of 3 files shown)


recipe-server/client/control_new/state/app/recipes/reducers.js
```diff
@@ -38,23 +38,31 @@ function history(state = new Map(), action) {
...
-      recipe = recipe
-        .set('action_id', recipe.getIn(['action', 'id'], null))
-        .set('latest_revision_id', recipe.getIn(['latest_revision', 'id'], null))
-        .set('approved_revision_id', recipe.getIn(['approved_revision', 'id'], null))
-        .remove('action')
-        .remove('latest_revision')
-        .remove('approved_revision');
+      recipes.forEach(receivedRecipe => {
+        newState = newState.set(receivedRecipe.get('id'), formatRecipe(receivedRecipe));
+      });
```
mythmon: As a drive comment, Immutable's [`withMutations` helper](https://facebook.github.io/immutable-js/docs/#/Map/withMutations) might help speed this up. It temporarily makes the data structure directly mutable, so you could write this as:

```js
const recipes = fromJS(action.recipes.results);
return state.withMutations(mutState => {
  recipe.forEach(receivedRecipe => {
    mutState.set(receivedRecipe.get('id'), formatRecipe(receivedRecipe));
  });
});
```

rehandalal:
holding off on merging in case you wanted to go with @mythmon's suggestion.

andymikulski:
@rehandalal I'll switch that out now, shall I merge once that change is put in?

recipe-server/client/control_new/state/app/recipes/reducers.js
```diff
@@ -38,23 +38,30 @@ function history(state = new Map(), action) {
...
-        .set('latest_revision_id', recipe.getIn(['latest_revision', 'id'], null))
-        .set('approved_revision_id', recipe.getIn(['approved_revision', 'id'], null))
-        .remove('action')
-        .remove('latest_revision')
-        .remove('approved_revision');
+    case RECIPE_PAGE_RECEIVE: {
+      const recipes = fromJS(action.recipes.results);
 
-      return state.set(action.recipe.id, recipe);
+      return state.withMutations(mutState => {
```
rehandalal: I think this is way weirder syntax. It's harder to read to me.

I was under the impression the `withMutations` stuff was going to apply to the `formatRecipe` and `formatRevision` methods.

rehandalal:
I believe the TC failure is unrelated.

recipe-server/client/control_new/state/app/recipes/reducers.js
mythmon: This is what I had in mind. What do you find harder to read about it?

Applying this only to the `formatRecipe` and `formatRevision` functions would have some of the performance  benefits, but wrapping the entire reducer transaction in `withMutation` (like done here) is better.

rehandalal:
needs to have latest master merged in (or be rebased).

andymikulski:
I rebased onto master, and also added a small commit that adds `withMutations` to the `format[Recipe|Revision]` functions. @rehandalal , mind double checking https://github.com/mozilla/normandy/pull/1042/commits/f3e7b3506bfed1dcafe170547c418c50c328feb4 ?

rehandalal:
TC failure is unrelated and happening on master.

PR #1047, opened 2017-09-07
#1039: Minified vendors in dev

### LINKED ISSUE #1039: Using minified copy of the development build of React and Redux
In prod, we get these errors in the new control interface:

> Warning: It looks like you're using a minified copy of the development build of React. When deploying React apps to production, make sure to use the production build which skips development warnings and is faster. See https://fb.me/react-minification for more details.
vendor.bundle.js:17:11414

> You are currently using minified code outside of NODE_ENV === 'production'. This means that you are running a slower development build of Redux. You can use loose-envify (https://github.com/zertosh/loose-envify) for browserify or DefinePlugin for webpack (http://stackoverflow.com/questions/30030031) to ensure you have the correct code for your production build.

This is likely related to our recent webpack work. Switching to a production build of React should help with our performance problems in #996.

andymikulski:
Fixes #1039 

- Adds a `production` check to the vendor webpack config which determines whether to compress or not
- ~Removes the `prod` flag from the npm `karma` script. I believe that our CI already runs everything with the prod flag, but I want to confirm on this PR.~
- ~Adds a `package-lock.json` because it keeps showing up in my repo and [npm says it should be tracked](https://docs.npmjs.com/files/package-lock.json)~

### DIFF
--- a/recipe-server/vendor.webpack.config.js
+++ b/recipe-server/vendor.webpack.config.js
@@ -5,6 +5,8 @@ var path = require('path');
 var UglifyJSPlugin = require('uglifyjs-webpack-plugin');
 var webpack = require('webpack');
 
+var production = process.env.NODE_ENV === 'production';
+
 var settings = require('./package.json');
 
 var ignoredVendors = [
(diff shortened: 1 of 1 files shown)


andymikulski:
@mythmon I've updated the PR to only change the prod flags and the prod-only uglifying. Ready for review.

mythmon:
TC failure is #757. This PR doesn't involve the add-on, so merging without TC passing.

PR #1053, opened 2017-09-08
#1051: Fix ExtensionSelect placeholder

### LINKED ISSUE #1051: SearchSelection for Extensions doesn't show placeholder
On stage and dev, the extension input doesn't show the placeholder that makes it clear that it is a search field, and not a simple drop down. This is confusing QA.

The placeholder element is in the DOM, but with `display: none` set on the element.

screenshot-2017-9-8 shield control panel

andymikulski:
Fixes #1051 

- Corrects bug where passing in inherited props broke the displayed placeholder.
  - Specifies which props to inherit, adds them to `propTypes` and `defaultProps` as appropriate
- Adds tests to determine that `ExtensionSelect` works and that the placeholder appears.
- Adds a `tests/mockStore.js` util file that allows us to wrap a component in a `Provider` with a mock `store`. This was needed in order to `mount` the component, since there is a nested Query component which is `connect`ed and requires a store etc.

### DIFF
--- a/recipe-server/client/control_new/components/extensions/ExtensionSelect.js
+++ b/recipe-server/client/control_new/components/extensions/ExtensionSelect.js
@@ -27,12 +27,14 @@ export default class ExtensionSelect extends React.Component {
     isLoadingSearch: PropTypes.bool.isRequired,
     onChange: PropTypes.func,
     size: PropTypes.oneOf(['small', 'large']),
+    value: PropTypes.any,
   };
 
   static defaultProps = {
     disabled: false,
     onChange: null,
     size: 'default',
+    value: null,
   };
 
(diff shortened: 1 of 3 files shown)


mythmon:
This looks like it should fix the problem @andymikulski described in irc.

andymikulski:
@Osmose pointed out that existing recipes were no longer showing the correct display value with this fix. I've pushed up another commit that resolves those final issues.

@mythmon , it's probably worth re-checking this PR. Thanks!

Osmose:
I've verified locally that the display value for existing recipes now works properly.

rehandalal:
This doesn't appear to touch any code that TC should care about, so ignoring the failure.

PR #1060, opened 2017-09-11
Fix #1058: Bind PreferenceExperiment driver functions.

### LINKED ISSUE #1058: Firefox Nightly: this.saveStartupPrefs is not a function
When Firefox Nightly starts, the following error is dumped to the console:

> this.saveStartupPrefs is not a function at   PreferenceExperiments.jsm:302

To fix this, add `.bind(PreferenceExperiments)` to line 175 (and also lines 176 - 180 just in case).

https://github.com/mozilla/normandy/blob/a2b399f966d15154fb751a2fad5cba2c55ede068/recipe-client-addon/lib/NormandyDriver.jsm#L173-L181

Osmose:
Mild oversight on my part.

### DIFF
--- a/recipe-client-addon/lib/NormandyDriver.jsm
+++ b/recipe-client-addon/lib/NormandyDriver.jsm
@@ -172,12 +172,23 @@ this.NormandyDriver = function(sandboxManager) {
 
     // Preference Experiment API
     preferenceExperiments: {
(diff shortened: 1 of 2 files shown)


Osmose:
This will need uplift to beta when we sync it since `saveStartupPrefs` got uplifted as part of the performance improvements landed last week.

PR #1062, opened 2017-09-14
Deprecate the old UI

rehandalal:
phew!

r?


Osmose:
TC failures are happening on master and unrelated, feel free to ignore them.

rehandalal:
TC failed but merged because TC is failing on master and also these code changes should not touch anything that affects the addon build.

PR #1064, opened 2017-09-19
Fix route that is missing trailing slash

rehandalal:
quick r?

### DIFF
--- a/recipe-server/client/control/routes.js
+++ b/recipe-server/client/control/routes.js
@@ -27,8 +27,10 @@ const routes = {
         crumb: 'Recipes Listing',
       },
       '/new': {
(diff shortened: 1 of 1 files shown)


recipe-server/client/control/routes.js
```diff
@@ -27,8 +27,10 @@ const routes = {
         crumb: 'Recipes Listing',
       },
       '/new': {
-        component: CreateRecipePage,
-        crumb: 'New Recipe',
+        '/': {
+          component: CreateRecipePage,
+          crumb: 'New Recipe',
+        }
```
andymikulski: I hit approve as I noticed this is missing a comma. I imagine CI will fail due to lint

PR #1071, opened 2017-09-22
Fix #957: Use UpdateUtils.jsm to determine the update channel.

### LINKED ISSUE #957: Use UpdateUtils.jsm to determine update channel
Currently, we use `Services.appinfo.defaultUpdateChannel` to determine the update channel used in filtering. However, this value changes from `beta` to `release` when the RC builds hit the Beta channel. This bit us by preemptively stopping a preference experiment on Beta.

Given that we ship changes to Firefox, we should probably use a more stable indicator for update channel. The update system uses [UpdateUtils.jsm](http://searchfox.org/mozilla-central/source/toolkit/modules/UpdateUtils.jsm) to get the update channel. However, this adds a bunch of extra modifiers to the channel. We should take this into account when figuring out how to write filters with this value:

- Should we add some transforms to compare the base channel?
- Should we manually remove the extra annotations and expose that as the channel value?

@Mossop @gijsk Can one of you double-check my thinking on this?

Osmose:
Small enough not to need two reviewers.

@gijsk I couldn't think of any good tests to add for this; mocking `getUpdateChannel` wouldn't really test much of anything, and I don't think a test that duplicated the logic in `getUpdateChannel` would help much either. I think what I really want is some sort of test that downloads Firefox from specific channels and checks the value, but that's way out of scope for the automation we have now. Thoughts?

### DIFF
--- a/recipe-client-addon/lib/ClientEnvironment.jsm
+++ b/recipe-client-addon/lib/ClientEnvironment.jsm
@@ -12,6 +12,7 @@ Cu.import("resource://gre/modules/XPCOMUtils.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "ShellService", "resource:///modules/ShellService.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "AddonManager", "resource://gre/modules/AddonManager.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "TelemetryArchive", "resource://gre/modules/TelemetryArchive.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "UpdateUtils", "resource://gre/modules/UpdateUtils.jsm");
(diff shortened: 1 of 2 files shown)


Osmose:
> Yeah, I don't think this is really testable. I also think it doesn't matter much - AIUI all this fixes is people who change the update channel via preference overrides in a distribution file (or through some other way change it on the default pref branch, but haven't recompiled) there's got to be very very few of those people, right? Or am I missing something?

Our current method of determining the channel returns `release` for release candidate builds on Beta, which has been interrupting experiments running on Beta that are planning their timelines around when release actually switches channels (vs the week beforehand that we send out rc builds).

gijsk:
Ugh. I think this means all other consumers of `defaultUpdateChannel` are also wrong (I think).

That doesn't matter much here except we also expose this in `NormandyDriver.jsm`. Shouldn't we update that, too?

I'll file some bugzilla stuff to remove the footgun. :-(

gijsk:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1402866 .

PR #1072, opened 2017-09-25
Add ExtensionSelect tests

### LINKED ISSUE #1050: Write tests for ExtensionSelect displaying the correct value and being selectable
As a followup to #1048, we should have some form of automated test that ensures the `ExtensionSelect` component displays the correct value as passed down via the `value` prop, and calls the `onChange` prop when a new value is selected.

andymikulski:
Fixes #1050 

- Adds tests for ExtensionSelect component

Apparently, it's a little difficult to test against Ant's dropdowns since it uses some weird CSS for positioning. As a result, the test against `onChange` feels a little weak, so I'm open to any ideas to expand on it.

Ready for review.

### DIFF
--- a/recipe-server/client/control/tests/components/extensions/test_ExtensionSelect.js
+++ b/recipe-server/client/control/tests/components/extensions/test_ExtensionSelect.js
@@ -1,12 +1,12 @@
+import { Select } from 'antd';
 import { mount } from 'enzyme';
-import { List } from 'immutable';
+import { fromJS, List } from 'immutable';
 import React from 'react';
 
 import { wrapMockStore } from 'control/tests/mockStore';
 import TestComponent from 'control/components/extensions/ExtensionSelect';
 
 const { WrappedComponent: ExtensionSelect } = TestComponent;
-
 
 describe('<ExtensionSelect>', () => {
   const props = {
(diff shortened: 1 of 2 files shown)


recipe-server/client/control/tests/components/extensions/test_ExtensionSelect.js
```diff
@@ -36,4 +36,42 @@ describe('<ExtensionSelect>', () => {
...
+          fromJS([{ xpi: '1', name: 'one' }, { xpi: '2', name: 'two' }])
+        }
+        onChange={val => {
+          selected = val;
+        }}
+      />,
+    ));
+
+    wrapper.find('.ant-select-selection__placeholder').simulate('click');
+    expect(wrapper.find(Select).props().children.size).toBe(2);
```
mythmon: Is children in this line the elements of the dropdown? Would it work to simulate('click') on one of them?

andymikulski: After playing with this a bit, I'm still at a wall.

- The 'children' reflect the options passed in, however selecting one (`.get(0)`) returns some sort of wrapper which does _not_ allow `simulate`ing events
- The children have `title` and `value` props, but searching the wrapper (`wrapper.find("[type='whatever']")` turns up empty
  - I suspect this is because Ant is rendering the dropdown menu someplace else on the page. For instance, if you look at the DOM, the dropdown menu is _not_ nested under the actual select box.
- Searching for the exact Option class names (`wrapper.find('.ant-select-dropdown-menu-item')`) also turns up nothing

At this point, I think I will have to dig into how Ant/rc-select performs their tests on this component.

andymikulski:
@mythmon mentioned we may be able to do something with enzyme's `render`, I will investigate that!

andymikulski:
Well, I ended up circling back around to what I had earlier, which was just targeting the internal `Select` component and testing that its `onChange` bubbled up through to the `ExtensionSelect` prop.

I'm still not in love with this, but I think it's close enough to land this PR. Other ExtensionSelect work needs to be done in the future, so we can touch on the tests again then.

PR #1076, opened 2017-09-27
#1054, #1055: Prevent dupe recipes in recent history

### LINKED ISSUE #1054: Duplicate recipes in sidebar.
One link goes to the `/recipe/{id}` page and the other goes to `/recipe/{id}/edit`.

screenshot-2017-9-7 shield control panel

### LINKED ISSUE #1055: Duplicate key error.
Error:
> Warning: flattenChildren(...): Encountered two children with the same key, \`.2:$/recipe/20/edit\`. [...]

This is probably an issue with `NavigationMenu.js`, specifically `recipeSessionHistory`.

andymikulski:
- Fixes #1054  - duplicate recipes in sidebar
- Fixes #1055 - duplicate React key warning

- ~Adds a check to snip `/edit`, `/clone`, and `/approval_request` from 'session views', ensuring that links in the 'recent recipes' section are to the `view recipe` page and never `edit recipe`, etc.~ 
Adds `slug` and `sessionSlug` properties to routes. If a route has a `sessionSlug`, that is the path used for session purposes (i.e. recently viewed nav list).
- Adds `getNamedPath` router util, which returns a URL given an internal route slug
- Adds shield identicons to menu nav since they were not implemented when landed
- Adds tests

Ready for review!

### DIFF
--- a/recipe-server/client/control/routerUtils.js
+++ b/recipe-server/client/control/routerUtils.js
@@ -6,10 +6,7 @@ export const searchRouteTree = (tree, name, currentUrl = '') => {
 
   // If the slug doesn't match, iterate over the given route tree until
   // we find one that matches what we need (if any exist).
-  const keys = Object.keys(tree);
-  for (let i = 0; i < keys.length; i += 1) {
-    const curPath = keys[i];
-
+  for (const curPath of Object.keys(tree)) {
     // A key beginning with '/' indicates it is a route tree, which should be searched.
     if (curPath.charAt(0) === '/') {
       const val = searchRouteTree(tree[curPath], name, currentUrl + curPath);
--- a/recipe-server/client/control/routerUtils.js
+++ b/recipe-server/client/control/routerUtils.js
@@ -1,10 +1,10 @@
 export const searchRouteTree = (tree, name, currentUrl = '') => {
-  // If we have the crumb, we have the complete route.
-  if (tree.crumb === name) {
+  // If we have the slug, we have the complete route.
+  if (tree.slug === name) {
     return currentUrl;
   }
 
-  // If the crumb doesn't match, iterate over the given route tree until
+  // If the slug doesn't match, iterate over the given route tree until
   // we find one that matches what we need (if any exist).
   const keys = Object.keys(tree);
   for (let i = 0; i < keys.length; i += 1) {
--- a/recipe-server/client/control/routerUtils.js
+++ b/recipe-server/client/control/routerUtils.js
@@ -1,0 +1,43 @@
+export const searchRouteTree = (tree, name, currentUrl = '') => {
+  // If we have the crumb, we have the complete route.
+  if (tree.crumb === name) {
+    return currentUrl;
+  }
+
+  // If the crumb doesn't match, iterate over the given route tree until
+  // we find one that matches what we need (if any exist).
+  const keys = Object.keys(tree);
+  for (let i = 0; i < keys.length; i += 1) {
+    const curPath = keys[i];
+
+    // A key beginning with '/' indicates it is a route tree, which should be searched.
+    if (curPath.charAt(0) === '/') {
+      const val = searchRouteTree(tree[curPath], name, currentUrl + curPath);
+
+      // Non-null `val` indicates the route was found.
+      if (val !== null) {
+        // The base route's slash sometimes leads to `//` showing up in the result URL.
+        return val.replace('//', '/');
+      }
+    }
+  }
+
+  // If we've gotten this far, the route has not been found in this route tree.
+  return null;
+};
+
+
+// Given a route (e.g. `/hello/:id/there`), finds params that need to be
+// populated (e.g. `:id`) and returns a string with populated values.
+export const replaceUrlVariables = (url, params) => {
+  let newUrl = url;
+  const urlParams = url.match(/:[a-z]+/gi);
+
+  if (urlParams) {
+    urlParams.forEach(piece => {
+      newUrl = newUrl.replace(piece, params[piece.slice(1)] || piece);
+    });
+  }
+
+  return newUrl;
+};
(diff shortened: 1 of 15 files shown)


recipe-server/client/control/state/app/session/actions.js
```diff
@@ -61,13 +61,17 @@ export function saveSession() {
...
 
-export function addSessionView(category, caption) {
+export function addSessionView(category, caption, identicon) {
   return async (dispatch, getState) => {
-    const url = getState().router.pathname;
+    let url = getState().router.pathname;
+
+    // Prevent exact subpages (e.g. edit, clone pages) from appearing in the nav.
+    // Instead, this will link to the 'view' page for that recipe/revision.
+    url = url.replace(/\/(edit|clone|approval_history)/, '');
```
mythmon: This seems fragile. It requires this function to know about all the rest of the code, and makes it less re-usable. I think this should be refactored so that callers pass the URL. That way each individual caller can manager how it's URLs need to be deduplicated.

Either that, or this sort of "parent page" information should be stored on the route, so it is stored in a central place with the routing information.

recipe-server/client/control/tests/state/session/test_actions.js
```diff
@@ -0,0 +1,60 @@
...
+    const defaultParams = ['category', 'caption', 'identicon'];
+    // Values = the 'default' dispatched values given our default params.
+    const defaultValues = { caption: 'caption', category: 'category', identicon: 'identicon' };
+
+
+    it('should dispatch a SESSION_INFO_HISTORY_VIEW event', async () => {
+      const meta = {
+        dispatch: () => {},
+      };
+      spyOn(meta, 'dispatch').and.callThrough();
```
mythmon: You can probably do this more cleanly with [`jasmine.createSpy`](https://jasmine.github.io/2.0/introduction.html#section-Spies:_%3Ccode%3EcreateSpy%3C/code%3E):

recipe-server/client/control/state/app/session/actions.js
andymikulski: I've gone with a bit of a combination of both options, seen in the latest commit: https://github.com/mozilla/normandy/pull/1076/commits/514f3359fbc23074975e6f28ed61baa1d2924604 

On the route definitions, I use an `ignoreSession` variable to determine if it should handle changing the reported URL. Then, in order to know what the 'correct' URL is, I put together a `getNamedRoute` function which traverses the route tree and finds the specified path based on the given crumb. For example:

```
getNamedRoute('New Extension') === '/extension/new/'
getNamedRoute('View Recipe') === '/recipe/:recipeId/'
getNamedRoute('View Recipe', { recipeId: 3 }) === '/recipe/3/'
```

This `getNamedRoute` function will actually be useful in a few places, so I quite like this approach. I still have to write tests, but this code/commit is certainly open for feedback.

recipe-server/client/control/state/app/session/actions.js
```diff
@@ -61,13 +62,23 @@ export function saveSession() {
...
   return async (dispatch, getState) => {
-    const url = getState().router.pathname;
+    const { router } = getState();
+    let url = router.pathname;
+    const ignoreSession = router.result && router.result.ignoreSession;
+
+    // Prevent exact subpages (e.g. edit, clone pages) from appearing in the nav.
+    // Instead, this will link to the proper 'View [Recipe|Extension]' page.
+    if (ignoreSession) {
+      url = getNamedRoute(`View ${capitalize(category)}`, router.params);
```
mythmon: I don't think that using user-facing names for url routing is a good idea. I think that routes should have another field (maybe slug?) that we can use for specifying routes. Specifically, needing to call capitalize seems like a bad thing here. It makes form and function too closely linked.

recipe-server/client/control/routes.js
```diff
@@ -42,30 +42,35 @@ const routes = {
           '/': {
             component: RecipeDetailPage,
             crumb: 'Revision',
+            ignoreSession: true,
```
mythmon: Can you explain the choice of name `ignoreSession`? I don't think I get it.

It might also be good to have a comment above the routes explaining the extra keys we are adding. They all make sense now, but they aren't documented anywhere, and the schema is pretty important.

andymikulski: The idea behind `ignoreSession` was to ignore the route for session purposes, but I agree it is/was too vague. I've switched over to this format:

```
...
'/clone': {
  '/': {
    component: CloneRecipePage,
    crumb: 'Clone Recipe',
    sessionSlug: 'recipe-view',
    slug: 'recipe-clone',
  },
},
...
```

With this approach, we can simply point to another slug and anything session-related will use that over the actual `slug`. This feels a bit better since the `routes.js` file now lets us define the redirect behavior without having to tie directly into the underlying breadcrumb/current URL stuff.

recipe-server/client/control/state/app/session/actions.js
andymikulski: Using `slug`s has changed this to ```url = getNamedRoute(`${category}-view`);```, good idea!

andymikulski:
This now utilizes `slug`s, `sessionSlug`s, and has all the appropriate tests. This should be ready for a final pass!

recipe-server/client/control/routerUtils.js
```diff
@@ -0,0 +1,43 @@
...
+  // If we have the slug, we have the complete route.
+  if (tree.slug === name) {
+    return currentUrl;
+  }
+
+  // If the slug doesn't match, iterate over the given route tree until
+  // we find one that matches what we need (if any exist).
+  const keys = Object.keys(tree);
+  for (let i = 0; i < keys.length; i += 1) {
+    const curPath = keys[i];
```
mythmon: Nit: I think this could be done as `for (const curPath of Object.keys(tree)) {`,

andymikulski:
Merging once CI is happy!

PR #1080, opened 2017-09-29
Bug 1402528 - recipe-client-addon: Watch prefs and enable/disable as needed.

mythmon:
This is a fairly large (in terms of lines) change to `RecipeRunner`'s startup and pref handling. I think it should be fairly straight forward though. It is mostly just pretty verbose.

### DIFF
--- a/recipe-client-addon/lib/RecipeRunner.jsm
+++ b/recipe-client-addon/lib/RecipeRunner.jsm
@@ -39,88 +39,147 @@ Cu.importGlobalProperties(["fetch"]);
 this.EXPORTED_SYMBOLS = ["RecipeRunner"];
 
 const log = LogManager.getLogger("recipe-runner");
-const prefs = Services.prefs.getBranch("extensions.shield-recipe-client.");
 const TIMER_NAME = "recipe-client-addon-run";
-const RUN_INTERVAL_PREF = "run_interval_seconds";
-const FIRST_RUN_PREF = "first_run";
 const PREF_CHANGED_TOPIC = "nsPref:changed";
+
+const TELEMETRY_ENABLED_PREF = "datareporting.healthreport.uploadEnabled";
+
+const SHIELD_PREF_PREFIX = "extensions.shield-recipe-client";
+const RUN_INTERVAL_PREF = `${SHIELD_PREF_PREFIX}.run_interval_seconds`;
+const FIRST_RUN_PREF = `${SHIELD_PREF_PREFIX}.first_run`;
+const SHIELD_ENABLED_PREF = `${SHIELD_PREF_PREFIX}.enabled`;
+const DEV_MODE_PREF = `${SHIELD_PREF_PREFIX}.dev_mode`;
+const API_URL_PREF = `${SHIELD_PREF_PREFIX}.api_url`;
+const LAZY_CLASSIFY_PREF = `${SHIELD_PREF_PREFIX}.experiments.lazy_classify`;
+
+const PREFS_TO_WATCH = [
+  RUN_INTERVAL_PREF,
+  TELEMETRY_ENABLED_PREF,
+  SHIELD_ENABLED_PREF,
+  API_URL_PREF,
+];
 
 this.RecipeRunner = {
   async init() {
-    if (!this.checkPrefs()) {
+    this.enabled = null;
+    this.checkPrefs();  // sets this.enabled
+    this.watchPrefs();
+
+    // Run if enabled immediately on first run, or if dev mode is enabled.
+    const firstRun = Services.prefs.getBoolPref(FIRST_RUN_PREF);
+    const devMode = Services.prefs.getBoolPref(DEV_MODE_PREF);
+
+    if (this.enabled && (devMode || firstRun)) {
+      await this.run();
+    }
+    if (firstRun) {
+      Services.prefs.setBoolPref(FIRST_RUN_PREF, false);
+    }
+  },
+
+  enable() {
+    if (this.enabled) {
+      return;
+    }
+    this.registerTimer();
+    this.enabled = true;
+  },
+
+  disable() {
+    if (!this.enabled) {
+      return;
+    }
+    this.unregisterTimer();
+    this.enabled = false;
+  },
+
+  /** Watch for prefs to change, and call this.observer when they do */
+  watchPrefs() {
+    for (const pref of PREFS_TO_WATCH) {
+      Services.prefs.addObserver(pref, this);
+    }
+
+    CleanupManager.addCleanupHandler(this.unwatchPrefs.bind(this));
+  },
+
+  unwatchPrefs() {
+    for (const pref of PREFS_TO_WATCH) {
+      Services.prefs.removeObserver(pref, this);
+    }
+  },
+
+  /** When prefs change, this is fired */
+  observe(subject, topic, data) {
+    switch (topic) {
+      case PREF_CHANGED_TOPIC: {
+        const prefName = data;
+
+        switch (prefName) {
+          case RUN_INTERVAL_PREF:
+            this.updateRunInterval();
+            break;
+
+          // explicit fall-through
+          case TELEMETRY_ENABLED_PREF:
+          case SHIELD_ENABLED_PREF:
+          case API_URL_PREF:
+            this.checkPrefs();
+            break;
+
+          default:
+            log.debug(`Observer fired with unexpected pref change: ${prefName}`);
+        }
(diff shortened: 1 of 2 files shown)


recipe-client-addon/test/browser/browser_RecipeRunner.js
```diff
@@ -356,22 +358,80 @@ decorate_task(
...
+      ["extensions.shield-recipe-client.api_url", "https://example.com"], // starts with "https://"
+    ],
+  }),
+  withStub(RecipeRunner, "run"),
+  withStub(RecipeRunner, "enable"),
+  withStub(RecipeRunner, "disable"),
+  withStub(CleanupManager, "addCleanupHandler"),
+
+  async function testPrefWatching(runStub, enableStub, disableStub, addCleanupHandlerStub) {
+    await RecipeRunner.init();
```
Osmose: Could we potentially have multiple watchers registered during this test (since this was already called at the start of the session)? Feels like that could cause some hard-to-trace test errors down the road.

recipe-client-addon/lib/RecipeRunner.jsm
```diff
@@ -39,88 +39,144 @@ Cu.importGlobalProperties(["fetch"]);
...
+      this.enable();
     } else {
-      log.debug(`Observer fired with unexpected pref change: ${prefName}`);
+      this.disable();
     }
   },
 
+  registerTimer() {
+    this.updateRunInterval();
+    CleanupManager.addCleanupHandler(() => timerManager.unregisterTimer(TIMER_NAME));
```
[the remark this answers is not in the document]
rhelmer: Shouldn't throw, just logs and moves on:
http://searchfox.org/mozilla-central/rev/e62604a97286f49993eb29c493672c17c7398da9/toolkit/components/timermanager/nsUpdateTimerManager.js#345

recipe-client-addon/lib/RecipeRunner.jsm
```diff
@@ -39,88 +39,144 @@ Cu.importGlobalProperties(["fetch"]);
...
+    if (this.enabled && (devMode || firstRun)) {
       await this.run();
-      prefs.setBoolPref(FIRST_RUN_PREF, false);
     }
+    if (firstRun) {
+      Services.prefs.setBoolPref(FIRST_RUN_PREF, false);
+    }
+  },
 
+  enable() {
```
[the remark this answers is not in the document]
rhelmer: Current code handles `registerTimer` being called multiple times:
http://searchfox.org/mozilla-central/rev/e62604a97286f49993eb29c493672c17c7398da9/toolkit/components/timermanager/nsUpdateTimerManager.js#319

Seems cleaner not to do it though 🎨

recipe-client-addon/test/browser/browser_RecipeRunner.js
mythmon: Since `enable` is stubbed out, and the watchers are set through that, no. I think we're ok here.

mythmon:
I updated enable and disable to check the current state before acting, which should fix the idempotency issues brought up in review.

recipe-client-addon/lib/RecipeRunner.jsm
```diff
@@ -39,88 +39,144 @@ Cu.importGlobalProperties(["fetch"]);
...
+  SHIELD_ENABLED_PREF,
+  API_URL_PREF,
+];
+
 this.RecipeRunner = {
   async init() {
-    if (!this.checkPrefs()) {
-      return;
-    }
+    this.enabled = null;
```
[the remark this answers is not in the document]
mythmon: I think the changes we want to enable and disable below necessitates keeping this bit of state. Sorry 'bout your mantra.

recipe-client-addon/lib/RecipeRunner.jsm
```diff
@@ -39,88 +39,152 @@ Cu.importGlobalProperties(["fetch"]);
...
+    if (firstRun) {
+      Services.prefs.setBoolPref(FIRST_RUN_PREF, false);
+    }
+  },
 
+  enable() {
+    if (this.enabled) {
+      return;
+    }
+    this.enabled = true;
```
Osmose: Minor: We should update this at the end of the function in case the changes throw and we're not actually enabled.

recipe-client-addon/lib/RecipeRunner.jsm
```diff
@@ -39,88 +39,152 @@ Cu.importGlobalProperties(["fetch"]);
...
-   */
-  observePrefChange(prefName) {
-    if (prefName === RUN_INTERVAL_PREF) {
-      this.updateRunInterval();
+  checkPrefs() {
+    let shouldEnable = true;
+
+    // Only run if Unified Telemetry is enabled.
+    if (!Services.prefs.getBoolPref(TELEMETRY_ENABLED_PREF)) {
+      log.debug("Disabling RecipeRunner because Unified Telemetry is disabled.");
```
Osmose: I like that this states what it's doing, can we update the other logs to also state that they're disabling Shield?

Osmose: I wonder if having multiple log messages claiming to be disabling the recipe runner would be confusing to, say, QA. Dunno if it's better to disable immediately and return, or list out the reasons like it is currently.

recipe-client-addon/lib/RecipeRunner.jsm
Osmose: We could probably use a test that multiple calls to enable/disable do not register multiple watchers.

mythmon:
Updated and rebased on master to fix merge conflicts.

mythmon:
I think my most recent fixes broke the TC failure. I'll look into that on Monday.

mythmon:
I couldn't reproduce the test failures locally. I re-pushed to trigger TC again. Maybe it was some sort of intermittent?

mythmon:
It doesn't seem like it was intermittent, but I don't really know what could have happened. @rhelmer do you have any ideas?

recipe-client-addon/lib/RecipeRunner.jsm
```diff
@@ -39,88 +39,147 @@ Cu.importGlobalProperties(["fetch"]);
...
+  /** Watch for prefs to change, and call this.observer when they do */
+  watchPrefs() {
+    for (const pref of PREFS_TO_WATCH) {
+      Services.prefs.addObserver(pref, this);
+    }
 
-    // Watch for the run interval to change, and re-register the timer with the new value
-    prefs.addObserver(RUN_INTERVAL_PREF, this);
-    CleanupManager.addCleanupHandler(() => prefs.removeObserver(RUN_INTERVAL_PREF, this));
+    CleanupManager.addCleanupHandler(this.unwatchPrefs.bind(this));
```
rhelmer: Hm. Are you sure this works? I thought this would give you a different reference, when you need the same for both add/remove... I've always done this like:

```
this.unwatchPrefs = this.unwatchPrefs.bind(this);
addListener(this.unwatchPrefs);
/* later... */
removeListener(this.unwatchPrefs);
```

recipe-client-addon/lib/RecipeRunner.jsm
```diff
@@ -39,88 +39,147 @@ Cu.importGlobalProperties(["fetch"]);
 this.EXPORTED_SYMBOLS = ["RecipeRunner"];
 
 const log = LogManager.getLogger("recipe-runner");
-const prefs = Services.prefs.getBranch("extensions.shield-recipe-client.");
```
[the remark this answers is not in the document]
mythmon: I did this because we are accessing prefs from more than just the `extensions.shield-recipe.client` sub-tree. It seemed awkward to use `prefs.getCharPref` in someplaces and `Services.prefs.getCharPref` in others, depending on which preference was being accessed.

recipe-client-addon/lib/RecipeRunner.jsm
mythmon: You're right that if we needed to remove a listener later, this would be a problem. But we never remove cleanup handlers, so we won't need this reference again. CleanupHandler is an Shield-ism. this will be the same in the callback (unwatchPrefs), soServices.prefs.removeObserver(pref, this)` (line 107) will work fine.

mythmon:
Doh. I forgot to update the preference name in the tests when I changed it in the code. I've also rebased this on master to pick up the TC fixes from #1110.

@rhelmer this should be ready for a final review, and if you approve it, it should be ready to merge.

PR #1085, opened 2017-10-03
Add Experimenter URL to Normandy API fixes #1084

### LINKED ISSUE #1084: Pass Experimenter URL to the web UI via the API
Here:

https://github.com/mozilla/normandy/blob/master/recipe-server/normandy/base/api/views.py#L59

We can pass the Experimenter URL to the React client so it knows where to call.

jaredkerim:


recipe-server/normandy/settings.py
```diff
@@ -358,6 +360,9 @@ def RAVEN_CONFIG(self):
 
     GITHUB_URL = values.Value('https://github.com/mozilla/normandy')
 
+    # URL that hosts Experimenter experiment definitions
+    EXPERIMENTER_PENDING_URL = values.URLValue('https://experimenter.services.mozilla.com/api/v1/experiments/?status=Pending')
```
mythmon: This setting will need to be added to the docs in `docs/ops/config.rst`.

recipe-server/normandy/settings.py
```diff
@@ -380,6 +385,9 @@ class Development(Base):
     SWAGGER_SETTINGS = Base.SWAGGER_SETTINGS
     SWAGGER_SETTINGS['VALIDATOR_URL'] = None
 
+    # URL that hosts Experimenter experiment definitions
+    EXPERIMENTER_PENDING_URL = values.URLValue('https://localhost/api/v1/experiments/?status=Pending')
```
mythmon: It probably makes sense either to add Experimenter to the docker-compose setup, or at least add a reference to the developer docs about how to set up an instance.

mythmon: Separately, I think this might be better as `EXPERIMENTER_API_URL` that points to `/api/v1` instead of pointing to a very specific API endpoint. the frontend can build the more specific URL from the less specific one, but it isn't very sound to go backwards, if we want to in the future.

recipe-server/normandy/settings.py
```diff
@@ -358,6 +363,9 @@ def RAVEN_CONFIG(self):
 
     GITHUB_URL = values.Value('https://github.com/mozilla/normandy')
 
+    # URL that hosts Experimenter experiment definitions
+    EXPERIMENTER_HOST = values.URLValue('https://experimenter.services.mozilla.com')
```
mythmon: This still needs docs.

mythmon: This probably also shouldn't have a default value. I doubt we want to use prod experimenter in the stage environment of Normandy.

recipe-server/normandy/settings.py
```diff
@@ -380,6 +388,9 @@ class Development(Base):
     SWAGGER_SETTINGS = Base.SWAGGER_SETTINGS
     SWAGGER_SETTINGS['VALIDATOR_URL'] = None
 
+    # URL that hosts Experimenter experiment definitions
+    EXPERIMENTER_HOST = values.URLValue('https://localhost')
```
mythmon: I can't imagine this is an accurate url for a local version of Experimenter. It should be optional.

We should also either include docs about how to setup Experimenter or add Experimenter to docker-compose to automatically be set up.

jaredkerim:
Closing this in favour of a more comprehensive PR.

PR #1090, opened 2017-10-04
Add OptGroup to ExtensionSelect to clarify what is being listed.

Osmose:
This makes it a bit clearer why only some options are shown in the dropdown.

Note that the "Recently uploaded" bit is a lie, because right now we're just fetching the first page of results, which is actually the oldest uploaded extensions. But IIRC the intention was to show recently-uploaded extensions, so that's the label I went with.

screen shot 2017-10-03 at 6 24 53 pm

screen shot 2017-10-03 at 6 25 00 pm


andymikulski:
TC failure looks unrelated, merging.

Issue #354, opened 2016-11-17
Modify dependency on telemetry preferences


Osmose:
There is a subset of actions that we want to abort early because they rely on unified telemetry, but some actions (like the 43 update notice) we want users to see even if they don't have telemetry enabled. So, instead of disabling the add-on, specific actions should disable themselves.

**Blocked by**: 
- [x] #353

Osmose:
After discussing things with @georgf, I think we need to change what telemetry preferences we're checking, and when. 

Currently:

- On startup, we check `toolkit.telemetry.unified`. If it is true, we continue. If it is false, we do not execute any recipes for the rest of the session.

Based on my conversation with @georgf, `datareporting.healthreport.uploadEnabled` is the correct pref to represent that Telemetry is available. I'd like to change our checks to:

- Always register the recipe run timer.
- On each attempt to run, check `datareporting.healthreport.uploadEnabled`. If it is true, continue. If it is false, do not download or execute recipes.

@MattGrimes Heartbeat recipes currently use Telemetry to track user interaction with the prompt. Do we want to still run those recipes even if Telemetry is disabled? I think the answer is yes, since users may still fill out a survey or otherwise benefit from the prompt, but want to confirm.

MattGrimes:
I say yes. Surveys are by nature opt-in. The user can always choose not to answer.

mythmon:
This will be fixed by #1094.

mythmon:
We'll need to revisit this more holistically when we implement system add-ons. I'm going to close this and we can deal with this issue more in depth in other issues.

PR #1094, opened 2017-10-04
#354 - Use the correct preference for telemetry.

mythmon:
CC @lonnen

### DIFF
--- a/recipe-client-addon/lib/RecipeRunner.jsm
+++ b/recipe-client-addon/lib/RecipeRunner.jsm
@@ -71,7 +71,7 @@ this.RecipeRunner = {
 
   checkPrefs() {
     // Only run if Unified Telemetry is enabled.
-    if (!Services.prefs.getBoolPref("toolkit.telemetry.unified")) {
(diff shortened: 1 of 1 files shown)


mythmon:
@rhelmer We are concerned about the transition of this, and are considering to a system-addon update to release with this change. Can you think about this transition extra hard for this review?

rhelmer:
Looks like this pref is set in a `const` in a few places in the codebase, with different names:
https://github.com/mozilla/normandy/blob/db5230a474f41daf1c3c45a33f103e90f2accd5a/recipe-client-addon/test/browser/browser_about_preferences.js#L6
https://github.com/mozilla/normandy/blob/935c526d77fb7892282e1883d5a5dd5e807b0101/recipe-client-addon/lib/ShieldPreferences.jsm#L24

Can it be shared? Other than that this seems ok... it's just reading the pref, it's sync, etc.

mythmon:
I don't think we have any way to share constants like this among files in Shield right now. I'd eventually like to have some sort of preferences manager that would let us do that, but for now I think we should leave the duplication.

PR #1096, opened 2017-10-04
Fix #1095 - Add cache headers to identicons.

### LINKED ISSUE #1095: Identicon API should have long lived cache headers.
The Identicon API is not supposed to change over time, so we should set cache headers to mark it as cacheable for a long time, and also mark it as immutable. This matches the behavior we use for static assets and other hash-addressed content, like action implementations.

mythmon:
r?


mythmon:
Another accidental branch deletion/closure. Oops. We still want this.

PR #1098, opened 2017-10-06
Fixes #917 Import Experiments from Experimenter into Normandy

### LINKED ISSUE #917: Import recipe definitions from Experimenter
https://github.com/mozilla/experimenter is a system for managing experiments in Firefox and wants to deploy those experiments using Shield Pref Studies.  To help automate the integration workflow, we can build a menu into the Shield control panel to query the experiment definitions from Experimenter and allow a Shield user to easily import them into the database.  This workflow lets us automate the import process without compromising any of the access control restrictions in place for Shield.

jaredkerim:
This PR encapsulates the core functionality to query experiment definitions from Experimenter (https://github.com/mozilla/experimenter) and import them into Normandy through the Normandy web UI.

This adds functionality to the Django backend to host the Experimenter API URL to the React frontend, state actions/reducers/selectors to receive that data and store it in the State, and components to display the experiments to the user.

Remaining TODO inside this PR:

* [ ] Look up 'actions' from the store by a stored constant (I'm not sure how we want to handle this, I added a note in the code)

* [x] Handle the case that a recipe sent to Normandy already exists (I think we want to add a richer response code to the recipe API that signals to the client that it already exists, which the client can check for and react accordingly)

* [x] Handle the case that the Normandy server sends a null value for the Experimenter API URL which the client can then gracefully handle.

Thank you for taking the time to review this!


recipe-server/client/control/tests/components/data/test_QueryExperiments.js
```diff
@@ -0,0 +1,40 @@
...
+  it('should work', () => {
+    const wrapper = () => shallow(<QueryExperiments {...props} />);
+
+    expect(wrapper).not.toThrow();
+  });
+  it('should call fetchExperiments on componentWillReceiveProps', () => {
+    let called = false;
+    const experimenterAPIUrl = 'https://localhost/api/v1';
+    const fetchExperiments = calledExperimenterAPIUrl => {
+      called = calledExperimenterAPIUrl === experimenterAPIUrl;
```
mythmon: You could probably use a stub for this, instead of rolling your own.

```js
const fetchExperimentsSpy = jasmine.createSpy();
/// mount
expect(fetchExperimentsSpy).not.toHaveBeenCalledWith(experimenterAPIUrl);
/// set props
expect(fetchExperimentsSpy).toHaveBeenCalledWith(experimenterAPIUrl);
```

recipe-server/client/control/tests/components/recipes/test_ImportExperimentsPage.js
```diff
@@ -0,0 +1,206 @@
...
+    };
+
+    const wrapper = shallow(<ExperimentRejectModal.WrappedComponent {...testProps} />);
+    expect(wrapper.find(Modal).length).toEqual(1);
+  });
+  it('should store the rejection reason in its state when edited', () => {
+    const reason = 'rejected for reasons.';
+    const wrapper = shallow(<ExperimentRejectModal.WrappedComponent {...props} />);
+    wrapper.instance().handleReasonChange({ target: { value: reason } });
+    expect(wrapper.state().reason).toEqual(reason);
```
mythmon: I didn't know that enzyme gave access to the state like this. That's cool.

recipe-server/client/control/tests/components/recipes/test_ImportExperimentsPage.js
```diff
@@ -0,0 +1,206 @@
...
+    wrapper.instance().handleCancel();
+  });
+  it('should call rejectExperiment with a reason when rejected', () => {
+    const reason = 'rejected for reasons.';
+    const experiment = ExperimentFactory.build();
+
+    const testProps = {
+      rejectExperiment: (rejectedExperiment, rejectionReason) => {
+        expect(rejectedExperiment.slug).toEqual(experiment.slug);
+        expect(rejectionReason).toEqual(reason);
```
mythmon: This type of call-assertion is a little less reliable, because this would also pass if the function is never called. This would probably be another good place to use a `jasmine.createSpy` and `expect(...).toHaveBeenCalledWith`.

recipe-server/client/control/tests/components/recipes/test_ImportExperimentsPage.js
```diff
@@ -0,0 +1,206 @@
...
+
+  it('should work', () => {
+    const wrapper = () => shallow(<ExperimentAcceptButton.WrappedComponent {...props} />);
+    expect(wrapper).not.toThrow();
+  });
+  it('should call importExperiment with provided experiment when clicked', () => {
+    const testProps = {
+      experiment,
+      importExperiment: calledExperiment => {
+        expect(calledExperiment.slug).toEqual(experiment.slug);
```
mythmon: Another spy here.

recipe-server/client/control/tests/components/recipes/test_ImportExperimentsPage.js
```diff
@@ -0,0 +1,206 @@
...
+
+  it('should work', () => {
+    const wrapper = () => shallow(<ExperimentRejectButton.WrappedComponent {...props} />);
+    expect(wrapper).not.toThrow();
+  });
+  it('should call selectRejectedExperiment with provided experiment when clicked', () => {
+    const testProps = {
+      experiment,
+      selectRejectedExperiment: calledExperiment => {
+        expect(calledExperiment.slug).toEqual(experiment.slug);
```
mythmon: Another spy here.

recipe-server/client/control/tests/factory.js
```diff
@@ -115,4 +115,14 @@ export class Factory {
     const product = new this(...args);
     return product.data;
   }
+
+  static buildMany(n, ...args) {
```
mythmon: giphy

recipe-server/client/control/tests/state/experimenter/index.js
```diff
@@ -0,0 +1,100 @@
...
+  items: new Map(),
+};
+
+function slugify(text) {
+  return text.toString().toLowerCase()
+    .replace(/\s+/g, '-')           // Replace spaces with -
+    .replace(/[^\w-]+/g, '')       // Remove all non-word chars
+    .replace(/--+/g, '-')         // Replace multiple - with single -
+    .replace(/^-+/, '')             // Trim - from start of text
+    .replace(/-+$/, '');            // Trim - from end of text
```
mythmon: Probably worth aligning these comments, or not offsetting them so much.

recipe-server/client/control/tests/state/experimenter/test_actions.js
```diff
@@ -0,0 +1,169 @@
...
+  EXPERIMENT_SELECT_REJECTED,
+  RECIPE_RECEIVE,
+} from 'control/state/action-types';
+
+const middlewares = [thunk];
+const mockStore = configureStore(middlewares);
+
+describe('Experimenter Actions fetchExperiments', () => {
+  beforeEach(() => {
+    jasmine.addMatchers(matchers);
```
mythmon: I imagine this could be done once at top level, instead of `beforeEach`. I think that would make the tests very slightly faster, and simpler.

recipe-server/client/control/tests/state/serviceInfo/index.js
```diff
@@ -13,6 +13,7 @@ export class ServiceInfoFactory extends Factory {
       user: new SubFactory(UserFactory),
       peer_approval_enforced: true,
       logout_url: '/logout/',
+      experimenter_api_url: 'https://localhost/api/v1',
```
mythmon: `example.com` is a better domain for test data.

recipe-server/client/control/components/recipes/ImportExperimentsPage.js
```diff
@@ -0,0 +1,257 @@
...
+  render() {
+    return (
+      <Button type="primary" onClick={this.handleClick}>Accept</Button>
+    );
+  }
+}
+
+@connect(null, { selectRejectedExperiment })
+@autobind
+export class ExperimentRejectButton extends React.PureComponent {
```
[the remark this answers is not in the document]
jaredkerim: Yeah I thought of that but between styling and their different behaviours it got hairy so I figured this was more straight forward.

recipe-server/client/control/components/recipes/ImportExperimentsPage.js
```diff
@@ -0,0 +1,257 @@
...
+    return (
+      <div>
+        <Card
+          title={experimentTitle}
+          extra={experimentButtons}
+        >
+          <div className="experiment-section">
+            <div className="experiment-population">
+              <span className="experiment-population-size">
+                {parseFloat(experiment.population_percent)}%
```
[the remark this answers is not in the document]
jaredkerim: It's because the Experimenter API stores and serializes percents as fixed precision Decimals so it's transmitted as a string so you don't wind up getting something like 90.50000001 of Release as a configuration when you meant exactly 90.5%, but it looks ugly to show 90.5000% when you only need to say 90.5%.

recipe-server/client/control/state/app/experimenter/reducers.js
```diff
@@ -0,0 +1,37 @@
...
+      return state.set('rejected_experiment', fromJS(action.experiment));
+    case EXPERIMENT_REJECTED:
+      return filterExperiments(state, action.experiment.experiment_slug);
+    default:
+      return state;
+  }
+}
+
+export default combineReducers({
+  items,
```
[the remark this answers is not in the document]
jaredkerim: I think I tried that originally but then it didn't make sense to have state.app.experiments.experiments rather than state.app.experiments.items.experiments.  It seemed like to be consistent with the rest of the application, I should name space my whole state into something called 'items', which I think @rehandalal recommended?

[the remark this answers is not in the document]
mythmon: Right, so `rejected_experiments` shouldn't be nested under items. `rejected_experiments` should be it's own reducer. Something like:

```js
function items(state = new List(), action) {
  // handles RECEIVE, ACCEPTED, ALREADY_ACCEPTED, and REJECTED
}

function rejected(state = null, action) {
  switch (action.type) {
    case EXPERIMENT_SELECT_REJECTED: {
      return fromJS(action.experiment);
    }
    default: {
      return state;
    }
}

export default combineReducers({
  items,
  rejected,
});
```

Then you'd have `state.app.experiments.items` and `state.app.experiments.rejected`. To me this makes more sense than nesting `rejected` under `items`.

This is a bigger change, but most of the time that we have a collection and then a selection into that collection, like here, we use a `Map` for the items, and instead of storing the entire item in `rejected`, it would be `rejectedId`, and just store the key into that map for the rejected item. I'm not sure if this would work well with Experimenter's data model though.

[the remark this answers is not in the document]
jaredkerim: Ya makes sense, sounds good! That helps keep things clearer.

jaredkerim:
Additional notes:

- send REQUEST_SEND/REQUEST_RECEIVE actions when we start and finish experimenter API calls so that the loading overlay kicks in properly

- add the action look up by slug for 'preference-experiment'

- add an error message and popup for the case that an api call fails to alert the user

mythmon:
@jaredkerim this PR is about 6 weeks old at this point. Do you have any concrete plans to finish this out? If not, I'm inclined to close this PR as the likelyhood that it hasn't bitrotted is approaching 0%. You could re-open this PR with an updated branch when you're ready to work on it again.

PR #1099, opened 2017-10-09
Add docs about testing filter expressions in the brower console.

mythmon:
These were originally written by @Osmose and put in a Gist: https://gist.github.com/Osmose/d59da84e9531487e5210cc0444f3eab9

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -74,7 +74,7 @@ language = None
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
-exclude_patterns = ['_build']
+exclude_patterns = ['_build', '.*']
 
 # The reST default role (used for this markup: `text`) to use for all
 # documents.
--- a/docs/user/filter_expressions.rst
+++ b/docs/user/filter_expressions.rst
@@ -526,3 +526,70 @@ This section lists some examples of commonly-used filter expressions.
    // plugin list returns `undefined`, which is a falsy value in JavaScript and
    // fails the match. Otherwise, it returns a plugin object, which is truthy.
    normandy.plugins['Shockwave Flash']
+
+
+Advanced: Testing Filter Expressions in the Browser Console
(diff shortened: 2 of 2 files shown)


docs/conf.py
```diff
@@ -74,7 +74,7 @@
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
-exclude_patterns = ['_build']
+exclude_patterns = ['_build', '.*']
```
mythmon: This makes Sphinx ignore hidden files, which helps it work better in my workflow.

andymikulski:
Would it be worth it to implement a little REPL into the site's front end to do this? It could handle `await`ing the promise response and all that, might make it more accessible. I can't imagine a ton of users would need/use it, though.

mythmon:
I think a REPL would be a useful feature, but it would be difficult to expose the actual Shield implementation to the site's frontend. We talked about doing this in the past for the preview feature, and couldn't figure out a safe way to do it, and so backed off. I've been pondering if we could get away with some sort of extension to do this, but I haven't come up with anything solid yet.

It would be pretty simple to make a REPL that executed filter expressions in the webpage, but it wouldn't include the extra details added by the recipe client, such as Firefox version or Telemetry  pings.

mythmon:
I think a REPL would be a useful feature, but it would be difficult to expose the actual Shield implementation to the site's frontend. We talked about doing this in the past for the preview feature, and couldn't figure out a safe way to do it, and so backed off. I've been pondering if we could get away with some sort of extension to do this, but I haven't come up with anything solid yet.

It would be pretty simple to make a REPL that executed filter expressions in the webpage, but it wouldn't include the extra details added by the recipe client, such as Firefox version or Telemetry  pings.

mythmon:
I probably accidentally deleted this branch, and that's why this got closed. It is definitely something I still want.

PR #1103, opened 2017-10-11
#1093: Cloned identicons

### LINKED ISSUE #1093: Cloned recipes retain the same identicon as the recipe they were cloned from
If I clone a recipe, the resulting new recipe has the same identicon as the old recipe (if I don't touch the identicon selector at all). From a user's perspective, I would expect the identicon to be unique per-recipe and change if I clone a new recipe, so as to avoid mixing up two recipes.

andymikulski:
- Fixes #1093 
- Prevents original identicon appearing while cloning a recipe
- Prevents issue where randomized identicon on creation form would change while waiting for requests.

### DIFF
--- a/recipe-server/client/control/components/recipes/RecipeForm.js
+++ b/recipe-server/client/control/components/recipes/RecipeForm.js
@@ -38,7 +38,7 @@ export default class RecipeForm extends React.PureComponent {
   static propTypes = {
     form: PropTypes.object.isRequired,
     isLoading: PropTypes.bool,
-    isCreating: PropTypes.bool,
+    isCreationForm: PropTypes.bool,
     onSubmit: PropTypes.func.isRequired,
     recipe: PropTypes.instanceOf(Map),
     selectedActionName: PropTypes.string.isRequired,
@@ -46,7 +46,7 @@ export default class RecipeForm extends React.PureComponent {
 
   static defaultProps = {
     isLoading: false,
-    isCreating: false,
+    isCreationForm: false,
     recipe: new Map(),
   };
 
@@ -56,6 +56,10 @@ export default class RecipeForm extends React.PureComponent {
     'preference-experiment': PreferenceExperimentFields,
     'opt-out-study': OptOutStudyFields,
   };
+
+  componentDidMount() {
+    this.defaultIdenticonSeed = this.defaultIdenticonSeed || IdenticonField.generateSeed();
+  }
 
   componentWillReceiveProps(newProps) {
     // Initial values are mostly handled via props, but if the recipe
@@ -67,7 +71,7 @@ export default class RecipeForm extends React.PureComponent {
 
   render() {
     const {
-      isCreating,
+      isCreationForm,
       isLoading,
       onSubmit,
       recipe,
@@ -78,8 +82,7 @@ export default class RecipeForm extends React.PureComponent {
 
     // If creating, the 'default' seed is randomly generated. We store it in memory
     // to prevent the form from generating a new identicon on each render.
-    this.defaultIdenticonSeed = this.defaultIdenticonSeed || IdenticonField.generateSeed();
-    const identiconSeed = isCreating ? this.defaultIdenticonSeed : null;
+    const identiconSeed = isCreationForm ? this.defaultIdenticonSeed : null;
 
     return (
       <Form onSubmit={onSubmit} className="recipe-form">
--- a/recipe-server/client/control/components/recipes/RecipeForm.js
+++ b/recipe-server/client/control/components/recipes/RecipeForm.js
@@ -57,9 +57,8 @@ export default class RecipeForm extends React.PureComponent {
     'opt-out-study': OptOutStudyFields,
   };
 
-  constructor(props) {
-    super(props);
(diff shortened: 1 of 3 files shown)


recipe-server/client/control/components/recipes/RecipeForm.js
```diff
@@ -57,6 +57,10 @@ export default class RecipeForm extends React.PureComponent {
     'opt-out-study': OptOutStudyFields,
   };
 
+  componentDidMount() {
+    this.defaultIdenticonSeed = this.defaultIdenticonSeed || IdenticonField.generateSeed();
```
mythmon: Should `defaultIdenticonSeed` be in state? It seems that since it is used in render(), if it changes it should cause a re-render. If not, it should be set to some value (such as null) in a constructor, which helps a) performance and b) readability.

andymikulski: I actually specifically _didn't_ want a re-render. Essentially, a bug I ran into was that on each render, the random 'default' seed would change, causing a weird identicon swap on page load (showed one first, then the other). Setting the `defaultIdenticonSeed` on the actual instance was meant to avoid the additional re-render (vs using state).

I've moved this `defaultIdenticonSeed` declaration to a constructor instead!

andymikulski:
TC failure looks unrelated, error message is regarding Rust:

```
...
We attempted to upgrade Rust to a modern version (1.21.0 or newer).
However, you appear to still have version 1.19.0.
...
```

recipe-server/client/control/components/recipes/RecipeForm.js
mythmon: I guess I'm still not getting something. Why do we store a default at all? Couldn't we leave the `initialValue` of the form item empty, and let it sort itself out? This seems like complexity that `IdenticonField` should be handling.

Also, if we keep it this way, the `this.defaultIdenticonSeed` will always be undefined in the constructor, so the set-if-not-exists dance isn't useful there.

mythmon:
The TC failures should be fixed with #1110.

Circle failed with 

```
<CloneRecipePage> should work - Firefox_56_0_0_(Ubuntu_0_0_0).<CloneRecipePage>

Expected function not to throw, but it threw TypeError: ke.removeAll is not a function.
```

recipe-server/client/control/components/recipes/RecipeForm.js
andymikulski: The identicon_seed field has its value set up like so:`value={recipe.get('identicon_seed')}` - the issue here is that for new/cloned recipes, the `identicon_seed` will be undefined, but we need a random one to be chosen so it appears on the 'create' form.

Doing something like `recipe.get('identicon', generateRandomIcon())` solves this in that it falls back to a random identicon if one is not set. But, the issue there is that on each re-render, `generateRandomIcon` is fired, so the identicon changes with each render. To solve that, I used `this.defaultSeed` - it's set once, and then remains that value for the rest of the component's lifespan. Subsequent re-renders still fall back on the already defined default.

The issue with having `IdenticonField` generating a default is the case in which the field does not have a value but will have one loaded later. Since it's a simple form component, value changes need to be registered through `onChange` (that is, the `value` is tracked as a prop, not internally in the component). From there, it steps into a weird territory where it fires `onChange` when there's no value yet, but if another value is loaded, then another `onChange` needs to fire. This leads to weird flashes of different identicons appearing during page load. And, speaking semantically, since it really wasn't changing but being set to an initial value, an `onChange` event shouldn't technically need to fire.

The reason why I had originally put the `defaultIdenticonSeed` initialization inside `didMount` was the idea that on each mount, the default seed identicon would change, preventing the same default identicon from being offered (say, if the user was creating multiple recipes in one session). I reckon the code being in the constructor function the same, with the only difference being we need to add the `super(props)` boilerplate to the constructor. You are correct about not needing the `thing = thing || otherThing` business in the constructor, though.

mythmon: Thanks for that explanation. I think I have a good handle of what's going on now. Who knew that randomized fields would be so complicated.

After all that, I think that setting the default in `didMount`, without preserving existing values (`this.defaultIdenticon = generateRandomIcon()`) would probably work well. It wouldn't affect recipes with existing identicons, and it would work even if component instances got re-used. (Is that a thing we need to worry about?)

That being said, I'm ok with any of the options we've gone through in this PR. If it fixes the clone bug, and doesn't flicker, I don't have any other concerns, and it seems like all of your solutions achieve that. I appreciate you working through it with me.

andymikulski:
Turns out it's because `removeAll` is in v4 of Immutable, and we're on v3. I've switched the code over to using two `remove`s, which I don't love, but eventually we can upgrade the package and revert it to use `removeAll`.

andymikulski:
This could probably stand to be rebased (to fix TC), plus I want to double check if there's a difference between using `didMount` vs the constructor. I will re-assign when this is ready!

andymikulski:
Landed on using `componentWillMount`; I realized that `didMount` fires _after_ the component renders for the first time, so it makes more sense to drop it into `willMount`.

Ready for review.

PR #1104, opened 2017-10-11
1088: ExtensionSelect losing focus

### LINKED ISSUE #1088: ExtensionSelect loses focus while typing due to search loads
When I try to type into the ExtensionSelect widget to select an extension for an opt-out study, the page gets blocked by an overlay while the widget tries to load results eagerly before I've finished typing. This prevents me from typing the rest of my query until after the load is finished, which is pretty noticeable locally, and very noticeable on prod. Also, if I'm deleting some of the characters when it loses focus, I end up triggering my browser to go back a few pages.

andymikulski:
- Fixes #1088 
- Fixes issue where typing into the `ExtensionSelect` component would result in the field losing focus to a `LoadingOverlay` taking over the page.
- This is also the only `LoadingOverlay` that did not specify the `requestIds` to listen for, so this issue should not come up anywhere else on the site.

### DIFF
--- a/recipe-server/client/control/components/common/LoadingOverlay.js
+++ b/recipe-server/client/control/components/common/LoadingOverlay.js
@@ -65,7 +65,9 @@ export class SimpleLoadingOverlay extends React.PureComponent {
 export default class LoadingOverlay extends React.PureComponent {
   static propTypes = {
     isLoading: PropTypes.bool.isRequired,
-    requestIds: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.string), PropTypes.string]),
+    requestIds: PropTypes.oneOfType([
+      PropTypes.arrayOf(PropTypes.string), PropTypes.string
+    ]).isRequired,
   };
 
   static defaultProps = {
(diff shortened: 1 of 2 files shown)


mythmon:
> This is also the only LoadingOverlay that did not specify the requestIds to listen for, so this issue should not come up anywhere else on the site.

Why does adding `requestId` prevent the loading add-on from stealing focus? Would it make sense to make `requestIds` a required prop?

mythmon:
Also, this should probably have  "Fixes #..." so that GH will autoclose things.

andymikulski:
>Why does adding requestId prevent the loading add-on from stealing focus? 

The issue was that a `LoadingOverlay` without `requestIds` listens for _any_ requests being made to the remote API. When the ExtensionSelect was fed input, it made an API call for filtered extensions, and the LoadingOverlay would swap some DOM elements accordingly. Hence, the fix for this bug was to specify the requests the Overlay should be listening for.

>Would it make sense to make requestIds a required prop?

Probably, this was the _only_ `LoadingOverlay` that didn't have its `requestIds` defined.

mythmon:
TC failure is #1110. The rest of this looks good.

PR #1110, opened 2017-10-16
Cachebust Taskcluster docker image.

mythmon:
This should fix our TC issues, since the Docker image is always built with the latest version of Rust.

I've built and pushed this updated version of the image already.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -22,7 +22,7 @@ tasks:
     payload:
       env:
       maxRunTime: 300 # 5 minutes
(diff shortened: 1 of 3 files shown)


andymikulski:
CI is unhappy about something related to the mock server, TC passed though!

mythmon:
I re-ran Circle, and it worked this time. I think it was an intermittent error.

mythmon:
This is a CI-only PR that has passed CI, and is a pretty normal maintenance task. I'm going to merge this without a final r+ from @andymikulski, to unblock other PRs.

PR #1114, opened 2017-10-24
Add high population marker to preference experiments.

mythmon:
I also added some nicer UI to the preference detail page, since this interacts with it a little bit.

### DIFF
--- a/recipe-client-addon/lib/PreferenceExperiments.jsm
+++ b/recipe-client-addon/lib/PreferenceExperiments.jsm
@@ -67,7 +67,7 @@ this.EXPORTED_SYMBOLS = ["PreferenceExperiments"];
 const EXPERIMENT_FILE = "shield-preference-experiments.json";
 const STARTUP_EXPERIMENT_PREFS_BRANCH = "extensions.shield-recipe-client.startupExperimentPrefs.";
 
-const MAX_EXPERIMENT_TYPE_LENGTH = 20;  // enforced by TelemetryEnvironment
+const MAX_EXPERIMENT_TYPE_LENGTH = 20; // enforced by TelemetryEnvironment
 const EXPERIMENT_TYPE_PREFIX = "normandy-";
 const MAX_EXPERIMENT_SUBTYPE_LENGTH = MAX_EXPERIMENT_TYPE_LENGTH - EXPERIMENT_TYPE_PREFIX.length;
 
--- a/recipe-client-addon/lib/PreferenceExperiments.jsm
+++ b/recipe-client-addon/lib/PreferenceExperiments.jsm
@@ -67,6 +67,10 @@ this.EXPORTED_SYMBOLS = ["PreferenceExperiments"];
 const EXPERIMENT_FILE = "shield-preference-experiments.json";
 const STARTUP_EXPERIMENT_PREFS_BRANCH = "extensions.shield-recipe-client.startupExperimentPrefs.";
 
+const MAX_EXPERIMENT_TYPE_LENGTH = 20;  // enforced by TelemetryEnvironment
+const EXPERIMENT_TYPE_PREFIX = "normandy-";
+const MAX_EXPERIMENT_SUBTYPE_LENGTH = MAX_EXPERIMENT_TYPE_LENGTH - EXPERIMENT_TYPE_PREFIX.length;
+
 const PREFERENCE_TYPE_MAP = {
   boolean: Services.prefs.PREF_BOOL,
   string: Services.prefs.PREF_STRING,
@@ -190,7 +194,7 @@ this.PreferenceExperiments = {
       TelemetryEnvironment.setExperimentActive(
         experiment.name,
         experiment.branch,
-        {type: experiment.experimentType}
+        {type: EXPERIMENT_TYPE_PREFIX + experiment.experimentType}
       );
 
       // Watch for changes to the experiment's preference
@@ -282,7 +286,7 @@ this.PreferenceExperiments = {
     preferenceValue,
     preferenceBranchType,
     preferenceType,
-    experimentType = "normandy-pref",
+    experimentType = "exp",
   }) {
     log.debug(`PreferenceExperiments.start(${name}, ${branch})`);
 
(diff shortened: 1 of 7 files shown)


recipe-server/client/control/components/recipes/PreferenceExperimentFields.js
```diff
@@ -49,7 +49,16 @@ export default class PreferenceExperimentFields extends React.PureComponent {
           >
             <DocumentUrlInput disabled={disabled} />
           </FormItem>
+
+          <FormItem
+            label="High volume recipe?"
+            name="arguments.isHighVolume"
+            initialValue={recipeArguments.get('isHighVolume')}
+          >
+            <Checkbox />
```
mythmon: This ends up looking pretty bad. I'm not sure how to make it look better. @andymikulski do you have any idea for how to improve this?

screenshot-2017-10-24 shield control panel

recipe-server/client/control/components/recipes/RecipeDetails.js
```diff
@@ -30,6 +30,50 @@ export default class RecipeDetails extends React.PureComponent {
     }
   }
 
+  renderArgumentValue(name, value) {
+    switch (name) {
+      case 'branches': {
+        const sumRatios = value.map(branch => branch.get('ratio')).reduce((a, b) => a + b) || 1;
+        return (
+          <table className="pref-experiment-branches">
```
mythmon: I'm pretty happy with how this turned out:

screenshot-2017-10-24 shield control panel 1

recipe-server/client/control/components/recipes/PreferenceExperimentFields.js
andymikulski: I think up until this point we've been using our custom [SwitchBox](https://github.com/mozilla/normandy/blob/master/recipe-server/client/control/components/forms/SwitchBox.js) component (which uses an Ant [Select](https://ant.design/components/switch/) under the hood). I think we'll end up having to do something similar for Checkboxes.

I don't imagine it'd be a huge thing to implement, most of the SwitchBox code would already apply. Alternatively, if you're interested in landing this sooner than later, you could simply swap the `Checkbox` here with a `SwitchBox` and we can change the components later once the checkbox stuff is written.

recipe-server/client/control/components/recipes/RecipeDetails.js
```diff
@@ -30,6 +30,50 @@ export default class RecipeDetails extends React.PureComponent {
     }
   }
 
+  renderArgumentValue(name, value) {
+    switch (name) {
```
andymikulski: I think these `switch`es could be simplified down to a couple `if`s. It's a little weird to read at first since there are two `switch`es in a row with only one `case` each.

It may be better without the big block of JSX, too - I'd suggest making the `<table>` its own component so reading through this function is a bit more concise.

mythmon: Yeah, the switches are overkill. I was imagining having more cases, and then never simplified after I realized I only had one case in each.

mythmon:
I've fixed the issues @andymikulski brought up and rebased this on master to pick up test changes. This still needs review for the system add-on and action changes.

recipe-server/client/control/components/recipes/RecipeDetails.js
```diff
@@ -68,3 +68,50 @@ export default class RecipeDetails extends React.PureComponent {
     );
   }
 }
+
+
+class ArgumentsValue extends React.PureComponent {
```
andymikulski: Neat, I like this way better!

mythmon:
@gijsk can you take a look at the add-on changes here? I still need to fix the merge conflicts, but I don't expect it will be anything major.

This is not blocking or related to 57, but it would be nice to have it soon, to ease analysis of telemetry data.

recipe-server/client/actions/preference-experiment/index.js
```diff
@@ -54,13 +59,15 @@ export default class PreferenceExperimentAction extends Action {
 
       // Otherwise, enroll!
       const branch = await this.chooseBranch(branches);
+      const experimentType = isHighPopulation ? 'normandy-exp-highpop' : 'normandy-exp';
```
gijsk: Hm, but you're validating for the `experimentType` starting with `normandy-pref` in `PreferenceExperiments.start()`. Neither of these 2 strings match that.

Unless I'm missing something, this means I'd like to see 2, maybe 3 changes:

1) fix the string `startsWith` check and/or these names so that this does pass
2) write a testcase that actually passes the 'high population' thing to the telemetry thing which would have thrown and shown something was wrong here (ie ensure we test the interaction between this code and the telemetry code, which apparently right now we don't or you would have noticed this!)
3) optionally, switch this to something that isn't so stringly-typed. Like, I get that this is JS, but if we had a set of value constants defined on `PreferenceExperiments` that we used throughout then this would be a harder mistake to make/introduce.

gijsk: Also, double quotes (or, where appropriate, template strings) please. :-)

(eslint can help with this, I believe?)

mythmon: About the quotes: This code is outside of the add-on, and so doesn't use the Firefox eslint configs. It is tied more closely to the rest of our Javascript  that for historical reasons prefers single quotes. It's a weird situation, but eslint helps 😄.

Nice catch on the test-mismatch though. I'll fix that up.

gijsk: Oh, right. Does that mean that my suggestion for a test would require testing server-to-client interaction? That might be a bit much... I don't know why I missed that the file wasn't in the add-on - maybe the `client` part of the path threw me off.

mythmon: Yeah, we don't have any tools set up to test actions against the actual add-on, unfortunately. That also means that change 3 is a bit out of reach. I think it would make sense to make this API a little less sensitive, and enforce the things we actually care about (like length limits and a normandy-prefix). I'll work on that.

mythmon:
I rebased this against master and added another commit that makes the experiment type of the easier less error prone. It also has slightly better tests now.

recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -306,9 +310,10 @@ this.PreferenceExperiments = {
       throw new Error(`Invalid value for preferenceBranchType: ${preferenceBranchType}`);
     }
 
-    if (!experimentType.startsWith("normandy-pref")) {
+    if (experimentType.length > MAX_EXPERIMENT_SUBTYPE_LENGTH) {
       throw new Error(
-        `Expected experimentType to begin with "normandy-pref". Found "${experimentType}".`
+        `experimentType must be less than ${MAX_EXPERIMENT_SUBTYPE_LENGTH} characters. ` +
+        `"${experimentType}" is ${experimentType.length} long.`
```
gijsk: Nit: missing a unit - 'characters' ? :-)

mythmon: I think that since the previous line gives the unit (characters) it is fine to not repeat it.

PR #1117, opened 2017-10-26
Fixes #1116 - Add copy buttons to argument values

### LINKED ISSUE #1116: Spaces Prepended to recipe copy
In the recipe view, if you double click one of the `<dd>` value fields and copy, you get four spaces of prepended text.  Looks fine in inspector, not sure what I'm missing.

This may sound minor, but I needed to copy a URL into a cloned survey and managed to miss the prepended spaces.

mythmon:
This depends on #1114, since it uses the `ArgumentsValue` component introduced there. The only commit specific to this PR is 76a91d1.

### DIFF
--- a/recipe-server/client/control/components/recipes/RecipeDetails.js
+++ b/recipe-server/client/control/components/recipes/RecipeDetails.js
@@ -1,8 +1,9 @@
-import { Card } from 'antd';
+import { Card, Icon } from 'antd';
 import { is, Map } from 'immutable';
 import PropTypes from 'prop-types';
 import React from 'react';
 import { connect } from 'react-redux';
+import CopyToClipboard from 'react-copy-to-clipboard';
 
 import { addSessionView } from 'control/state/app/session/actions';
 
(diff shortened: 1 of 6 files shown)


recipe-server/client/control/components/recipes/RecipeDetails.js
```diff
@@ -104,14 +104,29 @@ class ArgumentsValue extends React.PureComponent {
...
 
+    let inner;
+
     if (name === 'branches') {
-      return this.renderBranchTable(value);
+      inner = this.renderBranchTable(value);
+    } else if (name === 'extra_filter_expression') {
+      inner = <pre><code>{value}</code></pre>;
+    } else if (typeof value === 'boolean') {
+      inner = value ? 'True' : 'False';
```
andymikulski: I think it'd be a good idea to abstract out these `inner = ...` bits, like we have for `renderBranchTable`. This would allow us to write a bit more granular tests against the `ArgumentsValue` component, plus make the `render` function a bit more concise.

recipe-server/client/control/components/recipes/RecipeDetails.js
```diff
@@ -104,14 +104,29 @@ class ArgumentsValue extends React.PureComponent {
...
+    } else {
+      inner = value;
     }
 
-    if (typeof value === 'boolean') {
-      return <span>{value ? 'True' : 'False'}</span>;
-    }
+    const textToCopy = (value === undefined) ? "" : value.toString();
```
andymikulski: nit: This could be condensed down to `(value || '').toString()`

recipe-server/package.json
```diff
@@ -42,6 +42,7 @@
     "react": "15.3.1",
     "react-addons-css-transition-group": "15.3.1",
     "react-breadcrumbs": "1.3.16",
+    "react-copy-to-clipboard": "^5.0.1",
```
andymikulski: This package version should be pinned.

recipe-server/client/control/components/recipes/RecipeDetails.js
```diff
@@ -104,14 +104,29 @@ class ArgumentsValue extends React.PureComponent {
   render() {
     const { name, value } = this.props;
 
+    let inner;
```
andymikulski: I wrote the below comment first but looking at this now, I wonder if this whole section of logic that determines `inner` could be moved out to another function entirely. That may be too much abstraction when combined with the below comment, though. Open to thoughts.

mythmon: I think moving the `inner` logic into another function doesn't reduce complexity, it just shifts it to another function in the same context. I don't think that's beneficial. If there was a opportunity for re-use or simplification, I'd go for it, but I don't think there is here.

mythmon:
Thanks for pushing for tests on this. Instead of incurring more tech debt, I added tests. I also rebased this and squashed some commits to make it no longer dependent on #1114.

Issue #1100, opened 2017-10-09
When rolling back default-branch pref-studies, roll back to current default


mythmon:
Currently, when we roll back a defualt-branch pref study, we set the default branch's value to what it was when we started the study. Instead, we should set it to whatever it would be in the current version of Firefox had Shield not intervened.

For a concrete example, in Firefox 55, Screenshots was disabled by default. There was a default-branch pref study to enable it. In Firefox 56, Screenshots was enabled by default. The pref study only targeted 55. when a user upgraded from 55 to 56, Shield would unenroll them from the experiment, setting the value of the preference to disabled, since that is what it was before the study began. The correct behavior would be to set to enabled (ie, no change). [more details](https://mail.mozilla.org/pipermail/dev-shield/2017-September/000302.html).

Since the default branch is not persistent, users affected by this bug revert back to the desired behavior after restarting Firefox once.

PR #1119, opened 2017-10-27
PR #1100 - Revert study preferences to empty when appropriate.

mythmon:
This fixes an issue that came up in the QA of #1100. When the default value has changed during the experiment to an empty value (which happened for screenshots), this causes it to revert correctly.


mythmon:
Waiting to hear back from QA about this change before merging.

mythmon:
This has passed QA now.

PR #1121, opened 2017-10-31
Improved UI error handling

andymikulski:
- Adds `handleError` util function
  - Given a `context` and an Error, determines an appropriate message to display to the user.
  Ex: `handleError('Recipe creation failed.', someError)` returns `'Recipe creation failed. Please correct the form errors below.'`
  - Checks if the server failed, if the user is not currently connected, or if there was a form validation error.
  - Falls back to displaying the plain error message given from the server response.
- Fixes issue where validation errors would not appear in recipe argument fields

This could use some more fleshing out, for instance mythmon had mentioned catching expired login sessions. I think we can implement those features in the future, but what's in this PR should be a good start.

### DIFF
--- a/recipe-server/client/control/components/extensions/CreateExtensionPage.js
+++ b/recipe-server/client/control/components/extensions/CreateExtensionPage.js
@@ -19,7 +19,7 @@ import {
   },
 )
 @autobind
-export default class CreateExtensionPage extends React.PureComponent {
+export default class CreateExtensionPage extends React.Component {
   static propTypes = {
     createExtension: PropTypes.func.isRequired,
     push: PropTypes.func.isRequired,
--- a/recipe-server/client/control/components/extensions/CreateExtensionPage.js
+++ b/recipe-server/client/control/components/extensions/CreateExtensionPage.js
@@ -37,7 +37,7 @@ export default class CreateExtensionPage extends React.PureComponent {
   async handleSubmit(values) {
     const { createExtension, push } = this.props;
 
-    this.setState({ formErrors: undefined, });
+    this.setState({ formErrors: undefined });
 
     try {
       const extensionId = await createExtension(values);
--- a/recipe-server/client/control/components/extensions/CreateExtensionPage.js
+++ b/recipe-server/client/control/components/extensions/CreateExtensionPage.js
@@ -5,6 +5,7 @@ import React from 'react';
 import { connect } from 'react-redux';
 import { push as pushAction } from 'redux-little-router';
 
+import handleError from 'control/utils/handleError';
 import ExtensionForm from 'control/components/extensions/ExtensionForm';
 import {
   createExtension as createExtensionAction,
@@ -40,12 +41,9 @@ export default class CreateExtensionPage extends React.Component {
       message.success('Extension saved');
       push(`/extension/${extensionId}/`);
     } catch (error) {
-      message.error(
-        'Extension cannot be saved. Please correct any errors listed in the form below.',
-      );
-      if (error.data) {
-        this.setState({ formErrors: error.data });
-      }
+      handleError('Extension cannot be saved.', error);
+
+      this.setState({ formErrors: error.data });
     }
   }
 
--- a/recipe-server/client/control/components/extensions/CreateExtensionPage.js
+++ b/recipe-server/client/control/components/extensions/CreateExtensionPage.js
@@ -20,7 +20,7 @@ import {
   },
 )
 @autobind
-export default class CreateExtensionPage extends React.Component {
+export default class CreateExtensionPage extends React.PureComponent {
   static propTypes = {
     createExtension: PropTypes.func.isRequired,
     push: PropTypes.func.isRequired,
@@ -36,6 +36,9 @@ export default class CreateExtensionPage extends React.Component {
    */
   async handleSubmit(values) {
     const { createExtension, push } = this.props;
+
+    this.setState({ formErrors: undefined, });
+
     try {
       const extensionId = await createExtension(values);
       message.success('Extension saved');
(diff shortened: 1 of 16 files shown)


andymikulski:
TC failure is unrelated (`addon:lint`)

recipe-server/client/control/utils/handleError.js
```diff
@@ -0,0 +1,45 @@
...
+  const methods = { ...defaultMethods, ...methodOverrides };
+  let errMsg = '';
+
+  if (error) {
+    if (!methods.checkUserOnline()) {
+      errMsg = ERR_MESSAGES.NO_INTERNET;
+    } else if (error.message.indexOf(fetchFailureMessage) > -1) {
+      errMsg = ERR_MESSAGES.SERVER_FAILED;
+    } else if (error instanceof APIClient.APIError && error.data) {
+      errMsg = ERR_MESSAGES.FORM_VALIDATION;
```
mythmon: This isn't really accurate. You are pretty much ignoring the actual error from the server and assuming any time that the server responded with an error is a form validation error. I would like this to be much surer of itself before giving the form validation error message. It would be preferable to give uglier but more information dense answers instead of hiding details.

An example of this is a permission error. If you create a new user and don't give them any permissions, creating a recipe returns HTTP 403 and the body `{"detail":"You do not have permission to perform this action."}`. Compare to a form validation error: HTTP 400 with the body `{"name":["This field may not be blank."]}`. 

I think checking HTTP status codes is a necessary part of error handling that we don't do. A smaller change would be at least exposing the `detail` field if it exists.

andymikulski: The `APIClient` already handles determining the `message` to return with the server error, I assumed it was safe for `handleError` to inherit it from there:
```
// control/utils/api.js
...
if (!response.ok) {
  let message;
  let data;
  let err;

  try {
    data = await response.json();
    message = data.detail || response.statusText;
  } catch (error) {
    message = error.message;
    err = error;
  }

  throw new APIClient.APIError(message, data, err);
}
...
```

I agree that checking against the `data` prop for the form validation errors is weak, I can find a way to strengthen that. I can work on `api.js` to inspect the status codes, as well.

recipe-server/client/control/components/extensions/CreateExtensionPage.js
```diff
@@ -19,7 +20,7 @@ import {
   },
 )
 @autobind
-export default class CreateExtensionPage extends React.PureComponent {
```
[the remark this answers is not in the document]
andymikulski: The issue between `Pure/Components` is with regard to Ant's form/error handling. Ant uses plain JS objects for the `form` and `formErrors` props, which doesn't play well with `PureComponent`.

This page, however, shouldn't have been switched back to `Component`. I've gone through and corrected this. As for documenting when to use `Pure` vs regular components, I'd prefer to place that in some sort of centralized location, versus including the comment in each of the ~5 files. Any suggestions on that?

andymikulski:
After digging into not only the various error response codes, but the actual _display_ of the error messages, I've found some inconsistencies in our forms' validation/error handling.

For instance, we've been skipping front-end validation entirely, since all fields are marked as optional. This has worked fine since the back end validates on its own, which is why we haven't really noticed. There were also a handful of spots (around approval requests) where we didn't do any sort of validation/error handling. The more I dig into our forms, the more I'd like to refactor them. To prevent blowing up this PR, I will do open another with those changes.

This is ready for another pass!

mythmon:
TC failure is fixed on master.

andymikulski:
This is now passing CI, though needs a follow-up review by @mythmon .

recipe-server/client/control/components/extensions/CreateExtensionPage.js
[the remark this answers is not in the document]
mythmon: It would be good to have guidelines in the docs like "use pure components where possible" and "watch out for these known issues". That provides a good source of truth and a better onboarding experience.

However, I think also putting comments next to the components reminding us of the hard to see problems of pure components in some areas would also be good. This would prevent people from changing them in the future during refactoring. It would also ease code review, so I stop asking dumb questions like this. 😄

recipe-server/client/control/utils/handleError.js
```diff
@@ -0,0 +1,84 @@
...
+const handleAPIError = error => {
+  let message = '';
+
+  switch (error.data.status) {
+    case 400: // Bad Request
+      message = ERR_MESSAGES.FORM_VALIDATION;
+      break;
+
+    case 403: // Forbidden
+      message = checkLoginFailure(error) ? ERR_MESSAGES.NOT_LOGGED_IN : ERR_MESSAGES.NO_PERMISSION;
```
mythmon: Technically 403 only applies to a user that is logged in but not authorized to perform an action, whereas 401 is for users that aren't logged in.

We don't do this correctly in dev environments, and server 403 in all situations, which isn't good. I imagine we might serve 401s in prod some times though. We should probably include `case 401:` here as well.

andymikulski: Should the `checkLoginFailure` remain, or should this suffice:
```
    case 401: // Not logged in
      message = ERR_MESSAGES.NOT_LOGGED_IN;
      break;

    case 403: // Forbidden
      message = ERR_MESSAGES.NO_PERMISSION;
      break;
```

mythmon: That bit of code is technically correct, but not compatible with Normandy's server, unfortunately. I think that keeping the call to `checkLoginFailure` in the 403 branch, and also including the 401 branch would work better.

andymikulski:
TC failure is due to addon lint:

```
/root/mozilla-central/browser/extensions/shield-recipe-client/lib/RecipeRunner.jsm
  65:25  error  Multiple spaces found before '// sets this.e...'.  no-multi-spaces (eslint)
```

PR #1124, opened 2017-11-01
Update recipe-client-addon to use eslint 4 and latest eslint-plugin-mozilla.

Standard8:
In bug 1371293 we're working on updating mozilla-central to ESLint 4. Hence also looking at getting repositories that are imported into m-c updated as well. This updates the add-on to use eslint 4.

At the same time I'm updating a few of the libraries that it uses. Note the eslint-plugin-no-unsanitized addition is a peer dependency for eslint-plugin-mozilla.


recipe-client-addon/package.json
```diff
@@ -26,7 +27,6 @@
     "webpack": "3.1.0"
   },
   "dependencies": {
-    "mozjexl": "1.1.5",
```
mythmon: Why did you remove this?

Standard8: I think npm did it automatically - it is a duplicate of the current line 31.

PR #1128, opened 2017-11-09
Improve checks in response to 2017-11-07 outage

mythmon:
These code changes were made during the outage to alleviate problems found during recovery. None of them would have fixed the bug.

Currently hearbeat is disabled on production because it lacks these code changes. This leaves us very vulnerable to problems. These changes are high priority because of that. We'd like to deploy these changes today if possible.

@Osmose Rehan is busy with other high priority Pioneer things. Can you review this?

### DIFF
--- a/recipe-server/normandy/recipes/tests/test_checks.py
+++ b/recipe-server/normandy/recipes/tests/test_checks.py
@@ -5,6 +5,7 @@ import pytest
 from normandy.recipes import checks, signing
 from normandy.recipes.tests import RecipeFactory, SignatureFactory
 
+
 @pytest.mark.django_db
 class TestSignaturesUseGoodCertificates(object):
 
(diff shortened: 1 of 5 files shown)


recipe-server/normandy/recipes/tests/test_checks.py
```diff
@@ -0,0 +1,48 @@
+from datetime import timedelta
+
+import pytest
+
+from normandy.recipes import checks, signing
+from normandy.recipes.tests import RecipeFactory, SignatureFactory
+
+@pytest.mark.django_db
+class TestSignaturesUseGoodCertificates(object):
```
mythmon: I'm not sure why I resisted writing tests for the checks before. This was a lot easier than I expected it to be.

recipe-server/normandy/recipes/tests/test_checks.py
```diff
@@ -0,0 +1,49 @@
...
+        mock_verify_x5u = mocker.patch('normandy.recipes.checks.signing.verify_x5u')
+
+        def side_effect(x5u, *args):
+            if 'bad' in x5u:
+                raise signing.BadCertificate('testing exception')
+            return True
+        mock_verify_x5u.side_effect = side_effect
+
+        errors = checks.signatures_use_good_certificates(None)
+        assert not mock_verify_x5u.assert_called_once_with(recipe.signature.x5u, None)
```
Osmose: You'll want to remove the `assert not` here (and in the test below as well).

PR #1131, opened 2017-11-13
UI: Abstract Labeled Input from SwitchBox

andymikulski:
- Abstracts common `SwitchBox`/`CheckBox` code into a shared class, `LabeledInput`
- `LabeledInput` lets us wrap Ant's form components with the same label classes/styles we've established in the `SwitchBox`
- This allows us to more easily create input controls that fit within the conventions of our forms. For example, to make a labelled `Rating` input control, the code is simply:

```
import { Rate } from 'antd';
import autobind from 'autobind-decorator';

import LabeledInput from './LabeledInput';

@autobind
export default class Rating extends LabeledInput {
  static MAX_RATING = 5;

  getElement() {
    return Rate;
  }

  getElementProps() {
    return { value: this.props.value };
  }

  handleLabelClick() {
    let nextValue = (this.props.value || 0) + 1;
    nextValue = nextValue % (Rating.MAX_RATING + 1);

    this.props.onChange(nextValue);
  }
}
```

Currently, the `LabeledInput` component needs tests, but this is ready for feedback.

### DIFF
--- a/recipe-server/client/control/components/forms/LabeledInput.js
+++ b/recipe-server/client/control/components/forms/LabeledInput.js
@@ -1,6 +1,6 @@
 // eslint does not detect that aria roles come from `getLabelProps`,
 // ignoring the rule here to disable lint errors.
-/* eslint-disable jsx-a11y/interactive-supports-focus */
+/* eslint-disable jsx-a11y/no-static-element-interactions */
 
 import autobind from 'autobind-decorator';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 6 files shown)


andymikulski:
I've replaced the tests for the `SwitchBox` component with tests for `LabeledInput`. I don't totally feel the need to write tests for the `[Switch|Check]Box`components, but it would be trivial to throw those together, so that's an option.

Ready for review.

recipe-server/client/control/components/forms/LabeledInput.js
```diff
@@ -0,0 +1,59 @@
...
+
+import autobind from 'autobind-decorator';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+@autobind
+export default class LabeledInput extends React.Component {
+  static propTypes = {
+    children: PropTypes.node,
+    element: PropTypes.node,
```
andymikulski: I realize now that this prop is outdated/unnecessary.

PR #1132, opened 2017-11-13
Introduce yarn

andymikulski:


mythmon:
There are a few places in the docs that need updated as well:

```shell
~/src/normandy $ rg 'npm install' docs
docs/dev/recipe-client-addon/install.rst
31:      npm install

docs/dev/recipe-client-addon/workflow.rst
15:   npm install
35:      npm install --save ajv@5.2.2
152:   npm install

docs/dev/recipe-server/workflow.rst
38:   npm install

docs/dev/recipe-server/install.rst
70:      npm install
```

recipe-server/Dockerfile
```diff
@@ -11,16 +11,16 @@ RUN curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add -
     echo 'deb https://deb.nodesource.com/node_8.x jessie main' > /etc/apt/sources.list.d/nodesource.list && \
     echo 'deb-src https://deb.nodesource.com/node_8.x jessie main' >> /etc/apt/sources.list.d/nodesource.list && \
     apt-get update && apt-get install -y nodejs
-RUN npm install -g npm@5
+RUN npm install -g npm@5 && npm install -g yarn@1.0.0
```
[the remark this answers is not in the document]
mythmon: Also, to quote [the Yarn website](https://yarnpkg.com/en/docs/install#alternatives-tab):

> Installation of Yarn via npm is generally not recommended. When installing Yarn with Node-based package managers, the package is not signed, and the only integrity check performed is a basic SHA1 hash, which is a security risk when installing system-wide apps.
>
> For these reasons, it is highly recommended that you install Yarn through the installation method best suited to your operating system.

We can probably install via apt, just like we do for nodejs, using the [instructions here](https://yarnpkg.com/en/docs/install#linux-tab) (CI runs on Debian-based images).

andymikulski:
@mythmon Any idea why Circle is failing? Here's the very last bit of the error log:

```
Error removing intermediate container 0323957910e5: nosuchcontainer: No such container: 0323957910e51b1f53f948091702b1038b7a358ce550a254a566e9d6434be20a
Step 6 : COPY . /app
 ---> 64a76754647d
Error removing intermediate container 0323957910e5: nosuchcontainer: No such container: 0323957910e51b1f53f948091702b1038b7a358ce550a254a566e9d6434be20a
Successfully built 64a76754647d
~/normandy

./ci/circleci/bin/runner.sh dependencies returned exit code 1

Action failed: ./ci/circleci/bin/runner.sh dependencies
```

Alternatively, is there any way for me to run the CI tests locally?

mythmon:
Errors like `Error removing intermediate container 0323957910e5...` aren't concerning, they are due to a limitation in CircleCI's docker configuration.

There isn't a perfect way to run CI locally, but executing the scripts in `bin/ci/` can get pretty close. In this case you'd probably only need to run `bin/ci/dependencies`.

As for what the actual error is: it is buried in the rest of the output, and my usual strategy is to search for the string "fail:" (with the colon, it almost always matches our failures). With that, I found this:

```
Step 8 : RUN apt-get update && apt-get install yarn
 ---> Running in 3cb55cc0c75c
E: Type 'OK' is not known on line 1 in source list /etc/apt/sources.list.d/yarn.list
E: The list of sources could not be read.
The command '/bin/sh -c apt-get update && apt-get install yarn' returned a non-zero code: 100
FAIL: dependencies for recipe-server
```

which looks to me like something went wrong adding the yarn sources. You should be able to reproduce this failure just by building the docker container by running `docker build .` in the `normandy/recipe-server` directory.

recipe-server/Dockerfile
```diff
@@ -13,14 +13,18 @@ RUN curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add -
     apt-get update && apt-get install -y nodejs
 RUN npm install -g npm@5
 
+RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - \
+    echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
```
mythmon: The usage of `tee` in the docs was likely to make sudo usage possible. Since we don't need sudo, it would probably be best to do this the same way we do the nodesource repositories above, with simple shell redirection.

It would also be good to add this to the `RUN` command on line 10, so we don't add more steps to the docker file (which makes builds slower and the Docker image's file size larger)

recipe-server/Dockerfile
```diff
@@ -13,14 +13,18 @@ RUN curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add -
     apt-get update && apt-get install -y nodejs
 RUN npm install -g npm@5
```
mythmon: If we're using Yarn now, we can probably skip on installing npm5, yes?

andymikulski: I thought npm was still required under the hood, but I was incorrect. I'll remove it.

andymikulski:
I managed to get things running fine locally, though I had to also move a `COPY` command, let me know if there's an issue with that. The reason I had to move it was the `postinstall` command looks for `vendor.webpack.config.js`, but threw an error since it hadn't been copied to `app` yet. As far as I can tell this should be okay, but I may be wrong.

recipe-server/Dockerfile
```diff
@@ -10,17 +10,17 @@ RUN apt-get update && \
...
-    apt-get update && apt-get install -y nodejs
-RUN npm install -g npm@5
+    apt-get update && apt-get install -y nodejs && \
+    curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
+    echo 'deb https://dl.yarnpkg.com/debian/ stable main' > /etc/apt/sources.list.d/yarn.list && \
+    apt-get update && apt-get install yarn
 
-COPY ./requirements /app/requirements
-COPY ./package.json /app/package.json
+COPY . /app
```
mythmon: Moving app up this far will make the Docker layer cache miss all future steps most of the time, making working with Docker locally a lot slower (though it doesn't affect CircleCI). I think it would be better to continue copying only the things needed (`requirements/`, `package.json`, `package-lock.json`).

Do we need to keep the vendor post-install step? It can be convenient, but I think it mostly gets in my way personally. Alternatively, we could also include the webpack vendor file in this copy.

That being said, since we rarely do local development using Docker, the layer cache isn't actually saving us much work. If this is hard to fix, we can leave it as it is now. It is certainly simpler.

andymikulski: > Do we need to keep the vendor post-install step?

Not really. I had envisioned it being useful as we add new packages, but it indeed has become more of a hassle than anything. I've removed it from the `package.json` and reverted the `COPY` stuff in the Dockerfile. I also removed an errant `yarn` dependency I found in the `package.json`.

recipe-server/Dockerfile
```diff
@@ -15,11 +15,13 @@ RUN curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add -
     echo 'deb https://dl.yarnpkg.com/debian/ stable main' > /etc/apt/sources.list.d/yarn.list && \
     apt-get update && apt-get install yarn
 
-COPY . /app
+COPY ./requirements /app/requirements
+COPY ./package.json /app/package.json
```
mythmon: This will need `yarn.lock` too.

mythmon:
By the way, although I like this change, and it will likely be pretty nice for a lot of reasons, I don't expect it will impact CI times very much. Installing NPM dependencies is a minority of the time we spend, and a lot of Yarn's super powers come from really good caching which we don't get in CI. I expect we will get a small speedup for NPM  installs, and that the change to the overall CI time will be lost in the noise.

This is still something I'm glad to do. As I've worked more with NPM 5's package locking solution, the less impressed I've become. It will help local development, and I think it is overall a better tool for the job.

PR #1133, opened 2017-11-14
Squash migrations

### LINKED ISSUE #1023: Collapse migrations
We have a bunch of migrations. We should collapse most of them. This will speed up tests and new developer set up, and let us remove some older libraries.

We should probably keep migrations back to the last major deployed version (v55). Anything prior to that we can collapse.

rehandalal:
Fixes #1023.

r?


mythmon:
Everything worked this time, but the migrations failed lints. I thought we had a rule to ignore migrations in flake8. Maybe it doesn't understand the squashed version name? The TC failure is #1136.

PR #1138, opened 2017-11-17
Introduce Photon

andymikulski:
🎉

- Introduces `photon-ant` package
- Refactors existing custom styles to better fit the Photon ａｅｓｔｈｅｔｉｃ
- Swap icons for photon equivalents where possible

Note this is an in-progress development that will have to span the course of a few PR's. I'd like to open up this PR for feedback/review as-is, since it's already getting pretty large.

It's probably a good idea to load the site locally and check things out visually, as there are a lot of CSS changes.

---

screen shot 2017-11-17 at 12 25 32 pm
screen shot 2017-11-17 at 12 07 00 pm
screen shot 2017-11-17 at 12 07 08 pm


andymikulski:
Also should mention I'm certainly open to general visual feedback/critique, `photon-ant` is still pretty young so it has some iteration to go through.

andymikulski:
Circle's failure relates to security issues with some packages required through a `photon-ant` dependency. I will have to resolve that within the `photon-ant` package itself, however this PR is still good for review.

mythmon:
> Circle's failure relates to security issues with some packages required through a photon-ant dependency. I will have to resolve that within the photon-ant package itself, however this PR is still good for review.

Once we have Yarn set up, it may be possible to pin the version of the dependency in this project, instead of updating it in photon-ant. It is probably worth-while to update in both places though.

recipe-server/client/control/less/components/history-timeline.less
```diff
@@ -14,10 +14,10 @@
...
 }
 
-.ant-timeline-item-head-yellow {
-  border-color: @yellow-6;
-  color: @yellow-6;
-}
+// .ant-timeline-item-head-yellow {
+//   border-color: @yellow-6;
+//   color: @yellow-6;
+// }
```
mythmon: We shouldn't commit commented-out code.

mythmon:
This is very bright and shiny.

I don't see any major problems poking around the site, except that the section(?) around submit buttons looks very much like a text field:

screenshot-2017-11-20 shield control panel 1
screenshot-2017-11-20 shield control panel

I'm  also not a fan of how much the logout button stands out in the new colors. It is much higher contrast and visually heavier than most of the rest of the page. I think this also applies to a lesser extent to the delete branch buttons in the preference experiment arguments editor.

andymikulski:
>This is very bright and shiny.

I can't tell if this is good or bad.

>logout/delete branch buttons

Good point on these. I think we can replace the bright red buttons with 'danger' buttons which simply turn red on hover. I will see what options we have.

mythmon:
>> This is very bright and shiny.
>
> I can't tell if this is good or bad.

Jokingly: neither can I.

More seriously: I don't think it is bad. It definitely *looks* like Photon, which was the goal. It is certainly a change. The biggest thing I noticed is that the "you're on stage" warning is a very loud, bright yellow. I don't think that is a big problem though, since most users won't be on staging environments 🙂. The rest of the colors on the site actually work pretty nicely.

andymikulski:
Aight, I've gone through and addressed the feedback, plus fixed a handful of minor things I saw. Some feedback (specifically the radio button group) required actual `photon-ant` changes, which I have not published yet. I will get that package updated today and update this PR accordingly.

mythmon:
CI failures are #1143, which is an unrelated NSP error.

mythmon:
I'm not sure if the below is related to the photon-ant changes you mentioned, but I figured I'd call them out anyways.

The new boolean switches look kind of funny, color wise. Specifically the blue-on-red. I'd go with a neutral color for the text here:

screenshot-2017-11-27 shield control panel

There are some odd focus issues in the history sidebar. You can see these either by tabbing to them, or just by clicking them:

screenshot from 2017-11-27 15-34-22

I think the rest of the site looks pretty good. Nice work!

andymikulski:
The weird colors on the True/False buttons are `photon-ant` related, but I had not caught the link focus on the timeline. Good catch!

andymikulski:
This is probably going to need another round of feedback now that I've released `photon-ant@0.1.3`, I will assign to myself until it's ready.

andymikulski:
`photon-ant` changes have invalidated some of this PR, I will reopen this when the code is closer to landing.

PR #1143, opened 2017-11-27
Ignore NSP-532, since a fix is not currently available.

mythmon:
This is a security problem in the moment js library. I don't believe we are vulnerable to this problem, since we only use in client side. There is not fix available yet. I'm tracking this issue, and will file a PR to include a fix once it is published.

In the meantime, this is breaking the build, so ignoring it until there is a fix available seems to me the best way forwards.

### DIFF
--- a/recipe-server/.nsprc
+++ b/recipe-server/.nsprc
@@ -1,0 +1,5 @@
+{
+  "exceptions": [
+    "https://nodesecurity.io/advisories/532"
+  ]
+}


mythmon:
Updated to include a link to the moment issue. Hopefully TC passes this time.

mythmon:
@andymikulski it seems like NSP doesn't like comments in it's config file, and it was silently ignoring the config file. That made CI break. I've removed the comment, and added more info to the commit message. r?

PR #1144, opened 2017-11-30
Fix #1077 - Show all breadcrumbs

### LINKED ISSUE #1077: Breadcrumbs not showing items
Currently, when viewing a recipe, the breadcrumbs display as `Home/View Recipe`, when it should actually display as `Home/Recipe Listing/View Recipe`.

This is happening as a result of the change to trailing slashes in our URLs. Take this router snippet for example:

```
'/recipe': {
  '/': {
    component: RecipeListing,
    crumb: 'Recipes Listing',
  },
  '/new': {
    '/': {
      component: CreateRecipePage,
      crumb: 'New Recipe',
    },
  },
  ...
}
```

Say the user is on `/recipe/new/`. The breadcrumbs add a crumb for `New Recipe`, move up a parent, find there isn't a crumb on `/new`, skips it, doesn't find a crumb on `/recipe`, and continues to move on until there is a crumb to display.

There are a few options, but I think the crumbs need to smartly identify if there is an entry for `'/'` and if there is a breadcrumb attached to it.

mythmon:
I also wrote tests for all the other router selectors.

### DIFF
--- a/recipe-server/client/control/components/common/NavigationCrumbs.js
+++ b/recipe-server/client/control/components/common/NavigationCrumbs.js
@@ -5,58 +5,25 @@ import React from 'react';
 import { connect } from 'react-redux';
 import { Link } from 'redux-little-router';
 
-import { replaceUrlVariables } from 'control/routerUtils';
+import { getBreadcrumbs } from 'control/state/router/selectors';
 
 @connect(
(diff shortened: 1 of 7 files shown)


mythmon:
CI failure is #1143.

PR #1145, opened 2017-12-01
Update Moment for  NSP-532 and make Yarn use better.

mythmon:


recipe-server/package.json
```diff
@@ -35,7 +35,7 @@
     "json-loader": "0.5.4",
     "localforage": "1.4.3",
     "lodash.get": "4.4.2",
-    "moment": "2.14.1",
+    "moment": "^2.19.3",
```
mythmon: I propose that now that Yarn is pinning and hashing our versions, using semver in package.json is the correct thing to do, when we don't need an exact version for compatibility reasons.

PR #1148, opened 2017-12-08
Bug 1416153 - Part 0: Fix tests that uses BrowserTestUtils.waitForNew Tab to perform the registration of the next event handler instantly inside the new tab event handler. r=Gijs

mythmon:
This is a backport from [Bug 1416153](https://bugzilla.mozilla.org/show_bug.cgi?id=1416153). It helps with race conditions in tests that we hit when trying to sync v80.


mythmon:
bors r=gijsk

PR #1149, opened 2017-12-12
Fix high population switch not saving.

mythmon:
The Python tests are a sanity check that nothing is going wrong in the API. The changes in the JS should be better form code that doesn't accidentally throw away data.

@rehandalal r?


notriddle:
error: could not rename component directory from '/root/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/etc' to '/root/.rustup/tmp/hndv_r6y7cutu54x_dir/bk'

wat 🦆

mythmon:
@notriddle this is a problem we see commonly in our CI. It is something that would be a good match for bors-ng/bors-ng#157. We do full builds of Firefox, which rely on keeping an up-to-date version of Rust. Our build image is static, so about every 6 weeks we have to kick it to update Rust.

There may be a better solution to this, in the form of figuring out why rustup can't update itself, but I haven't looked into it yet.

mythmon:
bors r= @rehandalal

mythmon:
@rdalal The only failure was a simple lint problem. re-r?

rehandalal:
haz moar problems!

mythmon:
bors r= @rehandalal

PR #1150, opened 2017-12-13
Cachebust add-on builder Dockerfile to pick up new Rust version.

mythmon:
This will fix the TC build in #1149.


mythmon:
This is a routine change, so I'm going to self-r+ it.

bors r=me

PR #1153, opened 2017-12-18
#1151: "High Volume" toggle validation broken

### LINKED ISSUE #1151: "High Volume" toggle does not validate properly
If you wish to leave the "High Volume" toggle switch off when creating a preference experiment it thinks the switch is "empty" and tells you the field is required unless you switch it on and then off.

STR:
- Select preference experiment
- Fill in all required fields but do not touch the High Volume toggle switch
- Try to save

andymikulski:
Fixes #1151 

Initially thought this was an issue with one of the custom input components we made, turns out we were simply a missing a `false` Immutable fallback value.

Not totally sure how to write tests against this. It seemed a little excessive to write tests to specifically check if a form item has a certain default value, though maybe that's fine? Open to thoughts.


mythmon:
I don't think there is a reasonable way to test this in our current set up.

bors r+

PR #1154, opened 2017-12-18
Fix #1139 - Log study enrollment events via Telemetry Events

### LINKED ISSUE #1139: Log Study enrollment/unenrollment using Telemetry Events
It is possible to derive data about when a client enrolls and unenrolls from studies by watching the experiment table, but it is really hard. Since we implemented experiment marking in telemetry, Telemetry has added another new feature: Events. We should put enrollment and unenrollment, along with details about why the user enrolled/unenrolled in this new API.

mythmon:
This add documentation and implementation for enrollment events when studies start and stop. This includes both preferences and add-on studies.

@georgf Can you review this? The changes in `recipe-client-addon/` should be treated like Firefox code, and will eventually make its way into mozilla-central.

### DIFF
--- a/docs/user/data_collection.rst
+++ b/docs/user/data_collection.rst
@@ -83,3 +83,110 @@ reported under the key ``shield-recipe-client/recipe/<recipe id>``:
 .. js:data:: RECIPE_SUCCESS
 
    The recipe was executed successfully.
+
+
+Enrollment
+-----------
+Shield records enrollment and unenrollment of users into studies, and
+records that data using `Telemetry Events`_. All data is stored in the
+``normandy`` category.
+
+.. _Telemetry Events: https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/collection/events.html
+
+Preference Studies
+^^^^^^^^^^^^^^^^^^
+Enrollment
+   method
+      The string ``"enroll"``
+   object
+      The string ``"preference_study"``
+   value
+      The name of the study (``recipe.arguments.slug``).
+   extra
+      branch
+         The name of the branch the user was assigned to (example:
+         ``"control"`` or ``"experiment"``).
+      experimentType
+         The type of preference experiment. Currently this can take
+         values "exp" and "exp-highpop", the latter being for
+         experiments targetting large numbers of users.
+
+Unenrollment
+   method
+      The string ``"unenroll"``.
+   object
+      The string ``"preference_study"``.
+   value
+      The name of the study (``recipe.arguments.slug``).
+   extra
+      didResetValue
+         The string ``"true"`` if the preference was set back to its
+         original value, ``"false"`` if it was left as its current
+         value. This can happen when, for example, the user changes a
+         preference that was involved in a user-branch study.
+      reason
+         The reason for unenrollment. Possible values are:
+
+         * ``"recipe-not-seen"``: The recipe was no longer
+           applicable to this client This can be because the recipe
+           was disabled, or the user no longer matches the recipe's
+           filter.
+         * ``"user-preference-changed"``: The study preference was
+           changed on the user branch. This could mean the user
+           changed the preference, or that some other mechanism set a
+           non-default value for the preference.
+         * ``"user-preference-changed-sideload"``: The study
+           preference was changed on the user branch while Shield was
+           inactive. This could mean that the value was manually
+           changed in a profile while Firefox was not running.
+         * ``"unknown"``: A reason was not specificied. This should be
+           considered a bug.
+
+Add-on Studies
+^^^^^^^^^^^^^^
+Enrollment
+   method
+      The string ``"enroll"``
+   object
+      The string ``"addon_study"``
+   value
+      The name of the study (``recipe.arguments.slug``).
+   extra
+      addonId
+         The add-on's ID (example: ``"feature-study@shield.mozilla.com"``).
+      addonVersion
+         The add-on's version (example: ``"1.2.3"``).
+
+Unenrollment
+   method
+      The string ``"unenroll"``.
+   object
+      The string ``"addon_study"``.
+   value
+      The name of the study (``recipe.arguments.name``).
+   extra
+      addonId
+         The add-on's ID (example: ``"feature-study@shield.mozilla.com"``).
+      addonVersion
+         The add-on's version (example: ``"1.2.3"``).
+      reason
+         The reason for unenrollment. Possible values are:
+
+         * ``"install-failure"``: The add-on failed to install.
+         * ``"individual-opt-out"``: The user opted-out of this
+           particular study.
+         * ``"general-opt-out"``: The user opted-out of studies in
+           general.
+         * ``"recipe-not-seen"``: The recipe was no longer applicable
+           to this client. This can be because the recipe was
+           disabled, or the user no longer matches the recipe's
+           filter.
+         * ``"uninstalled"``: The study's add-on as uninstalled by some
+           mechanism. For example, this could be a user action or the
+           add-on self-uninstalling.
+         * ``"uninstalled-sideload"``: The study's add-on was
+           uninstalled while Shield was inactive. This could be that
+           the add-on is no longer compatible, or was manually removed
+           from a profile.
+         * ``"unknown"``: A reason was not specificied. This should be
+           considered a bug.
--- a/recipe-client-addon/lib/AddonStudies.jsm
+++ b/recipe-client-addon/lib/AddonStudies.jsm
@@ -91,8 +91,9 @@ function getStore(db) {
  * Mark a study object as having ended. Modifies the study in-place.
  * @param {IDBDatabase} db
  * @param {Study} study
+ * @param {String} reason Why the study is ending.
  */
-async function markAsEnded(db, study, reason = "unknown") {
+async function markAsEnded(db, study, reason) {
   if (reason === "unknown") {
     log.warn(`Study ${study.name} ending for unknown reason.`);
   }
@@ -321,11 +322,12 @@ this.AddonStudies = {
   /**
    * Stop an active study, uninstalling the associated add-on.
    * @param {Number} recipeId
+   * @param {String} reason Why the study is ending. Optional, defaults to "unknown".
    * @throws
    *   If no study is found with the given recipeId.
    *   If the study is already inactive.
    */
-  async stop(recipeId, reason) {
+  async stop(recipeId, reason = "unknown") {
     const db = await getDatabase();
     const study = await getStore(db).get(recipeId);
     if (!study) {
--- a/recipe-client-addon/lib/AddonStudies.jsm
+++ b/recipe-client-addon/lib/AddonStudies.jsm
@@ -35,10 +35,9 @@ XPCOMUtils.defineLazyModuleGetter(this, "FileUtils", "resource://gre/modules/Fil
 XPCOMUtils.defineLazyModuleGetter(this, "IndexedDB", "resource://gre/modules/IndexedDB.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "AddonManager", "resource://gre/modules/AddonManager.jsm");
 XPCOMUtils.defineLazyModuleGetter(this, "Addons", "resource://shield-recipe-client/lib/Addons.jsm");
-XPCOMUtils.defineLazyModuleGetter(
-  this, "CleanupManager", "resource://shield-recipe-client/lib/CleanupManager.jsm"
-);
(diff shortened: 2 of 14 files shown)


recipe-client-addon/lib/TelemetryEvents.jsm
```diff
@@ -0,0 +1,61 @@
...
+
+      unenroll: {
+        methods: ["unenroll"],
+        objects: ["preference-study", "addon-study"],
+        extra_keys: ["reason", "resetValue"],
+        record_on_release: true,
+      },
+    });
+
+    Services.telemetry.setEventRecordingEnabled(TELEMETRY_CATEGORY, true);
```
georgf: You don't need this for events that are registered at runtime through `registerEvents()`.
This is only needed for events shipping in Firefox code (to avoid collecting all the time).

recipe-client-addon/lib/TelemetryEvents.jsm
```diff
@@ -0,0 +1,61 @@
...
+    isStarted = true;
+
+    CleanupManager.addCleanupHandler(() => {
+      Services.telemetry.setEventRecordingEnabled(TELEMETRY_CATEGORY, false);
```
georgf: Ditto, this is not needed.

recipe-client-addon/lib/TelemetryEvents.jsm
```diff
@@ -0,0 +1,61 @@
...
+      isStarted = false;
+    });
+  },
+
+  sendEvent(method, object, value, extra) {
+    if (isStarted) {
+      Services.telemetry.recordEvent(TELEMETRY_CATEGORY, method, object, value, extra);
+    } else {
+      log.warn(
```
georgf: Telemetry already logs errors when you try to record events that are not registered.
Maybe that is sufficient?

mythmon:
Thanks for the review, I've unassigned you, and will remove the unneeded parts you pointed out.

@gijsk can you review the rest of the code? Most of it should be pretty straight forward. I'm also working with PI to do manual QA of this feature, to make sure it matches the spec laid out in the docs.

docs/user/data_collection.rst
```diff
@@ -83,3 +83,106 @@ reported under the key ``shield-recipe-client/recipe/<recipe id>``:
...
+         preference that was involved in a user-branch study.
+      reason
+         The reason for unenrollment. Possible values are:
+
+         * ``"recipe-not-seen"``: The recipe was no longer
+           applicable to this client This can be because the recipe
+           was disabled, or the user no longer matches the recipe's
+           filter.
+         * ``"user-preference-changed"``: The study preference was
+           changed on the user branch. This ocould mean the user
```
gijsk: Nit: `ocould`

docs/user/data_collection.rst
```diff
@@ -83,3 +83,106 @@ reported under the key ``shield-recipe-client/recipe/<recipe id>``:
...
+      The string ``"enroll"``
+   object
+      The string ``"addon-study"``
+   value
+      The name of the study (``recipe.arguments.slug``).
+   extra
+      addonId
+         The add-on's ID (example: ``"feature-study@shield.mozilla.com"``).
+      addonVersion
+         The add-on's ID (example: ``"1.2.3"``).
```
gijsk: Nit: version

docs/user/data_collection.rst
```diff
@@ -83,3 +83,106 @@ reported under the key ``shield-recipe-client/recipe/<recipe id>``:
...
+      The string ``"unenroll"``.
+   object
+      The string ``"addon-study"``.
+   value
+      The name of the study (``recipe.arguments.name``).
+   extra
+      addonId
+         The add-on's ID (example: ``"feature-study@shield.mozilla.com"``).
+      addonVersion
+         The add-on's ID (example: ``"1.2.3"``).
```
gijsk: Same here.

docs/user/data_collection.rst
```diff
@@ -83,3 +83,106 @@ reported under the key ``shield-recipe-client/recipe/<recipe id>``:
...
+      reason
+         The reason for unenrollment. Possible values are:
```
gijsk: From the code, looks like this could also be `"unknown"` ? But that value is not listed here...

recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -377,8 +382,11 @@ this.PreferenceExperiments = {
       observer() {
         const newValue = getPref(UserPreferences, preferenceName, preferenceType);
         if (newValue !== preferenceValue) {
-          PreferenceExperiments.stop(experimentName, false)
-                               .catch(Cu.reportError);
+          PreferenceExperiments.stop(experimentName, {
+            resetValue: false,
+            reason: "user-preference-changed",
+          })
+          .catch(Cu.reportError);
```
gijsk: I would put this on the previous line, but ymmv.

docs/user/data_collection.rst
```diff
@@ -83,3 +83,106 @@ reported under the key ``shield-recipe-client/recipe/<recipe id>``:
...
+   method
+      The string ``"unenroll"``.
+   object
+      The string ``"preference-study"``.
+   value
+      The name of the study (``recipe.arguments.slug``).
+   extra
+      resetValue
+         The string ``"true"`` if the preference was set back to its
+         original value, ``"false"`` if it was left as its current
```
gijsk: Out of curiosity, why strings that contain booleans? Why not just booleans?

Also, in the epitome of nitpicks, the name of this property makes me think its value is the value the preference was reset to, not whether or not it was reset. Dunno if there are obvious better choices. `didResetValue` or something?

recipe-client-addon/lib/ShieldPreferences.jsm
```diff
@@ -73,22 +73,24 @@ this.ShieldPreferences = {
     let prefValue;
     switch (prefName) {
       // If the FHR pref changes, set the opt-out-study pref to the value it is changing to.
-      case FHR_UPLOAD_ENABLED_PREF:
+      case FHR_UPLOAD_ENABLED_PREF: {
```
gijsk: Huh, are the braces a new eslint thing? I don't mind either way, just surprised...

recipe-client-addon/lib/TelemetryEvents.jsm
```diff
@@ -0,0 +1,61 @@
...
+
+    CleanupManager.addCleanupHandler(() => {
+      Services.telemetry.setEventRecordingEnabled(TELEMETRY_CATEGORY, false);
+      isStarted = false;
+    });
+  },
+
+  sendEvent(method, object, value, extra) {
+    if (isStarted) {
+      Services.telemetry.recordEvent(TELEMETRY_CATEGORY, method, object, value, extra);
```
gijsk: @georgf when, if ever, does `recordEvent` throw?

recipe-client-addon/test/browser/browser_AddonStudies.js
```diff
@@ -218,7 +226,7 @@ decorate_task(
   AddonStudies.withStudies(),
   async function testStopNoStudy() {
     await Assert.rejects(
-      AddonStudies.stop("does-not-exist"),
+      AddonStudies.stop("does-not-exist", "test"),
```
gijsk: Out of curiosity, what breaks if you don't provide a reason here? I thought there was a default value for the argument?

gijsk: To be clear, it makes sense to want to test the propagation of the reason, but that doesn't happen here - you've just updated all the existing consumers and I'm not entirely sure why.

recipe-client-addon/test/browser/browser_PreferenceExperiments.js
```diff
@@ -611,19 +647,36 @@ decorate_task(
...
+    await PreferenceExperiments.stop("test", {reason: "test"});
     ok(setInactiveStub.calledWith("test", "branch"), "Experiment is unregistered by stop()");
+
+    Assert.deepEqual(
+      sendEventStub.getCall(0).args,
+      ["enroll", "preference-study", "test", {
+        experimentType: "exp",
+        branch: "branch",
+      }],
+      "start should send the correct telemetry event"
```
gijsk: I like to have unique failure messages so I can quickly find out where things failed, but it's up to you.

docs/user/data_collection.rst
mythmon: According to [the docs](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/collection/events.html#recordevent), API only allows "string keys & values", so booleans aren't an option.

I think you're right about the name of `resetValue`. There isn't a lot of outside-of-the-code history to this name, so now is a good time to change it.

recipe-client-addon/lib/ShieldPreferences.jsm
mythmon: They aren't an eslint thing. I personally prefer them because it creates a scope for `let` and `const`. It doesn't apply here, but doing it consistently makes sense to me.

recipe-client-addon/test/browser/browser_AddonStudies.js
mythmon: To help with my development, I had made all the functions that take `reason` require it, and throw if it was not passed. Then I made all the tests pass, including new tests that included reason propagation. After I got it working, I removed the error throwing and provided a default value to make this code less likely to throw errors.

Thinking about it more, I should have either left the error reporting in, or gone through and removed the no-op reasons in the tests. Do you have opinions about which would be better?

gijsk: I don't really feel all that strongly either way. I think not throwing (and so removing the no-op reasons) might be a slightly safer bet given the few cases in the last 6 months or so where we had issues on startup. It might be nice to have a test that ensures that we end up with the default "unknown" value when no reason is passed.

mythmon:
I rebased this on master to fix the merge conflicts, and incorporated the feedback from @georgf and @gijsk. Apologies that I wasn't able to make it a second commit for a good interdiff, the rebase was not kind to workflow.

@gijsk Can you give this another pass and, if it is ready, tell bors r+?

I don't know if this is needed, or if you already have access, but it can't hurt:

bors delegate= @gijsk

recipe-client-addon/lib/PreferenceExperiments.jsm
```diff
@@ -454,8 +461,11 @@ this.PreferenceExperiments = {
    *   If there is no stored experiment with the given name, or if the
    *   experiment has already expired.
    */
-  async stop(experimentName, resetValue = true) {
+  async stop(experimentName, {resetValue = true, reason = "unknown"} = {}) {
```
gijsk: Nit: please update the docs.

recipe-client-addon/lib/AddonStudies.jsm
```diff
@@ -304,17 +325,17 @@ this.AddonStudies = {
    *   If no study is found with the given recipeId.
    *   If the study is already inactive.
    */
-  async stop(recipeId) {
+  async stop(recipeId, reason) {
```
gijsk: Nit: please update the docs.

recipe-client-addon/lib/AddonStudies.jsm
```diff
@@ -93,11 +92,21 @@ function getStore(db) {
  * @param {IDBDatabase} db
  * @param {Study} study
  */
-async function markAsEnded(db, study) {
+async function markAsEnded(db, study, reason = "unknown") {
```
gijsk: Docs here too

recipe-client-addon/test/browser/browser_PreferenceExperiments.js
```diff
@@ -752,8 +805,8 @@ decorate_task(
     mockPreferences.set("fake.preference", "changed value");
     await PreferenceExperiments.init();
     ok(stopStub.calledWith("test"), "Experiment is stopped because value changed");
-    is(Preferences.get("fake.preference"), "changed value", "Preference value was not changed");
-  }
+    ok(Preferences.get("fake.preference"), "changed value", "Preference value was not changed");
```
gijsk: Err, this is a strange change. `ok` only takes 1 test argument, and its value should be truthy for it to pass. Is this just a mistake? :-)

mythmon:
> as I don't want to accidentally get it to merge this before these last tiny bits of feedback are addressed.

Good call. Bors actually does the merging and doesn't understand r+wc. So had you told it r+, it would have done exactly what you feared. I've updates the docs, so I think this is good to go now.

bors r= @gijsk, @georgf

PR #1155, opened 2017-12-18
#1111: Paused enrollment indicator

### LINKED ISSUE #1111: Show paused status in preview
I've paused enrollment on a few studies and sent them to other folks for review. The only way for them to see the change is hit "edit" first. It would be nice if we showed the paused/live status in the main recipe view.

andymikulski:
Fixes #1111 (mostly)

Ticket #1111 requests an indicator on the recipe details page for 'paused enrollment', however this data already exists on the page: 
screen shot 2017-12-18 at 12 41 15 pm

The ticket requested adding this info to the listing page, which this PR does:
screen shot 2017-12-18 at 2 03 39 pm

This should, for the most part, close #1111, as the 'paused' status is now available on the `listing`, `details`, and `edit` pages.

### DIFF
--- a/recipe-server/client/control/components/common/EnrollmentStatus.js
+++ b/recipe-server/client/control/components/common/EnrollmentStatus.js
@@ -51,7 +51,7 @@ export default class EnrollmentSatus extends React.Component {
     } = this.props;
 
     return (
-      <Link href={`/recipe/${recipe.id}/`} className={cx(!recipe.enabled && 'is-lowkey')}>
+      <Link href={`/recipe/${recipe.id}/`} className={cx('status-link', !recipe.enabled && 'is-lowkey')}>
         <Icon
           className={cx('status-icon', this.getColor())}
           type={this.getIcon()}
--- a/recipe-server/client/control/components/common/EnrollmentStatus.js
+++ b/recipe-server/client/control/components/common/EnrollmentStatus.js
@@ -1,0 +1,83 @@
+import { Icon } from 'antd';
+import PropTypes from 'prop-types';
+import React from 'react';
+import { Link } from 'redux-little-router';
+import { connect } from 'react-redux';
+
+@connect((state, { record }) => {
+  const isEnabled = record.enabled;
+  const isPaused = !isEnabled || !!record.arguments.isEnrollmentPaused;
+
+  return {
+    record,
+    isEnabled,
+    isPaused,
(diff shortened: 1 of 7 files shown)


mythmon:
@MattGrimes do these screenshots look good to you?

recipe-server/client/control/components/common/EnrollmentStatus.js
```diff
@@ -0,0 +1,83 @@
+import { Icon } from 'antd';
+import PropTypes from 'prop-types';
+import React from 'react';
+import { Link } from 'redux-little-router';
+import { connect } from 'react-redux';
+
+@connect((state, { record }) => {
```
mythmon: This isn't actually using `state` anywhere, as far as I can see. We probably don't want to `@connect` things that don't read from Redux. I would expect things like `isEnabled` and `isPaused` to be calculated in `render`, since they are derived from this component's props.

andymikulski: That's actually how it started out, but for testing purposes I felt it was more straightforward to expose the `isEnabled`/`isPaused` props. I can disconnect it.

mythmon: I think using `isEnabled`/`isPaused` directly in tests skips out on testing an important bit of logic, that is, how those fields are determined.

andymikulski:
CI error is due to tests using `.find(Icon)`. I should probably figure out why that works locally for me but not on Circle.

mythmon:
> CI error is due to tests using .find(Icon). I should probably figure out why that works locally for me but not on Circle.

I imagine that you and Circle are using different versions of Firefox. Could that do it? It may also be a Linux vs OSX thing?

andymikulski:
I've removed the `connect` bits, and refactored some duplicate code.

I have also decided to remove the filtering for the `paused` column after hitting some friction getting it to work consistently. I imagine we can revisit this piece when we implement the back end functionality.

Ready for another pass!

mythmon:
This looks good now, thanks for the changes!

bors r+

mythmon:
I'm not sure why the test timed out, but it also has test failures like this that need fixing:

```
✗ should print `paused` if the recipe is enabled and paused
    Expected '<Connect(L) />' to contain 'Paused'.
```

Issue #1112, opened 2017-10-18
Ability to Pause pref flips


MattGrimes:
We've run into a couple of situations where we have hit our enrollment numbers for pref flips but needed to keep the recipe live for an additional period of time. It would be nice if we could pause enrollment for pref-flips the same way we can for native studies so we don't over-enroll.

mythmon:
Implementation note: opt-out studies pause enrollment in the action, not in the driver, so this is something we can fix without a system add-on update. This should  be a pretty easy change. The opt-out studies version of this is just a few lines: https://github.com/mozilla/normandy/blob/4a7bf8083cac71b521e909ee3e55bda4c318c67e/recipe-server/client/actions/opt-out-study/index.js#L40-L41

PR #1159, opened 2018-01-04
#1112: Add enrollment pausing to pref-experiments

andymikulski:


mythmon:
```
TOTAL: 1 FAILED, 287 SUCCESS

1) should do nothing if enrollment is paused
     PreferenceExperimentAction execute
     Failed: D.studies is undefined
@client/actions/tests/index.js:1:40841
```

Additionally, it seems we have a bug in Circle that if the JS tests fail, it hangs forever.

andymikulski:
This is ready for review, though it looks like TC failed immediately. I'll see if I pushed something weird or if it's a greater TC problem in general.

mythmon:
I think Taskcluster is ready now

bors r+

PR #1161, opened 2018-01-09
Fixes #1140 - Be more permissive with uploaded file names, including allowing @

### LINKED ISSUE #1140: "@" is being dropped from XPI filenames
When we upload an extension if it has an "@" in the filename, which is common given our naming structure, this "@" is dropped.

i.e.: `something@shield.mozilla.org.xpi` becomes `somethingshield.mozilla.org.xpi`.

mythmon:
I spend some quality time with the debugger to see what was stripping the `@` out of file names, but I eventually tracked it down.


rehandalal:
Were you able to confirm this actually works with S3? 

Also I wonder if this is too permissive.

As per [Amazon's S3 docs](https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html):

>**Characters That Might Require Special Handling**
> 
>The following characters in a key name may require additional code handling and will likely need to be URL encoded or referenced as HEX. Some of these are non-printable characters and your browser might not handle them, which also requires special handling: 
>
> . | . | .
>-- | -- | --
>Ampersand ("&") | Dollar ("$") | ASCII character ranges 00–1F hex (0–31 decimal) and 7F (127 decimal)
>'At' symbol ("@") | Equals ("=") | Semicolon (";")
>Colon (":") | Plus ("+") | Space – Significant sequences of spaces may be lost in some uses >(especially multiple                                        spaces)
>Comma (",") | Question mark ("?") |  
>
> **Characters to Avoid**
>
>You should avoid the following characters in a key name because of significant special handling for consistency across all applications. 
>
> . | . | . 
>-- | -- | --
>Backslash ("\") | Left curly brace ("{") | Non-printable ASCII characters (128–255 decimal                                        characters)
>Caret ("^") | Right curly brace ("}") | Percent character ("%")
>Grave accent / back tick ("`") | Right square bracket ("]") | Quotation marks
>'Greater Than' symbol (">") | Left square bracket ("[") | Tilde ("~")
>'Less Than' symbol ("<") | 'Pound' character ("#") | Vertical bar / pipe ("\|")

mythmon:
Good catch. I modified the storage mixin to remove some special characters (the second set from your link). I think that will be enough for us, since our storage driver will do the extra handling needed for the other characters.

PR #1164, opened 2018-01-11
Fix #1162 - Allow filtering recipes by action name

### LINKED ISSUE #1162: Make filtering the recipe API by action easier
Currently we can filter recipes in the API by the action id, which is very awkward, mostly because we don't expose action ids directly in most API endpoints. We should make this filter accept action names as well as ids.

mythmon:
The first commit here is quite large, and is just be moving code around because there was no structure or order to the recipe API tests.

The second commit is the real meat of things, allowing both the v1 and v2 recipe APIs to be filtered by action name, like `/api/v2/recipe/?action=preference-experiment`.


mythmon:
bors r= @rehandalal

mythmon:
Taskcluster failure is #1168.

mythmon:
bors r= @rehandalal

PR #1166, opened 2018-01-12
1074: Immutable copy/paste values

### LINKED ISSUE #1074: Button to dump recipe branch names to string array?
I'm writing a quick spark script to count clientIds overall and by branch for a pref-flip study and realized it would be a little faster if there was a button in the UI to dump these 9 branch names to an array in a textbox or directly to the clipboard.

e.g. `["branch1", "branch2", ..., "branchN"]`

andymikulski:
Fixes #1074

- Changes Immutable copy/paste values (on Recipe Details page) to appropriate JSON strings
  - e.g. `List [ Map { "ratio": 1, ...` is now `[{"ratio": 1, ...`
- Increases size of copy/paste icons and adds a tooltip with helper text.
- Tests

### DIFF
--- a/recipe-server/client/control/components/recipes/RecipeDetails.js
+++ b/recipe-server/client/control/components/recipes/RecipeDetails.js
@@ -1,5 +1,5 @@
-import { Card, Icon } from 'antd';
-import { is, Map } from 'immutable';
+import { Card, Icon, Tooltip } from 'antd';
+import { is, List, Map } from 'immutable';
 import PropTypes from 'prop-types';
 import React from 'react';
 import { connect } from 'react-redux';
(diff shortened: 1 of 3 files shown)


recipe-server/client/control/components/recipes/RecipeDetails.js
```diff
@@ -80,6 +80,15 @@ export class ArgumentsValue extends React.PureComponent {
     name: null,
   };
 
+  static stringifyImmutable(value) {
+    return JSON.stringify(value.toJS());
```
mythmon: I think Immutable has some special hooks so that you can just do `JSON.stringify(value)`. It will give the same result as calling `.toJS()` first. If that's true, I don't think we even need this method.

andymikulski: >All Immutable Collections also implement toJSON() allowing them to be passed to JSON.stringify directly.

Neat, TIL you can customize the `stringify` output if you implement your own `toJSON`.

andymikulski:
>Thanks for doing this! I'm not sure if the immutable/json change will be helpful or not...

IMO, the value should be as 'generic' as possible, and the `List/Map` inclusions seemed a little too coupled to the implementation. I'll switch the Immutable `stringify` business now and tell bors to land!

andymikulski:
bors r= @mythmon

PR #1170, opened 2018-01-23
Upgrade to eslint-plugin-mozilla 0.6.0, enable ESLint of jsm files and fix issues.

Standard8:
I'm preparing to change m-c's no-unused-vars rule for .jsm files to include global variables as well as normal ones. As a result of early tests, I picked up some issues in SheildRecipeClient.jsm - there's some unused variables that don't appear to be used anywhere.

Also included in this patch is an upgrade the plugin to 0.6.0 to get the necessary extra rules. As a result mozilla/use-services is now defined by default.

I also realised that the repository wasn't linting the .jsm files (though they're all pretty good!), so I fixed that as well.

Finally, there's a couple of comma-dangle fixes that Lint was complaining about (one was already there, not sure why automation hadn't picked it up).

### DIFF
--- a/recipe-client-addon/.eslintrc.js
+++ b/recipe-client-addon/.eslintrc.js
@@ -19,10 +19,13 @@ module.exports = {
     "indent-legacy": ["warn", 2, {SwitchCase: 1}],
     "mozilla/no-aArgs": "warn",
     "mozilla/balanced-listeners": 0,
-    "mozilla/use-services": "error",
     "no-console": "warn",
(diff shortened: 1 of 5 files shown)


mythmon:
Thanks! I appreciate the work to deal with our external existence. In the near future we're going to be moving normandy-client canonically to m-c, so you won't have to keep updating our rules.

bors r+

Standard8:
No problem. A move is nice to know about as well.

PR #1171, opened 2018-01-26
Support generalTime format when validating signatures

mythmon:
Autograph recently started using a new certificate chain that includes certificates with a new timestamp format. This adds support for the new timestamp when validating signatures on objects.


mythmon:
CC @milescrabill

PR #1172, opened 2018-01-29
#1033: Make publish button stand out

### LINKED ISSUE #1033: Make publish button stand out?
@MattGrimes had some trouble finding the publish button in the new UI until I pointed it out to him. I think there's two minor improvements we could make here:

1. After approving a revision, users have to go back to the recipe details page to publish the recipe, although publishing immediately after reviewing is a common user flow. Maybe we should redirect users back to that page after approving a review? Or maybe the review page should show a publish button if the recipe is eligible to be published?

2. On the recipe details page, the publish button blends in with the edit/clone buttons a bit too well. Maybe we should relocate it, or maybe make it green?

andymikulski:
Fixes #1033 

- Moves 'publish' button above more common functions
- Adds shiny button decoration (*ooh, ahh*)

### DIFF
--- a/recipe-server/client/control/components/recipes/DetailsActionBar.js
+++ b/recipe-server/client/control/components/recipes/DetailsActionBar.js
@@ -107,20 +107,36 @@ export default class DetailsActionBar extends React.PureComponent {
     } = this.props;
 
     return (
-      <div className="details-action-bar clearfix">
-        <Link href={`${routerPath}clone/`} id="dab-clone-link">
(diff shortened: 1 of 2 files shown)


rehandalal:
I don't know if I love this solution it makes the button area look really weird to me. I think changing the color is a better solution than moving it to a new line. And while the shine is cool I think it might be overkill especially if we're not doing that consistently throughout the UI.

andymikulski:
I don't think the control interface will exist much longer anyway, so I will just close this PR and re-open later if necessary.

Issue #1173, opened 2018-01-30
[meta] Break up the monorepo


mostlygeek:
This is following up on discussions we had about splitting this mono repo up.  

The main tasks I recall are: 

1.  recipe-client, moving to the Tree and deployed on Firefox train schedule
1.  recipe-server, stays here and this repo becomes recipe-server. Or we can make a new repo named normandy-recipe-server? 
1.  normandy-admin, deprecate / phase out in favor of all UI living in [Delivery Dashboard](https://github.com/mozilla/delivery-dashboard)
1.  mozjexl -> new repo 
1. shield icon generator (svg)
1. ??? others?

There are a lot of details on actually accomplishing any of the above.  Creating this issue to discuss at a high level.  I think The Breakup should be a high priority and done soon as possible. 

[Project](https://github.com/mozilla/normandy/projects/14)

mostlygeek:
Since this is happening in [Project 14](https://github.com/mozilla/normandy/projects/14) closing this Proposal.

PR #1175, opened 2018-02-02
Bug 1435210 - Provide a reason when stopping preference studies from actions.

mythmon:
This should explain the `reason: unknown` data for preference study unenrollment we've seen in early Telemetry data. The `stop` method called here is [this method on `PreferenceExperiments.jsm](https://github.com/mozilla/normandy/blob/b9ebeeb4867e61b59dc10331820a44c702e1d35e/recipe-client-addon/lib/PreferenceExperiments.jsm#L469).


mythmon:
TC failure is new lint rules in mozilla-central. Lets ignore that for now, and I'll move forward faster with removing the recipe-client from this repo (part of #1173).

rehandalal:
oops...
bors r+

(

mythmon:
#1178 is on its way to landing, so this PR is unblocked.

bors r= @rehandalal

PR #1181, opened 2018-02-06
Import Experiments from Experimenter into Normandy

jaredkerim:
This is a big one!  This allows Normandy admins to import experiment definitions from Experimenter into Normandy.  The work flow is: that 

1) a Normandy admin goes to Normandy
2) goes to the Import Experiments page
3) Normandy reads the list of Pending experiments from Experimenter
4) Normandy lists all the experiments with all their details
5) the Normandy admin can click either Accept or Reject
6) Accept will insert the experiment into Normandy as if the admin had entered it manually into the new Recipe form and send a message back to Experimenter to say it has been accepted
7) Reject will send a message back to Experimenter saying it was rejected along with a reason and it will not send any information to Normandy

Testing this locally involves also standing up an Experimenter instance, where can be found here:

https://github.com/mozilla/experimenter

It's fully dockerized so it should be fairly straight forward to setup locally, I can provide help with that.

Thanks for taking the time to review this!


recipe-server/client/control/state/app/experimenter/actions.js
```diff
@@ -0,0 +1,111 @@
...
+  return `${experimenterAPIUrl}/experiments/?status=Pending`;
+}
+
+export function escapeClientMatching(experiment) {
+  // TODO JEXL has no ability to add comment strings so we have to
+  // massage the incoming client matching string into something it
+  // will accept
+  const cleanedMatching = experiment.client_matching.replace(/"/gm, "'").replace(/(\r\n|\n|\r)/gm, '" &&\n"');
+  return `"TODO - CONVERT TO JEXL" &&
+"Browser: ${parseFloat(experiment.population_percent)}% of Firefox ${experiment.firefox_version} ${experiment.firefox_channel}" &&
```
rehandalal: I'm curious about what a typical string looks like and what you're cleaning out here?

mythmon:
@rehandalal @jaredkerim what's the status of this? Is it going to land in this repo, or migrate to delivery-console?

rehandalal:
@mythmon I'm handing over final review to @andymikulski since I haven't add any bandwidth.

andymikulski:
>Is it going to land in this repo, or migrate to delivery-console?

I think what's going to have to happen is this PR lands here (do those review items above need fixing?) and then I will manually move the changes over to `delivery-console`.

---

I tried to set this up locally and it seems I can't create anything but projects within Experimenter:

screen shot 2018-03-20 at 4 01 03 pm

As a result, normandy only shows this page:

screen shot 2018-03-20 at 4 02 11 pm

I'm not sure if I pulled the experimenter repo at a bad time or if I simply have something misconfigured. @jaredkerim Any idea on how I can populate even some fake data into Experimenter for me to import?

jaredkerim:
@andymikulski Thanks for taking a look at this!

A simple way to create some test data is to open a shell in the Experimenter root directory and do the following:

make shell

Then enter this snippet:

https://gist.github.com/jaredkerim/91e5abc464debed4b32e94beb221fc8c

Then refresh the Shield page and if it's configured properly to talk to your local experimenter you should see some experiments show up.  Let me know if you can't get it working and I'll help!

recipe-server/client/control/components/common/NavigationMenu.js
```diff
@@ -8,27 +8,44 @@ import { Link } from 'redux-little-router';
...
+      recipeSessionHistory,
+      router,
+    } = this.props;
     const { pathname, search } = router;
 
+    let importExperiments;
+    if (experimenterConfigured) {
+      importExperiments = (
+        <Item key="/recipe/import/">
+          <Link href="/recipe/import/">Import Experiments</Link>
```
andymikulski: This should be `Import from Experimenter` or similar.

recipe-server/client/control/components/common/NavigationMenu.js
```diff
@@ -56,6 +73,8 @@ export default class NavigationMenu extends React.PureComponent {
                 </Item>),
               )
             }
+
+            {importExperiments}
```
andymikulski: This should be moved to the top of the menu - currently it is lost in the 'most recently viewed' list.

recipe-server/client/control/routes.js
```diff
@@ -50,7 +51,12 @@ const routes = {
           slug: 'recipe-new',
         },
       },
-
+      '/import': {
+        '/': {
+          component: ImportExperimentsPage,
+          crumb: 'ImportExperiments',
```
andymikulski: This should be `Import Experiments`.

recipe-server/client/control/state/app/experimenter/actions.js
jaredkerim: We want to communicate the different types of filtering that will go here, but the Experimenter user won't actually formulate it as a JEXL statement, they'll just write it in natural language like 'Please filter for locale en-US only'.  That then needs to be communicated to the Shield/Normandy admin, but because JEXL doesn't support actual comments, the best I could seem to manage was just making it a string literal which the Shield/Normandy admin will then need to convert to a proper JEXL statement.

jaredkerim:
@andymikulski Thanks that's really good feedback, it's a lot of information and it's hard to figure out everything it's trying to say.

I moved the layout around to try to make things make more sense, tell me if this is more grokkable:

screenshot 2018-04-03 14 47 34

That's a real experiment that just launched recently, so that's how it would actually look to a reviewer with the real experiment parameters.  

I addressed the code feedback, fixed another little bug (the query object didn't seem to be firing when I navigated to the page), so things should be in a pretty good state.

When you get another chance please take another look and let me know if there's anything else I need to change.

Thanks again!

andymikulski:
My feedback's been addressed, assigning this back over to @rehandalal (based on discussion with @mostlygeek) to figure out the delivery-console part of landing this PR.

mythmon:
I'm not sure what the future of this integration is, but all UI work is moving out of this repo. 
Sorry for moving the goal posts on you, again, @jaredkerim. I expect that the future of this will be related to mozilla/delivery-console, but exactly what that relationship is like is up in the air.

PR #1182, opened 2018-02-08
correct link to gecko docs

peterbe:
(my first PR! Yay!) Not sure what the triaging and assignment and reviewers practice is.


mythmon:
For triaging, I think it will be fine to leave it blank unless you have an idea for who should be reviewing your PR. I can take care of assigning people for review.

Issue #1091, opened 2017-10-04
ExtensionSelect shows the oldest uploaded add-ons instead of the most recently uploaded ones


Osmose:
IIRC @andymikulski 's intent for the ExtensionSelect was to show a list of the most recently uploaded extensions, but currently the widget just looks for the first page of results, which by default orders by ID, and thus by creation date.

PR #1186, opened 2018-02-21
Fix #1165 - Allow ordering the extension API.

### LINKED ISSUE #1165: Add ordering to extension listing
Required to resolve #1091 

Currently, there is no sorting/ordering enabled for the Extension Listing pages and related API endpoints. This functionality is required in a couple spots, particularly the `ExtensionSelect` component.

mythmon:
This adds the ability to change the sort order of the extensions API, and adds tests for sorting the extensions API and the recipes API (which had sorting, but no tests for sorting).

### DIFF
--- a/recipe-server/normandy/recipes/tests/api/v1/test_api.py
+++ b/recipe-server/normandy/recipes/tests/api/v1/test_api.py
@@ -1,4 +1,5 @@
 import hashlib
+from datetime import datetime, timedelta
 from unittest.mock import patch
 
 from django.db import connection
@@ -571,6 +572,32 @@ class TestRecipeAPI(object):
             res = api_client.get(f'/api/v1/recipe/?action=nonexistant')
             assert res.status_code == 200
             assert res.data == []
+
+        def test_order_last_updated(self, api_client):
+            now = datetime.now()
+            yesterday = now - timedelta(days=1)
+            r1 = RecipeFactory(updated=yesterday)
+            r2 = RecipeFactory(updated=now)
+
+            res = api_client.get(f'/api/v2/recipe/?ordering=last_updated')
+            assert res.status_code == 200
+            assert [r['id'] for r in res.data['results']] == [r1.id, r2.id]
+
+            res = api_client.get(f'/api/v2/recipe/?ordering=-last_updated')
+            assert res.status_code == 200
+            assert [r['id'] for r in res.data['results']] == [r2.id, r1.id]
+
+        def test_order_name(self, api_client):
+            r1 = RecipeFactory(name="a")
+            r2 = RecipeFactory(name="b")
+
+            res = api_client.get(f'/api/v2/recipe/?ordering=name')
+            assert res.status_code == 200
+            assert [r['id'] for r in res.data['results']] == [r1.id, r2.id]
+
+            res = api_client.get(f'/api/v2/recipe/?ordering=-name')
+            assert res.status_code == 200
+            assert [r['id'] for r in res.data['results']] == [r2.id, r1.id]
 
     @pytest.mark.django_db
     class TestSigned(object):
--- a/recipe-server/normandy/recipes/tests/api/v1/test_api.py
+++ b/recipe-server/normandy/recipes/tests/api/v1/test_api.py
@@ -1,5 +1,4 @@
 import hashlib
-from datetime import datetime, timedelta
 from unittest.mock import patch
 
 from django.db import connection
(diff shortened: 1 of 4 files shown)


peterbe:
How urgent is this?
Without having opened the code, would I be of use just reviewing the code itself without fully understanding how it all works together, or even test it locally?

mythmon:
This isn't urgent. It blocks #1091, but that is a front-end issue that I think isn't likely to get fixed until Delivery Console is up and running (though that's really @rehandalal's call).

That being said, it would be nice if you were up to speed enough to work on Normandy stuff soon, and to me review is a good way to do that.

recipe-server/normandy/recipes/tests/api/v1/test_api.py
```diff
@@ -572,6 +573,32 @@ def test_list_filter_action(self, api_client):
             assert res.status_code == 200
             assert res.data == []
 
+        def test_order_last_updated(self, api_client):
+            now = datetime.now()
+            yesterday = now - timedelta(days=1)
+            r1 = RecipeFactory(updated=yesterday)
+            r2 = RecipeFactory(updated=now)
+
+            res = api_client.get(f'/api/v2/recipe/?ordering=last_updated')
```
peterbe: Why an f-string?

recipe-server/normandy/recipes/tests/api/v2/test_api.py
```diff
@@ -541,6 +542,32 @@ def test_list_filter_action(self, api_client):
...
+            r2 = RecipeFactory(name="b")
+
+            res = api_client.get(f'/api/v2/recipe/?ordering=name')
+            assert res.status_code == 200
+            assert [r['id'] for r in res.data['results']] == [r1.id, r2.id]
+
+            res = api_client.get(f'/api/v2/recipe/?ordering=-name')
+            assert res.status_code == 200
+            assert [r['id'] for r in res.data['results']] == [r2.id, r1.id]
+
```
peterbe: I think there should be a test that tests ordering by `xxxxx` or something. Will the API reject it or ignore it?

recipe-server/normandy/recipes/tests/api/v1/test_api.py
```diff
@@ -1,4 +1,5 @@
 import hashlib
+from datetime import datetime, timedelta
```
peterbe: nit: Why not just importing the module like the line above?

recipe-server/normandy/recipes/tests/api/v2/test_api.py
mythmon: The API should currently ignore it, though rejecting it would be nice.

recipe-server/normandy/recipes/tests/api/v1/test_api.py
mythmon: `datetime.datetime` is awkward, and the names `datetime` and `timedelta` are well known and unlikely to collide with other things. Also, this is the style used in the rest of the codebase.

recipe-server/normandy/recipes/tests/api/v1/test_api.py
mythmon: This was a copy and paste failure, the test that I duplicated actually needed the f-string for including a recipe id. Will fix.

recipe-server/normandy/recipes/tests/api/v1/test_api.py
peterbe: Consistency > personal preferences.

peterbe:
I still see f-strings where not needed. 
And I still use mentions of "api/v2" in the v1 test_api.py files.

peterbe:
That Test Summary thing in Circle was cool!
screen shot 2018-02-23 at 12 27 00 pm
Never seen that before. 

I read through the whole PR again and it looks good in terms of code and stuff and I think you can figure out the test failure since it's probably fixable locally.

peterbe:
What do you think about raising a 400 error if the ordering string is not valid? I think it could definitely help if some poor user has a typo. I.e. `ordering=lastupdated`.
Perhaps it's something we can tackle separately later.

mythmon:
It's a good idea. It makes the API easier to use. I don't think it is in scope for this issue though, because we are currently just using ordering primitives exposed by DRF, and to do what you propose we'd have to dig further into those. This PR just covers adding sorting to a particular endpoint.

Also, sorry for all the back and forth with CI. I swear I run tests locally most of the time... :face_with_head_bandage:

mythmon:
@peterbe I think this is ready to go now. Sorry about all the noise!

peterbe:
Let's do this!

bors r+

PR #1202, opened 2018-03-05
enable CORS for REST endpoints, fixes #954

### LINKED ISSUE #954: Enable CORS on Normandy servers
https://normandy.cdn.mozilla.net (our public, read-only address) should add CORS headers allowing anyone to use the data on other web pages.

peterbe:
This way the most primitive CORS is enabled for all things `/api/*`. You can test it like this:

```
▶ curl -v -k https://localhost:8000/api/v2/recipe/ -H "Origin: http://example.com"  2>&1 | grep Access
< Access-Control-Allow-Origin: *
```
and
```
▶ curl -v -k https://localhost:8000/api/v2/recipe/ -H "Origin: http://example.com" -X OPTIONS 2>&1 | grep Access
< Access-Control-Allow-Origin: *
< Access-Control-Allow-Headers: accept, accept-encoding, authorization, content-type, dnt, origin, user-agent, x-csrftoken, x-requested-with
< Access-Control-Allow-Methods: DELETE, GET, OPTIONS, PATCH, POST, PUT
< Access-Control-Max-Age: 86400
```

For the production instance the Origin header needs to match a specific hardcoded list. I only put the production admin UI URL in there for now. This will need to change at some point to also include the domain for the production Delivery Console. 

If we land this, I guess our Dev and Stage uses the `Production` DJANGO_CONFIGURATION so we're going to need to manually add whatever domains we use for doing prototypes. E.g. rdalal.github.io or something.


peterbe:
For the record, GitHub use `*` for their API. 
https://developer.github.com/v3/#cross-origin-resource-sharing

I'm not entirely sure why we need to mess around with a whitelist of domains but it's somehow comforting to be the gentlest possible to start.

recipe-server/normandy/settings.py
```diff
@@ -389,6 +399,13 @@ class Production(Base):
...
 
+    # Custom CORS settings that overrides the CORS class's configuration.
+    # In production we harden it down a bit extra.
+    CORS_ORIGIN_ALLOW_ALL = values.BooleanValue(False)
+    CORS_ORIGIN_WHITELIST = values.ListValue([
+        'https://normandy-admin.prod.mozaws.net'
+    ])
+
 
 class ProductionReadOnly(Production):
```
mythmon: I would prefer if the `ReadOnly` variants of the configs set `CORS_ALLOW_METHODS = ['GET', 'OPTIONS']` (ie, only non-writing, idempotent methods). This is not strictly necessary, but it further hardens the API, and it presents more accurate information in `Access-Control-Allow-Methods`.

Generally, the read only servers don't have write access to the DBs, and the APIs don't allow any write methods. This is mostly a reflection of that current policy.

recipe-server/normandy/settings.py
```diff
@@ -162,7 +164,15 @@ def CSP_DEFAULT_SRC(self):
     CSRF_COOKIE_NAME = 'csrftoken-20170707'
 
 
-class Base(Core):
+class CORS:
```
mythmon: :+1: I like having a mixin for this kind of setting. It gives a nice bit of organization.

recipe-server/normandy/settings.py
```diff
@@ -162,7 +164,15 @@ def CSP_DEFAULT_SRC(self):
...
+    """Default settings related to setting CORS headers."""
+
+    CORS_ORIGIN_ALLOW_ALL = values.BooleanValue(True)
+
```
mythmon: As a very minor nit, I don't think this blank line is needed. Feel free to leave it the way it is though.

recipe-server/normandy/settings.py
```diff
@@ -389,6 +399,13 @@ class Production(Base):
     DEFAULT_FILE_STORAGE = values.Value('normandy.base.storage.S3Boto3PermissiveStorage')
     AWS_S3_FILE_OVERWRITE = False
 
+    # Custom CORS settings that overrides the CORS class's configuration.
+    # In production we harden it down a bit extra.
+    CORS_ORIGIN_ALLOW_ALL = values.BooleanValue(False)
+    CORS_ORIGIN_WHITELIST = values.ListValue([
+        'https://normandy-admin.prod.mozaws.net'
```
mythmon: I think we should leave this list empty by default. We're going to be customizing this per-environment with environment variables anyways, and this value won't actually be needed. Same-origin requests don't use CORS and so don't need any of these headers.

I'd keep the line setting `ALLOW_ALL` to `False` though. That's good.

mythmon:
>  For the record, GitHub use * for their API.
>
> I'm not entirely sure why we need to mess around with a whitelist of domains but it's somehow comforting to be the gentlest possible to start.

I do generally agree with using GitHub as an example API. One major difference between us is that they are "self service", in that you don't have to talk to a human before you can use it. It also doesn't really have the same impact over GitHub that Normandy has over Firefox.

Since one of the major concerns of using SPAs to access Normandy opens us up to certain types of attacks, I think being extra cautious with this is the better choice. Especially because changing an environment variable to add a new site is fairly easy, and each site will need its own Auth0 credentials anyways, so there is already per-site work to be done.

> If we land this, I guess our Dev and Stage uses the Production DJANGO_CONFIGURATION so we're going to need to manually add whatever domains we use for doing prototypes. E.g. rdalal.github.io or something.

We should run stage and prod the same way. That is, I don't think we should be adding random testing sites to stage. We have normandy.dev.mozaws.net for that. Once Delivery Console has a real staging site, I think it is fine to give *that* access to normandy.stage.mozaws.net.

I'd be ok setting normandy.dev.mozaws.net to accept `*` though, since it would ease development.

mythmon:
@milescrabill can you take a quick look at these docs, and make sure they make sense?

mythmon:
> It looks like I'll need to override DJANGO_CORS_ORIGIN_WHITELIST for each environment (they're all using Production/ProductionReadOnly) which is fine.

For now, we should leave this empty on prod and stage admin servers, until we have consumers for those servers. The dev server, the prod readonly server, and the stage readonly server should have `DJANGO_CORS_ORIGIN_ALLOW_ALL` set to `True`.

recipe-server/normandy/settings.py
peterbe: I contemplated calling it `CORSMixin` but thought it was a bit overkill considering that this isn't much object oritentation, but more a classes-as-groupings.

recipe-server/normandy/settings.py
peterbe: sure thing. consistency.

peterbe:
@mythmon Hopefully all feedback addressed.

docs/ops/config.rst
```diff
@@ -309,6 +309,27 @@ in other Django projects.
...
+   needs the ``Origin`` header needs to match ``DJANGO_CORS_ORIGIN_WHITELIST``.
+   In all environments other than ``Production`` this is set to True.
+
+   .. note::
+
+      The CORS headers only apply to URLs that match the regex ``/api/*``.
+
+.. envvar:: DJANGO_CORS_ORIGIN_WHITELIST
+
+   :default: ``['https://normandy-admin.prod.mozaws.net']``
```
mythmon: This needs updated.

recipe-server/normandy/settings.py
```diff
@@ -389,6 +398,11 @@ class Production(Base):
     DEFAULT_FILE_STORAGE = values.Value('normandy.base.storage.S3Boto3PermissiveStorage')
     AWS_S3_FILE_OVERWRITE = False
 
+    # Custom CORS settings that overrides the CORS class's configuration.
+    # In production we harden it down a bit extra.
+    CORS_ORIGIN_ALLOW_ALL = values.BooleanValue(False)
+    CORS_ORIGIN_WHITELIST = values.ListValue([])
```
mythmon: We should include this value in `Base` (or rather in `Cors`), so that other environments can override it with envvars too. For example, I can't set `DJANGO_CORS_ORIGIN_WHITELIST` in local development, because the `Development` class doesn't read the envvar.

recipe-server/normandy/settings.py
```diff
@@ -405,6 +419,10 @@ class ProductionReadOnly(Production):
         'security.W017',  # Check CSRF cookie http only
     ])
 
+    # In ReadOnly mode you have no business using the API to do any writes
+    # anyway. This gives early feedback to that.
+    CORS_ALLOW_METHODS = values.ListValue(['GET', 'OPTIONS'])
```
mythmon: Since this is controllable via envvars, it should also be in `Base`, so it is available to all environments. It should also be in the documentation.

peterbe:
Thanks for all the great feedback @mythmon !

Issue #1192, opened 2018-02-26
Detect invalid order by keys.


mythmon:
In #1165, @peterbe pointed out that it would be nice if invalid order by keys would return a client error, instead of silently doing nothing. This sounds like a good idea that would make our API easier to develop against, and we should do it.

PR #1203, opened 2018-03-05
[WIP] Move enabled status to recipe revisions

rehandalal:
**DO NOT MERGE THIS!**

It's a little more complicated than I originally thought so looking for feedback to make sure I'm not going in the wrong direction.

f?

TODO: 
- [ ] Migrations
- [ ] Tests
- [ ] Update API
- [ ] Update UI

### DIFF
--- a/recipe-server/normandy/recipes/decorators.py
+++ b/recipe-server/normandy/recipes/decorators.py
@@ -1,14 +1,15 @@
-class CurrentRevisionProperty(object):
-    """
-    A class to help create class properties that return None if an instance of the parent class
-    has no `current_revision`.
-    """
-    def __init__(self, method):
+class GenericRevisionProperty(object):
+    related_field = None
+
+    def __init__(self, method, default=None):
         self.method = method
+        self.default = default
         self.__doc__ = method.__doc__
 
     def __get__(self, instance, owner):
-        if not instance.current_revision:
+        if not self.related_field:
+            raise NotImplementedError('You must set a related field.')
+        if not getattr(instance, self.related_field):
             return None
         return self.method(instance)
 
@@ -19,8 +20,35 @@ class CurrentRevisionProperty(object):
         raise AttributeError('This property cannot be deleted.')
 
 
-def current_revision_property(method):
+class CurrentRevisionProperty(GenericRevisionProperty):
     """
-    A decorator that will return None if the instance has no `current_revision`
+    A class to help create class properties that return a default value if an instance of the
+    parent class has no `current_revision`.
     """
-    return CurrentRevisionProperty(method)
+    related_field = 'current_revision'
+
+
+def current_revision_property(default=None):
+    """
+    A decorator that will return a default value if the instance has no `current_revision`
+    """
+    def decorator(method):
+        return CurrentRevisionProperty(method, default)
+    return decorator
+
+
+class ApprovedRevisionProperty(GenericRevisionProperty):
+    """
+    A class to help create class properties that return a default value if an instance of the
+    parent class has no `approved_revision`.
+    """
+    related_field = 'approved_revision'
+
+
+def approved_revision_property(default=None):
+    """
+    A decorator that will return a default value if the instance has no `approved_revision`
+    """
+    def decorator(method):
+        return ApprovedRevisionProperty(method, default)
+    return decorator
--- a/recipe-server/normandy/recipes/decorators.py
+++ b/recipe-server/normandy/recipes/decorators.py
@@ -1,14 +1,15 @@
-class GenericRevisionProperty(object):
-    related_field = None
-
-    def __init__(self, method, default=None):
+class RelatedFieldProperty(object):
+    """
+    A class to help create class properties that return a default value if an instance of the
+    parent class has an empty related field.
+    """
+    def __init__(self, method, related_field, default=None):
         self.method = method
         self.default = default
+        self.related_field = related_field
         self.__doc__ = method.__doc__
 
     def __get__(self, instance, owner):
-        if not self.related_field:
-            raise NotImplementedError('You must set a related field.')
         if not getattr(instance, self.related_field):
             return None
         return self.method(instance)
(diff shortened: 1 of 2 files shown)


recipe-server/normandy/recipes/decorators.py
```diff
@@ -1,14 +1,15 @@
-class CurrentRevisionProperty(object):
-    """
-    A class to help create class properties that return None if an instance of the parent class
-    has no `current_revision`.
-    """
-    def __init__(self, method):
+class GenericRevisionProperty(object):
+    related_field = None
```
mythmon: I think if related_field was passed as a parameter to __init__, then you could get rid of all the subclasses, and just create instances of this field in the decorators. It would be less lines of code, but I'm not sure if there is something I'm missing that it would break.

mythmon:
This isn't really what I had expected. This creates an interesting parallel history for a recipe, sort of? I don't think there is anything wrong with it, and it has the nice property of not creating extra revisions when you enabled/disable something, which has its advantages. It also nicely side-steps the problem of the interaction between enabled and approved.

(Though maybe we **should** change so that approval (and thus another user) is required for enable/disable. That's a bigger change though)

I worry a bit about the performance of this. Since `EnabledState` is a 1-to-many, it means making a lot of DB calls when getting the history of a revision, right? This doesn't trip any of our "reasonable number of DB calls" tests currently, but that's because none of our hot-path views use history revisions. It might be worth writing a test like that for the revision listing page too. Or at least test it with something like django-debug-toolbar.

I think you'll also need to trigger a re-signing when you enable a recipe, since `enabled` is [a part of the signed blob](https://normandy.cdn.mozilla.net/api/v1/recipe/signed/). Or maybe that happens automatically somewhere else.

mythmon:
I though about this some more, and I don't think it's going to work. In this setup, each revision has one value for enabled, even though it may have multiple `EnabledState`s. That means that we can't tell intra-revision changes. 

For example,. consider if a recipe is created, then the first revision is approved and enabled, and then the revision is disabled. The only revision in history will have enabled=false, even though it was enabled for a while. That won't give analysts the datas try need, since they need to know for what time range was the recipe/each revision enabled.

rehandalal:
> This isn't really what I had expected. This creates an interesting parallel history for a recipe, sort of?

Yeah it wasn't what I was originally thinking but then I started running through workflows and this seems like the best way to handle it.

> (Though maybe we should change so that approval (and thus another user) is required for enable/disable. That's a bigger change though)

This is an interesting idea and if it's something we want we should just do this now rather than adapt how enabled/disabled works again later. And that's why I wanted to post this before continuing down this road.

> I worry a bit about the performance of this.

I do too. I will need to do some testing.

> I think you'll also need to trigger a re-signing when you enable a recipe, since enabled is a part of the signed blob. Or maybe that happens automatically somewhere else.

Yeah I'll investigate this.

rehandalal:
> For example,. consider if a recipe is created, then the first revision is approved and enabled, and then the revision is disabled. The only revision in history will have enabled=false, even though it was enabled for a while. That won't give analysts the datas try need, since they need to know for what time range was the recipe/each revision enabled.

The way this is set up you would never have more than one enabled entry and more than one disabled entry per revision. Once it's been disabled it clears the `approved_revision` field meaning to re-enable you would need to create a new revision and then go through the enable dance again. I imagined in the serializer for revisions I would return the entire enable state history with the revision. So someone looking at this data should have the entire history.

That being said since it's only ever a maximum of two entries I'm now starting to think maybe I should have two one-to-one fields on the revision for `enabled_state` and `disabled_state` and that might be better for performance too.

mythmon:
I talked with @MattGrimes about this a bit, and I learned a couple things:

* Disable being permanent isn't actually desirable. There are times when someone will want to re-enable a recipe after disabling, and making a no-op change to the recipe in order to do so is not great.
* The current flow of approve and then enable, with two key approval and single user enable/disable works well.

This makes me think that having many `EnabledStates` per revision might actually be what we want. As long as we expose all the enabled states in the API, instead of only the latest, it should work. I think?

I was also concerned about having multiple histories (when a revision was created and when it was enabled/disabled being separate event streams), but thinking about it more, we already have this in the form of approval. I don't think it's an issue anymore, after some more thought.

All of that is to say, f+. I think the design of this can work, as long as we have the right data exposed in the API, and don't introduce any major performance problems. It nicely maintains our current semantics, while adding the extra data needed.

rehandalal:
Closing because it's stale but there are probably some good ideas here for the future.

rehandalal:
This should be ready for review. The UI should still work as it did and since we're not making major UI updates to the product anymore, I have cut that from the TODO list. 

r?

normandy/recipes/tests/api/v2/test_api.py
```diff
@@ -733,21 +733,27 @@ def test_it_can_enable_recipes(self, api_client):
...
 
         def test_cannot_enable_unapproved_recipes(self, api_client):
-            recipe = RecipeFactory(enabled=False)
+            recipe = RecipeFactory()
+
+            res = api_client.post('/api/v2/recipe/%s/enable/' % recipe.id)
+            assert res.status_code == 409
+            assert res.data['error'] == 'Cannot enable a recipe that is not approved.'
+
+        def test_cannot_reenable_recipes(self, api_client):
```
mythmon: To me "reenable" sounds like "enable after disabling". Would a better name for this be `test_cannot_enable_enabled_recipe`?

rehandalal: Makes sense.

rehandalal:
I haven't checked the number of queries. I'll go do that!

rehandalal:
> I think it makes sense land this as is though (assuming the database isn't crippled), and add more info the API in another PR. What do you think?

This was my thinking. As it stands we will collect enable history but not serve it up to users.

normandy/recipes/tests/api/v1/test_api.py
```diff
@@ -499,17 +505,11 @@ def test_filtering_by_enabled_fuzz(self, api_client):
                 "<%2fscript><svg%2fonload%3d'%2b%2f'%2f%2b"
             )
             res = api_client.get(url)
-            assert res.status_code == 400
-            assert res.data == {
-                'messages': [
-                    "'javascript:/*</script><svg/onload='+/'/+' "
-                    "value must be either True or False.",
```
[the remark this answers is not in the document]
mythmon: It would be nice, if it is easy. Most of the API works like this though: unrecognized fields or unexpected values in certain fields are just ignored, which isn't great. #1192 is related. Maybe some sort of larger filter schema checker is in order.

rehandalal:
> In the previous conversation, we talked about checking the number of database queries on the history page with and without this. Have you done that?

Before: 4
After: 4

normandy/recipes/tests/api/v1/test_api.py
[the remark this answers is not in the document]
rehandalal: I'm gonna leave this as is, if it's consistent with the behavior on the rest of the API. We should address this in v3.

mythmon:
:tada: :tada: 

bors r+

PR #1207, opened 2018-03-12
Wrap JEXL validation in try/catch

### LINKED ISSUE #1059: Crashing the JEXL parser can result in a 500 code in /api/v2/recipe/
If you POST an invalid JEXL string like

extra_filter_expression: '"invalid string'

You can crash the JEXL parser and 500 the response:

VisitationError at /api/v2/recipe/
SyntaxError: EOL while scanning string literal (<unknown>, line 1)

Parse tree:
<Node called "string" matching ""TODO - CONVERT TO JEXL
        Browser: 10% of Firefox 57.0 Nightly
        Additional: Locales: slovak, en_us, dansk, so_dj"">  <-- *** We were here. ***
    <RegexNode matching ""TODO - CONVERT TO JEXL
            Browser: 10% of Firefox 57.0 Nightly
            Additional: Locales: slovak, en_us, dansk, so_dj"">

Request Method: POST
Request URL: http://localhost:8443/api/v2/recipe/
Django Version: 1.10.7
Python Executable: /usr/local/bin/python3.6
Python Version: 3.6.0

This should probably be handled more gracefully.

glasserc:
Fixes #1059.

### DIFF
--- a/recipe-server/normandy/recipes/api/v2/serializers.py
+++ b/recipe-server/normandy/recipes/api/v2/serializers.py
@@ -154,7 +154,17 @@ class RecipeSerializer(serializers.ModelSerializer):
         jexl.add_transform('preferenceIsUserSet', lambda x: x)
         jexl.add_transform('preferenceExists', lambda x: x)
 
-        errors = list(jexl.validate(value))
+        try:
+            errors = list(jexl.validate(value))
+        except Exception as exc:
+            # The JEXL parser can occasionally throw exceptions when
+            # called to validate certain invalid inputs.
+            # Catch them and at least indicate the field that failed,
+            # even if we can't explain exactly what the problem was.
+            # See https://github.com/mozilla/normandy/issues/1059.
+            error = 'The JEXL parser failed to validate {}'.format(value)
+            raise serializers.ValidationError([error])
+
         if errors:
             raise serializers.ValidationError(errors)
 
--- a/recipe-server/normandy/recipes/api/v2/serializers.py
+++ b/recipe-server/normandy/recipes/api/v2/serializers.py
@@ -162,7 +162,7 @@ class RecipeSerializer(serializers.ModelSerializer):
             # Catch them and at least indicate the field that failed,
             # even if we can't explain exactly what the problem was.
             # See https://github.com/mozilla/normandy/issues/1059.
-            error = 'The JEXL parser failed to validate {}'.format(value)
+            error = f'The JEXL parser failed to validate {value}'
             raise serializers.ValidationError([error])
 
         if errors:
--- a/recipe-server/normandy/recipes/api/v1/serializers.py
+++ b/recipe-server/normandy/recipes/api/v1/serializers.py
@@ -166,7 +166,6 @@ class RecipeSerializer(serializers.ModelSerializer):
             error = f'The JEXL parser failed to validate {value}'
             raise serializers.ValidationError([error])
 
-
         if errors:
             raise serializers.ValidationError(errors)
 
--- a/recipe-server/normandy/recipes/api/v1/serializers.py
+++ b/recipe-server/normandy/recipes/api/v1/serializers.py
@@ -155,7 +155,18 @@ class RecipeSerializer(serializers.ModelSerializer):
         jexl.add_transform('preferenceIsUserSet', lambda x: x)
         jexl.add_transform('preferenceExists', lambda x: x)
 
-        errors = list(jexl.validate(value))
+        try:
+            errors = list(jexl.validate(value))
+        except Exception as exc:
+            # The JEXL parser can occasionally throw exceptions when
+            # called to validate certain invalid inputs.
+            # Catch them and at least indicate the field that failed,
+            # even if we can't explain exactly what the problem was.
+            # See https://github.com/mozilla/normandy/issues/1059.
+            error = f'The JEXL parser failed to validate {value}'
+            raise serializers.ValidationError([error])
+
+
         if errors:
             raise serializers.ValidationError(errors)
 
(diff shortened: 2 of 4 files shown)


recipe-server/normandy/recipes/api/v2/serializers.py
```diff
@@ -154,7 +154,17 @@ def validate_extra_filter_expression(self, value):
...
-        errors = list(jexl.validate(value))
+        try:
+            errors = list(jexl.validate(value))
+        except Exception as exc:
+            # The JEXL parser can occasionally throw exceptions when
+            # called to validate certain invalid inputs.
+            # Catch them and at least indicate the field that failed,
+            # even if we can't explain exactly what the problem was.
+            # See https://github.com/mozilla/normandy/issues/1059.
+            error = 'The JEXL parser failed to validate {}'.format(value)
```
mythmon: Normandy runs with Python 3.6, which means we have access to f-strings. That means you could write this as:

```python
error = f'The JEXL parser failed to validate {value}'
```

which is exactly equivalent.

Note that this is a style change. Since it is not enforced by the linter, you can write the code however you like. I just mention it because f-strings are a new Python feature that not many projects get to use.

glasserc:
Writing a test for this feature is a little challenging because there is no specific "problematic string" that will reliably cause `validate()` to throw an exception (and indeed, any that we find should be reported as bugs to upstream, along the lines of https://github.com/Osmose/pyjexl/pull/12). I've instead written a test that mocks the JEXL parser to throw an exception, which seems like the cleanest way to test the behavior of our serializer in the case that the JEXL parser throws an exception.

mythmon:
Looks like there was a test failure in the v2 api:

```
        with mock.patch('normandy.recipes.api.v2.serializers.JEXL', jexl_mock):
            assert not serializer.is_valid()
>           assert serializer.errors['extra_filter_expression'] == [
                'The JEXL parser failed to validate aces'
            ]
E           KeyError: 'extra_filter_expression'
```

Also a lint error:

```
normandy/recipes/api/v1/serializers.py:170:9: E303 too many blank lines (2)
```

glasserc:
Oops, that was sloppy on my part, sorry.

mythmon:
This is why we have CI. :smiley: It isn't fair to ask humans to always be correct, so we use robots to make sure. Thanks for fixes!

bors r+

recipe-server/normandy/recipes/api/v2/serializers.py
```diff
@@ -154,7 +154,17 @@ def validate_extra_filter_expression(self, value):
...
-        errors = list(jexl.validate(value))
+        try:
+            errors = list(jexl.validate(value))
+        except Exception as exc:
+            # The JEXL parser can occasionally throw exceptions when
+            # called to validate certain invalid inputs.
+            # Catch them and at least indicate the field that failed,
+            # even if we can't explain exactly what the problem was.
+            # See https://github.com/mozilla/normandy/issues/1059.
+            error = f'The JEXL parser failed to validate {value}'
```
peterbe: I know this is merged but my python-code-smell-alarm buzzed when I saw this. 
I guess there isn't a way to re-wrap standard exception classes into specific ones inside `jexl`. 

Also the variable `exc` doesn't get used. It might be worth calling it's str method as part of the `error` string. E.g.
```python
error = f'The JEXL parser failed to validate {value} ({exc})'
```

mythmon: That's actually a good point, @peterbe. I probably should have spotted `except Exception`, and our linter should have caught the unused `exc`.

Including the exception in the output seems a bit problematic, as it may expose server state. Ultimately, I think patching upstream, like @glasserc did, is the correct response.

Maybe we should narrow the scope of this to only catch the specific error we saw in #1059? So it would be `except jexl.VisitationError`. That could allow other errors to slip out and cause server 500s, but it is more hygienic. Thoughts @peterbe, @glasserc?

glasserc: The reason the code smells bad is because we are trying to work around a bug rather than actually fixing it. And not even just one specific bug, but a hypothetical future bug that we expect the `pyjexl` library to manifest, which we can't point to but only imagine may come about based on our experience using this library. That bug, just to be explicit, is that the `validate()` function may throw an exception rather than generate a list of errors in the input. This would be a violation of the contract that the `validate()` function has. Because it would be a violation of the contract, and again because we have no idea what such a violation would look like, it's impossible to write clean code that addresses it. We have no reason to suspect that this future bug would even raise the same `jexl.VisitationError` instead of e.g. `KeyError` or `AttributeError`. A bug in a library could raise essentially any error. As indicated in PEP8, the way to catch "any exception that signifies an error" is to catch `Exception`.

If the bad-smelling code bothers you, I'd recommend reverting the PR. We could clean it up a tiny bit (removing the unused variable), but again, it can't really address the problem, just paper over it. Instead, a better use of time might be to try to improve the robustness of the `jexl` library so that it doesn't ever have this kind of bug. For example, I have opened https://github.com/Osmose/pyjexl/pull/13 as a first step in this direction.

peterbe: Is it so bad to have 500s? We have Sentry and stuff. 
But I'm unsure what the context of all of this is. 
Personally I like @glasserc where you instead take this as an indicator that it's time to go deep down inside the `.validate` method instead. If you pass in invalid data to a function called `validate` it should probably not crash and return false. 

Perhaps, as a first step is to focus on the inner problem that it can raise errors. 
Then secondly, write down specific pyjexl'esque exceptions that *might* happen and shouldn't wake people up at night.

PR #1219, opened 2018-04-10
Allow for client side actions fixes 1193

peterbe:
So many questions, but it feels like a good start...

See:
screen shot 2018-04-10 at 2 40 23 pm

I never figured out how to POST a new recipe with `curl` so I wrote a hack using python `requests`. It looks like this:

```python
r = requests.post(
    'https://localhost:8000/api/v2/recipe/',
    json={
        'action_id': 8,
        'name': 'Peter Loves Requests',
        'extra_filter_expression': 'true',
        'arguments': {
            'message': 'Some Message Here'
        }
    },
    headers={
        'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.13; rv:61.0) Gecko/20100101 Firefox/61.0',
        # 'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8',
         'Accept': 'application/json',
        'Cookie': 'csrftoken-20170707=CgPuYkTYmevLzeCx6xOzSoZKjygzwH5OLND9kgK8uHsCir7VqwWRw7jg5cW988Xb; sessionid=yfe2eyn6q7fc4vorxy3p5plnats95kzu; csrftoken=GKq8mHx2woQg0S2WJy060c4OGlwpcRCiwtuzEXyp9iheCv9RfIoGrdyp57UkudMA; tabstyle=raw-tab',
        'Referer': 'https://localhost:8000/api/v2/recipe/',
        'X-CSRFToken': 'CgPuYkTYmevLzeCx6xOzSoZKjygzwH5OLND9kgK8uHsCir7VqwWRw7jg5cW988Xb',
    },
    verify=False,
    auth=HTTPBasicAuth('peterbe', 'test123')
)
print(r.status_code)
# r.raise_for_status()

print(json.dumps(r.json()))
```
That worked. 

- [x] Add [`mozilla-normandy-action-argument-schemas`](https://www.npmjs.com/package/mozilla-normandy-action-argument-schemas) to `package.json`
- [x] Ability to save an `Action` model **without** an implementation
- [x] Ability to run `./manage.py update_actions` and read from [`mozilla-normandy-action-argument-schemas`](https://www.npmjs.com/package/mozilla-normandy-action-argument-schemas) to pick up schema
- [x] Ability to create a RecipeRevision from the API (v1 and v2?) with `curl -XPOST`
- [ ] Document about "remote actions" in http://normandy.readthedocs.io/en/latest/user/actions/index.html


recipe-server/normandy/recipes/api/v2/serializers.py
```diff
@@ -18,6 +18,10 @@
 
 class ActionSerializer(serializers.ModelSerializer):
     arguments_schema = serializers.JSONField()
+    # NOTE(peterbe): We *could* make the serializer NOT include
+    # this field if the action has no 'implementation'.
+    # Arguably it's nice when APIs are consistent with the keys.
+    # If the action does not have an implementation, this value will be None.
```
peterbe: I played for a bit with the idea of NOT including this field if the Action doesn't have an implementation. Just got deeper in the rabbit hole with little success. Then, upon reflection I feel more strongly about it always being included and left as `None` if there is no implementation. 
Worth considering deleting this comment.

recipe-server/normandy/recipes/management/commands/update_actions.py
```diff
@@ -23,7 +23,8 @@ def add_arguments(self, parser):
 
     @transaction.atomic
     def handle(self, *args, **options):
-        action_names = settings.ACTIONS.keys()
+        action_names = list(settings.ACTIONS.keys())
+        action_names += list(settings.REMOTE_ACTIONS.keys())
         if options['action_name']:
             action_names = [name for name in action_names if name in options['action_name']]
```
peterbe: There is [a test that says it's OK to not match any action](https://github.com/peterbe/normandy/blob/7a289c7103b4b8df694a2032e42626f8fe751d9d/recipe-server/normandy/recipes/tests/test_commands.py#L140). I don't like this but didn't want to kick up a fuzz right now. 
If someone types `./manage.py update_actions preff-rolloutt` by accident it doesn't even complain. I think it should. Perhaps not very important.

recipe-server/normandy/recipes/tests/__init__.py
```diff
@@ -49,7 +49,8 @@ class Meta:
 
     @factory.lazy_attribute
     def implementation_hash(action):
-        return sri_hash(action.implementation.encode(), url_safe=True)
+        if action.implementation is not None:
+            return sri_hash(action.implementation.encode(), url_safe=True)
```
peterbe: Perhaps you can take a look at this. Perhaps I don't understand how this `@factory.lazy_attribute` works. Arguably, the method shouldn't even be called/triggered for the actions without an implementation.

recipe-server/normandy/settings.py
```diff
@@ -153,6 +153,20 @@ def CSP_DEFAULT_SRC(self):
         'opt-out-study': os.path.join(BASE_DIR, 'client/actions/opt-out-study'),
     }
 
+    # Mapping of named actions that don't exist in the repo but exists
+    # in the installed mozilla-normandy-action-argument-schemas node package.
+    REMOTE_ACTIONS = {
+        # Name we use: Name inside mozilla-normandy-action-argument-schemas
+        'console-log2': 'consoleLog',
+    }
```
peterbe: This is the most important change. Feedback welcome. 
First of all, I picked the name "remote action" to mean the actions that don't have an implementation locally. 

The name `console-log2` obviously doesn't exist. I made that up. But I needed to call it something. 

So, `console-log` is specifically problematic because `console-log` is an exact name that exists in **two** places at the moment. 

* Perhaps we should call this one `console-log` and make it override the "local action" called `console-log`. 
* Perhaps we rename `console-log2` to `console-log` the day we do `git rm -r recipe-server/client/actions/console-log/` and make `console-log`, from that day onwards, a remote one. 
* What do you think about the mapping? Somewhere, we need to write down the name of the action as it appears in `mozilla-normandy-action-argument-schemas`.

recipe-server/normandy/recipes/management/commands/update_actions.py
```diff
@@ -57,14 +58,26 @@ def handle(self, *args, **options):
...
     with open(implementation_path) as f:
         return f.read()
 
 
 def get_arguments_schema(action_name):
-    action_directory = settings.ACTIONS[action_name]
-    with open(os.path.join(action_directory, 'package.json')) as f:
-        action_metadata = json.load(f)
-        return action_metadata['normandy']['argumentsSchema']
+    if action_name in settings.REMOTE_ACTIONS:
```
peterbe: Note the order. If an action (e.g. `console-log`) appears in *both* `settings.ACTIONS` and `settings.REMOTE_ACTIONS`, then the first (and only) place it will look for the argument schema is via `settings.REMOTE_ACTIONS`. Makes sense. It's the future.

peterbe:
As a general comment. We ought to take a look at the duplication of API v1 and v2. I gather that old Firefoxes might use the old v1 API, but things like the Normandy Admin UI (soon aka. Delivery Console) should only use v2. Right?
So we could weed out any v1 API stuff that is only relevant for admin UI and kill it. What do you think?

recipe-server/normandy/settings.py
rehandalal: I'm curious about why you chose to implement it this way?

Wouldn't it be better to just add a new field to the `Action` model? Maybe something along the lines of:
```python
remote_mapping = models.CharField(max_length=32, null=True)
```

In this case `remote_mapping` would be set to `'consoleLog'`. And if an action does not have a `remote_mapping` we know it is local.

peterbe: Before the Action can be created, we need to know what it's called in mozilla-central. Technically, what it's called in mozilla-normandy-action-argument-schemas. 

One grander possibility is to NOT have `settings.REMOTE_ACTIONS` at all. Instead, we can do something like this:
```python
with open('mozilla-normandy-action-argument-schemas/schemas.json') as f:
  schemas = json.load(f)
  for key, schema in schemas:  # e.g. prefRollout, consoleLog, etc. 
      try:
          action = Action.objects.get(name=key)
          action.implementation = None
          action.schema = json.dumps(schemas[key])
          action.save()
      except Action.DoesNotExist:
          Action.objects.create(
              name=key,
              schema=json.dumps(schemas[key]),
              implementation=None,  # default anyway
          )
```

Then, to add new actions, you `yarn upgrade mozilla-normandy-action-argument-schemas --latest && ./manage.py update_actions`

This way, the actions are not listed in advance. If we add another action to mozilla-central (and not forget to add its JSON schema to mozilla-normandy-action-argument-schemas`) we won't have to mess with `recipe-server/normandy/settings.py`. The *name* in `Action.name` will be the names as they appear in the JS client code. E.g. `consoleLog` or `preferenceRollout`.

rehandalal: > One grander possibility is to NOT have settings.REMOTE_ACTIONS at all.

That was sort of what I was getting at. It seems like it would be pretty annoying to have to maintain this list manually. I think it would be better to just use `update_actions` to generate the list of actions as we do now. And the more I think about it the less we would seem to need a mapping.

peterbe:
Perhaps premature or off-topic but instead of relying on an npm package for the schemas, could we not draw them directly from mozilla-central via mercurial (possibly by "curling" from hg.mozilla.org). At the moment, every time someone changes the schema in mozilla-central, someone has to remember how to publish a new `mozilla-normandy-action-argument-schemas` out of mozilla-central to npmjs.com. 

This is a continuation-thought about the idea that @rehandalal brought up about NOT hardcoding the local actions in Firefox (aka. what I called "remote actions").

recipe-server/normandy/settings.py
mythmon: I like the grand idea of not maintaining a mapping for actions in `settings.py`. I'd go so far as to remove the other mapping too, if we could get away with it.

There is a problem here though. The names "remote actions" and "local actions" are going to get confused really quick. In the client, and in most of the things I've been writing, "remote actions" means an action with an implementation on the server, and "local action" means an action with an implementation in the client. Re-using these names as-is would be confusing, but swapping them on the server is also confusing. I think that we need to make a distinction on the server, it should be between actions with and without implementations. I don't think we should use the words "local" or "remote" at all in the server-side code.

The problem of an action existing both in `mozilla-normandy-action-argument-schemas` and in `mozilla/normandy` is going to come up over and over. Every action we migrate (i.e. all of them) will be in the same situation as `console-log` eventually, and the pattern of `console-log2` isn't going to fly. I've touched on this elsewhere, but one idea that comes to mind is to take the schema from the npm package, but keep any implementation we have locally. This also complicates things though, if we change the schema before fully deprecating the server-side implementation.

I think a good way forward would be to use the new schemas, and then whenever we make new schemas, be careful to make them backwards compatible. Perhaps when updating schemas we should print a warning if it causes any pre-existing (enabled?) recipes to have invalid arguments?

mythmon:
> Perhaps premature or off-topic but instead of relying on an npm package for the schemas, could we not draw them directly from mozilla-central via mercurial (possibly by "curling" from hg.mozilla.org). At the moment, every time someone changes the schema in mozilla-central, someone has to remember how to publish a new mozilla-normandy-action-argument-schemas out of mozilla-central to npmjs.com.

I'd like to not do this, at least for now. `npm` is a nice stable API, it we have enough checks and balances to make sure that everything stays up to date, I think. Curling from hg.mozilla.org is about half a step above just screen scraping, and I don't trust that to be stable. It's also pretty antithetical to Normandy's general approach of [pinning all the things](https://memegen.link/xy/pin/all_the_things.jpg).

peterbe:
> Curling from hg.mozilla.org is about half a step above just screen scraping

Yeah, I know. It's not neat. Either way, if we HTTP scrape or do `hg clone hg.mozilla.org/...` we have to somehow consider pinning. Instead of by semver we'd have to use a mercurial sha. And that's a pain too. Suppose we did find a good solution for the scraping, we'd still be just with JS code. The `schemas.json` only exists thanks to npm. 

Neither relying on `mozilla-normandy-action-argument-schemas` or `hg.mozilla.org` makes me hot and bothered, but I think we should keep it mulling in our heads.

recipe-server/normandy/recipes/api/fields.py
```diff
@@ -33,6 +33,8 @@ def __init__(self, view_name='recipes:action-implementation', **kwargs):
         super().__init__(view_name=view_name, **kwargs)
 
     def get_url(self, obj, view_name, request, format):
+        if not obj.implementation:
+            return
```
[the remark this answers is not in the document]
peterbe: `return` is the same as `return None`. Perhaps not universally known fact.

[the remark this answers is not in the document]
mythmon: I know that `return` and `return None` have the same effect. Luckily Python isn't like Javascript with separate `null` and `undefined`. They certainly read differently to me though. To me `return` is "early exit from a function that doesn't return anything" and `return None` is "I have nothing to return, when normally there could be something".

This is probably my leaning towards heavily typed languages. In Rust a plain `return` is only valid in a "void" function, whereas `return None` would be a function that return `Option<T>` (ie, explicitly either a value or nothing). The opposite of `return None` in Rust is `return Some(foo)`.

This is not a blocker, mostly a readability thing. It is really minor so I'm fine leaving this the way it is now.

peterbe:
Note-to-self (from Vidyo chat with @mythmon)

1. Merge actions! If there is already an action called "foo" that has an implementation, and we encounter an action from mozilla-normandy-action-argument-schemas  called "foo", keep the implementation, but extract the schema from the "foo" in mozilla-normandy-action-argument-schemas 
2. Get rid of settings.ACTIONS and get rid of settings.REMOTE_ACTIONS
3. All the old actions can be gotten from listing the dirs in `normandy/recipe-server/client/actions/` (exclude `tests`)
4. For the exception of `consoleLog`. When mapping that one to its Action.name in Django, rewrite it to `console-log`. Document the legacy explanation. 
5. Consider printing a warning when the schema changes during `./manage.py update_actions` to help us keep in check that we don't break old clients who rely on the server-side implementation. Only a warning print because adding new optional fields shouldn't break things.

recipe-server/normandy/recipes/api/fields.py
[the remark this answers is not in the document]
rehandalal: +1 I had the same reaction but figured I'd see if anyone else brought it up

recipe-server/normandy/recipes/tests/api/v1/test_api.py
```diff
@@ -224,6 +241,22 @@ def test_it_can_create_recipes(self, api_client):
...
+            res = api_client.post('/api/v1/recipe/', {
+                'name': 'Test Recipe',
+                'action': action.name,
+                'arguments': {},
+                'extra_filter_expression': 'whatever',
+                'enabled': True
+            })
+            assert res.status_code == 201
+
+            recipe, = Recipe.objects.all()
```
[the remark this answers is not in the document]
peterbe: What's what `firstitem, = somelist` does too. If `len(Recipe.objects.all())` is  anything but exactly 1 you get an IndexError.

recipe-server/normandy/recipes/api/fields.py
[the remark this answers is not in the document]
peterbe: If two people had the same reaction, I'll concede :)

recipe-server/normandy/recipes/management/commands/update_actions.py
peterbe: It felt easy to do since I was already in there messing around. Now you get:
```
▶ ./manage.py update_actions neverheardof
Raven is not configured (logging is disabled). Please see the documentation for more information.
INFO 2018-04-17 15:14:10,767 raven.contrib.django.client.DjangoClient Raven is not configured (logging is disabled). Please see the documentation for more information.
CommandError: No actions matching ['neverheardof']
```

peterbe:
It's hopefully updated accordingly now. Some of the review nits belonged to code I completely rewrote. 

When I run `./manage.py update_actions` the the `console-log` action will have its old `implementation` (and `implementation_hash`) but its `.arguments_schema` comes from `node_modules/mozilla-normandy-action-argument-schemas/schemas.json`. I know this because I edited that file manually and made sure it got picked up. When it noticed the difference in schema it looked like this:
```
▶ ./manage.py update_actions
Updating action show-heartbeat...Done
Updating action opt-out-study...Done
Updating action preference-experiment...Done
Updating action console-log...Action $console-log is changing implementation AND argument schema. You might want to manually check that the new argument schema is compatible with the old implementation.
Done
```

mythmon:
Uh, I hope that is a temporary failure...

bors retry

PR #1227, opened 2018-04-16
Ditch hashin

rehandalal:
r?


recipe-server/requirements/default.txt
```diff
@@ -72,9 +72,9 @@ gevent==1.2.1 \
 gunicorn==19.4.5 \
     --hash=sha256:c57f1b005a4b90933303c8deed9bedeb509331aa6a0a990023a5796e52bd8988 \
     --hash=sha256:53b58044764ad79d732af18c580b1a54b724adf4d290ec19c4ca78ab22a1ee0d
-hashin==0.7.2 \
-    --hash=sha256:28213aa9cc1cbce1d33f813877e079f1c963a95c9f76c501faacb1a80723e114 \
-    --hash=sha256:dd38d9356e2810ccae928de8fa1aa023d885566a7f3c4b6840ba55ecd7d7bd35
+hashin==0.13.0 \
```
peterbe: I would recommend removing this entirely. Install it locally or with pipsi instead.

rehandalal:
> I think you should remove hashin entirely from the requirements. And all its dependencies.

Yeah I never really understood why it was in here. @mythmon are we cool to ditch from requirements?

> By the way, is pyparsing a dependency of packaging??

I believe so.

mythmon:
I'm fine with ditching it. For my personal workflow, it is nice to have it in the repo, but we don't keep it up to date, and tools like pipsi makes things a lot easier now anyways.

rehandalal:
bors r=@peterbe

PR #1230, opened 2018-04-17
upgrade to psycopg2==2.7.4

peterbe:
It's about time to upgrade. Also, since my homebrew now gives me postgres 10.3 installed locally, I can't even install `psycopg2==2.6.1` in my mac. This upgrade fixes that too.


mythmon:
[From pypi](https://pypi.org/project/psycopg2-binary/):

> The [psycopg2-binary] package is a practical choice for development and testing but in production it is advised to use the package built from sources.

I don't  think we should switch to psycopg2-binary, even though it would be nice.

peterbe:
Good catch. I guess we just need to upgrade `psycopg2` itself.

PR #1232, opened 2018-04-17
Avoid bare exceptions fixes #1231

### LINKED ISSUE #1231: Avoid bare exceptions
I.e. [here](https://github.com/mozilla/normandy/blob/2d87b95c5185ae0cc785cb846526934833dc2e4c/recipe-server/normandy/recipes/api/v1/serializers.py#L178) and [here](https://github.com/mozilla/normandy/blob/2d87b95c5185ae0cc785cb846526934833dc2e4c/recipe-server/normandy/recipes/api/v2/serializers.py#L177)

peterbe:
This got messier than I had intended :)

The validation of the `arguments` shouldn't really happen until the action validation has succeeded. Ideally, the validation of the `arguments` should start after the `action` (or `action_id` for v2) has succeeded. Otherwise it's validating the wrong thing. But if I raise a `ValidationError` inside a `validate()` method, the error message isn't "namespaced" on the `arguments`. 

In this patch, the validation of `arguments` only kicks in if you have a valid `Action` instance found.

### DIFF
--- a/recipe-server/normandy/recipes/tests/api/v2/test_api.py
+++ b/recipe-server/normandy/recipes/tests/api/v2/test_api.py
@@ -195,10 +195,36 @@ class TestRecipeAPI(object):
             assert recipes.count() == 1
 
         def test_creation_when_action_does_not_exist(self, api_client):
-            res = api_client.post('/api/v2/recipe/', {'name': 'Test Recipe',
-                                                      'action_id': 1234,
-                                                      'arguments': '{}'})
+            res = api_client.post('/api/v2/recipe/', {
+                'name': 'Test Recipe',
+                'action_id': 1234,
+                'arguments': '{}',
+            })
             assert res.status_code == 400
+            assert res.json()['action_id']
+
+            recipes = Recipe.objects.all()
+            assert recipes.count() == 0
+
+        def test_creation_when_action_id_is_missing(self, api_client):
+            res = api_client.post('/api/v2/recipe/', {
+                'name': 'Test Recipe',
+                'arguments': '{}'
+            })
+            assert res.status_code == 400
+            assert res.json()['action_id']
+
+            recipes = Recipe.objects.all()
+            assert recipes.count() == 0
+
+        def test_creation_when_action_id_is_invalid(self, api_client):
+            res = api_client.post('/api/v2/recipe/', {
+                'name': 'Test Recipe',
+                'action_id': 'a string',
+                'arguments': '{}',
+            })
+            assert res.status_code == 400
+            assert res.json()['action_id']
 
             recipes = Recipe.objects.all()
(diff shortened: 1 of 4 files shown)


recipe-server/normandy/recipes/tests/api/v2/test_api.py
```diff
@@ -195,10 +195,36 @@ def test_it_can_create_disabled_recipes(self, api_client):
...
-            res = api_client.post('/api/v2/recipe/', {'name': 'Test Recipe',
-                                                      'action_id': 1234,
-                                                      'arguments': '{}'})
+            res = api_client.post('/api/v2/recipe/', {
+                'name': 'Test Recipe',
+                'action_id': 1234,
+                'arguments': '{}',
+            })
+            assert res.status_code == 400
+            assert res.json()['action_id']
```
peterbe: I didn't want to make an assertion of the **value** because it's not our code really. It's Django REST Framework's wording.

mythmon: DRF exposes these as constants, so the comparison could be something like:

```python
assert res.json()['action_id'] == serializers.SlugRelatedField.default_error_message['does_not_exist'].format(pk_value=1234)
```

This is kind of wordy though. I think it is worth making sure that we got the right error out of this *somehow* though. This could be failing because `action_id` is an invalid key for all we know.

recipe-server/normandy/recipes/tests/api/v2/test_serializers.py
```diff
@@ -64,17 +64,27 @@ def test_it_works(self, rf):
             'identicon_seed': Whatever.startswith('v1:'),
         }
 
+    def test_validation_with_invalid_action(self):
```
mythmon: How is this test different than `test_validation_with_wrong_action`, below?

peterbe: this time the `action_id` is a string. If you let that through to `Action.objects.get(pk='some string')` you get a ORM `ValueError`.

mythmon: It's a string in both cases, just in the other test it is the string '9999'.

peterbe: The `'9999'` gets converted to an int. Check my followup commit that really checks the error messages using `default_error_messages`. One sec.

peterbe:
There. Now checking the error messages properly.

recipe-server/normandy/recipes/tests/api/v2/test_serializers.py
mythmon: Weird. Thanks for talking through this with me.

PR #1240, opened 2018-04-24
Argument validation should happen at the object level

rehandalal:
Phew! This was a little trickier than expected with having to deal with `PATCH` requests.

r?

### DIFF
--- a/recipe-server/normandy/recipes/api/v1/serializers.py
+++ b/recipe-server/normandy/recipes/api/v1/serializers.py
@@ -219,7 +219,6 @@ class RecipeSerializer(serializers.ModelSerializer):
                             currentLevel[path] = error.message
 
             if errorResponse:
-                print(errorResponse)
                 raise serializers.ValidationError({'arguments': errorResponse})
 
         return data
--- a/recipe-server/normandy/recipes/api/v1/serializers.py
+++ b/recipe-server/normandy/recipes/api/v1/serializers.py
@@ -161,58 +161,68 @@ class RecipeSerializer(serializers.ModelSerializer):
 
         return value
 
-    def validate_arguments(self, value):
-        # This in an invariance validation. It depends on the action_id
-        # being valid.
-        action_id = self.initial_data.get('action')
-        if not action_id:
-            # If this is the case, it will be caught by the basic
-            # validation.
-            return
+    def validate(self, data):
+        action = data.get('action')
 
-        # Ensure the value is a dict
-        if not isinstance(value, dict):
-            raise serializers.ValidationError('Must be an object.')
+        if action is None:
+            if not self.instance:
+                raise serializers.ValidationError({
+                    'action': serializers.PrimaryKeyRelatedField.default_error_messages['required']
+                })
+            action = self.instance.action
 
-        # Get the schema associated with the selected action
-        try:
-            schema = Action.objects.get(name=action_id).arguments_schema
-        except Action.DoesNotExist:
-            raise serializers.ValidationError('Could not find arguments schema.')
+        arguments = data.get('arguments')
 
-        schemaValidator = JSONSchemaValidator(schema)
-        errorResponse = {}
-        errors = sorted(schemaValidator.iter_errors(value), key=lambda e: e.path)
+        if arguments is None and not self.instance:
+            raise serializers.ValidationError({
+                'arguments': serializers.PrimaryKeyRelatedField.default_error_messages['required']
+            })
 
-        # Loop through ValidationErrors returned by JSONSchema
-        # Each error contains a message and a path attribute
-        # message: string human-readable error explanation
-        # path: list containing path to offending element
-        for error in errors:
-            currentLevel = errorResponse
+        if arguments is not None:
+            # Ensure the value is a dict
+            if not isinstance(arguments, dict):
+                raise serializers.ValidationError({'arguments': 'Must be an object.'})
 
-            # Loop through the path of the current error
-            # e.g. ['surveys'][0]['weight']
-            for index, path in enumerate(error.path):
-                # If this key already exists in our error response, step into it
-                if path in currentLevel:
-                    currentLevel = currentLevel[path]
-                    continue
-                else:
-                    # If we haven't reached the end of the path, add this path
-                    # as a key in our error response object and step into it
-                    if index < len(error.path) - 1:
-                        currentLevel[path] = {}
+            # Get the schema associated with the selected action
+            try:
+                schema = Action.objects.get(pk=action.id).arguments_schema
+            except Action.DoesNotExist:
+                raise serializers.ValidationError({'arguments': 'Could not find arguments schema.'})
+
+            schemaValidator = JSONSchemaValidator(schema)
+            errorResponse = {}
+            errors = sorted(schemaValidator.iter_errors(arguments), key=lambda e: e.path)
+
+            # Loop through ValidationErrors returned by JSONSchema
+            # Each error contains a message and a path attribute
+            # message: string human-readable error explanation
+            # path: list containing path to offending element
+            for error in errors:
+                currentLevel = errorResponse
+
+                # Loop through the path of the current error
+                # e.g. ['surveys'][0]['weight']
+                for index, path in enumerate(error.path):
+                    # If this key already exists in our error response, step into it
+                    if path in currentLevel:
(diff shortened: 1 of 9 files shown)


recipe-server/normandy/recipes/api/v1/serializers.py
```diff
@@ -161,58 +161,68 @@ def validate_extra_filter_expression(self, value):
...
+                        if index < len(error.path) - 1:
+                            currentLevel[path] = {}
+                            currentLevel = currentLevel[path]
+                            continue
+                        # If we've reached the final path, set the error message
+                        else:
+                            currentLevel[path] = error.message
+
+            if errorResponse:
+                print(errorResponse)
```
mythmon: Oops?

recipe-server/normandy/recipes/tests/api/v2/test_api.py
```diff
@@ -285,6 +285,29 @@ def test_creation_when_arguments_is_a_string(self, api_client):
...
+                    'type': 'object',
+                    'properties': {'message': {'type': 'string'}},
+                    'required': ['message']
+                }
+            )
+            data = {
+                'name': 'Test Recipe',
+                'enabled': True,
+                'extra_filter_expression': 'true',
+                'action_id': '{}'.format(action.id),
```
mythmon: Would it be simpler to do one of either `str(action.id)`, or `f'{action.id}`?

mythmon:
CI has problems with two tests: 

```
    def test_validation_with_wrong_arguments(self):

>       assert serializer.errors['arguments'] == {
            'surveyId': 'This field may not be blank.',
            'surveys': {
                0: {'title': 'This field may not be blank.'},
                2: {'weight': '0 is less than the minimum of 1'},
                3: {'weight': '\'lorem ipsum\' is not of type \'integer\''}
            }
        }
E       KeyError: 'arguments'
```

```
self = <normandy.recipes.tests.api.v1.test_serializers.TestRecipeSerializer object at 0x7f7076348160>

    def test_validation_with_wrong_action(self):
        serializer = RecipeSerializer(data={
            'action': 'action-that-doesnt-exist', 'arguments': {}
        })
    
        with pytest.raises(serializers.ValidationError):
    
>       assert serializer.errors['arguments'] == ['Could not find arguments schema.']
E       KeyError: 'arguments'

normandy/recipes/tests/api/v1/test_serializers.py:77: KeyError
```

normandy/recipes/tests/api/v1/test_serializers.py:103: KeyError

rehandalal:
Test failure is:
```
self = <normandy.recipes.tests.test_signing.TestVerifyX5u object at 0x7f839068ef28>
```

This was fixed by #1241

peterbe:
Uh?
screen shot 2018-04-25 at 9 28 19 am

peterbe:
By the way, I tested it with `requests.post` and this...:
```python
r = requests.post(
    'https://localhost:8000/api/v2/recipe/',
    json={
        'action_id': '9',
        'name': 'Peter Prefers Rollouts III',
        'extra_filter_expression': 'true',
        'arguments': {
#            'slug': 'myslug',
#            'preferences': [
 #             {'preferenceName': 'danger', 'value': True},
 #           ]
        }
    },
    headers={
        'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.13; rv:61.0) Gecko/20100101 Firefox/61.0',
         'Accept': 'application/json',
        'Cookie': 'csrftoken-20170707=oAwciGrk0VyFA5l4aXSrSposI6c3sYfCmA40rlHEzQhKBpKrl64A9bc9NpPlV0ry; csrftoken=GKq8mHx2woQg0S2WJy060c4OGlwpcRCiwtuzEXyp9iheCv9RfIoGrdyp57UkudMA; tabstyle=raw-tab; sessionid=8yt4$
        'Referer': 'https://localhost:8000/api/v2/recipe/',
        #'X-CSRFToken': 'CgPuYkTYmevLzeCx6xOzSoZKjygzwH5OLND9kgK8uHsCir7VqwWRw7jg5cW988Xb',
        'X-CSRFToken': 'oAwciGrk0VyFA5l4aXSrSposI6c3sYfCmA40rlHEzQhKBpKrl64A9bc9NpPlV0ry',
    },
    verify=False,
    auth=HTTPBasicAuth('peterbe', secret)
)
```
...worked. Note the `'action_id': '9'`. It yields:
`{"arguments": {"preferences": "This field may not be blank.", "slug": "This field may not be blank."}}`

peterbe:
`RecipeForm.cleanRecipeData` from within `CreateRecipePage.formAction` is `undefined`. 

I've actually not pondered it before but your [syntax is right](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static#Calling_static_methods) but I guess it doesn't work in the world of React and decorator and async/await. 
The `RecipeForm` from `import RecipeForm from 'control/components/recipes/RecipeForm';` is not a class. It's a function.

rehandalal:
Yeah I had it working but then i decided to move the `cleanRecipeData` function to a static method of `RecipeForm` and didn't test that. I forgot the class had decorators and was wrapped so accessing static methods becomes complicated. Reverted back to exporting `cleanRecipeData`. Should be working again.

recipe-server/normandy/recipes/api/v2/serializers.py
```diff
@@ -160,55 +160,64 @@ def validate_extra_filter_expression(self, value):
...
+        if arguments is not None:
+            # Ensure the value is a dict
+            if not isinstance(arguments, dict):
+                raise serializers.ValidationError({'arguments': 'Must be an object.'})
+
+            # Get the schema associated with the selected action
+            try:
+                schema = Action.objects.get(pk=action.id).arguments_schema
+            except Action.DoesNotExist:
+                raise serializers.ValidationError({'arguments': 'Could not find arguments schema.'})
```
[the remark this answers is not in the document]
rehandalal: Good catch!

peterbe:
screen shot 2018-04-25 at 10 50 29 am

I don't really know where this should be fixed but I see `const errorString = error instanceof Array ? error.join(' ') : error;` in https://github.com/mozilla/normandy/blob/cb17c665391c03f88980035e0faeec142d9b7f98/recipe-server/client/control/components/forms/FormItem.js#L95 trying to "convert" the `error` to `errorString`. It could be something like this instead:
```javascript
let errorString = error;
if (error instanceof Array) errorString = error.join(' ')
if (!error instanceof String) errorString = JSON.stringify(error)
```
or something that actually works :)

Even better would be to take the JSON from the 400 message and there massage the values of that object into strings. 

The 400 thing is actually *working* in that it prevents me from a successful submission. And the JSON blog textarea gets a red border. Just no message.
screen shot 2018-04-25 at 11 01 24 am

peterbe:
It works though. Yay!

I'm not sure this is something we need to worry ourselves with right now, but it looks kinda odd in the "View recipe" page:
screen shot 2018-04-25 at 11 04 26 am

As you can see, my `arguments` dict looks like this:
screen shot 2018-04-25 at 11 07 01 am
I.e. I have two `.preferences` being an array of dicts. 
Perhaps the display on View recipe could be some version of `json.dumps(arguments)`

mythmon:
The display on the view page isn't something  we're worrying about right now. There is some custom code that renders arguments, but I don't think we're going to be improving that at this time in Normandy. Delivery Console can do better in the future.

mythmon:
The remaining CI failure is #1241, and is fixed on master. If bors merges this, it will pass the tests, so it is safe to "bors r+" this now (once @peterbe is happy with the changes).

mythmon:
bors r=mythmon, @peterbe

mythmon:
Oops, lint problems:

```
normandy/recipes/api/v2/serializers.py:169:100: E501 line too long (102 > 99 characters)
normandy/recipes/tests/api/v1/test_serializers.py:77:100: E501 line too long (107 > 99 characters)
```

rehandalal:
bors r=@mythmon, @peterbe

peterbe:
Even the guy who wrote `therapist` accidentally let a flake8 error slip through :)

When I worked on [Tecken](https://github.com/mozilla-services/tecken/) I used `pytest-flake8` so it would check flake8 on all files and all test files simply when running `pytest`. It also meant CI could kill two birds with one stone instead of having to wait and do a linting thing.

PR #1243, opened 2018-04-26
Make revision ids stable, and unrelated to revision contents

### LINKED ISSUE #1212: Convert revisions ids to simple autoincrement instead of hashes
Currently `RecipeRevision` ids are not the normal Django pattern of a simple autoincrementing integer key. Instead they are using a hash of some of the data in the revision. The original goal was to have a git-like system and to avoid duplicating revisions.

We either didn't implement these features, or have removed them. The hashes are a point of complexity we don't need, and tech debt we should get rid of. Specifically this makes it very difficult to change the way we deal with filtering. When trying to migrate from simple filters to filter objects, it isn't reasonable to make filter objects correspond exactly with their simple filter counterparts. It also doesn't allow us to change the representation of filter objects in the future.

One issue is that these revisions are used in URLs, and making this change will change the URLs. I briefly talked to S&I, and although they use these URLs, they don't save or bookmark individual links, so if the format changes it won't affect them.

This issue covers:

* Removing the revision hashing
* Switching to auto-incrementing IDs for revisions
* A database migration to move existing revisions over to the new system
* Checking that the Control interface continues to work (I suspect it won't need any changes).

mythmon:
This isn't what we wanted. We wanted to change to a simple autoincremented integers for ids, but that has proven to be very difficult to migrate to. This is a compromise.

Fixes #1212.

### DIFF
--- a/normandy/recipes/models.py
+++ b/normandy/recipes/models.py
@@ -2,6 +2,7 @@ import hashlib
 import json
 import logging
 from collections import defaultdict
+from random import random
 
 from django.conf import settings
 from django.contrib.auth.models import User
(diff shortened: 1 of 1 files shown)


mythmon:
Rehan is going to explore an alternate solution to this problem.

rehandalal:
Got side-tracked will work on that this evening.

rehandalal:
Closing in favor of #1246

PR #1246, opened 2018-05-02
Migrate revisions from hashes to incremental IDs

### LINKED ISSUE #1212: Convert revisions ids to simple autoincrement instead of hashes

rehandalal:
Fixes #1212 

r?


normandy/recipes/migrations/0003_tmpreciperevision.py
```diff
@@ -0,0 +1,223 @@
...
+
+    if revision.approved_for_recipe.count():
+        tmp.approved_for_recipe.add(revision.approved_for_recipe.get())
+
+    if revision.latest_for_recipe.count():
+        tmp.latest_for_recipe.add(revision.latest_for_recipe.get())
+
+    try:
+        revision.approval_request.tmp_revision = tmp
+    except ApprovalRequest.DoesNotExist:
```
mythmon: Can this be done with an if instead of a try?

```python
if revision.approval_request:
    revision.approval_request.tmp_revision = tmp
```

normandy/recipes/migrations/0003_tmpreciperevision.py
```diff
@@ -0,0 +1,223 @@
...
+            field=models.ForeignKey(null=True, on_delete=django.db.models.deletion.SET_NULL,
+                                    related_name='latest_for_recipe',
+                                    to='recipes.TmpRecipeRevision'),
+        ),
+        migrations.AlterField(
+            model_name='approvalrequest',
+            name='revision',
+            field=models.OneToOneField(null=True, on_delete=django.db.models.deletion.SET_NULL,
+                                       related_name='approval_request',
+                                       to='recipes.RecipeRevision'),
```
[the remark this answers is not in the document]
rehandalal: yep

mythmon:
> As a side note: I'm not sure if we need a migration to get fresh signatures after this change? I suspect yes?

In the past I think we would have, since the revision id is part of the signed blob. However, nowadays we forcibly resign everything on every deploy anyways, so we don't have to make those migrations anymore.

normandy/recipes/migrations/0003_tmpreciperevision.py
mythmon: Weird.

mythmon:
I tested this migration back and forth. It seems to handle data well, and I don't see any issues with the code. Nice job! It will be really great to have a normal id scheme here.

One issue I ran into is that the Normandy control interface can't view old revisions any more. They show up in the history side bar, but clicking on them goes to a page with no data. It is still possible to view the latest revision and edit it. I didn't check approving new revisions.

We aren't working on new features for the control interface, but I'd prefer not to break it. At the very least I'd like to explain what's going on, in case it is a problem with the migration. Though every thing looked fine to me in the API, so I don't think that's the case.

rehandalal:
I'll investigate

mythmon:
```
$ eslint normandy client karma.conf.js webpack.config.js *.webpack.config.js

/app/client/control/components/recipes/CloneRecipePage.js
  16:10  error  'getUrlParam' is defined but never used  no-unused-vars

/app/client/control/components/recipes/DetailsActionBar.js
  28:3  error  'getUrlParam' is defined but never used  no-unused-vars

/app/client/control/components/recipes/RecipeDetailPage.js
  20:10  error  'getUrlParam' is defined but never used  no-unused-vars
```

mythmon:
Besides the lint errors, lgtm.

PR #1250, opened 2018-05-04
Update python dependencies

rehandalal:
wheeee!

~I still need to update some packages in constraints.txt.~

And I'll do django in another PR


requirements/default.txt
```diff
@@ -1,10 +1,33 @@
...
+    --hash=sha256:a005ed6163490988711ff732386b08effcbf8df62ae93dd1e5bda0714fad8afb \
+    --hash=sha256:ae35dbcb6b011af6c840893b32399252d81ff57d52c13e12422e16b5fea1d0fb \
+    --hash=sha256:b1e8491c6740f21b37cca77bc64677696a3fb9f32360794d57fa8477b7329eda \
+    --hash=sha256:c906bdb482162e9ef48eea9f8c0d967acceb5c84f2d25574c7d2a58d04861df1 \
+    --hash=sha256:cb18ffdc861dbb244f14be32c47ab69604d0aca415bee53485fcea4f8e93d5ef \
+    --hash=sha256:d86da365dda59010ba0d1ac45aa78390f56bf7f992e65f70b3b081d5e5257b09 \
+    --hash=sha256:e22f0997622e1ceec834fd25947dc2ee2962c2133ea693d61805bc867abaf7ea \
+    --hash=sha256:f2fe545d27a619a552396533cddf70d83cecd880a611cdfdbb87ca6aec52f66b \
+    --hash=sha256:f7fd3ed3745fe6e81e28dc3b3d76cce31525a91f32a387e1febd6b982caf8cdb \
+    --hash=sha256:f9210820ee4818d84658ed7df16a7f30c9fba7d8b139959950acef91745cc0f7
```
mythmon: I imagine you generated this with hashin. That means that these hashes include hashes for Python 2.7, and other things we don't want. You can avoid that with hashin's `-p 3.6` flag, which will only include hashes for Python 3.6.

normandy/recipes/tests/api/v1/test_serializers.py
```diff
@@ -52,7 +52,7 @@ def test_it_works(self, rf):
             'locales': [locale.code],
             'is_approved': False,
             'latest_revision_id': recipe.latest_revision.id,
-            'approved_revision_id': recipe.approved_revision_id,
```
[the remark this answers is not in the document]
rehandalal: It didn't actually change. `recipe.approved_revision_id` is `None` but I felt like this was a little clearer.

I made the change while trying to figure out why this test was failing. The actual reason for the failure was:
https://github.com/mozilla/normandy/pull/1250/files/b07e5ab421c9e1be8db49b8da072b3e1418ccd40..e89ad3d44c946183808b1b0266c23c603e12d9cc#diff-5e1738ebe896bccd02dcf8290b09c7b8R108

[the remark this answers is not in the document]
mythmon: Cool, that makes sense :+1:.

rehandalal:
I think this is ready for primetime.

mythmon:
I got a minor warning when running the tests:

```
normandy/studies/tests/api/v2/test_api.py::TestExtensionAPI::()::test_order_bogus
  /home/mythmon/src/normandy/.direnv/python-3.6.5/lib/python3.6/site-packages/rest_framework/pagination.py:198:
    UnorderedObjectListWarning: Pagination may yield inconsistent results with an unordered object_list: <class 'normandy.studies.models.Extension'> QuerySet.
    paginator = self.django_paginator_class(queryset, page_size)

-- Docs: http://doc.pytest.org/en/latest/warnings.html
```

mythmon:
Besides the test warning above, this looks good to me.

rehandalal:
Warnings should be fixed now too!

normandy/studies/models.py
```diff
@@ -8,6 +8,9 @@ class Extension(models.Model):
     name = models.CharField(max_length=255)
     xpi = models.FileField(upload_to='extensions')
 
+    class Meta:
+        ordering = ('name',)
```
mythmon: What do you think about making this `-id` instead of `name`? I think that would better match the other APIs, and would also nicely fix #1091.

rehandalal: amended

PR #1322, opened 2018-05-04
[WIP] auth by bearer authorization header, fixes 1245

peterbe:
Here's the code that proved the prototype: https://github.com/mozilla/normandy/compare/mozilla:master...peterbe:auth-by-bearer-authorization-header-fixes-1245

* [ ] Tune the code and resolve all `XXX ...` comments
* [ ] Unit tests
* [ ] Wonder if this should only work for certain paths like `/api/v2`. Some whitelist might be a good idea. 

Also, we have two other concurrent issues to solve. The delivery console needs to correctly send the access_token. 
And our Nginx Lua code doesn't allow any requests through (even if the header already has a `Authorization`). We need to find a way to bypass that. Ideally with some conditionals rather than having to edit the Lua code.


peterbe:
The PR for getting delivery console to play ball is here: https://github.com/mozilla/delivery-console/pull/157

peterbe:
> What is the benefit of this?

Just because I'm not as familiar with what you could reach beyond the django rest framework. Suppose that we disallow any auth-by-bearer-token for anything *other* than `/api/v2/*` then we have less of a surface to worry about.

normandy/base/middleware.py
```diff
@@ -128,3 +136,109 @@ def process_request(self, request):
...
+def default_username(email):
+    """Generate username for the Django user.
+
+    :arg str/unicode email: the email address to use to generate a username
+
+    :returns: str/unicode
+
+    """
+    # bluntly stolen from django-browserid
+    # store the username as a base64 encoded sha224 of the email address
```
mostlygeek: code says sha1

mostlygeek:
> And our Nginx Lua code doesn't allow any requests through (even if the header already has a Authorization). We need to find a way to bypass that. Ideally with some conditionals rather than having to edit the Lua code.

This PR should be moving all authorization / token verification into python. We wouldn't need the nginx/lua anymore.  

We can discuss migrating user to this new setup in a different issue.

normandy/base/middleware.py
```diff
@@ -128,3 +136,109 @@ def process_request(self, request):
...
+
+class OIDCAccessTokenAuthenticationMiddleware:
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+        # XXX Here we *could* do something like:
+        # if not settings.OIDC_ACCESS_TOKEN_ENABLED:
+        #    from django.core.exceptions import MiddlewareNotUsed
+        #    raise MiddlewareNotUsed
+        self.oidc_user_endpoint = settings.OIDC_USER_ENDPOINT
```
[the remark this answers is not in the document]
peterbe: Just a nice feeling that once the middleware is instanciated, it's complete and doesn't need to re-read other sources.

normandy/base/middleware.py
```diff
@@ -128,3 +136,109 @@ def process_request(self, request):
...
+        self.UserModel = auth.get_user_model()
```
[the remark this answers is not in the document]
peterbe: If there's a cost of calling `auth.get_user_model()` it's only paid once per python process.

normandy/base/middleware.py
```diff
@@ -128,3 +136,109 @@ def process_request(self, request):
...
+    return smart_text(base64.urlsafe_b64encode(
+        hashlib.sha1(force_bytes(email)).digest()
+    ).rstrip(b'='))
+
+
+class OIDCAccessTokenAuthenticationMiddleware:
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+        # XXX Here we *could* do something like:
```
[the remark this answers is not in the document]
peterbe: My PR doesn't introduce a on-off switch for using authentication by bearer tokens. Should it? Is it overkill? Why would you want to switch it off, ever?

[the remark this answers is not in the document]
mythmon: I don't think there needs to be a separate on-off switch besides just removing the middleware, which we can manage well with environment variables.

normandy/settings.py
```diff
@@ -212,6 +212,7 @@ def AUTHENTICATION_BACKENDS(self):
 
     OIDC_REMOTE_AUTH_HEADER = values.Value('HTTP_REMOTE_USER')
     OIDC_LOGOUT_URL = values.Value(None)
+    OIDC_USER_ENDPOINT = values.URLValue('https://auth.mozilla.auth0.com/userinfo')
```
mythmon: We may want to either distinguish this new auth method from the previous "OIDC" settings/middlewares here. Having a setting `USE_OIDC` that isn't actually related to the OIDC middle ware sounds confusing.

I don't think we can completely remove it yet though. It may be good to have two separate config presets though.

normandy/base/middleware.py
```diff
@@ -128,3 +136,109 @@ def process_request(self, request):
...
+    :returns: str/unicode
+
+    """
+    # bluntly stolen from django-browserid
+    # store the username as a base64 encoded sha224 of the email address
+    # this protects against data leakage because usernames are often
+    # treated as public identifiers (so we can't use the email address).
+    return smart_text(base64.urlsafe_b64encode(
+        hashlib.sha1(force_bytes(email)).digest()
+    ).rstrip(b'='))
```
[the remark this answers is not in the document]
peterbe: I don't actually see where our current implementation uses the email as the username. In https://github.com/mozilla/normandy/blob/master/normandy/base/migrations/0001_copy_email_to_username.py the username becomes "peter" if the email was "peter@example.com". 

With this proposed change, for all *new* users the username becomes as a hash version of the email. This allows for peter@example.com and peter@contractor.com to both be able to log in distinctly.

normandy/base/middleware.py
```diff
@@ -128,3 +136,109 @@ def process_request(self, request):
...
+    def __call__(self, request):
+        response = self.process_request(request)
+        if not response:
+            response = self.get_response(request)
+        return response
+
+    def process_request(self, request):
+        try:
+            access_token = re.findall(
+                'Bearer\s+(.*)',  # Is this right?
```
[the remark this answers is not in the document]
peterbe: Could it ever tack on more things. Like `Bearer: deadbeef123; charset=utf8` or something like that? Probably not likely. I agree that the `(.*)` is pretty solid as is.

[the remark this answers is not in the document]
mythmon: > Could it ever tack on more things. Like Bearer: deadbeef123; charset=utf8 or something like that?

That would not be standards compliant. I don't think we should worry about it.

peterbe:
NOTE! This PR only really kicks in if the request has an `HTTP_AUTHORIZATION` header. If the header is *present* the outcome is *only* one of these:

1. 403 Forbidden
2. Signed in for the duration of the request.

If you however request *without* the header at all, you *could* do anonymous requests. Meaning, things like `curl localhost:8000/api/v2/service-info` will `200 OK` and the JSON would have `user=null` in it. 

We *could* change approach and do something like this:
```python
        try:
            access_token = re.findall(
                'Bearer\s+(.*)',  # Is this right?
                request.META.get('HTTP_AUTHORIZATION') or '',
                re.I
            )[0]
        except IndexError:
            # No authorization header, no soup!
            raise PermissionDenied('no authorization bearer token')
```
*That* would be the equivalent of doing something like this in Nginx:
```
location / {
    auth_basic "Private Stats";
    auth_basic_user_file /etc/nginx/.htpasswd;
    proxy_pass http://127.0.0.1:5000;
}
```

If we want to replace the Lua solution we ought to disallow all anonymous requests. But also, what does that mean in terms of a transition phase from the old to the new? What do you think @mythmon ?

normandy/base/middleware.py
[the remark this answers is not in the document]
peterbe: I'll remove that `# Is this right?` comment.

normandy/base/middleware.py
```diff
@@ -128,3 +136,109 @@ def process_request(self, request):
...
+                request.META.get('HTTP_AUTHORIZATION') or '',
+                re.I
+            )[0]
+        except IndexError:
+            # That's cool. If it's not there, it's not there.
+            # This middleware will not be able to authenticate the user
+            # for the duration of this request.
+            return
```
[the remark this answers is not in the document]
peterbe: In your snippet of code there you didn't use `re.IGNORE_CASE`. Was that deliberate?  The RFC doesn't appear to say it's case-insensitive. 
I think I'll steal this variant (instead of using `except IndexError`).

normandy/base/middleware.py
```diff
@@ -128,3 +136,109 @@ def process_request(self, request):
...
+            # Awesome! Turn this into a Django User
+            user_profile = response.json()
+            if 'email' not in user_profile:
+                raise ImproperlyConfigured(
+                    "The HTTP Authorization header had an access token that "
+                    "was created without the 'email' scope."
+                )
+            email = user_profile['email']
+            if not email:
+                # XXX Not sure how this could happen.
```
peterbe: This is totally possible. If someone does this:
```
new auth0.WebAuth({
    domain: OIDC_DOMAIN,
    clientID: OIDC_CLIENT_ID,
    redirectUri: OIDC_CALLBACK_URL,
    responseType: 'token id_token',
    scope: 'openid profile',  // <--- NOTE! No 'email' scope
  });
```

We should definitely raise a `PermissionDenied` but I don't know if whatever message we put in there will be visible to the client if they debug the response body. Either way, this would be a good opportunity to explain the OIDC scope thing at least.

mythmon:
> If we want to replace the Lua solution we ought to disallow all anonymous requests. But also, what does that mean in terms of a transition phase from the old to the new? What do you think @mythmon ?

The fact that the current Lua solution doesn't allow anonymous requests was a limitation, not something that we did on purpose. Enabling anonymous requests would be nice resolution to this PR.

normandy/base/middleware.py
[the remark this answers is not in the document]
mythmon: Uh, no the lack of `re.IGNORE_CASE` was an accident. You should keep using that.

normandy/base/middleware.py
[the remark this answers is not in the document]
mythmon: It also means that testing is harder, since a test can't easily change a global setting and then test the results. In general, I think that storing the values from settings mostly leads to trouble.

normandy/base/middleware.py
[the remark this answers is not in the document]
peterbe: Removed comment.

normandy/base/middleware.py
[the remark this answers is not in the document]
peterbe: No, the middleware is instanciated for every test. So if you change `settings.OIDC_USER_ENDPOINT` in just one test it changes for that request (if the test makes a request)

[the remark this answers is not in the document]
peterbe: I had to check, out of curiosity. If you run django's runserver, it creates and stores instances of the middleware classes between requests.

normandy/settings.py
peterbe: Naming things are hard. 

I like you point that it is confusing that *this* OIDC stuff we're doing now is muddled with the old OIDC stuff. And we can't delete all the old stuff yet, can we?

One nice start would be to create a new class, similar to [CORS](https://github.com/peterbe/normandy/blob/b82b74618cfde4b7ee014d213f45cbbc7cfa1958/normandy/settings.py#L176) but dedicated to this OIDC-authentication-by-access_token. Then it's at least not near the "old OIDC" stuff. 
Ideas for a name (of the class)?

mythmon:
> So this is where I have to defer to your judgement. Perhaps Normandy is so security paranoid that no anonymous access should be allowed.

Normandy has historically been pretty security paranoid. However I think that not allowing anonymous requests will actually worsen our security overall. If we rely on all requests being authed, then we may not look at each view carefully enough. If we know that some users may be anonymous, we'll have things like the permission classes here: https://github.com/mozilla/normandy/blob/848fba594e92396eeb483e29c1fafb36a2f62133/normandy/recipes/api/v2/views.py#L77-L80

I think anonymous requests are ok.

mythmon:
> So this is where I have to defer to your judgement. Perhaps Normandy is so security paranoid that no anonymous access should be allowed.

Normandy has historically been pretty security paranoid. However I think that not allowing anonymous requests will actually worsen our security overall. If we rely on all requests being authed, then we may not look at each view carefully enough. If we know that some users may be anonymous, we'll have things like the permission classes here: https://github.com/mozilla/normandy/blob/848fba594e92396eeb483e29c1fafb36a2f62133/normandy/recipes/api/v2/views.py#L77-L80

I think anonymous requests are ok.

mostlygeek:
Note: OAuth access tokens are for __authorization__ not __authentication__!  

+1 for anonymous access where it is appropriate.  
+1 for decorators adding the authorization middleware around appropriate views 

> It's perhaps altogether a rather silly conversation since the whole admin API is protected by VPN in the first place :)

We should design this like the VPN and the nginx-lua doesn't exist.

normandy/base/middleware.py
```diff
@@ -128,3 +136,122 @@ def process_request(self, request):
...
+            if (
+                'family_name' in user_profile and 'given_name' in user_profile and
+                (
+                    user_profile['family_name'] != user.last_name or
+                    user_profile['given_name'] != user.first_name
+                )
+            ):
+                user.first_name = user_profile['given_name']
+                user.last_name = user_profile['family_name']
+                user.save()
```
mostlygeek: updating the first/last name of the user in the database introduces a side effect.  Seems weird to have it here.

normandy/base/middleware.py
```diff
@@ -128,3 +136,122 @@ def process_request(self, request):
...
+        if response.status_code == 200:
+            # Awesome! Turn this into a Django User
+            user_profile = response.json()
+            if 'email' not in user_profile:
+                raise ImproperlyConfigured(
+                    "The HTTP Authorization header had an access token that "
+                    "was created without the 'email' scope."
+                )
+            email = user_profile['email']
+            if not email:
```
mostlygeek: if ther ewasn't an `email` scope, wouldn't the `if 'email' not in user_profile` catch it first?

mythmon: No, because if the email is `''` (the empty string) then it is `in user_profile`, but is not truthy, and so the code will take the `if not email:` branch.

normandy/base/middleware.py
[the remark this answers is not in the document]
mythmon: For those playing along at home, Peter figured out that he read the code in the migration wrong. The username would become "peter@example.com" in his example. The code he was looking at was both a forward and backards migration, which doesn't make sense when read straight through.

@peterbe how would you like to proceed here?

normandy/base/middleware.py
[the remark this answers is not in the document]
mythmon: Accessing settings isn't an expensive operation. Storing settings off to the side creates different views into what the settings actually are. This is a minor issue here, but code that relies on settings not changing is fragile and has the potential to become a maintenance issue later. I'd prefer not to do it, even if it would work in the short term.

normandy/base/middleware.py
peterbe: I think I'm going to collapse these and only use the `PermissionDenied`.

The verbose comment below is still useful but now it's actually situated in the wrong place. 
If someone does, correctly, use the `email` scope and super-unlikely come with a blank email then we simply don't want to let them in. Perhaps there's something in the RFC of OAuth2.0 about the email having to be a real email if you set the scope but I'm too lazy to dig into that. No email, permission denied. Using `ImproperlyConfigured` was wrong.

normandy/base/middleware.py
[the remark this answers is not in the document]
peterbe: I don't care. The max_length of `username` (150 chars) is different than `email` (254 chars) but who has an email that is longer than 150?! 
Either way, simplicity rules and this project doesn't have gazillions of users.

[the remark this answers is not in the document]
mostlygeek: +1 for using just the email address and not hashing it.

normandy/base/middleware.py
```diff
@@ -117,3 +122,150 @@ def process_request(self, request):
...
+    4. We extract the email from the user profile (and first- and last name)
+    5. Turn the email into a Django User model instance.
+    6. If possible, extracts the first- and last name and updates the User instance persistently.
+    7. Sets that user as the user of the request thus enabling `request.user.is_active`.
+
+    NOTE! When the bearer token is present, it does a network request to the OIDC provider
+    every single time. There is no caching. Even for identical access tokens used
+    frequently. The justification for this at the Normandy Admin UI requests are
+    not high-traffic to performance critical. Also, we don't have a cache framework
+    set up.
```
mythmon: > Also we don't have a cache frame work set up

IIRC, Django defaults to an in-memory cache, which would still be beneficial.

normandy/base/middleware.py
```diff
@@ -117,3 +122,150 @@ def process_request(self, request):
...
+    """
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+        self.oidc_user_endpoint = settings.OIDC_USER_ENDPOINT
```
mythmon: I feel bad harping on this, but I want to draw a hard line here. Storing values from settings for long term use is something I want to not do in Normandy. Can you take this out?

normandy/base/middleware.py
```diff
@@ -117,3 +122,150 @@ def process_request(self, request):
...
+
+    Note, it only signs in the user's request. Not session. Since this is not
+    an authentication backend, there is no 'Set-Cookie' header returned in the
+    response.
+
+    It's admittedly confusing to mix the terms authentication and authorization.
+    Arguably the client has already *authenticated* and basically converted her
+    password for an access token. Here in the context of Normandy we kill two
+    birds with one stone: check that access token and use the user profile that
+    comes out to sign the user into the rest of the Django request.
```
mythmon: I don't think this paragraph adds much to a reader who wasn't a part of our debate. I'd like to just take it out entirely.

normandy/base/middleware.py
```diff
@@ -117,3 +122,150 @@ def process_request(self, request):
...
+            # access token did work but not sufficient for our needs.
+            raise PermissionDenied("User profile lacks 'email' scope.")
+
+        # Minimal normalization. Also forces all emails to be lower case
+        # which makes the UserModel lookup by email faster.
+        email = email.strip().lower()
+
+        # Awesome! Turn this email into a Django User instance.
+        try:
+            user = self.UserModel.objects.get(email=email)
```
mythmon: To me this is the crux of the authn vs authz debate. Because this code will never deny the user *after this point*, I don't think authz is happening. To me authz can only happen once you have decided who the user is, and aren't doing anything like that after this point in this middleware. (I don't think we need any changes from this comment, it was just a good visual aid for my point).

normandy/base/middleware.py
```diff
@@ -117,3 +122,150 @@ def process_request(self, request):
...
+        ):
+            user.first_name = user_profile['given_name'].strip()
+            user.last_name = user_profile['family_name'].strip()
+            user.save()
+
+            # We *could* do something like this too:
+            # if user_profile.get('picture'):
+            #     pf = user.get_profile()
+            #     pf.picture = user_profile['picture']
+            #     pf.save()
```
mythmon: I'd like to either do this or remove the commented code.

normandy/base/middleware.py
```diff
@@ -117,3 +122,150 @@ def process_request(self, request):
...
+
+        # Now, let's "become" this user for the rest of the request.
+        request.user = user
+        user_logged_in.send(sender=user.__class__, request=request, user=user)
+
+    @backoff.on_exception(
+        backoff.expo,
+        requests.exceptions.RequestException,
+        max_tries=5,
```
mythmon: I dug into the code, and I figured out that this will wait up to about 30 seconds. Most of the requests that will be authenticated will (I suspect) be interactive, so this seems too long. Can we change one of the parameters so that this takes less time to fail in the case we get no response? For example, setting `base=1.3` would make 5 requests take about 10 seconds, which seems better to me.

I've submitted a request to Auth0 support to see if the have any standard guidance about this. I also checked the docs, but I didn't find anything there.

normandy/base/middleware.py
```diff
@@ -117,3 +122,150 @@ def process_request(self, request):
...
+    )
+    def _fetch_oidc_email(self, access_token):
```
mythmon: This says `_fetch_oidc_email`, but it doesn't returns a profile, right? Can we change the name?

normandy/settings.py
```diff
@@ -186,7 +186,20 @@ class CORS:
...
-class Base(Core, CORS):
+class OIDC:
+    """Settings related to talking to an OIDC provider for authorizing
+    access tokens received from the client."""
+
+    # Note! We *could* just have a setting called 'OIDC_DOMAIN' and then use
+    # https://$OIDC_DOMAIN/.well-known/openid-configuration and extract the
+    # 'userinfo_endoint' value from that during startup or something.
+    # As of May 2018, the likelyhood of this URL changing and the fact that it's
+    # the only URL we need, let's just make the setting thee URL that we need
```
mythmon: Typo: "thee"

normandy/settings.py
```diff
@@ -483,3 +497,4 @@ class Test(Base):
     AUTOGRAPH_URL = None
     AUTOGRAPH_HAWK_ID = None
     AUTOGRAPH_HAWK_SECRET_KEY = None
+    OIDC_USER_ENDPOINT = 'https://auth.example.com/userinfo'
```
mythmon: Is this a duplicate? It seems like it unconditionally overrides the entire `CORS` class's contribution.

docs/ops/config.rst
```diff
@@ -338,6 +338,15 @@ in other Django projects.
    reduced to "read-only" methods when using the ``ProductionReadOnly``
    configuration.
 
+.. envvar:: DJANGO_OIDC_USER_ENDPOINT
+
+   :default: ``https://auth.mozilla.auth0.com/userinfo``
+
+   URL where we sent access tokens received as an authorization bearer token.
+   This URL needs to match the OIDC domain used by the client to authenticate.
+   The endpoint URL is actually extracted from reading ``/.well-known/openid-configuration``
```
mythmon: I think this may be confusing, and a reader might think that Normandy will make the request to the well-known URL. Maybe something like "The value for this setting is usually listed in `/.well-known/openid-configuration` on the OIDC provider."?

normandy/base/middleware.py
```diff
@@ -117,3 +122,150 @@ def process_request(self, request):
...
         return response
+
+
+class OIDCEndpointRequestError(Exception):
+    """Happens when the server-to-server communication with the OIDC endpoint
+    succeeds but the OIDC endpoints responds with a status code > 200 and less
+    than 500."""
+
+
+class OIDCAccessTokenAuthorizationMiddleware:
```
mythmon: I'm actually ok with the use of the word authorization here, because it is dealing specifically with the authorization HTTP header.

normandy/base/middleware.py
peterbe: I'm just not convinced it's worth it. But perhaps you have a better feeling for how the Admin UI works. I don't expect there to be that much traffic anyway. I'd rather stall and incrementally add this later if we find that it's slowing 
things down. 

We'd need to try to make sure the TTL on that caching matches the TTL on the access token itself. I think it's set to 2h according to the idToken I get in delivery console. 

Mind you, the in-memory cache is per Python process and if we have many workers per EC2 there's little cache hits.

normandy/base/middleware.py
peterbe: Funny, I heard another interesting argument about the authn vs authz bikeshed today. **Authn is global and authz is when there's an action.**

Going to the will-call booth to get your ticket is authn. But actually walking through the security gate, showing your ticket, is authz. The action is "Walking into the secure area". 

Arguably, calling `GET /api/v2/service-info` is not action but `POST -d <data> /api/v2/recipes` is.

normandy/base/middleware.py
peterbe: Agreed.

normandy/base/middleware.py
peterbe: I'd rather we switch to `backoff.constant` then. With max_tries=5 it means a max of 5 seconds.

normandy/base/middleware.py
peterbe: Done.

normandy/settings.py
peterbe: Not sure what you mean. The point of this is to make sure the tests never actually reach out to auth.mozilla.auth0.com. This way the tests aren't at the mercy of the environment. 

Also, the domain auth.example.com will not resolve so if your unit tests accidentally *don't* mock requests, you'd find out immediately.

docs/ops/config.rst
peterbe: That's a much better phrase. Thanks.

peterbe:
So the remaining things are:

1. Should we use django's in-memory cache to cache the OIDC requests.get? 
2. Where do we write about bearer token authorization in the docs?

normandy/base/middleware.py
peterbe: Another argument for calling it authz here in this PR is that it's stateless. Authn is usually state-full. I.e. turning credentials into something that lasts, e.g. a session cookie, or an access token in localStorage.

peterbe:
>  If I wanted to test this locally, what is the best way to do that?

To quick answer is that you need to checkout out https://github.com/mozilla/delivery-console/compare/master...peterbe:send-authorization-header-to-normandy-fixes-140 and then run `yarn start` and click the Login button on `localhost:3000`. 

What I did was I "Copy to cURL" in the Network panel so I can repeatedly test this on the command line. Also gave me the chance to manually mess with the access token to assert that it would 403 correctly. 

A longer answer is that you can open the Web console and type `copy(JSON.parse(localStorage.session).accessToken)` then paste that into
```
curl -s -k -v 'https://localhost:8000/api/v2/service_info/'  -H 'authorization: bearer ll8W2vhqM2EaO16lbeOSWfmtcAic56q1'
```

normandy/settings.py
mythmon: I totally misread this on my first pass. I thought this class was the base settings, and missed that it was the test settings. It is good as is, Sorry!

normandy/base/middleware.py
mythmon: :+1: to `backoff-constant`.

normandy/base/middleware.py
mythmon: I'm not arguing for or against storing things in the cache. I'd be happy to leave the code as it is. I was just objecting to that last sentence, as it isn't strictly true.

mythmon:
> Where do we write about bearer token authorization in the docs?

It may be a mistake, but I think I'd like to opt to just not write about them *right now*, and wait until the dust settles, so to speak. Once we have the whole pipeline figured out (delcon, nginx, normandy, auth0, etc.), lets make a cohesive guide for what authing to Normandy's API looks like. Does that sound ok? If not, probably just writing down the docs literally anywhere that shows up on readthedocs is good enough for me.

normandy/base/middleware.py
peterbe: Sentence removed.

normandy/base/middleware.py
peterbe: Do you want me to search-and-replace authz for authn? I'm not wedded to either. The fact that it arose a "debate" by several experienced developers means it's tricky notion to get right.

peterbe:
bors r= @mythmon

PR #1337, opened 2018-05-08
Switch to python-dockerflow

### LINKED ISSUE #1334: Migrate from mozilla-cloud-services-logger to python-dockerflow
[mozilla-cloud-services-logger](https://github.com/mozilla/mozilla-cloud-services-logger) is no longer maintained and not compatible with Django >= 2.0. 

The README for the project suggests using [mozilla-services/python-dockerflow](https://github.com/mozilla-services/python-dockerflow).

rehandalal:
Fixes #1334.

r?


rehandalal:
Looks like I need to tweak some circle things

bin/ci/dependencies
```diff
@@ -8,10 +8,9 @@ sudo apt-get install firefox
 sudo apt-get remove firefox-mozilla-build
 
 # store version info in app directory, for Docker to pick up
-mkdir -p __version__
-git rev-parse HEAD > __version__/commit
-git describe --tags --exact-match > __version__/tag || true # may fail if not on a tag
-date '+%s' > __version__/build_time
+printf '{"commit":"%s","version":"%s","source":"https://github.com/%s/%s","build":"%s"}\n' \
+  "$1" "$2" "$3" "$4" "$5" \
```
mythmon: I think this new version is a lot less clear. It mostly just moves the problem of `$CIRCLE_TAG` to a place that doesn't care. I think this version would be more readable, and should pass CI.

```bash
printf '{"commit":"%s","version":"%s","source":"https://github.com/%s/%s","build":"%s"}\n' \
  "$CIRCLE_SHA1" \
  "${CIRCLE_TAG-}" 
  "$CIRCLE_PROJECT_USERNAME" \
  "$CIRCLE_PROJECT_REPONAME" \
  "$CIRCLE_BUILD_URL" \
  > version.json
```

Note the syntax `${CIRCLE_TAG-}`. The `-` substitutes a value for an unbound variable, in this case the empty string. We could also include a default value, or an alternate way to get a "version", like:

```bash
${CIRCLE_TAG-$(git describe)}
```

This would be equivalent to the JS `CIRCLE_TAG || git_describe()`.

PR #1343, opened 2018-05-09
Use pip 10's "pip check" command to check dependency compatibility

mythmon:
This would have caught the error that caused #1339 to fail in a more easily understandable way. It will also detect more subtle package incompatibilities, should we find them in the future.


mythmon:
Done.

bors r= @rehandalal

PR #1349, opened 2018-05-10
Organize requirements files to simplify usage

mythmon:
In #1338, I changed the way the docs are built so that filters could be documented with Sphinx's autodoc. [That broke readthedocs](https://readthedocs.org/projects/normandy/builds/7170535/), since the requirements file it read didn't include any of our other requirements. Since autodoc works by importing our Python files, that means that without installing all of the dependencies, Sphinx can't work.

To fix this, I reorganized the requirements files a bit. To start, I made the constraints automatically included from the requirements. I also updated the docs to reflect that. Then I made the docs requirements include the default requirements, so Sphinx will have what it needs. Next I added the constraints that Sphinx needed.

Finally, once this merges I can update readthedocs' config to point to the new requirements file.


requirements/default.txt
```diff
@@ -1,3 +1,5 @@
+-c constraints.txt
```
rehandalal: 🎉

PR #1352, opened 2018-05-10
Add Python 3.5 hashes to packages, for readthedocs

mythmon:
#1349 got the RTD builds farther, [but now they fail](https://readthedocs.org/projects/normandy/builds/7170900/) because RTD runs Python 3.5, and we only have Python 3.6 hashes for some things. I've added the Python 3.5 hashes for these packages, which are the ones that failed in that recent build.

Hopefully third time's the charm.


mythmon:
Updated

PR #1370, opened 2018-05-15
Allow all origins via CORS

### LINKED ISSUE #1369: Open up our CORS settings to allow JS usage of the Normandy API
When we added CORS support to Normandy, [I argued that we should have a whitelist of domains](https://github.com/mozilla/normandy/pull/1202#issuecomment-370585368), instead of allowing any origin. When that happened, I had more in mind things like Delivery Console, and not dashboards, such as those built by S&I.

Hamilton pinged me today, trying to get "around CORS" when working with Normandy. This is a silly situation. We aren't using the same-origin policy as a security mechanism *in this case*, because all GET requests are idempotent, and we don't have sensitive information in our responses. These are the two major attacks the same-origin policy defends against. We also have CSP, which prevents other times of cross-origin attacks.

In short, we should set the allowed origins for CORS to `*`.

mythmon:
This aligns better with how we want the API to be used eventually.

Fixes #1369.

### DIFF
--- a/normandy/settings.py
+++ b/normandy/settings.py
@@ -426,10 +426,6 @@ class Production(Base):
     DEFAULT_FILE_STORAGE = values.Value('normandy.base.storage.S3Boto3PermissiveStorage')
     AWS_S3_FILE_OVERWRITE = False
 
-    # Custom CORS settings that overrides the CORS class's configuration.
-    # In production we harden it down a bit extra.
-    CORS_ORIGIN_ALLOW_ALL = values.BooleanValue(False)
-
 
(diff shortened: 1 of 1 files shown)


peterbe:
bors r= @peterbe

mythmon:
For posterity, @peterbe's codepen has this code in it:

```js
fetch('https://localhost:8000/api/v2/servjeoifjeoifj/', {
  headers: {
    'Authorization': 'Bearer junkhere'
  }
}).then(response => console.log(response.status))
.catch(ex => console.warn(ex.toString()))
```

PR #1378, opened 2018-05-17
Carry over enabled state on approval

rehandalal:
r?


normandy/recipes/models.py
```diff
@@ -521,9 +523,19 @@ def approve(self, approver, comment):
...
+
+        # Check if the recipe is enabled we should carry over it's enabled state
+        carryover_enabled = None
+        if recipe.enabled:
+            carryover_enabled = recipe.approved_revision.enabled_state
+
         recipe.approved_revision = self.revision
         recipe.save()
 
+        # Note: We need to update the approved_revision before we can enable the new revision
```
mythmon: This comment is referring to the two lines prior, right? I find it a bit hard to understand as it is. Maybe "Note: enabling the new revision has to happen after updating the approved revision because ..."?

rehandalal:
bors r=@mythmon

PR #1389, opened 2018-05-22
Comment field for recipes

### LINKED ISSUE #1082: Add general notes field to recipes
Every recipe should have a comments field that recipe editors can save arbitrary notes about the recipe in. This could be related bugs or documents, or explanation of what's going on. This should not be visible to users (public API is fine), related to any particular action, or have any other purpose.

rehandalal:
Fixes #1082.

r?


mythmon:
Cool. Any idea when we can expect to see this in Delivery console?

PR #1392, opened 2018-05-24
Add bug number to revisions

### LINKED ISSUE #1198: Add bug number metadata field to recipe revision
Recipe revisions should include a bug number that the recipe is related to. This should be an optional field, and won't be included in the minimal recipe serializer used in `/api/v1/recipe/signed/`.

mythmon:
Fixes #1198


mythmon:
I dropped the recipe current revision property when I rebased this on master. I fixed that and force pushed.

bors r= @rehandalal

PR #1393, opened 2018-05-24
Validate that rollbacks refer to existing rollouts

### LINKED ISSUE #1248: Validate preference rollback slugs
In preference roll*back* arguments, there is a key, `rolloutSlug` that refers to an existing rollout to undo. We should validate, server side, that a rollout with that slug actually exists.

mythmon:
Fixes #1248


mythmon:
Probably worth looking at this one with whitespace changes ignored, since I re-arranged the tests a bunch. (they added it as a real feature now, in the top right of the diff view).

PR #1398, opened 2018-06-06
Fix OIDC authorization

### LINKED ISSUE #1397: OIDC Authorization does not work for POST requests
The bearer token authorization works fine for GET requests but seems to return a 403 error when making POST requests to the API.

rehandalal:
Fixes #1397.

r?


normandy/base/api/authentication.py
```diff
@@ -0,0 +1,128 @@
...
+            # Ideally the client should review their config. The
+            # access token did work but not sufficient for our needs.
+            raise exceptions.AuthenticationFailed("User profile lacks 'email' scope.")
+
+        # Minimal normalization. Also forces all emails to be lower case
+        # which makes the UserModel lookup by email faster.
+        email = email.strip().lower()
+
+        # Awesome! Turn this email into a Django User instance.
+        # TODO: Users created here do not have requisite permissions. Should we scrap this?
```
rehandalal: @mythmon thoughts?

peterbe:
The HTTP POST fails because of CSRF, not because of authz.

E.g. 
```
▶ curl 'http://localhost:8000/api/v2/recipe/' -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.13; rv:62.0) Gecko/20100101 Firefox/62.0' -H 'Accept: application/json' -H 'Accept-Language: en-US,en;q=0.5' --compressed -H 'Referer: http://localhost:3000/recipe/new/' -H 'content-type: application/json' -H 'x-csrftoken: undefined' -H 'authorization: Bearer SJrw517T_W-0qEMH8bRgJQYBHkwxQ-iG' -H 'origin: http://localhost:3000' -H 'DNT: 1' -H 'Connection: keep-alive' -H 'Pragma: no-cache' -H 'Cache-Control: no-cache' --data '{"action_id":"1","name":"SOmething","identicon_seed":"v1:az11tsaydlb","extra_filter_expression":"true","arguments":{"message":"Hi"}}'
{"detail":"CSRF Failed: CSRF cookie not set."}%
```

I used the "Copy as cURL" from the Network panel when I tried to make a recipe.

peterbe:
I think we should focus on that first before ditching a middleware and switching to a django-rest-framework authentication backend solution.

peterbe:
An auth backend solution means it will respond with `set-cookie` and it will be stateful. A session cookie will be stored in Postgres. 

In the delivery-console we use `headers.append('X-CSRFToken', Cookie.get('csrftoken-20170707'));` which I don't know where it's expecting to get that cookie from. 

If we're going to use CSRF for the Normandy API we need to first fetch a CSRF token but I don't know what the point of that is. 

What we probably really ought to do is ditch any form of session auth and only use bearer tokens. 

1. Disable any auth backend (even the default one) in django rest framework
2. Delete the CSRF middleware

Basically, we're going to make ALL cookies obsolete and just use bearer tokens.

rehandalal:
> The HTTP POST fails because of CSRF, not because of authz.
> 
> E.g. 
> ```
> ▶ curl 'http://localhost:8000/api/v2/recipe/' -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.13; rv:62.0) Gecko/20100101 Firefox/62.0' -H 'Accept: application/json' -H 'Accept-Language: en-US,en;q=0.5' --compressed -H 'Referer: http://localhost:3000/recipe/new/' -H 'content-type: application/json' -H 'x-csrftoken: undefined' -H 'authorization: Bearer SJrw517T_W-0qEMH8bRgJQYBHkwxQ-iG' -H 'origin: http://localhost:3000' -H 'DNT: 1' -H 'Connection: keep-alive' -H 'Pragma: no-cache' -H 'Cache-Control: no-cache' --data '{"action_id":"1","name":"SOmething","identicon_seed":"v1:az11tsaydlb","extra_filter_expression":"true","arguments":{"message":"Hi"}}'
> {"detail":"CSRF Failed: CSRF cookie not set."}%
> ```
> 
> I used the "Copy as cURL" from the Network panel when I tried to make a recipe. 

I thought that might be the case but if you run the original failing test it is failing because the user that is being created by the middleware does not have appropriate permissions.

rehandalal:
> An auth backend solution means it will respond with set-cookie and it will be stateful. A session cookie will be stored in Postgres.

I shouldn't have used the phrase "Auth backend" this is a Django Rest Framework authentication class and it does not create a session. At first I was running into a weird thing where it seemed like DRF was not getting the user info on POST so I switched away from the middleware. But I kept that change because I actually think that it is a better way to authenticate API requests than using a middleware.

normandy/base/api/authentication.py
```diff
@@ -0,0 +1,90 @@
...
+        user_profile = self.fetch_oidc_user_profile(access_token)
+
+        email = user_profile.get('email', '').strip().lower()
+        if not email:
+            # This would happen if someone has requested an access token
+            # from their OIDC provider *without the 'email' scope*.
+            raise exceptions.AuthenticationFailed("User profile lacks 'email' scope.")
+
+        # Turn this email into a Django User instance.
+        # TODO: Users created here do not have requisite permissions. Should we scrap this?
```
mythmon: To respond to a now-diseased comment, the current process is that permissions are granted manually by a super user (usually me). This is only done after the authentication method makes the user, so `get_or_create` is a good pattern to have here.

Later, I'd like to automatically derive permissions/groups based on LDAP groups or Doorman. I think we should do that later though, and leave this code as is.

normandy/base/api/authentication.py
```diff
@@ -0,0 +1,90 @@
...
+            email=email, defaults={'username': email[:150]})
+
+        # Sync user data with OIDC profile
+        family_name = user_profile.get('family_name', '').strip()
+        given_name = user_profile.get('given_name', '').strip()
+        if family_name or given_name:
+            if family_name != user.last_name:
+                user.last_name = family_name
+            if given_name != user.first_name:
+                user.first_name = given_name
```
mythmon: Do the `if A != B` checks here actually have any effect? Updating `user` with the same value doesn't have an affect, as far as I know.

peterbe: Ideally it should be 
```python
if family_name != user.last_name or given_name != user.first_name:
    user.last_name = family_name
    user.first_name = given_name
    user.save()
```

Otherwise you run risk of calling `.save()` every time `if family_name or given_name:` is true even if there's no need to change anything.

peterbe:
> the user that is being created by the middleware does not have appropriate permissions.

That's interesting that we didn't notice that! 
In Tecken, where I did similar middleware solution I deliberately overrode the [`user.has_perm`](https://github.com/mozilla-services/tecken/blob/9263992e3dd80894a8f01c6268eff0d712cd7b94/tecken/tokens/middleware.py#L55) because the list of permissions shouldn't be tied to the user, it should be tied to the token that was used since you can make a token with many-to-many permissions to make the tokens less dangerous. 

I really think we can solve this similarly but we don't have a token to get permissions from, we have the permissions in the ORM based on the user. 

It's very possible that the reason the permissions didn't work was because of some sort of lazy-loading optimization or something in Django. 

I'll try to take your failing test and see if I can resolve it based on the middleware solution.

mythmon:
> In Tecken, where I did similar middleware solution I deliberately overrode the user.has_perm because the list of permissions shouldn't be tied to the user, it should be tied to the token that was used since you can make a token with many-to-many permissions to make the tokens less dangerous.
>
> I really think we can solve this similarly but we don't have a token to get permissions from, we have the permissions in the ORM based on the user.

Please don't do this. I'd really like to get delivery console unblocked as soon as possible, because it is have a large impact on Normandy features. The current permissions model is suboptimal, yes. It requires manual work for each user, and isn't responsive to changes in LDAP, but *it works*. This is not a critical failure, considering it is the third in a line of defense-in-depth checks for accessing Normandy.

I would very strongly like to avoid feature creep on PR, and get *signing in* working. We can revisit permissions and group management once we actually have anything running in prod and have unblocked frontend work for Normandy features.

rehandalal:
this is ready for review

normandy/base/tests/api/test_authentication.py
```diff
@@ -0,0 +1,159 @@
...
+    """Bearer token authentication"""
+
+    @pytest.fixture
+    def csrf_api_client(self):
+        return APIClient(enforce_csrf_checks=True)
+
+    @pytest.fixture
+    def mock_oidc(self, settings, requestsmock):
+        def fn(user=None, **kwargs):
+            if user:  # If content kwarg is not empty this user will be ignored
```
mythmon: I don't really understand this comment. Is it still correct?

normandy/base/api/authentication.py
```diff
@@ -0,0 +1,97 @@
...
+        family_name = user_profile.get('family_name', '').strip()
+        given_name = user_profile.get('given_name', '').strip()
+        if family_name or given_name:
+            if family_name != user.last_name or given_name != user.first_name:
+                user.last_name = family_name
+                user.first_name = given_name
+                user.save()
+
+        if not user.is_active:
+            raise exceptions.AuthenticationFailed('User inactive or deleted.')
```
mythmon: I expect we'll never use this, and will probably in fact replace this with a check to auth0/Doorman. I guess it's probably a good thing to have until then though.

normandy/base/api/authentication.py
```diff
@@ -0,0 +1,97 @@
...
+
+        return (user, access_token)
+
+    def authenticate_header(self, request):
+        return self.keyword
```
mythmon: I might be good to include a comment or docstring explaining this function.

normandy/base/tests/api/test_authentication.py
```diff
@@ -0,0 +1,159 @@
...
+urlpatterns = [
+    url(
+        r'^bearer/$',
+        MockView.as_view(authentication_classes=[BearerTokenAuthentication])
+    ),
+]
+
+
+@pytest.mark.django_db
+@pytest.mark.urls('normandy.base.tests.api.test_authentication')
```
mythmon: Huh. I've never seen this before. That's really useful

normandy/base/tests/api/test_authentication.py
rehandalal: It's not still correct. I'll get rid of that.

mythmon:
Looks good to me. I'm not sure if you wanted to wait for Peter or not, so I'm not going to merge it. Feel free to merge this if you don't need Peter's review.

bors delegate= @rehandalal

rehandalal:
bors r=@mythmon, @peterbe

PR #1401, opened 2018-06-08
Bug  1467164 - Correctly check arguments for new preference experiments

mythmon:
Without this check, it isn't possible to enable a revision that only has one revision.


mythmon:
I had an extra line left over from a merge, and also some weird logout bug. I rebased on master and fixed the extra line. This should work now.

PR #1405, opened 2018-06-14
Ignore .node-version

rehandalal:
tiny r?


rehandalal:
oh yeah i guess i could use excludes. it's generated by nodenv.

rehandalal:
BAI FELICIA!

PR #1411, opened 2018-06-22
Only sign enabled recipes, and include less data in the signed payload

### LINKED ISSUE #758: Only sign approved recipes
Only approved recipes should be capable of being signed. Trying to sign a recipe that is not approved should raise an exception.

mythmon:
I did some updating of our signing code.

I reduced the number of fields in the data that gets signed for recipes further, to be only exactly what Normandy client needs. This should make signatures need to be regenerated less often (though I didn't change how often we actually regenerate them.)

I also made it so we don't sign a recipe unless it is enabled. Since we shouldn't be running recipes that aren't enabled, this shouldn't hurt anything, and increase security.

Finally, I removed the bulk sign methods implemented on Recipe and Action querysets. They were a performance optimization, but maintaining two signing flows didn't seem worth it.

Fixes #758.

### DIFF
--- a/normandy/recipes/models.py
+++ b/normandy/recipes/models.py
@@ -227,6 +227,10 @@ class Recipe(DirtyFieldsMixin, models.Model):
             self.signature = None
             return
 
+        # Don't sign recipe that aren't enabled
+        if not self.enabled:
+            return
+
         logger.info(
             f'Requesting signature for recipe with id {self.id} from Autograph',
             extra={'code': INFO_REQUESTING_RECIPE_SIGNATURES, 'recipe_ids': [self.id]}
--- a/normandy/recipes/models.py
+++ b/normandy/recipes/models.py
@@ -80,37 +80,6 @@ class RecipeQuerySet(models.QuerySet):
     def only_disabled(self):
         return self.exclude(approved_revision__enabled_state__enabled=True)
 
-    @transaction.atomic
-    def update_signatures(self):
-        """
-        Update the signatures on all Recipes in the queryset.
-        """
-        # Convert to a list because order must be preserved
-        recipes = list(self)
-
-        try:
-            autographer = Autographer()
-        except ImproperlyConfigured:
-            for recipe in recipes:
-                recipe.signature = None
-                recipe.save()
-            return
-
-        recipe_ids = [r.id for r in recipes]
-        logger.info(
-            f'Requesting signatures for recipes with ids [{recipe_ids}] from Autograph',
-            extra={'code': INFO_REQUESTING_RECIPE_SIGNATURES, 'recipe_ids': recipe_ids}
-        )
-
-        canonical_jsons = [r.canonical_json() for r in recipes]
-        signatures_data = autographer.sign_data(canonical_jsons)
-
-        for recipe, sig_data in zip(recipes, signatures_data):
-            signature = Signature(**sig_data)
-            signature.save()
-            recipe.signature = signature
-            recipe.save()
-
 
 class Recipe(DirtyFieldsMixin, models.Model):
     """A set of actions to be fetched and executed by users."""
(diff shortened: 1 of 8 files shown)


normandy/recipes/models.py
```diff
@@ -227,6 +196,10 @@ def update_signature(self):
             self.signature = None
             return
```
peterbe: Why not a `.save()` here? It used to be a .save here when it was do in `update_signatures()`.

normandy/recipes/tests/test_models.py
```diff
@@ -420,6 +410,36 @@ def test_signatures_update_correctly_on_enable(self, mocked_autograph):
...
+
+        sign_data_mock.side_effect = Exception("Can't sign yet")
+        recipe = RecipeFactory(name='unchanged', action=action)
+        assert not recipe.enabled
+        assert not recipe.is_approved
+        assert recipe.signature is None
+
+        # Updating does not generate a signature
+        recipe.revise(name='changed')
+        assert recipe.signature is None
```
peterbe: The above comment about the lack of `.save()` should get caught here if you do `recipe.refresh_from_db()`, no?

normandy/recipes/models.py
```diff
@@ -227,6 +196,10 @@ def update_signature(self):
...
 
+        # Don't sign recipe that aren't enabled
+        if not self.enabled:
+            return
```
peterbe: This should probably raise an error instead. I checked how the management command works and it only ever uses `Recipe.objects.only_approved()` so the management command should never get here. 

However, I'm guessing it *could* happen if you use the API to sign a recipe individually and manually and that should trigger some sort of error. But can you? Can you sign an individual recipe with the API?? If so, it should be explicit that the request was "bad".

mythmon:
For some background, here's the cases where a Recipe can get signed currently (I'm going to ignore Actions for now):

* During deploys, we run `./manage.py update_signatures --force`, resigning all signed recipes, and unsigning recipes that shouldn't be signed.
* On a cron job, once a day, we run `./manage.py update_signatures`, which resigns any recipes whose signature is about to expire, and unsigns any that shouldn't be signed.
* In `Recipe.save`, we call update_signature *every time a recipe is saved*.

That last part is important, and I think is the thing that you missed since I didn't touch it in this commit. That's why `Recipe.update_signature` doesn't call `.save()` on the model: it is already being saved.

normandy/recipes/models.py
mythmon: As I mentioned in my other comment, `Recipe.update_signature` is called during `Recipe.save`. `RecipeQuerySet.update_signatures` was not.

normandy/recipes/tests/test_models.py
mythmon: This is a good point. I'll add some extra checks to these tests to make sure that changes are being persisted correctly.

normandy/recipes/models.py
mythmon: I think eventually raising an error is a good idea, but we aren't quite there yet. I want to see how this shakes out a little before changing the signing logic more. After that, we could make this raise instead of return early.

Also, recipes can be approved but not enabled, so there are plenty of times that an approved recipe should not be signed.

Finally, the API does not expose signing directly. It is done implicitly. Previously basically everything was signed, though after this patch, (hopefully) only recipes that should be going to users will be signed.

normandy/recipes/models.py
peterbe: Now I see where the save happens, it happens here: https://github.com/mythmon/normandy/blob/0db4bd58b1303bcc2f8be6a45652b0029ec7cb3e/normandy/recipes/management/commands/update_recipe_signatures.py#L39

mythmon:
bors r= @peterbe

PR #1415, opened 2018-07-03
Add a deprecation alert message to UI

rehandalal:
We should hold off on merging this but I thought I'd file it now so we're ready.

### DIFF
--- a/client/control/components/common/DeprecationAlert.js
+++ b/client/control/components/common/DeprecationAlert.js
@@ -1,0 +1,86 @@
+import { Alert } from 'antd';
+import PropTypes from 'prop-types';
+import React from 'react';
+import { connect } from 'react-redux';
+
+const ENV_PROD = 'ENV_PROD';
+const ENV_STAGE = 'ENV_STAGE';
+const ENV_DEV = 'ENV_DEV';
+
+const DC_URLS = {
+  [ENV_PROD]: 'https://delivery-console.prod.mozaws.net/',
+  [ENV_STAGE]: 'https://delivery-console.stage.mozaws.net/',
+  [ENV_DEV]: 'https://delivery-console.dev.mozaws.net/',
+};
+
+/**
+ * Simple component which tells the user whether they are viewing a dev or staging
+ * environment. On production, nothing is shown.
+ */
+@connect(() => ({
+  currentUrl: window.location.href,
+}))
+export default class EnvAlert extends React.PureComponent {
+  static propTypes = {
+    currentUrl: PropTypes.string.isRequired,
+  };
+
+  static productionFragments = ['normandy-admin.prod.'];
+  static stageFragments = ['normandy-admin.stage.'];
+
+  /**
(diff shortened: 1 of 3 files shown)


client/control/components/common/DeprecationAlert.js
```diff
@@ -0,0 +1,86 @@
...
+
+const DC_URLS = {
+  [ENV_PROD]: 'https://delivery-console.prod.mozaws.net/',
+  [ENV_STAGE]: 'https://delivery-console.stage.mozaws.net/',
+  [ENV_DEV]: 'https://delivery-console.dev.mozaws.net/',
+};
+
+/**
+ * Simple component which tells the user whether they are viewing a dev or staging
+ * environment. On production, nothing is shown.
```
mythmon: Comment is out of date.

client/control/components/common/DeprecationAlert.js
```diff
@@ -0,0 +1,86 @@
...
+ */
+@connect(() => ({
+  currentUrl: window.location.href,
```
mythmon: Isn't there a selector that does path information? Or is that only in DC?

Also, if this is the only thing being passed it, I don't think it needs to be a connected component.

client/control/components/common/DeprecationAlert.js
```diff
@@ -0,0 +1,86 @@
...
+  /**
+   * Given a URL and an array of strings, determines if that URL contains at least
+   * one of the strings.
+   *
+   * @param  {String}        url       URL to search over.
+   * @param  {Array<String>} fragments Collection of strings to find in the URL.
+   * @return {Boolean}                 True if URL contains at least one fragment.
+   */
+  static findFragmentsInURL(url, fragments) {
+    return !!fragments.find(piece => url.indexOf(piece) > -1);
```
mythmon: The world is good, instead of `.indexOf() > -1`, we can now do `url.includes(piece)`.

client/control/components/common/DeprecationAlert.js
```diff
@@ -0,0 +1,86 @@
...
+};
+
+/**
+ * Simple component which tells the user whether they are viewing a dev or staging
+ * environment. On production, nothing is shown.
+ */
+@connect(() => ({
+  currentUrl: window.location.href,
+}))
+export default class EnvAlert extends React.PureComponent {
```
mythmon: This is still called `EnvAlert`, and should probably be renamed.

rehandalal:
> Do we have a milestone or list of issues that we need to fix before deprecating this?

We do not have a list of issues however I am going to sending out an email to S&I asking them to do a few days of testing on DC. I suspect we will get a bunch of issues out of that and will have a better timeline for deprecation.

mythmon:
I think we're ready for this.

rehandalal:
bors 🙈

mythmon:
gah! sorry. I should figure out how to not allow that.

PR #1417, opened 2018-07-03
Cache the OIDC profile

### LINKED ISSUE #1416: Hitting Auth0's rate limits when verifying the bearer token
We sometimes hit [Auth0's rate limits](https://auth0.com/docs/policies/rate-limits) when making lots of requests to the API using bearer token authentication. This is because we verify the token on every request. We should implement some kind of caching to prevent this.

rehandalal:
Fixes #1416.

r?

### DIFF
--- a/normandy/base/api/authentication.py
+++ b/normandy/base/api/authentication.py
@@ -1,5 +1,6 @@
 from django.conf import settings
 from django.contrib.auth import get_user_model
+from django.core.cache import cache
 
 import backoff
 import requests
@@ -81,11 +82,19 @@ class BearerTokenAuthentication(BaseAuthentication):
 
     @backoff.on_exception(backoff.constant, requests.exceptions.RequestException, max_tries=5)
     def fetch_oidc_user_profile(self, access_token):
+        cache_key = f"oidc-profile-{access_token}"
+        cached_response = cache.get(cache_key)
+
+        if cached_response:
+            return cached_response
+
         url = settings.OIDC_USER_ENDPOINT
         response = requests.get(url, headers={"Authorization": f"Bearer {access_token}"})
 
         if response.status_code == 200:
-            return response.json()
+            profile = response.json()
+            cache.set(cache_key, cache_key, 15)
+            return profile
         elif response.status_code == 401:
             # The OIDC provider did not like the access token.
             raise exceptions.AuthenticationFailed("Unauthorized access token")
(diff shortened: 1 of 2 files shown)


rehandalal:
There are a couple things I need to fix up.

normandy/base/api/authentication.py
```diff
@@ -81,11 +82,19 @@ def authenticate_header(self, request):
...
+        if cached_response:
+            return cached_response
+
         url = settings.OIDC_USER_ENDPOINT
         response = requests.get(url, headers={"Authorization": f"Bearer {access_token}"})
 
         if response.status_code == 200:
-            return response.json()
+            profile = response.json()
+            cache.set(cache_key, cache_key, 15)
```
mythmon: I just commented on this in the issue, but I think we shouldn't use a flat 15 second cache time. Instead I think we should read the rate limit information provided by the response (in the headers `X-RateLimitRemaining` and `X-RateLimit-Reset`), and adjust the cache based on that.

normandy/base/api/authentication.py
```diff
@@ -81,11 +82,19 @@ def authenticate_header(self, request):
 
     @backoff.on_exception(backoff.constant, requests.exceptions.RequestException, max_tries=5)
     def fetch_oidc_user_profile(self, access_token):
+        cache_key = f"oidc-profile-{access_token}"
```
mythmon: Is it safe to use the access token as the cache key? I'm not sure what else we would use though. Maybe only a portion of it, like the first half of the token (16 chars)?

rehandalal: or i can just hash it.

normandy/base/api/authentication.py
```diff
@@ -92,8 +98,12 @@ def fetch_oidc_user_profile(self, access_token):
         response = requests.get(url, headers={"Authorization": f"Bearer {access_token}"})
 
         if response.status_code == 200:
+            now = int(time.mktime(datetime.utcnow().timetuple()))
+            ratelimit_expires_in = int(response.headers.get("X-RateLimit-Reset", 0)) - now
```
mythmon: A default of 0 is probably going to do weird things, like set the cache to expire in negative 50 years. Maybe the default should be now Or now + 60?

rehandalal:
bors r=@mythmon

PR #1418, opened 2018-07-05
update docs on pip install

peterbe:


docs/dev/install.rst
```diff
@@ -50,7 +50,7 @@ Installation
 
    .. code-block:: bash
 
-      pip install -r requirements/default.txt
+      pip install -r requirements/default.txt -c requirements/constraints.txt
```
mythmon: This isn't needed, since the first line in `requirements/default.txt` is `-c constraints.txt`. That causes pip to use the constraints file with less typing.

docs/dev/install.rst
```diff
@@ -193,4 +193,4 @@ code. This will save time when submitting pull requests.
 If you ever need to bypass Therapist, you can do so by passing
 ``--no-verify`` to your ``git commit`` command.
 
-.. _Therapist: http://therapist.readthedocs.io/en/latest/overview.html
\ No newline at end of file
+.. _Therapist: http://therapist.readthedocs.io/en/latest/overview.html
```
mythmon: Git tells me there is a change here, but I can't tell what it is. Did you change anything here?

docs/dev/install.rst
peterbe: Neat! I totally missed that.

docs/dev/install.rst
peterbe: I suspect my editor makes sure the file has a last final linebreak.

PR #1435, opened 2018-08-06
CircleCI 2.0

### LINKED ISSUE #1196: Transition Normandy to CircleCI 2.0
CircleCI [has announced](https://circleci.com/blog/sunsetting-1-0/) they are going to stop supporting CircleCI 1.0 on August 31st. Normandy currently uses 1.0. We'll need to migrate to 2.0 before the end of August.

peterbe:
Fixes #1196 

- [x] Make it work
- ~~Move sections into dedicated bash `.sh` files~~ Actually, I like having it all in one `run.sh` file.
- [x] Remove old circle stuff
- [x] Update docs to explain how to run what circle does on your laptop
- [x] Push to dockerhub script restored


peterbe:
So much research and so little success. Basically, I can't think of any way to get files *out* of the container after it has run (with `docker-compose`). 

I posted https://discuss.circleci.com/t/how-can-i-extract-the-junit-xml-files-from-within-a-docker-container-in-docker/24089 in extreme optimism. 

Unless some magical hack appears, I think we have to be prepared to miss out of junit in CircleCI 2.0.

Dockerfile
```diff
@@ -25,9 +25,10 @@ RUN pip install --upgrade --no-cache-dir -r requirements/pip.txt && \
 COPY . /app
 RUN NODE_ENV=production yarn build && \
     DJANGO_CONFIGURATION=Build ./manage.py collectstatic --no-input && \
-    mkdir -p media && chown app:app media && \
-    mkdir -p /test_artifacts && \
-    chmod 777 /test_artifacts
+    mkdir -p media && chown app:app media
+
+# RUN chown -R 10001:10001 /app
```
mythmon: Should probably remove this commented out line, since it is duplicated below.

docs/dev/ci.rst
```diff
@@ -0,0 +1,42 @@
...
+
+The steps in ``.circleci/config.yml`` should be sufficiently self-explanatory. Each step should
+work to just copy and paste into your own terminal.
+
+There are some exceptions though. For example, when a step is set up to run with
+``background: true`` it means you can run it in its own dedicated terminal and run the
+other commands in another terminal.
+
+Also, steps with ``when: always`` means they can be run after any other step. In particular
+we use this to extract content from containers that have finished successfully or failed.
```
mythmon: I think "they can be run after any other step" is confusing. It makes me think that the step in question is somehow run multiple times after every other step, or something else. I think that the comment in the CircleCI config next to `when: always` is sufficient, and we don't need to mention it here.

Dockerfile
```diff
@@ -37,6 +38,8 @@ ENV DJANGO_SETTINGS_MODULE=normandy.settings \
     NEW_RELIC_CONFIG_FILE=newrelic.ini
 EXPOSE $PORT
 
+ENTRYPOINT ["/bin/bash", "/app/ci/run.sh"]
+
 CMD $CMD_PREFIX \
```
mythmon: I'm a little nervous about having both `ENTRYPOINT` and `CMD`, but after reading some docs and looking closely at `run.sh`, I think it will be fine. It might be nice to have a comment here or in `run.sh`'s `*)` case explaining whats going on. I'm also not 100% sure the `$CMD_PREFIX` stuff will work.

@sciurus can you remind me what we're using `$CMD_PREFIX` for? and do you think that this along with the `*)` case in `run.sh` will work out fine with it?

sciurus: This is a little convoluted but it should work.

everything is started as an argument to run.sh. For a few arguments (e.g. `lint`) it has a predefined set of commands to run. Anything it doesn't recognize it will just pass through to the shell.

CMD_PREFIX is undefined except for on the APM node. There it is defined as new relic's wrapper script.

mythmon: I'm mostly concerned about the situation in which `CMD_PREFIX` is not undefined. Will `exec "$@"` work fine in that case? I guess it depends on what `CMD_PREFIX` is set to?

mythmon: PS: I'd still like to see a comment, probably in both the `Dockerfile` and `run.sh` about what's going on. It may also make sense to move run.sh out of `ci/`. Maybe keeping it all in `bin/` is a better idea?

sciurus: Regarding when CMD_PREFIX is defined, sure, it should work if that's a valid command. In our case it would be the difference between `exec gunicorn` and `exec newrelic-admin run-program gunicorn`

sciurus: @peterbe the dockerfile is just started via a systemd unit file which declares

```
ExecStart=/usr/bin/docker run \
        --net host -m 0b --env-file /etc/dockerflow/app.txt -v /usr/share/GeoIP/GeoIP2-Country.mmdb:/app/GeoIP2-Country.mmdb --log-driver=none  \
        --name app \
        hub.prod.mozaws.net/pipelines/normandy:v97-422 \
```

for the current version. So you can see it runs the default `CMD`.

I am thinking in a similar vein to you. It would be simpler if anything we anticipated running was defined in `run.sh`, and `CMD` just specified one of those to make it the default. In `run.sh` we could have

``
web)
$CMD_PREFIX \
    gunicorn  \
    --log-file - \
    --worker-class ${GUNICORN_WORKER_CLASS:-sync} \
    --max-requests ${GUNICORN_MAX_REQUESTS:-0} \
normandy.wsgi:application
;;
```

and in the Dockerfile `CMD` could change to

`CMD web`

We could bikeshed the name `web`.

Comment (peterbe) [reply Dockerfile]: There's another subtle refactoring. The `Dockerfile` does `yarn build` and `manage.py collectstatic` also. Which is repeated in `ci/run.sh`. 

I was in love with the idea that the `ci/` directory was "self contained" and was only used for CI and if you one day decide to not run CI anymore we could just delete this directory and nothing would change on the side of running things. 

Perhaps we `git mv ci/run.sh bin/run.sh` (and fix other relative paths accordingly) and make that `run.sh` the home for all bash logic such as `collectstatic`, karma tests, pytests etc. 

We *could* add a couple of entries inside that `bin/run.sh` that calls `ci/run.sh` instead. 

In conclusion; 
* less bash in `Dockerfile`, move it to `run.sh`
* use the current `ci/run.sh` for other non-CI related things such as "web"
* change the `CMD` line to be just `CMD web`.

By the way, this is pretty much how [Tecken does it](https://github.com/mozilla-services/tecken/blob/786b1da12a8163d66a6e86eb6a258661b29b34b1/Dockerfile#L72-L74).

Comment (peterbe) [reply Dockerfile]: So @mythmon @sciurus Do you like that plan? I.e. remove the extra bash from the `Dockerfile` and move it into `bin/run.sh` and edit `bin/run.sh` so that it can execute the other `docker-compose` commands we run in CI.

Comment (sciurus) [reply Dockerfile]: 👍

Comment (mythmon) [reply Dockerfile]: I think this sounds good.

**Comment by peterbe** [comment]
@mythmon refactored. Ready for review again.

File: bin/run.sh [hunk 4]
```diff
@@ -51,13 +51,18 @@ case $1 in
...
-  start)
-    NODE_ENV=production yarn build
-    DJANGO_CONFIGURATION=Build ./manage.py collectstatic --no-input
+  first-start)
+    echo "Starting the gunicorn server the first time"
     ./manage.py migrate
     ./manage.py update_actions
-    ./bin/download_geolite2.sh
-    DJANGO_CONFIGURATION=ProductionInsecure gunicorn  \
+    $CMD_PREFIX gunicorn \
```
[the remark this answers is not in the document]
peterbe: A bash function is probably best. Will do.

ci/docker-compose.yml
```diff
@@ -38,4 +38,4 @@ services:
       "
 
 volumes:
-  test-artifacts:
\ No newline at end of file
+  ? test-artifacts
```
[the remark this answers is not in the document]
peterbe: uh wat?! I didn't write that.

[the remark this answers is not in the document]
mythmon: Maybe it came from a YAML autoformatter?

mythmon:
@sciurus Can you review this and approve/request changes? After that I think we're ready to go.

mythmon:
@peterbe thanks for the work. I'm really excited to have a better foundation to build our CI on.

peterbe:
@sciurus I'm going to land this if you approve.

sciurus:
Sorry I missed that this was ready for review again! Looking now.

Dockerfile
```diff
@@ -25,9 +25,9 @@ RUN pip install --upgrade --no-cache-dir -r requirements/pip.txt && \
 COPY . /app
 RUN NODE_ENV=production yarn build && \
     DJANGO_CONFIGURATION=Build ./manage.py collectstatic --no-input && \
-    mkdir -p media && chown app:app media && \
-    mkdir -p /test_artifacts && \
-    chmod 777 /test_artifacts
+    mkdir -p media && chown app:app media
```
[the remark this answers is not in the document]
mythmon: Media is the place where the webapp can write files, especially uploaded files. This doesn't matter in our prod builds, since they go to S3, but it is important for other uses of this docker file.

peterbe:
@sciurus Fixed the unnecessary `chown`.

mythmon:
Alright. Lets do this.

bors r= @mythmon, @sciurus

PR #1437, opened 2018-08-06
Upgrade Dockerfile from Python 3.6.0 to 3.6.6

### LINKED ISSUE #1434: Use a more recent 3.6 for Docker
We're currently using [3.6.0](https://github.com/mozilla/normandy/blob/148ba346fea32afa9107e8f7e8f91553e89e5ebe/Dockerfile#L1) and although I think it's up-to-date on security fixes it might be good to use a more recent version of 3.6. At the time of writing 3.6.6 is available but we could just use 3.6 and use Renovate to track a specific sha.

mythmon:
Fixes #1434

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.6.0-slim
+FROM python:3.6.6-slim
 WORKDIR /app
 RUN groupadd --gid 1001 app && useradd -g app --uid 1001 --shell /usr/sbin/nologin app
 RUN apt-get update && \


peterbe:
Apparently not as easy as we thought :)

I had the same problem with Tecken. Technically, [similar problem](https://github.com/mozilla-services/tecken/pull/928/files) where lots of Debian-esque stuff broke between versions. 

If you want to, we can tackle this after the CircleCI upgrade since the new CircleCI stuff will be faster to iterate on.

mythmon:
Drat, I hoped it would be just that easy. Lets leave this until after your CI work is done.

PR #1442, opened 2018-08-10
Do not use X-Forwarded-Host

sciurus:
We started stripping this header at Nginx due to https://bugzilla.mozilla.org/show_bug.cgi?id=1433134

There is no reason to keep it enabled in the normandy codebase.

### DIFF
--- a/normandy/settings.py
+++ b/normandy/settings.py
@@ -399,7 +399,6 @@ class Development(Base):
 class Production(Base):
     """Settings for the production environment."""
 
-    USE_X_FORWARDED_HOST = values.BooleanValue(False)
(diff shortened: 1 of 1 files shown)


normandy/settings.py
```diff
@@ -399,7 +399,7 @@ class Development(Base):
 class Production(Base):
     """Settings for the production environment."""
 
-    USE_X_FORWARDED_HOST = values.BooleanValue(True)
+    USE_X_FORWARDED_HOST = values.BooleanValue(False)
```
mythmon: `False` is the default in the `Base` config. Lets just delete the line instead of duplicating it here.

PR #1451, opened 2018-08-15
Use JSON array format for Dockerfile CMD

mythmon:
Besides being a generally good idea, this is required when using both CMD and ENTRYPOINT.


mythmon:
Ok, one more time.

PR #1452, opened 2018-08-15
resilience against rounding error in timestamp in test_caching fixture

### LINKED ISSUE #1423: Caching test for bearer token authentication intermittently fails
The test:
`normandy.base.tests.api.test_authentication.TestBearerTokenAuthentication.test_caching`

We currently use `sleep` to add a delay that is 1 second longer than the time in which the cache should expire but this appears to intermittently cause issues and can hopefully be done in a better way.

peterbe:
Fixes #1423 

Hopefully my comment is clear enough. A way to "visualize" it is to create [a script like this](https://gist.github.com/peterbe/699e6826a82717e896a8ddb9b0a53fde).

### DIFF
--- a/normandy/base/tests/api/test_authentication.py
+++ b/normandy/base/tests/api/test_authentication.py
@@ -83,8 +83,22 @@ class TestBearerTokenAuthentication(object):
             {"email": user.email, "given_name": user.first_name, "family_name": user.last_name}
         ).encode("utf-8")
 
-        ratelimit_reset = int(time.mktime(datetime.utcnow().timetuple())) + 1
-
+        # The reason for the `+ 2` here is that if we only add +1 second, there's a small
+        # chance that the `time.mktime(utcnow)` done here is *different* from the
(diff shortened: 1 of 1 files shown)


peterbe:
I've actually never used it but I've reviewed some code that uses [freezegun](https://github.com/spulec/freezegun). We used it in socorro to mock dates but I don't know if it can be used in this case. Would be cool. It might resolve the icky problem that a single unit tests 2+ seconds to run when it could be done in 0.02 seconds in theory.

peterbe:
Perhaps a follow-up issue to use freezegun. Do we have a lot of other unit tests that could benefit from it?

peterbe:
ping @rehandalal

peterbe:
Judging by ripgrep, we only have 1 single instance of using `time.sleep`. Just park this idea of using freezegun somewhere in the back.

rehandalal:
> Just park this idea of using freezegun somewhere in the back.

Yeah it's not high priority but getting the extra seconds back would be nice.

PR #1457, opened 2018-08-20
python:3.6-slim with hash

### LINKED ISSUE #1434: Use a more recent 3.6 for Docker

peterbe:
Fixes #1434 

I don't know why Renovate hasn't suggested something like this yet. Perhaps it's in our renovate config. Either way, I see no risk of using 3.6 instead of 3.6.6. Especially since we use a hash.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,10 @@
-FROM python:3.6.0-slim@sha256:ed16e4e6ee97ad893c5be1a22aae0de95b996b0a21e544d9b8a4625f66d1d49e
+FROM python:3.6-slim@sha256:5a96684a1729acd0680b39a3c24ef33f36bb6951873f77cde1e227b059a0f881
 WORKDIR /app
 RUN groupadd --gid 1001 app && useradd -g app --uid 1001 --shell /usr/sbin/nologin app
 RUN apt-get update && \
     apt-get upgrade -y && \
     apt-get install -y --no-install-recommends \
-    gcc libpq-dev curl apt-transport-https libffi-dev openssh-client
+    gcc libpq-dev curl apt-transport-https libffi-dev openssh-client gnupg
 
 # Install node from NodeSource
 RUN curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add - && \


Dockerfile
```diff
@@ -1,10 +1,10 @@
...
+FROM python:3.6-slim@sha256:5a96684a1729acd0680b39a3c24ef33f36bb6951873f77cde1e227b059a0f881
 WORKDIR /app
 RUN groupadd --gid 1001 app && useradd -g app --uid 1001 --shell /usr/sbin/nologin app
 RUN apt-get update && \
     apt-get upgrade -y && \
     apt-get install -y --no-install-recommends \
-    gcc libpq-dev curl apt-transport-https libffi-dev openssh-client
+    gcc libpq-dev curl apt-transport-https libffi-dev openssh-client gnupg
 
 # Install node from NodeSource
```
peterbe: Are the steps below really the "best" way to do this? It doesn't look much like the various Dockerfiles I found for [node 8 for stretch](https://github.com/nodejs/docker-node/blob/master/8/stretch/Dockerfile).

peterbe: For example, when you build the image you get warning messages like this:
```
Warning: apt-key output should not be parsed (stdout is not a terminal)
```

peterbe: It makes sense, but it's still sadface to have to look at warnings.

mythmon: We aren't parsing the contents of apt-key. That error happens any time that the output doesn't go to an interactive terminal, and our CI setup is not interactive.

I'm sure these steps aren't the best way to do this. The best way would be to maintain a combination Python-and-Node image and pull that instead of hacking up the Python image to add Node. But it works, so I'm not inclined to change it. If you'd like to do so, go ahead.

peterbe:
I agree. Let's configure Renovate first so they give use heads-up when the Docker images change. Is there a reason we haven't set it up yet? Mind if I go ahead and request it from the mozilla github team?

mythmon:
Sure, lets go ahead and get Renovate set up.

mythmon:
I think we should close this in favor of #1464. @peterbe, what do you think?

peterbe:
> I think we should close this in favor of #1464. @peterbe, what do you think?

No, this PR is actually two things. Pinning and upgrading from python-3.6.0 to python-3.6. I can rebase this PR after #1464 has landed so the sha is correct and so that we can use 3.6.latest.

peterbe:
Conflicting files. Wanna rebase?

peterbe:
Oh heck! I spoke too soon. This is *my* PR :)

PR #1461, opened 2018-08-24
Store revision author

### LINKED ISSUE #1460: The creator of a revision is not stored
STR: This test fails

```python
def test_creating_recipes_stores_the_user(self, api_client):
    action = ActionFactory()
    res = api_client.post(
        "/api/v2/recipe/",
        {
            "name": "Test Recipe",
            "action_id": action.id,
            "arguments": {},
            "extra_filter_expression": "whatever",
        },
    )
    recipe = Recipe.objects.get()
    assert recipe.latest_revision.user is not None
```

This is not just a testing problem, none of the recent revisions on the server have authorship information. It isn't just a serialization problem, we don't have the data in the DB. It is a problem in prod.

We still get data from recipe approvals, which probably correlates well with recipe authorship. This is still really bad though, especially considering [the recently fixed authentication problem](https://bugzilla.mozilla.org/show_bug.cgi?id=1485148).

rehandalal:
Fixes #1460.

r?

### DIFF
--- a/normandy/recipes/api/v2/serializers.py
+++ b/normandy/recipes/api/v2/serializers.py
@@ -129,7 +129,7 @@ class RecipeSerializer(serializers.ModelSerializer):
 
     def update(self, instance, validated_data):
         request = self.context.get("request")
-        if "user" not in validated_data and request and request.user:
+        if request and request.user:
             validated_data["user"] = request.user
 
         instance.revise(**validated_data)
@@ -137,7 +137,7 @@ class RecipeSerializer(serializers.ModelSerializer):
 
     def create(self, validated_data):
         request = self.context.get("request")
-        if "user" not in validated_data and request and request.user:
+        if request and request.user:
             validated_data["user"] = request.user
 
         if "identicon_seed" not in validated_data:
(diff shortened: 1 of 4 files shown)


mythmon:
This generally looks good. I have two questions though:

1. Does this allow me to specify the author direclty via the API? Could I sign in as me and make a revision under your name? That probably wouldn't be good, but since it prefers user from `validated` data, maybe that could happen?

2. Is there a reason why we allow `null` users on revisions at all? Should we update the model to fix this? Of course, this causes problems with migrations...

rehandalal:
> 1. Does this allow me to specify the author direclty via the API? Could I sign in as me and make a revision under your name? That probably wouldn't be good, but since it prefers user from `validated` data, maybe that could happen?

Yes you could potentially do that if the data posted included the `user`. We could change this behavior if that's not what we want or until we find a reason to allow other values for `user`.

> 2. Is there a reason why we allow `null` users on revisions at all? Should we update the model to fix this? Of course, this causes problems with migrations...

I think it had something to do with the transition from reversion. I think they allowed `null` values and there might have been some in the revisions. We could backfill with an anonymous user and prevent future `null` values.

mythmon:
Being able to post data as another user sounds like a security problem. I don't think we should allow that. 

I think that adding an anonymous user is kind of a hack, but having a DB level guarantee this won't happen in the future would be nice. We could do this in another PR if you'd like, it isn't as urgent.

rehandalal:
> I think that adding an anonymous user is kind of a hack, but having a DB level guarantee this won't happen in the future would be nice. We could do this in another PR if you'd like, it isn't as urgent.

+1 for doing that in another PR.

PR #1469, opened 2018-08-24
Make all words required when searching

mythmon:
This should make searching a bit nicer. I didn't realize how bad "any word matches" was without a scoring step. I considered adding a scoring step, but I don't know if that is reasonable at our scale.


peterbe:
Is to correct that if you type "wordthatappears wordthatneverappears" should now return 0 recipes?

mythmon:
> Is to correct that if you type "wordthatappears wordthatneverappears" should now return 0 recipes?

Yes, that is correct. I think this will be a better behavior, short of full text search.

> However, perhaps a much better approach is to use a full text search instead. It doesn't even require an index first.

This is probably a good way forward. I don't really know much about it though, so I'm not sure how best to proceed.

mythmon:
bors r= @peterbe

PR #1505, opened 2018-08-27
Deduplicate docker-compose file config in CircleCI config

mythmon:
Using an environment variable lets us stop repeating the docker-compose config file for every line.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,11 @@ jobs:
   build_test_publish:
     docker:
       - image: ubuntu:18.04
+
     working_directory: ~/normandy
+
+    environment:
+        COMPOSE_FILE: ci/docker-compose.yml
 
     steps:
       - run:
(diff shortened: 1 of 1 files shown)


mythmon:
> line is no longer copy-n-pastable to the command line locally.

How often do you find yourself doing this? This doesn't seem very useful to me, since this is a config file and not documentation. I also don't find myself needing to run these kind of tests outside of CI very often. Usually there are easier/more efficient ways to do things for the tasks I need to do (mostly not using Docker).

peterbe:
I use virtualenv when doing local dev but when I worked on migrating to circleci 2.0 I ran those commands locally. A LOT. Also, because of the darn volume mounting problem I had to inject the volume mount thing into the `docker-compose.yml` run and test things and before `git commit && git push ...` I had to remove that volume mount stuff again. Painful. 

It's only a matter of time until we have a CI problem that only happens in docker in CI and not when running `pytest` locally in a virtulenv. Experience speaks :)

mythmon:
I definitely agree that we'll have issues that won't be reproducible locally. In my experience on Normandy, those usually happen when we make changes to CI, in which case the person working on it is generally very familiar with what's going on.

I consider problems that only show up in CI to be a bug in our testing setup. I endeavor to eliminate the underlying cause whenever it happens. They also happen infrequently enough that I'm already going to be reading docs to figure out how to recreate the environment used in CI, so having a note that says the commands in the config rely on an environment variables.

I don't think "I can't copy and paste this" is a good reason to not improve this code. It would be nice if there were a better way to reproduce Circle's without pushing to Github. I don't think that manually copying lines out of our Circle CI config is a good way to do that though. Even with Circle CI 2.0, we can't reliably reproduce their environment locally.

I'm not adamant about this particular change, but I don't agree with your concerns.

mythmon:
[bors's status page](https://app.bors.tech/repositories/1262) still shows this PR as "awaiting review". I think it may have been down when Peter issued his command. Lets try again

bors r= @peterbe

PR #1509, opened 2018-08-28
Use prebuilt browser image for JS tests

mythmon:


mythmon:
I'm not sure this is going to work, but it is worth a try.

mythmon:
It worked! I had some trouble figuring out how the networking worked between the test runner and the browser container, but I think the final solution is pretty simple, and mirrors the contract tests nicely.

The two successful runs above both ran in less than 5 minutes. The last big block is building the docker image, which I don't think we'll be able to improve much without Docker layer caching.

@peterbe what do you think about this approach? I like it, but I'm biased since it I was able to achieve my goals of using a common docker image for the browser. I'm not sure if it is better than what we had before, or if there is any way we could clean it up.

.circleci/config.yml
```diff
@@ -51,18 +53,19 @@ jobs:
...
-          name: Start Django/Gunicorn server
-          command: docker-compose up web
-          background: true
-
       - run:
           name: Contract tests
-          command: docker-compose run web contracttest
+          command: |
+            docker-compose up -d web
+            docker-compose run web contracttest
```
mythmon: I thought about making a separate Docker Compose for `contracttest` as well, so that this could have the exact same shape as the JS tests, but I'm not sure that's worth it.

bin/wait-for-it.sh
```diff
@@ -1,182 +0,0 @@
-#!/usr/bin/env bash
```
mythmon: I think this file was an older version of the script that we weren't using. All of the CI was using the copy in the `ci` directory, as far as I could tell.

.circleci/config.yml
```diff
@@ -41,7 +41,9 @@ jobs:
 
       - run:
           name: Build
-          command: docker-compose build web
+          command: |
+            docker-compose build
+            docker-compose pull db
```
peterbe: What does that do? I guess it happens automatically by `docker-compose up -d web` but I guess by doing it here you're "preparing" to have the download done so it doesn't take so damn long to start the web container thing later. 

Is that right? In fact, if it is right I think you should jot this down as a comment. At least I had a reaction of "Why is this done???"

.circleci/config.yml
```diff
@@ -51,18 +53,19 @@ jobs:
...
-          command: docker-compose up web
-          background: true
-
       - run:
           name: Contract tests
-          command: docker-compose run web contracttest
+          command: |
+            docker-compose up -d web
+            docker-compose run web contracttest
+            docker-compose kill web
```
peterbe: 👍

.circleci/config.yml
```diff
@@ -51,18 +53,19 @@ jobs:
...
 
       - run:
           name: JavaScript tests
-          command: docker-compose run --user root web js-tests
+          command: |
+            docker-compose up -d js-tests-browser
+            docker-compose up js-tests
+            docker-compose kill js-tests-browser
```
peterbe: Cool! It matches the `Contract test` step above.

bin/wait-for-it.sh
peterbe: oops! Good catch

ci/docker-compose.yml
```diff
@@ -34,8 +34,24 @@ services:
...
+      - ARTIFACTS_PATH=/artifacts
+    volumes:
+      - test-artifacts:/artifacts
+    command: js-tests
+
+  js-tests-browser:
+    build:
+      context: .
+      dockerfile: js-tests-browser.Dockerfile
+    command: ./wait-for-it.sh -t 30 js-tests:9876 -- firefox js-tests:9876 --headless
```
peterbe: Fancy! I'm impressed.

.circleci/config.yml
mythmon: You're right, this is to get all the prep work out of the way first thing, so that the test steps are "purer", and have more consistent timing. I'll add a comment.

mythmon:
bors r= @peterbe

Issue #1524, opened 2018-09-10
Remove RecipeRevision.[countries|locales|channels] manytomany fields


peterbe:
These have been migrated to `filter_object_json` where appropriate.

peterbe:
Whoaaa! This led me deep into a rabbit hole of hacking that just started tasting worse per edit. 

The `countries`, `locales`, and `channels` is deeply integrated in the code. There are lots of complex tests that depend on it being a ManyToManyField. For example, the `test_models.py` depends on doing things like this: https://github.com/mozilla/normandy/blob/d17dbe00ca32df582a732ac0cf121dc7f72313bb/normandy/recipes/tests/test_models.py#L470-L486

There's also the deeply complex migrations *and* their unit test code that tests migrating from these ManyToManyFields to filter_objects. 

Until all of these things have been refactored, carefully, then can we worry about removing the many-to-many tables.

mythmon:
Would it make sense to remove the fields from the API, so we don't get any new recipes with these fields set?

peterbe:
Rough guestimate of work...

1. Remove unit tests that mess with revising the `countries`, `locales` and `channels`. 
2. Don't make it possible to filter the API (all 3 versions, sigh) by locales, countries or channels. (@mythmon is that ok?)
3. Remove unit tests that test filtering (and serializing?) around these three fields.
4. MOre aggressive stuff. 

If this is sane, let's do it gently with one thing at a time.

peterbe:
> Would it make sense to remove the fields from the API, so we don't get any new recipes with these fields set?

Uh? So there are still some that come in an populate the database for the many-to-many tables? I.e. stuff that goes against that big data migration when filter_object_json was launched.

mythmon:
> Uh? So there are still some that come in an populate the database for the many-to-many tables? I.e. stuff that goes against that big data migration when filter_object_json was launched.

In practice, no. There shouldn't be any new data, since none of the tools that write to the API expose these fields. Sorry if I gave you false alarm. In theory it is possible though, since those fields exist.

peterbe:
I think I need to remind myself that we shall not undo the `Channels`, `Locales` and `Countries` models. Maybe we do at some point but it's out of context for this issue.

shell1:
slow many-to-many db fields that never ended up getting used that are slow.  so should just delete them

mythmon:
I took a stab at this in #1943, and I've decided it is not actually a `good-first-bug`.

PR #1527, opened 2018-09-11
remove tests that query by mcountry,channels,locales m2m fields

peterbe:
Part of #1524 

Once this lands we can remove the serialization of Channel, Locale and Countries from the `recipes/api/v*/serializers.py`.


normandy/recipes/tests/test_models.py
```diff
@@ -447,79 +405,11 @@ def test_recipe_revise_partial(self):
...
-        recipe.revise(name="changed", action=a2, channels=[c])
+        recipe.revise(name="changed", action=a2)
         assert recipe.action == a2
         assert recipe.name == "changed"
         assert recipe.arguments == {"message": "something"}
-        assert recipe.filter_expression == (
-            "(normandy.channel in ['beta']) && " "(something !== undefined)"
-        )
-
-    def test_recipe_doesnt_revise_when_clean(self):
```
mythmon: I think this test is still valid, just without the channels part.

peterbe: Restored, fixed and rebased (new commit message)

mythmon: Maybe there was a different typo that I didn't see, but I think "mcountry" in the commit message isn't right.

peterbe:
Now with a separate commit to remove the serialization.

normandy/recipes/tests/test_models.py
peterbe: There. Fixed now.

mythmon:
:red_circle: :red_circle: :red_circle: 

Thanks for taking this all out, I'm excited about the rest of it too.

peterbe:
bors retry

PR #1536, opened 2018-09-14
1535 test coverage holes in api

peterbe:
To test this I ran:

```bash
pip install coverage
coverage run -m pytest normandy && coverage html --skip-covered
open htmlcov/index.html
```
I also had to create a `.coveragerc` which I don't know if we can add to the repo since we don't do coverage testing regularly. But I had to create this:

`.coveragerc`:
```
[run]
source = normandy
branch = True
data_file = /tmp/.coverage

[report]
omit =
    tests/*
    normandy/settings.py
    normandy/wsgi.py
    *management*
    *management/commands*

[xml]
output = /tmp/coverage.xml
```

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,13 @@
+[run]
+source = normandy
+branch = True
+data_file = /tmp/.coverage
+
+[report]
+omit =
+    tests/*
+    normandy/settings.py
+    normandy/wsgi.py
+
+[xml]
+output = /tmp/coverage.xml
--- a/docs/dev/coverage.rst
+++ b/docs/dev/coverage.rst
@@ -1,0 +1,35 @@
+=============
+Test Coverage
+=============
+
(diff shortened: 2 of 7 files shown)


mythmon:
I think that adding the coverage file to the repository is a good idea. My opinions on test coverage have softened a bit lately. One question I have about that coverage file though, why exclude management commands? Especially because management commands are used as cron job entry points, they are an important part of our infrastructure that should probably have tests (at least some of them do already).

peterbe:
> why exclude management commands

because I copied [the file from Tecken](https://github.com/mozilla-services/tecken/blob/master/.coveragerc) and just replaced the word "tecken" with "normandy". Kinda.

peterbe:
> I think I'm just going to stop trying to merge things this week.

It's also 5pm my time so I'm in no rush. I'll do some retries on these next week.

peterbe:
bors retry

PR #1548, opened 2018-09-19
v3 api recipe creation without action|arguments test coverage

### LINKED ISSUE #1547: v3 api recipe creation without action|arguments is not tested
See:
screen shot 2018-09-19 at 2 26 17 pm
that's `v3/serializers.py`

peterbe:
Fixes #1547

There are *now* two unit tests that try to create a recipe with the v3 api but with `action_id` or `arguments` omitted from the JSON posted. That's actually, "natively", tested in django rest framework.

### DIFF
--- a/normandy/recipes/api/v3/serializers.py
+++ b/normandy/recipes/api/v3/serializers.py
@@ -153,18 +153,10 @@ class RecipeSerializer(serializers.ModelSerializer):
     def validate(self, data):
         data = super().validate(data)
         action = data.get("action")
-        required_error_msg = serializers.PrimaryKeyRelatedField.default_error_messages["required"]
-
         if action is None:
-            if not self.instance:
-                raise serializers.ValidationError({"action_id": required_error_msg})
             action = self.instance.action
 
         arguments = data.get("arguments")
-
-        if arguments is None and not self.instance:
-            raise serializers.ValidationError({"arguments": required_error_msg})
-
         if arguments is not None:
             # Ensure the value is a dict
             if not isinstance(arguments, dict):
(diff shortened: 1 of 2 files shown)


normandy/recipes/api/v3/serializers.py
```diff
@@ -153,18 +153,10 @@ def validate_extra_filter_expression(self, value):
     def validate(self, data):
         data = super().validate(data)
         action = data.get("action")
-        required_error_msg = serializers.PrimaryKeyRelatedField.default_error_messages["required"]
-
         if action is None:
-            if not self.instance:
-                raise serializers.ValidationError({"action_id": required_error_msg})
```
mythmon: The old validation here strikes me as really odd, so I tracked down when it was added. It was in commit 49dd7adbdeb997f46bf3c6c7764d15b4de2bf783 / PR #1240, though I don't really understand why. I think we haven enough test coverage in this area though, so I'm just going to chalk it up to code is weird sometimes. @rehandalal you wrote that PR. Do you know why this is the way it is?

peterbe:
(at the moment, there's a bug in GitHub or something because I can't "reply" to your comment above. So I'll try to add a general comment instead)

I definitely remember being involved in that stuff. I think I wrote the first version and then Rehan improved it. The tricky thing was that you can't validate "arguments" and "actions" separately. They depend on each other. Of course. So I suspect it was when we moved the invariant validation in the `validate` method that we left a trail of junk.

mythmon:
> invariant validation in the validate method that we left a trail of junk.

I'm not sure that's true. I didn't see anything like this in the previous version that had the validation checks in separate methods.

normandy/recipes/api/v3/serializers.py
rehandalal: my brain appears to have dropped all info on why i did whatever i did

mythmon: Well, in that case I'll just chalk it up to a mystery. Lets go ahead and land this.

PR #1549, opened 2018-09-20
docs: Explain filter object shape better

mythmon:


normandy/recipes/filters.py
```diff
@@ -6,6 +6,18 @@
 Each filter below defines a rule, and when it matches a user. Filter parameters
 are represented as JSON. Most users will interact with filters at a higher
 level, such as a web interface that allows building a filter with a form.
+
+Filter objects are generally specified as a JSON object with at least a
+"type" field, and other fields determined by that type. For example,
+ChannelFilter` below has a type of "channel" and requires a ``channels``
```
peterbe: Should that be "ChannelFilter" with two backticks? ...since this gets included in Sphinx as rst.

normandy/recipes/filters.py
```diff
@@ -76,9 +107,9 @@ def to_jexl(self):
 class BucketSampleFilter(BaseFilter):
     """
     Sample a portion of the users by defining a series of buckets, evenly
-    distributing users into those buckets, and then selecting a range of those
-    buckets. This is stable: a given Firefox profile will always match or always
-    not match.
+    distributing users into those buckets, and then selecting a range of
+    those buckets. This is stable: a given set of inputs will always match or
+    always not match.
```
peterbe: I don't know why, but I struggle to parse that sentence; *"a given set of inputs will always match or not match."*
It feels like "duh. of course." 
To me, when you say "this is stable" I imagine it means that it's the opposite of unpredictable. If a certain Firefox Profile runs through this filter a second time, the outcome won't be different.

normandy/recipes/filters.py
```diff
@@ -190,12 +237,15 @@ def to_jexl(self):
...
-    "version": VersionFilter,
+    f.type: f
+    for f in [
+        ChannelFilter,
+        LocaleFilter,
+        CountryFilter,
+        BucketSampleFilter,
+        StableSampleFilter,
+        VersionFilter,
+    ]
```
peterbe: nice!

normandy/recipes/filters.py
mythmon: Eh, yeah this isn't right. I had originally tried to make a reference, but couldn't figure out so deleted the markdown. I apparently missed some though.

normandy/recipes/filters.py
mythmon: I struggle to explain this actually. It is unpredictable (given the inputs, the only way to know the answer is to do the hashing), but not random (the same inputs always produce the same outputs). This is explained well by the functional concept of "purity", but I don't think that's a good thing to write in this documentation. I'll take another stab at this.

mythmon:
bors r= @peterbe

PR #1561, opened 2018-09-27
Add no-cache headers to classify_client.

### LINKED ISSUE #1559: classify-client endpoint should disallow caching
The classify-client endpoint should never be cached. Currently we ensure this by using a separate domain name that routes traffic directly to a load balancer instead of through Cloudfront. We could simplify the infrastructure and use a single domain if Normandy would send `Cache-Control: no-cache, no-store, must-revalidate` for requests to classify-client.

mythmon:
Fixes #1559

### DIFF
--- a/normandy/recipes/api/v1/views.py
+++ b/normandy/recipes/api/v1/views.py
@@ -1,7 +1,7 @@
 from django.conf import settings
 from django.db import transaction
 from django.db.models import Q
-from django.views.decorators.cache import cache_control
+from django.views.decorators.cache import never_cache
 
 import django_filters
 from rest_framework import generics, permissions, views, viewsets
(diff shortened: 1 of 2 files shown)


normandy/recipes/api/v1/views.py
```diff
@@ -168,6 +169,7 @@ class ClassifyClient(views.APIView):
     permission_classes = []
     serializer_class = ClientSerializer
 
+    @cache_control(no_cache=True, no_store=True, must_revalidate=True)
```
sciurus: Oops, my comment didn't go through with the review.

THis looks fine, but I think you could just use the [never_cache](https://docs.djangoproject.com/en/2.1/topics/http/decorators/#caching) decorator.

PR #1568, opened 2018-10-04
Schedule Renovate updates for only Monday mornings

mythmon:


mythmon:
"Monday mornings" is one of the presets of Renovate, and seemed like a reasonable default. I didn't use the preset directly because it didn't make the Timezone clear, so I did it manually, and tuned it to "before East coast working hours".

PR #1577, opened 2018-10-08
black 18.9b0

peterbe:


mythmon:
I think we should do this as a part of the normal Pyup update PR, #1576.

PR #1578, opened 2018-10-08
validate filter objects' locales, countries and channels

### LINKED ISSUE #1544: Filter object's locales, countries and channels is not validated
I just managed to save a recipe's filter object, using my [unfinished branch](https://github.com/mozilla/delivery-console/issues/176), that set the filter object to something that is definitely not in the `Country` model. 

E.g.
```
In [7]: r.refresh_from_db()

In [8]: r.filter_object
Out[8]:
[{'type': 'locale', 'locales': ['sv-SE']},
 {'type': 'country', 'countries': ['SE', 'junkjunkjunk']}]

```

peterbe:
Fixes #1544 

I'm generally very reluctant to fixtures but perhaps we could add some that add `Channel.objects.create(slug="beta, name="Beta"); Channel.objects.create(slug="release", name="Release")`

### DIFF
--- a/normandy/recipes/tests/api/v2/test_api.py
+++ b/normandy/recipes/tests/api/v2/test_api.py
@@ -12,10 +12,13 @@ from pathlib import Path
 from normandy.base.api.permissions import AdminEnabledOrReadOnly
 from normandy.base.tests import UserFactory, Whatever
 from normandy.base.utils import canonical_json_dumps
-from normandy.recipes.models import ApprovalRequest, Channel, Country, Locale, Recipe
+from normandy.recipes.models import ApprovalRequest, Recipe
 from normandy.recipes.tests import (
     ActionFactory,
     ApprovalRequestFactory,
+    ChannelFactory,
+    CountryFactory,
+    LocaleFactory,
     RecipeFactory,
     RecipeRevisionFactory,
     fake_sign,
@@ -159,7 +162,7 @@ class TestRecipeAPI(object):
 
         def test_it_can_create_recipes_with_only_filter_object(self, api_client):
             action = ActionFactory()
-            Channel.objects.get_or_create(slug="beta", name="Beta")
+            channel = ChannelFactory()
 
             res = api_client.post(
                 "/api/v2/recipe/",
@@ -168,7 +171,7 @@ class TestRecipeAPI(object):
                     "action_id": action.id,
                     "arguments": {},
                     "extra_filter_expression": "   ",
-                    "filter_object": [{"type": "channel", "channels": ["beta"]}],
+                    "filter_object": [{"type": "channel", "channels": [channel.slug]}],
                     "enabled": True,
                 },
             )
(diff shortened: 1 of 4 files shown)


normandy/recipes/tests/api/v2/test_api.py
```diff
@@ -159,6 +159,7 @@ def test_it_can_create_recipes(self, api_client):
 
         def test_it_can_create_recipes_with_only_filter_object(self, api_client):
             action = ActionFactory()
+            Channel.objects.get_or_create(slug="beta", name="Beta")
```
mythmon: First, we shouldn't `get_or_create` in tests, since that implies we aren't sure of the default state of the DB at the beginning of test, which isn't good.

Second, we should probably be using a Factory for this like we do other models, instead of creating objects by hand. One way that could help is to not think about the slug/name, and just do something like:

```python
channel = ChannelFactory()

res = api_client.post(
    "/api/v2/recipe/",
    {
        ...
        "filter_object": [{"type": "channel", "channels": [channel.slug]}],
    }
)
```

peterbe: Hmm... This isn't actually that nice. I appreciate the desire to be consistent, i.e. if recipes use factories, that we should do that for Country, Locale and Channel too. 
However, the code becomes more complex looking.

Before:
```python
def test_country_works(self, api_client):
    Country.objects.get_or_create(code="US", name="United States of America")
    Country.objects.get_or_create(code="DE", name="Germany")
    res = self.make_recipe(
        api_client, filter_object=[{"type": "country", "countries": ["US", "DE"]}]
    )
    assert res.status_code == 201, res.json()
    recipe_data = res.json()

    Recipe.objects.get(id=recipe_data["id"])
    assert recipe_data["filter_expression"] == (
        '(normandy.country in ["US","DE"]) && (true)'
    )
```

After:
```python
def test_country_works(self, api_client):
    country1 = CountryFactory()
    country2 = CountryFactory(code="DE")
    res = self.make_recipe(
        api_client,
        filter_object=[{"type": "country", "countries": [country1.code, country2.code]}],
    )
    assert res.status_code == 201, res.json()
    recipe_data = res.json()

    Recipe.objects.get(id=recipe_data["id"])
    assert recipe_data["filter_expression"] == (
        f'(normandy.country in ["{country1.code}","{country2.code}"]) && (true)'
    )
```

I think the code before *read* more easily. It's more natural and there's definitely less noise. Not to mention that now I have to use these weird looking variable names like `country2` and if I call it `country_de` that's almost worse. 

But I'm not going to kick up a fuzz. Just registering my observation.

mythmon: A lot of our factories by default use a random value for fields. If that were the csae here, then you wouldn't need to specify the code on the second factory, and could in fact do `country1, country2 = CountryFactory.build_batch(2)`. That doesn't get around having to refer to `country2.code` instead of just assuming "DE" thought.

It would also be fine to just hard code `CountryFactory(name="Germany", code="DE")` and then using "DE" explicitly throughout the rest of the test. Using these factories equivalently to the `get_or_create` you had before is also fine.

mythmon:
If you want to tweak the factories more, feel free. Otherwise this is ready to merge.

bors delegate+

peterbe:
I think it's good enough. 

bors r+

PR #1581, opened 2018-10-11
Expose enable_state id in v3 API

mythmon:
The `carry_over` field of `EnabledStateSerializer` gives back an ID, but the ID of enabled states wasn't exposed anymore. This fixes that, and also adds a test.


normandy/base/tests/__init__.py
```diff
@@ -29,9 +29,13 @@ def contains(cls, *values):
...
-            # Will throw is it does not match
-            datetime.strptime(s, "%Y-%m-%dT%H:%M:%S.%fZ")
-            return True
+            if not isinstance(s, str):
+                return False
+            try:
+                datetime.strptime(s, "%Y-%m-%dT%H:%M:%S.%fZ")
+                return True
+            except ValueError:
+                return False
```
mythmon: This change makes the errors that come out of this method easier to understand. It gives back something like `"foo" != <Whatever named "datetime">` instead of throwing a `ValueError`.

normandy/recipes/tests/api/v3/test_serializers.py
```diff
@@ -181,3 +183,55 @@ def test_it_uses_cdn_url(self, rf, settings):
         action = ActionFactory()
         serializer = ActionSerializer(action, context={"request": rf.get("/")})
         assert serializer.data["implementation_url"].startswith(settings.CDN_URL)
+
+
+@pytest.mark.django_db()
+class TestEnabledStateSerializer:
+    def test_it_works(self):
```
mythmon: This test is a little gross, since it goes through the whole process of creating and enabling two revisions of a recipe. It was easier than trying to create `EnabledState` objects out of whole cloth though.

rehandalal: i think (unfortunately) this is the best way to test thees.

PR #1591, opened 2018-10-18
No migrations left behind, in CI

### LINKED ISSUE #1590: No migrations left behind, in CI
If you have a PR which changes the models but forgot to generate the migrations for it, it should break CI. 

See https://github.com/mozilla/normandy/issues/1588#issuecomment-431135260 for background.

peterbe:
Fixes #1590

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,6 +58,10 @@ jobs:
           command: docker-compose run web python-tests
 
       - run:
+          name: Missing migrations
+          command: docker-compose run web migrations-check
+
+      - run:
           name: Contract tests
           command: |
             docker-compose up -d web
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     working_directory: ~/normandy
 
     environment:
-        COMPOSE_FILE: ci/docker-compose.yml
+      COMPOSE_FILE: ci/docker-compose.yml
```
peterbe: My vscode fixes yaml files automatically and although this is unrelated I don't think it hurts to correct.

mythmon: In the future it would be good to only include these kind of fixes to files you fixed already. This change is more correct so lets keep it, but if we're going to have automatic fixes for these kind of files, we should add them as rules to Therapist instead of ad-hoc changes.

PR #1593, opened 2018-10-18
correct escape for regex

peterbe:
One lesser `DeprecationWarning` can't hurt:
```
/Users/peterbe/dev/MOZILLA/NORMANDY/normandy/normandy/control/views.py:17: DeprecationWarning: invalid escape sequence \.
  match = re.search("normandy\.(\w+)\.mozaws", hostname, re.I)
```

### DIFF
--- a/normandy/control/views.py
+++ b/normandy/control/views.py
@@ -14,7 +14,7 @@ def index(request):
     hostname = request.get_host()
 
     delivery_console_url = delivery_console_urls["prod"]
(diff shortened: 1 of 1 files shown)


peterbe:
> reading r strings in Python as "regex" instead of raw strings

I don't even know what a "raw string" is. I could google it and learn but I'm just making the observation that I simply don't know and I've never used for *anything* other than regex stuff.

PR #1613, opened 2018-11-09
Add command to update addon URLs in opt-out-study arguments.

### LINKED ISSUE #1597: Script to rewrite addonUrls stored in DB
I've realized the full addonUrl is stored in Normandy's database. The values there refer to an S3 bucket. For example,

```
normandy=> select arguments_json from public.recipes_reciperevision where name = 'T12Clone';
                                                                                               arguments_json                                                                        
                        
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
------------------------
 {"name": "T12OptStudy", "description": "T12OptDescription", "addonUrl": "https://net-mozaws-stage-us-east-1-normandy.s3.amazonaws.com/extensions/vk_scrobbler-2.1.1.1-fx.xpi", "isEn
rollmentPaused": false}
(1 row)
```

After the GCP migration, we'll want to rewrite them to refer to the domain we'll use to front our GCS bucket so that we can retire the S3 bucket. To do this I think a python script that 

1. finds all the recipe revisions with addonURL in their arumgent_json
1. and updates them to replace the domain with a new value (either provided as an argument to the script or read from whatever setting we implement for #1567)

would be handy.

mythmon:
This isn't a migration since it won't be applicable to all environments.

Fixes #1597.

### DIFF
--- a/normandy/recipes/management/commands/update_addon_urls.py
+++ b/normandy/recipes/management/commands/update_addon_urls.py
@@ -1,0 +1,52 @@
+import json
+
+from urllib.parse import urlparse, urlunparse
+from django.core.management.base import BaseCommand
+
+from normandy.recipes.models import RecipeRevision
+
+
+class Command(BaseCommand):
+    """
+    Rewrite all revisions to update the URL of add-ons stored in Normandy's
+    file storage to a new hostname.
+    """
+
+    help = "Updates add-on URL in revisions"
+
+    def add_arguments(self, parser):
+        parser.add_argument("new_hostname", help="Hostname to convert add-on urls to")
+
+    def handle(self, *args, new_hostname, **options):
+        target_revisions = RecipeRevision.objects.filter(action__name="opt-out-study")
+        update_count = 0
+        for rev in target_revisions:
+            if "addonUrl" not in rev.arguments:
+                self.stderr.write(
+                    f"Warning: Recipe {rev.recipe.id} revision {rev.id} has action=opt-out-study, "
+                    f"but no addonUrl",
+                    ending="\n",
+                )
+                continue
+            # Pull into a local variable to modify the arguments since
+            # `rev.arguments` is actually a property that parses JSON, not a
+            # real attribute of the object
+            arguments = rev.arguments
+            parsed_url = urlparse(arguments["addonUrl"])
+            # _replace is not a private method. parsed_url is a named tuple,
+            # and named tuples expose all their public methods with underscores
+            # to avoid colliding with data keys.
+            new_url_parts = parsed_url._replace(netloc=new_hostname)
+            new_url = urlunparse(new_url_parts)
+            if new_url != arguments["addonUrl"]:
+                update_count += 1
+                arguments["addonUrl"] = new_url
+                rev.arguments_json = json.dumps(arguments)
+                rev.save()
+
+        if update_count == 0:
+            self.stdout.write("No revisions updated")
+        elif update_count == 1:
+            self.stdout.write("One revision updated")
+        else:
+            self.stdout.write(f"{update_count} revisions updated")
--- a/normandy/recipes/tests/test_commands.py
+++ b/normandy/recipes/tests/test_commands.py
@@ -1,3 +1,4 @@
+import hashlib
 import json
 from unittest.mock import patch
 from datetime import timedelta
@@ -7,7 +8,7 @@ from django.core.management import call_command, CommandError
 import pytest
 
 from normandy.base.tests import UserFactory
-from normandy.recipes.models import Action
+from normandy.recipes.models import Recipe, Action
 from normandy.recipes.tests import ActionFactory, RecipeFactory
 
 
@@ -257,3 +258,42 @@ class TestUpdateActionSignatures(object):
         call_command("update_action_signatures")
         a.refresh_from_db()
         assert a.signature.signature == "original signature"
+
+
+@pytest.mark.django_db
+class TestUpdateAddonUrls(object):
+    def test_it_works(self):
+        action = ActionFactory(name="opt-out-study")
+        recipe = RecipeFactory(
+            action=action,
+            arguments={"addonUrl": "https://before.example.com/extensions/addon.xpi"},
+        )
+        call_command("update_addon_urls", "after.example.com")
+        recipe = Recipe.objects.get(id=recipe.id)
+        assert recipe.arguments["addonUrl"] == "https://after.example.com/extensions/addon.xpi"
+
+    def test_signatures_are_updated(self, mocked_autograph):
+        action = ActionFactory(name="opt-out-study")
+        recipe = RecipeFactory(
+            action=action,
+            arguments={"addonUrl": "https://before.example.com/extensions/addon.xpi"},
+            approver=UserFactory(),
+            enabler=UserFactory(),
+            signed=True,
+        )
+        # preconditions
+        assert recipe.signature is not None
+        assert recipe.signature.signature == hashlib.sha256(recipe.canonical_json()).hexdigest()
+        signature_before = recipe.signature.signature
+
+        call_command("update_addon_urls", "after.example.com")
+        recipe.refresh_from_db()
+
+        assert recipe.signature is not None
+        assert recipe.signature != signature_before
+        assert recipe.signature.signature == hashlib.sha256(recipe.canonical_json()).hexdigest()
+
+    def test_new_hostname_is_required(self):
+        with pytest.raises(CommandError) as err:
+            call_command("update_addon_urls")
+        assert "the following arguments are required: new_hostname" in str(err)


normandy/recipes/management/commands/update_addon_urls.py
```diff
@@ -0,0 +1,52 @@
...
+    Rewrite all revisions to update the URL of add-ons stored in Normandy's
+    file storage to a new hostname.
+    """
+
+    help = "Updates add-on URL in revisions"
+
+    def add_arguments(self, parser):
+        parser.add_argument("new_hostname", help="Hostname to convert add-on urls to")
+
+    def handle(self, *args, new_hostname, **options):
```
peterbe: I've always done ...

```python
def handle(self, **options):
    new_hostname = options["new_hostname"]
```

But come to think of it I can't see what the difference is. 

[The docs](https://docs.djangoproject.com/en/2.1/howto/custom-management-commands/) use `options`.

normandy/recipes/management/commands/update_addon_urls.py
```diff
@@ -0,0 +1,52 @@
...
+        target_revisions = RecipeRevision.objects.filter(action__name="opt-out-study")
+        update_count = 0
+        for rev in target_revisions:
+            if "addonUrl" not in rev.arguments:
```
peterbe: If you have `arguments = {'addonUrl': None}` this line wouldn't notice but later it would fail on `urlparse(None)`.

Does that matter?

normandy/recipes/tests/test_commands.py
```diff
@@ -257,3 +258,42 @@ def test_it_does_not_resign_up_to_date_actions(self, settings, mocked_autograph)
...
+        signature_before = recipe.signature.signature
+
+        call_command("update_addon_urls", "after.example.com")
+        recipe.refresh_from_db()
+
+        assert recipe.signature is not None
+        assert recipe.signature != signature_before
+        assert recipe.signature.signature == hashlib.sha256(recipe.canonical_json()).hexdigest()
+
+    def test_new_hostname_is_required(self):
```
peterbe: This test is not needed. It's just testing Django. Not your code.

normandy/recipes/management/commands/update_addon_urls.py
```diff
@@ -0,0 +1,52 @@
...
+
+    help = "Updates add-on URL in revisions"
+
+    def add_arguments(self, parser):
+        parser.add_argument("new_hostname", help="Hostname to convert add-on urls to")
+
+    def handle(self, *args, new_hostname, **options):
+        target_revisions = RecipeRevision.objects.filter(action__name="opt-out-study")
+        update_count = 0
+        for rev in target_revisions:
```
peterbe: Are there many of these in the database where this matters? Would it be worth changing this line to `for rev in target_revisions.iterator():` which will do the same thing but only make iterated lists of 2,000 instances at a time.

normandy/recipes/management/commands/update_addon_urls.py
```diff
@@ -0,0 +1,52 @@
...
+                    f"Warning: Recipe {rev.recipe.id} revision {rev.id} has action=opt-out-study, "
+                    f"but no addonUrl",
+                    ending="\n",
+                )
+                continue
+            # Pull into a local variable to modify the arguments since
+            # `rev.arguments` is actually a property that parses JSON, not a
+            # real attribute of the object
+            arguments = rev.arguments
+            parsed_url = urlparse(arguments["addonUrl"])
```
peterbe: Instead of creating the `new_url` you count "exit early" here by doing...:

```python
if parsed_url.netloc == new_hostname:
    continue
```

normandy/recipes/management/commands/update_addon_urls.py
```diff
@@ -0,0 +1,52 @@
...
+                arguments["addonUrl"] = new_url
+                rev.arguments_json = json.dumps(arguments)
+                rev.save()
+
+        if update_count == 0:
+            self.stdout.write("No revisions updated")
+        elif update_count == 1:
+            self.stdout.write("One revision updated")
+        else:
+            self.stdout.write(f"{update_count} revisions updated")
```
peterbe: This line is all you need :)

normandy/recipes/tests/test_commands.py
```diff
@@ -257,3 +258,42 @@ def test_it_does_not_resign_up_to_date_actions(self, settings, mocked_autograph)
...
+
+
+@pytest.mark.django_db
+class TestUpdateAddonUrls(object):
+    def test_it_works(self):
+        action = ActionFactory(name="opt-out-study")
+        recipe = RecipeFactory(
+            action=action,
+            arguments={"addonUrl": "https://before.example.com/extensions/addon.xpi"},
+        )
```
peterbe: Perhaps it'd be worth adding something like:
```python
    other_action = ActionFactory(name="anything-else")
    other_recipe = RecipeFactory(
        action=other_action,
        arguments={"addonUrl": "https://before.example.com/extensions/addon.xpi"},
    )
    call_command("update_addon_urls", "after.example.com")
    other_recipe.refresh_from_db()
    # Should *not* change.
    assert other_recipe.arguments["addonUrl"] == "https://before.example.com/extensions/addon.xpi"
```

normandy/recipes/management/commands/update_addon_urls.py
mythmon: That's a good point. I'll add a check for that (as improbably as it is).

normandy/recipes/tests/test_commands.py
mythmon: I actually wrote this test before I wrote the command. It is mostly testing that I correctly remembered how to set `required=True` on the argparser. I'd prefer to keep it.

normandy/recipes/management/commands/update_addon_urls.py
mythmon: I've never seen `.iterator()` before, that's interesting. I don't think it will matter but it is semantically more what we want here.

I checked [the api](https://normandy.cdn.mozilla.net/api/v3/recipe_revision/) and there are currently 1989 revisions. So if `.iterator()` makes chunks of 2000, then it won't have any effect.

normandy/recipes/management/commands/update_addon_urls.py
mythmon: We can make nice UIs that have good pluralization behavior. "1 revisions updated" grates on me. I guess I could probably use an actual pluralization tool from Django to do this though.

normandy/recipes/tests/test_commands.py
mythmon: :+1:

normandy/recipes/management/commands/update_addon_urls.py
```diff
@@ -0,0 +1,53 @@
...
+            if arguments.get("addonUrl") is None:
+                self.stderr.write(
+                    f"Warning: Recipe {rev.recipe.id} revision {rev.id} has action=opt-out-study, "
+                    f"but no addonUrl",
+                    ending="\n",
+                )
+                continue
+
+            parsed_url = urlparse(arguments["addonUrl"])
+            if parsed_url.netlock == new_hostname:
```
peterbe: haha! Thank the lords of CI :)

mythmon:
For some reason `recipe.refresh_from_db()` wasn't working, so I reverted. @peterbe if you have any insight into that, I'd be very curious.

normandy/recipes/management/commands/update_addon_urls.py
```diff
@@ -0,0 +1,51 @@
...
+
+    help = "Updates add-on URL in revisions"
+
+    def add_arguments(self, parser):
+        parser.add_argument("new_hostname", help="Hostname to convert add-on urls to")
+
+    def handle(self, *args, new_hostname, **options):
+        target_revisions = RecipeRevision.objects.filter(action__name="opt-out-study")
+        update_count = 0
+        for rev in target_revisions.iterator():
```
peterbe: There's no need for this now. If there's only about ~2k records anyway, the iterator doesn't help. Not important though. I brought it because I wasn't sure how many of these there are in the DB and if it was a genuinely large number it'd help to use `.iterator()`.

normandy/recipes/management/commands/update_addon_urls.py
```diff
@@ -0,0 +1,51 @@
...
+            # Pull into a local variable to modify the arguments since
+            # `rev.arguments` is actually a property that parses JSON, not a
+            # real attribute of the object
+            arguments = rev.arguments
+
+            if arguments.get("addonUrl") is None:
```
peterbe: What if it's empty? Like `{'addonUrl': ''}`?

normandy/recipes/tests/test_commands.py
```diff
@@ -257,3 +258,54 @@ def test_it_does_not_resign_up_to_date_actions(self, settings, mocked_autograph)
...
+        call_command("update_addon_urls", "after.example.com")
+
+        recipe = Recipe.objects.get(id=recipe.id)
```
peterbe: So, if you replace this line with `recipe.refresh_from_db()` the next line below fails?

normandy/recipes/tests/test_commands.py
peterbe: Wait, what?! What "real problem".

mythmon: The real problem was that I trying to update the revision arguments directly, like`revision.arguments["updateUrl"] = ...`. That made nothing work at all, and was very confusing, so I was trying random things.

normandy/recipes/tests/test_commands.py
mythmon: Yeah. I'm really not sure why. I don't think we are doing any caching or anything. I didn't dig too deep though.

mythmon:
I believe I've fixed all the issues.

bors r= @peterbe, @sciurus

PR #1616, opened 2018-11-12
Publish recipes on Remote Settings

leplatrem:
~~**This is a draft**~~

This is the counterpart of https://bugzilla.mozilla.org/show_bug.cgi?id=1506175


normandy/recipes/exports.py
```diff
@@ -0,0 +1,53 @@
...
+class RemoteSettings(object):
+    """
+    Interacts with a RemoteSettings service.
+
+    """
+    def __init__(self):
+        self.check_config()
+        self.cid = settings.REMOTE_SETTINGS_COLLECTION_ID
+
+    @cached_property
```
mythmon: `cached_property` [stores the result on `self`](https://docs.djangoproject.com/en/2.1/ref/utils/#django.utils.functional.cached_property). Since you only ever use this by something like `RemoteSettings().publish()`, the cache isn't going to help at all. Perhaps it should be cached in a module level variable? What is the cost of recreating a Kinto client?

normandy/recipes/exports.py
```diff
@@ -0,0 +1,53 @@
...
+                raise ImproperlyConfigured(msg)
+
+    def publish(self, revision):
+        from normandy.recipes.api.v3.serializers import RecipeRevisionSerializer
+
+        serializer = RecipeRevisionSerializer(revision)
+        as_dict = serializer.data.copy()
+        as_dict["id"] = str(revision.recipe.id)
+        del as_dict["recipe"]
+        del as_dict["approval_request"]
```
mythmon: In the current Normandy client, we use a specialized version of the v1 serializer that serves only the needed fields to run the recipe. I'm not sure what the right approach is. Should the new version use a similar approach? Or do we want it to be more general and include much more information?

Additionally, I think there may be an incompatibility here. The client code you put in didn't change how the recipes were handled, so I'd expect it would be looking for a v1 style recipe, which has some differences to the v3 style given here. I think we probably ought to make the client accept a v3 style recipe, so we can eventually deprecate v1, but that's a bigger change on the client.

normandy/settings.py
```diff
@@ -357,6 +357,13 @@ def RAVEN_CONFIG(self):
     AUTOGRAPH_SIGNATURE_MAX_AGE = values.IntegerValue(60 * 60 * 24 * 7)
     AUTOGRAPH_X5U_CACHE_BUST = values.Value(None)
 
+    # Remote Settings settings
+    REMOTE_SETTINGS_ENABLED = values.BooleanValue(True)
+    REMOTE_SETTINGS_URL = values.Value("http://localhost:8888/v1")
+    REMOTE_SETTINGS_USERNAME = values.Value("normandy")
+    REMOTE_SETTINGS_PASSWORD = values.Value("n0rm4ndy")
+    REMOTE_SETTINGS_COLLECTION_ID = values.Value("normandy-recipes")
```
mythmon: For the final version, you'll probably want to leave these as blanks (like `REMOTE_SETTINGS_URL = values.Value()`) since I don't think we can supply reasonable defaults.

normandy/recipes/tests/test_exports.py
```diff
@@ -0,0 +1,187 @@
...
+
+        mock_kinto_request.return_value = mock.MagicMock(status_code=200)
+        recipe = RecipeFactory(name="Test", approver=UserFactory(), enabler=UserFactory())
+
+        remotesettings = exports.RemoteSettings()
+        remotesettings.publish(recipe)
+
+        auth = (settings.REMOTE_SETTINGS_USERNAME, settings.REMOTE_SETTINGS_PASSWORD)
+        headers = {"User-Agent": KINTO_USER_AGENT, "Content-Type": "application/json"}
+        url = settings.REMOTE_SETTINGS_URL + "/buckets/main-workspace/collections/normandy-recipes"
```
peterbe: Shouldn't that be something like ...

```python
url = settings.REMOTE_SETTINGS_URL + f"/bucket/${REMOTE_SETTINGS_BUCKET}/collections/${settings.REMOTE_SETTINGS_COLLECTION_ID}"
```

normandy/recipes/tests/test_exports.py
```diff
@@ -0,0 +1,187 @@
...
+        # Assert the correct request was made
+        payload = json.dumps({"data": exports.recipe_as_record(recipe)})
+
+        assert mock_kinto_request.call_count == 2
+        mock_kinto_request.assert_any_call(
+            "put", record_url, data=payload, auth=auth, headers=headers
+        )
+        mock_kinto_request.assert_any_call(
+            "patch", url, data='{"data": {"status": "to-sign"}}', auth=auth, headers=headers
+        )
```
peterbe: Perhaps we should be using `requestsmock` since the `requests` that `kinto_client` uses is the same one.

leplatrem:
Thanks for your comments!

@mythmon I consider this version reviewable :)

normandy/recipes/tests/test_models.py
```diff
@@ -579,6 +579,57 @@ def test_disable(self):
...
+        settings.REMOTE_SETTINGS_USERNAME = "normandy"
+        settings.REMOTE_SETTINGS_PASSWORD = "n0rm4ndy"
+
+        recipe = RecipeFactory(name="Test")
+
+        approval_request = recipe.latest_revision.request_approval(creator=UserFactory())
+        approval_request.approve(approver=UserFactory(), comment="r+")
+        recipe.approved_revision.enable(user=UserFactory())
+
+        mocked_remotesettings.return_value.publish.assert_called_with(recipe)
```
peterbe: Nice!

normandy/recipes/exports.py
```diff
@@ -0,0 +1,96 @@
...
+
+import kinto_http
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
+
+
+logger = logging.getLogger(__name__)
+
+REMOTE_SETTINGS_BUCKET = "main-workspace"
+NB_RETRY_REQUESTS = 3
```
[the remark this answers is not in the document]
leplatrem: NB was a dumb abbreviation of NumBer :D

normandy/recipes/tests/test_exports.py
```diff
@@ -0,0 +1,187 @@
...
+        settings.REMOTE_SETTINGS_COLLECTION_ID = None
+        with pytest.raises(ImproperlyConfigured) as exc:
+            exports.RemoteSettings()
+        assert "REMOTE_SETTINGS_COLLECTION_ID" in str(exc)
+
+        # Include everything
+        settings.REMOTE_SETTINGS_URL = self.test_settings["URL"]
+        settings.REMOTE_SETTINGS_USERNAME = self.test_settings["USERNAME"]
+        settings.REMOTE_SETTINGS_PASSWORD = self.test_settings["PASSWORD"]
+        settings.REMOTE_SETTINGS_COLLECTION_ID = "normandy-recipes"
```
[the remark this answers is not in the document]
leplatrem: Yes but we set it to `None` above

normandy/recipes/exports.py
```diff
@@ -0,0 +1,96 @@
...
+        self.client.update_record(data=record)
+        self.client.patch_collection(id=self.cid, data={"status": "to-sign"})
+        logger.info(f"Published record '{recipe.id}' for recipe {recipe}")
+
+    def unpublish(self, recipe):
+        """
+        Unpublish the specified `recipe` by deleted its associated records on the remote server.
+        """
+        try:
+            self.client.delete_record(id=str(recipe.id))
```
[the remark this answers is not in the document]
leplatrem: Nope, that's why we have those quotes elsewhere :)

normandy/recipes/exports.py
```diff
@@ -0,0 +1,96 @@
...
+                raise ImproperlyConfigured(msg)
+
+    def publish(self, recipe):
+        """
+        Publish the specified `recipe` on the remote server by upserting a record.
+        """
+        record = recipe_as_record(recipe)
+        self.client.update_record(data=record)
+        self.client.patch_collection(id=self.cid, data={"status": "to-sign"})
+        logger.info(f"Published record '{recipe.id}' for recipe {recipe}")
```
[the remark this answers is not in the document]
leplatrem: well, the `__str__` for recipe is `recipe.name`, which is safe I believe

normandy/recipes/exports.py
```diff
@@ -0,0 +1,96 @@
+import logging
+
+import kinto_http
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
+
+
+logger = logging.getLogger(__name__)
+
+REMOTE_SETTINGS_BUCKET = "main-workspace"
```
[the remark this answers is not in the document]
leplatrem: It's a constant, but yes we can put in settings

normandy/recipes/tests/test_exports.py
leplatrem: Done :)

normandy/recipes/exports.py
```diff
@@ -0,0 +1,96 @@
...
+        client = kinto_http.Client(
+            server_url=str(settings.REMOTE_SETTINGS_URL),
+            auth=(str(settings.REMOTE_SETTINGS_USERNAME), str(settings.REMOTE_SETTINGS_PASSWORD)),
+            bucket=REMOTE_SETTINGS_BUCKET,
+            collection=self.cid,
+            retry=NB_RETRY_REQUESTS,
+        )
+        return client
+
+    def check_config(self):
```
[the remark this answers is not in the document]
leplatrem: Done, thanks for pointing this out!

normandy/recipes/tests/test_exports.py
```diff
@@ -0,0 +1,187 @@
...
+        "URL": "https://remotesettings.example.com/v1",
+        "USERNAME": "normandy",
+        "PASSWORD": "n0rm4ndy",
+    }
+
+    def test_default_settings(self, settings):
+        """Test default settings values."""
+
+        assert not settings.REMOTE_SETTINGS_ENABLED
+        assert settings.REMOTE_SETTINGS_COLLECTION_ID == "normandy-recipes"
```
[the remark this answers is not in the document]
leplatrem: Why?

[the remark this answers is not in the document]
peterbe: The test is about checking that `REMOTE_SETTINGS_ENABLED` is false by default. But this last line checks something for `REMOTE_SETTINGS_COLLECTION_ID` which doesn't feel related to the context.

docs/dev/install.rst
```diff
@@ -194,3 +194,69 @@ If you ever need to bypass Therapist, you can do so by passing
 ``--no-verify`` to your ``git commit`` command.
 
 .. _Therapist: http://therapist.readthedocs.io/en/latest/overview.html
+
+.. _remotesettings-install:
+
+RemoteSettings
```
peterbe:
```suggestion
Remote Settings
```

docs/dev/install.rst
```diff
@@ -194,3 +194,69 @@ If you ever need to bypass Therapist, you can do so by passing
...
+      curl -X PUT ${SERVER}/accounts/normandy \
+           -d '{"data": {"password": "n0rm4ndy"}}' \
+           -H 'Content-Type:application/json'
+
+      curl -X PUT ${SERVER}/buckets/main-workspace/collections/normandy-recipes \
+           -H 'Content-Type:application/json' \
+           -u 'normandy:n0rm4ndy'
+
+With the configuration in place, Normandy should start without error. When a recipe is enabled, it is published and becomes visible
+on the RemoteSettings server:
```
peterbe: Sometimes you write `RemoteSettings` and sometimes you write `Remote Settings`.

normandy/recipes/tests/test_exports.py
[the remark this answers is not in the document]
leplatrem: the test says "Test default settings values"

normandy/recipes/apps.py
```diff
@@ -11,4 +12,5 @@ class RecipesApp(AppConfig):
 
     def ready(self):
         checks.register()
+        RemoteSettings().check_config()
```
leplatrem: Where should I add a test for this?

docs/dev/install.rst
```diff
@@ -194,3 +194,69 @@ If you ever need to bypass Therapist, you can do so by passing
...
+2. Add the following configuration to ``.env`` (create the file
+   if it does not exist yet):
+
+   .. code-block:: ini
+
+      DJANGO_REMOTE_SETTINGS_URL=http://localhost:8888/v1
+      DJANGO_REMOTE_SETTINGS_USERNAME=normandy
+      DJANGO_REMOTE_SETTINGS_PASSWORD=n0rm4ndy
+
+3. Add the following configuration to ``server.ini`` and restart the Remote Settings container:
```
mythmon: It isn't clear that `server.ini` and `.env` belong in different projects. Can you mention where these files should go?

normandy/recipes/apps.py
mythmon: I don't think we need to test the `ready` method. I think the tests you have for `check_config` below are good enough.

normandy/recipes/exports.py
```diff
@@ -0,0 +1,120 @@
...
+    record["id"] = str(recipe.id)
+    return record
+
+
+class RemoteSettings:
+    """
+    Interacts with a RemoteSettings service.
+
+    Basically, a recipe becomes a record in the dedicated collection on Remote Settings.
+    When it is disabled, the associated record is deleted.
```
mythmon: Will Remote Settings mind if we create a recipe, enable it, disable, and then re-enable it again? Does Remote Settings have the concept of undeleting a record? Or will it create a new record?

normandy/recipes/exports.py
```diff
@@ -0,0 +1,120 @@
...
+        try:
+            self.client.delete_record(id=str(recipe.id))
+
+        except kinto_http.KintoException as e:
+            if e.response.status_code == 404:
+                logger.warning(f"The recipe '{recipe.id}' was never published. Skip.")
+                return
+            raise
+
+        self.client.patch_collection(id=self.collection_id, data={"status": "to-sign"})
```
mythmon: What affect does this line have? I assume it is related to the special signing bypass you mentioned above. Can you add a comment explaining it?

normandy/recipes/tests/test_models.py
```diff
@@ -579,6 +579,42 @@ def test_disable(self):
...
+
+        recipe.approved_revision.disable(user=UserFactory())
+
+        mocked_remotesettings.return_value.unpublish.assert_called_with(recipe)
+
+        # Unpublishes once when disabled twice.
+        with pytest.raises(EnabledState.NotActionable):
+            recipe.approved_revision.disable(user=UserFactory())
+
+        assert mocked_remotesettings.return_value.publish.call_count == 1
```
mythmon: Can you add two more tests?

1. A test about when a new revision is approved on a recipe that is already enabled? It should automatically publish the newly approved revision without it needing to be enabled again.
2. A test of a recipe that is enabled, disabled, and then re-enabled?

normandy/recipes/tests/test_exports.py
```diff
@@ -0,0 +1,246 @@
...
+            f"/collections/{settings.REMOTE_SETTINGS_COLLECTION_ID}"
+            f"/records/{recipe.id}"
+        )
+        requestsmock.request("put", record_url, status_code=503)
+
+        remotesettings = exports.RemoteSettings()
+        with pytest.raises(kinto_http.KintoException):
+            remotesettings.publish(recipe)
+
+        assert requestsmock.call_count == 4  # retried thrice.
```
mythmon: It might be nice to check the value of `settings.REMOTE_SETTINGS_RETRY_REQUESTS` to figure out this value, in case we change in the future. It also makes it clearer why we retry three times.

normandy/recipes/exports.py
leplatrem: When the Remote Settings client fetches the changes from the server, there is no distinction between a record creation from update. 

If the client had sync and caught the deletion, it will be interpreted as a creation. If the client hadn't caught the deletion, the re-creation will be interpreted as an update.

Does it make sense?

normandy/recipes/exports.py
```diff
@@ -0,0 +1,120 @@
...
+    def publish(self, recipe):
+        """
+        Publish the specified `recipe` on the remote server by upserting a record.
+        """
+        if self.client is None:
+            return  # no-op if disabled.
+
+        record = recipe_as_record(recipe)
+        self.client.update_record(data=record)
+        self.client.patch_collection(id=self.collection_id, data={"status": "to-sign"})
```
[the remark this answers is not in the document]
leplatrem: Good catch, I will see if I can do this elegantly ;) we have no API to rollback staged changes (yet)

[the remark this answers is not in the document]
leplatrem: Note: There is no reason for it to fail, plus we have retry > 0...

leplatrem:
I added some safety checks in cf481b9.

This is very unlikely to happen and makes the code a bit heavier/uglier.

Why unlikely? It only affect a situation where a record is published and the review approval fails. The RS server configuration for review settings is checked thoroughly on Normandy startup. The network requests are retried 3 times in case of error. So this could happen if the RS configuration is changed and the specific config to disable multi-signoff for Normandy is lost.

 I wouldn't mind if we'd revert the commit. Your call :)

peterbe:
> I added some safety checks in cf481b9.

> This is very unlikely to happen and makes the code a bit heavier/uglier.

I'm still a newbie when it comes to Kinto :) 
So what it attempts to do (if we ignore the rollback ad-hoc stuff) is that it updates Kinto with a record then approves? Why can't it be auto-approved? Why does it need approvals at all? 

Does approval involve signing? (If so, the `patch_collection` might depend on autograph which might fail)

> The RS server configuration for review settings is checked thoroughly on Normandy startup

That checks that it's set up correctly but that won't prevent operational errors from happening, right? Like, in symbols.mozilla.org there's a startup check that checks that the server is configured to be able to access the S3 bucket but sometimes, in runtime, I do get the occasional strange operational error (e.g. SSLError).

What I'm worried about is that I don't understand the risks involved with that second command. Is it more complicated than the `update_record()`? Or is it just a fancy wrapper for a single `UPDATE records SET ... WHERE ...=...`?

Like, in Web-Performance, you worry about the speed of the DNS lookup of the *first* image. But if that works, image 2, image 3, image N, is extremely likely to work and be fast (hi keep-alive connections). But it does happen sometimes that downloading images 1-37 worked fine but image 38 just suddenly, and rather randomly, got stuck. 

I actually think it's your call @leplatrem You have a better feeling for the risks and what plan for how we can audit collections put into RS that aren't approved and how to deal with that.

normandy/recipes/exports.py
```diff
@@ -0,0 +1,172 @@
...
+        self.client.update_record(data=record)
+        try:
+            # 2. Approve the changes immediately (multi-signoff is disabled).
+            self.client.patch_collection(id=self.collection_id, data={"status": "to-sign"})
+
+        except kinto_http.exceptions.KintoException:
+            # Approval failed. Cancel changes and raise.
+            try:
+                # Restore the record already published.
+                in_prod = self.client.get_record(id=record["id"], bucket=MAIN_BUCKET_ID)
```
peterbe: Why is `bucket=MAIN_BUCKET_ID` needed in `get_record` but not in `update_record`?

normandy/recipes/exports.py
```diff
@@ -0,0 +1,172 @@
...
+                self.client.update_record(data=in_prod)
+            except kinto_http.exceptions.KintoException as e:
+                # Record was never published, delete it.
+                if e.response.status_code == 404:
+                    self.client.delete_record(id=record["id"])
+            raise
```
peterbe: This is the wrong exception. This will potentially raise the second exception.

normandy/recipes/exports.py
```diff
@@ -0,0 +1,172 @@
...
+        try:
+            # 2. Approve the changes immediately (multi-signoff is disabled).
+            self.client.patch_collection(id=self.collection_id, data={"status": "to-sign"})
+
+        except kinto_http.exceptions.KintoException:
+            # Approval failed. Cancel changes and raise.
+            try:
+                # Restore the record already published.
+                in_prod = self.client.get_record(id=record["id"], bucket=MAIN_BUCKET_ID)
+                self.client.update_record(data=in_prod)
```
peterbe: I don't understand this line. You're updating it (on line 127) and that worked. They you're getting the record out by ID and updating it again. What's the point of that?

normandy/recipes/exports.py
mythmon: I'm less concerned about the client and more about the server. Since I know that kinto uses tombstones for deleted records, I wanted to check that the server won't mind if we try and resurrect a tombstone. It sounds like you're not worried though, so I expect we're fine here.

normandy/recipes/exports.py
mythmon: How could this be the second exception? A bare `raise` in an exception handler always raises the exception that triggered the `except` block. I tested it here:

```python
>>> def foo():
...   try:
...     raise Exception("first")
...   except:
...     try:
...       raise Exception("second")
...     except:
...       pass
...     raise
... 
>>> foo()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 3, in foo
Exception: first
```

mythmon:
I'm happy with this, though I'd also like to see @peterbe's comments covered as well. For what its worth, I'm happy with a normal number of retries (instead of the loop hammer), but I do think we need logic to detect that the second request failed and roll back the first. It may be rare, but it seems like it would be an important step.

Do we also want a management command that would ensure consistency, in case something slips through the incremental cracks? In the best case it would do nothing, but it would be great to have it available. I'd suggest we run it on a daily cron, as a backup. We could also use this as the initial sync when we enable RS integration. We could land that in a separate commit, if you like.

leplatrem:
> I'm happy with this, though I'd also like to see @peterbe's comments covered as well. 

I added more comments. Hopefully it covers Peter's questions

> Why is bucket=MAIN_BUCKET_ID needed in get_record but not in update_record?

This is kinto-http. If you specify the bucket or the collection, it overrides the bucket and collection from the client attributes (see instantiation). 

> I do think we need logic to detect that the second request failed and roll back the first.

At some point there should be an API to revert the changes that were not approved, and reset the main-workspace collection to the same state as the public one. We don't have it yet https://github.com/Kinto/kinto-signer/issues/260

> I'm still a newbie when it comes to Kinto :)
>

It's not obvious. But this is more about remote settings, and more precisely kinto-signer, which mixes the notion of publication workflows and signing... 

> So what it attempts to do (if we ignore the rollback ad-hoc stuff) is that it updates Kinto with a record then approves? Why can't it be auto-approved? Why does it need approvals at all?
>

Yes, it creates the records in `main-workspace/normandy-recipes` and flipping `status: "to-sign"` will copy and sign to `main/normandy-recipes`.
The current code relies on auto-approval already. With approval, there would be other intermediary steps, like a review request, that a reviewer (different from changes author) has to approve etc.

> Does approval involve signing? (If so, the patch_collection might depend on autograph which might fail)

Yes it does.

It hasn't failed in more than two years, but it could indeed.

>> The RS server configuration for review settings is checked thoroughly on Normandy startup
>>
> That checks that it's set up correctly but that won't prevent operational errors from happening, right? 

Yes. I mentioned that because the main reason for the patch operation to fail is related to configuration (permissions, group membership etc.)

> You have a better feeling for the risks and what plan for how we can audit collections put into RS that aren't approved and how to deal with that.

At first I was very confident, since we haven't seen any error for a long time. But I realized there was a situation that we would like to avoid at all cost, and covering it would justify complexifying a bit the code. Consider for example the following scenario without safety check:

- Alice enables a recipe A. Auto-approve fails. The UI shows an error. The recipe A is in the main-workspace but not approved (clients don't have it).
- Alice does not retry to enable, and opens a bugzilla ticket to ask for feedback
- In the mean time, Bob enables another recipe. Auto-approves succeeds this time.
- The whole content of main-workspace is signed and published. 
- Recipe A and recipe B are now enabled!! Even if recipe A is not marked as enabled in the Normandy DB!

Again, this is very unlikely to happen, but the probability is not zero since there is no «transaction» around the two HTTP operations (create + approve). In that sense, your hammering code was a lot simpler, and BTW I would not be against replacing my approach with it.

We are not in a hurry to land this code, I'd like you to fully understand every corner case before it gets merged :)

> Do we also want a management command that would ensure consistency, in case something slips through the incremental cracks? In the best case it would do nothing, but it would be great to have it available.

Yes, as you said, we would need it to initialize the RS data once at least. 
For the operational aspect, if that gives us confidence to have a cron job, let's do it! It should probably compare the two API content then...

> We could land that in a separate issue and PR, if you like. 

I don't mind stack more stuff here if you have intermediate reviews... we can squash the commits before merging anyway....

normandy/recipes/exports.py
peterbe: Great. I had old Python 2.7 in my head. I remember Socorro having this and it was really wrong. Socorro was/is 2.7.

docs/dev/install.rst
```diff
@@ -194,3 +194,70 @@ If you ever need to bypass Therapist, you can do so by passing
...
+
+      SERVER=http://localhost:8888/v1
+
+      curl -X PUT ${SERVER}/accounts/normandy \
+           -d '{"data": {"password": "n0rm4ndy"}}' \
+           -H 'Content-Type:application/json'
+
+      curl -X PUT ${SERVER}/buckets/main-workspace/collections/normandy-recipes \
+           -H 'Content-Type:application/json' \
+           -u 'normandy:n0rm4ndy'
```
peterbe: These failed for me. Output:

```
{"permissions":{"write":["account:normandy"]},"data":{"password":"$2b$12...notsureitmatters...Zff5b4Mcem","id":"normandy","last_modified":1543510196665}}

{"code":403,"errno":121,"error":"Forbidden","message":"This user cannot access this resource."}
```

I followed the instructions above. I added the two `kinto.signer...` lines under the `[app:main]` section and made sure I stopped and started the server.

leplatrem: Looks like autograph can't be reached

peterbe: :)

I think the pressing issue is that the (normandy) instructions don't explain how to set up autograph. Perhaps it's because normandy already has some and I'm just failing that independent of remote-settings :)

mythmon: Normandy's docs for setting up Autograph are here: https://normandy.readthedocs.io/en/latest/dev/install.html#autograph. You'll need to configure remote-settings to talk to Autograph as well.

peterbe:
Mayhaps I'm missing something obvious but the way I understand it, to test this full you need Django running, talking to remote settings (port :8888), which needs to talk to autograph. 

If I start three terminals with:

* `./manage.py runserver`
* `docker run -p 8765:8000 -it mozilla/autograph`
* ```docker run -v `pwd`/server.ini:/etc/kinto.ini \
           -v `pwd`/attachments:/tmp/attachments \
           -e KINTO_INI=/etc/kinto.ini \
           -p 8888:8888 \
           mozilla/kinto-dist```

that won't even work because the networking isn't working. That kinto server is, by default, expected to talk to `autograph-server:8000` (or something configurable). 

The only way I know to do docker networking is by putting all of these things in a `docker-compose` and let the service names automatically become the hostnames. Trying to get this set up and working. 

I've got something sorta working but I'm not there yet. A `docker-compose.yml` exclusively for local development that has all bits working integrated.

peterbe:
It works!!
Short version...

**Exhibit A**
screen shot 2018-11-30 at 3 15 51 pm

**Exhibit B**
screen shot 2018-11-30 at 3 16 24 pm

I'll explain more what I had to do to test this...

peterbe:
To test the full monty, you need normandy talking to `remotesettings:8888` and `remotesettings:8888` talking to `autograph:8000`. To accomplish this I created my own `docker-compose.yml` file and it looks like this:

```yml
version: "2"

services:
  db:
    image: postgres:9.4

  web:
    build:
      context: .
      dockerfile: Dockerfile
    image: normandy:web
    environment:
      - DJANGO_SETTINGS_MODULE=normandy.settings
      - DJANGO_CONFIGURATION=Development

      - DATABASE_URL=postgres://postgres@db/postgres
      - DJANGO_SECRET_KEY=vattevve

      - DJANGO_REMOTE_SETTINGS_URL=http://remotesettings:8888/v1
      - DJANGO_REMOTE_SETTINGS_USERNAME=normandy
      - DJANGO_REMOTE_SETTINGS_PASSWORD=n0rm4ndy

      - DJANGO_AUTOGRAPH_URL=http://autograph:8000/
      - DJANGO_AUTOGRAPH_HAWK_ID=normandev
      - DJANGO_AUTOGRAPH_HAWK_SECRET_KEY=3dhoaupudifjjvm7xznd9bn73159xn3xwr77b61kzdjwzzsjts
    env_file:
      - .env
    depends_on:
      - db
      - autograph
      - remotesettings
    ports:
      - "8000:8000"
    volumes:
      - $PWD:/app
    command: dev-start

  remotesettings:
    image: mozilla/kinto-dist:latest
    ports:
      - "8888:8888"
    environment:
      - KINTO_INI=/etc/kinto.ini
    volumes:
      - $PWD/remote-settings/server.ini:/etc/kinto.ini
      - $PWD/remote-settings/attachments:/tmp/attachments

  autograph:
    image: mozilla/autograph:latest
    ports:
      - "8765:8000"
```

I also had to add a new script. Or rather, I added a new command (`dev-start`) to the existing `bin/run.sh` and it looks like this:

```bash
  dev-start)
    echo "Setting up remote-settings"
    SERVER=http://remotesettings:8888/v1
    curl -X PUT ${SERVER}/accounts/normandy \
      -d '{"data": {"password": "n0rm4ndy"}}' \
      -H 'Content-Type:application/json'
    curl -X PUT ${SERVER}/buckets/main-workspace \
      -H 'Content-Type:application/json' \
      -u 'normandy:n0rm4ndy'
    curl -X PUT ${SERVER}/buckets/main-workspace/collections/normandy-recipes \
      -H 'Content-Type:application/json' \
      -u 'normandy:n0rm4ndy'


    echo "Starting the dev server"
    ./manage.py migrate
    ./bin/download_geolite2.sh
    ./manage.py update_actions
    ./manage.py initial_data
    ./manage.py runserver 0.0.0.0:8000
    ;;
```

You start the whole thing like this:

```bash
$ docker-compose build web
$ docker-compose up autograph remotesettings web
```

The bash stuff could probably be refactored a bunch so you don't have to do all those set-up'y things every time. E.g. executing `./bin/download_geolite2.sh`.

The tricky thing that stumbled me a couple of times is that remotesettings default `server.ini` (as per the installation instructions) uses the default in-memory storage so the creation of user, bucket, and collection are all "empheral" and if you try to do it a second time you get 4xx errors I think. 
Perhaps, to solve this, we can let the `remotesettings` server use the same Postgres database, in docker-compose, as the normandy server uses. 

Another complication was the superuser creation. The way I solved it was ugly. I used delivery-console to log in at least once. That creates a user. Then I used this:

```bash
$ docker-compose run web bash
app@a50c237f44c0:/app$ ./manage.py shell
>>> from django.contrib.auth.models import *
>>> User.objects.filter(is_superuser=False).update(is_superuser=True)
```
Crude but works. 

There were some other small changes necessary to `.dockerignore` for the sake of the static assets and remote actions schemas to not disappear.

Is there interest in keeping my `docker-compose` tricks in the repo so it becomes easier to do this kind of full end-to-end testing? If so I'll file a followup PR later. But it's out-of-context for this PR.

peterbe:
By the way, I can confirm that unpublishing it in Delivery Console correctly removed it from the remote settings collection. yay.

docs/dev/install.rst
```diff
@@ -194,3 +194,74 @@ If you ever need to bypass Therapist, you can do so by passing
...
+
+.. code-block:: bash
+
+      curl ${SERVER}/buckets/main/collections/normandy-recipes/records
+
+On the Firefox side, pulling the recipes from the local instance of Remote Settings will require these preferences to
+be set:
+
+- ``services.settings.server``: ``http://localhost:8888/v1``
+- ``services.settings.verify_signature``: false
```
peterbe: I don't have this. 
screen shot 2018-11-30 at 3 59 58 pm

docs/dev/install.rst
```diff
@@ -194,3 +194,74 @@ If you ever need to bypass Therapist, you can do so by passing
...
+- ``app.normandy.remotesettings.enabled``: true
```
peterbe: I don't have this either
screen shot 2018-11-30 at 4 01 11 pm

normandy/recipes/exports.py
```diff
@@ -0,0 +1,207 @@
...
+    Since Normandy already has the required approval/signoff features, the integration
+    bypasses the one of Remote Settings (leveraging a specific server configuration for this
+    particular collection).
+
+    .. notes::
+
+        Remote Settings signoff workflow relies on several buckets (see kinto-signer API).
+
+        The `main-workspace` is only readable and writable by authorized accounts.
+        The `main` bucket is read-only, but publicly readable. The Remote Settings
```
peterbe: If docstrings are restructured text this should be double backticks. (and the line above)

normandy/recipes/exports.py
```diff
@@ -0,0 +1,207 @@
...
+        clients pull data from there.
+
+        Since the review step is disabled for Normandy, publishing data is done in two steps:
+
+        1. Create, update or delete records in the `main-workspace` bucket
```
peterbe: ditto

normandy/recipes/exports.py
```diff
@@ -0,0 +1,207 @@
...
+        2. Approve the changes by flipping the ``status`` field to ``to-sign``
+           in the collection metadata
+        3. The server will sign and publish the new data to the `main` bucket.
```
peterbe: ditto

normandy/recipes/exports.py
```diff
@@ -0,0 +1,207 @@
...
+
+    """
+
+    MAIN_BUCKET_ID = "main"
+    """The name of the publicly readable bucket of Remote Settings where the approved
+    records are published.
+    """
```
peterbe: Why is this a string and not a regular `# comment...`?

normandy/recipes/exports.py
```diff
@@ -0,0 +1,207 @@
...
+        # Kinto is the underlying implementation of Remote Settings. The client
+        # is basically a tiny abstraction on top of the requests library.
+        self.client = (
+            kinto_http.Client(
+                server_url=str(settings.REMOTE_SETTINGS_URL),
+                auth=(
+                    str(settings.REMOTE_SETTINGS_USERNAME),
+                    str(settings.REMOTE_SETTINGS_PASSWORD),
+                ),
+                bucket=str(settings.REMOTE_SETTINGS_BUCKET_ID),
```
peterbe: Why all these `str()` here? Django-configurations makes sure they have the right types.

normandy/recipes/exports.py
```diff
@@ -0,0 +1,207 @@
...
+                collection=self.collection_id,
+                retry=int(settings.REMOTE_SETTINGS_RETRY_REQUESTS),
```
peterbe: No need to cast to int here. It's already done by django-configurations.

normandy/recipes/exports.py
```diff
@@ -0,0 +1,207 @@
...
+                self.client.update_record(data=in_prod)
+
+            except kinto_http.exceptions.KintoException as e:
+                if e.response.status_code == 404:
+                    # It was a new record that was never published in the `main` bucket,
+                    # revert its creation in the `main-workspace` bucket by deleting it.
+                    self.client.delete_record(id=record["id"])
+            raise
+
+        logger.info(f"Published record '{recipe.id}' for recipe {recipe}")
```
peterbe: Shouldn't that be `{recipe.name}` like it is on line 207?

I for one think we should rely on repr for both these things. Feels more solid. Like:

```python
logger.info(f"Published record {recipe.id!r} for recipe {recipe.name!r}")
```

since otherwise the "name" becomes a part of the English text.

normandy/recipes/management/commands/check_remote_settings.py
```diff
@@ -0,0 +1,81 @@
...
+
+    help = "Check or sync Remote Settings content"
+
+    def add_arguments(self, parser):
+        parser.add_argument("--sync", action="store_true", help="Synchronized remote content")
+
+    def handle(self, *args, sync=False, **options):
+        remote_settings = RemoteSettings()
+
+        local_recipes = [r for r in Recipe.objects.all() if r.enabled]
```
peterbe:
```python
local_recipes = Recipe.objects.filter(approved_revision__enabled_state__enabled=True)
```
should be the same thing. More efficient.

normandy/recipes/management/commands/check_remote_settings.py
```diff
@@ -0,0 +1,81 @@
...
+        # Lookup the recipes locally, those that are published but should not.
+        to_unpublish = []
+        for rid in remote_by_id.keys():
+            try:
+                to_unpublish.append(Recipe.objects.get(id=rid))
+            except Recipe.DoesNotExist:
+                to_unpublish.append(Recipe(id=rid))
+
+        # If there is nothing to do. Exit.
+        if len(to_publish) + len(to_update) + len(to_unpublish) == 0:
```
peterbe: nit: more pythonic would be

```python
if not to_publish and not to_update and not to_unpublish:
    ...
```

normandy/recipes/management/commands/check_remote_settings.py
```diff
@@ -0,0 +1,81 @@
...
+                    to_update.append(local_recipe)
+            else:
+                to_publish.append(local_recipe)
+        # Lookup the recipes locally, those that are published but should not.
+        to_unpublish = []
+        for rid in remote_by_id.keys():
+            try:
+                to_unpublish.append(Recipe.objects.get(id=rid))
+            except Recipe.DoesNotExist:
+                to_unpublish.append(Recipe(id=rid))
```
peterbe: This is really clever! You pretend to have it, by ID, just so you can send something to the `unpublish` method.

normandy/recipes/management/commands/check_remote_settings.py
```diff
@@ -0,0 +1,81 @@
...
+        if sync:
+            # Sync is enabled, recipes are un/published.
+            for r in to_publish + to_update:
+                remote_settings.publish(r)
+            for r in to_unpublish:
+                remote_settings.unpublish(r)
+
+        else:
+            # Show differences on stdout.
+            style = self.style.SUCCESS if len(to_publish) == 0 else self.style.MIGRATE_LABEL
```
peterbe: neat! but you *could* do `style = self.style.SUCCESS if not to_publish else self.style.MIGRATE_LABEL` 
or `style = to_publish and self.style.MIGRATE_LABEL or self.style.SUCCESS`

normandy/recipes/management/commands/check_remote_settings.py
```diff
@@ -0,0 +1,81 @@
...
+            style = self.style.SUCCESS if len(to_update) == 0 else self.style.MIGRATE_LABEL
+            self.stdout.write(style(f"{len(to_update)} recipes to update:"))
+            for r in to_update:
+                self.stdout.write(f" * {r} (id={r.id})")
+
+            style = self.style.SUCCESS if len(to_unpublish) == 0 else self.style.MIGRATE_LABEL
+            self.stdout.write(style(f"{len(to_unpublish)} recipes to unpublish:"))
+            for r in to_unpublish:
+                name = r.name if r.name else self.style.WARNING("Unknown locally")
+                self.stdout.write(f" * {name} (id={r.id})")
```
peterbe: Perhaps this is nit-pickery but I think the file should be called `sync_remote_settings.py` and it should have a `--dry-run` flag instead. 

All this printing stuff, which is great, should *always* happen. And then, the final thing of the whole function should be:

```python
if not options["dry_run"]:
    # Sync is enabled, recipes are un/published.
    for r in to_publish + to_update:
        remote_settings.publish(r)
    for r in to_unpublish:
        remote_settings.unpublish(r)
```
That way, the default functionality of this script is without an optional flag and that's the command we might one day put into a cron job. Just `./manage.py sync_remote_settings`.

normandy/recipes/models.py
```diff
@@ -449,18 +450,24 @@ def _create_new_enabled_state(self, **kwargs):
...
         self.recipe.save()
 
+    @transaction.atomic
     def enable(self, user, carryover_from=None):
         if self.enabled:
             raise EnabledState.NotActionable("This revision is already enabled.")
 
         self._create_new_enabled_state(creator=user, enabled=True, carryover_from=carryover_from)
 
+        RemoteSettings().publish(self.recipe)
```
peterbe: Why this? That means the `self.client` class instance object needs to be re-created for every recipe. Perhaps that's immaterial since it's so rare and cheap.

normandy/recipes/exports.py
leplatrem: I think it's the way docstrings at the module level work

leplatrem:
**TBD**

- [ ] Write general RS overview docs
- [ ] Add sync command docs
- [ ] Unit test sync command

mythmon:
Lets do it!

bors r+

PR #1626, opened 2018-11-19
Create user API endpoint

rehandalal:
r?

### DIFF
--- a/normandy/base/api/permissions.py
+++ b/normandy/base/api/permissions.py
@@ -21,10 +21,10 @@ class AdminEnabledOrReadOnly(AdminEnabled):
         return super().has_permission(request, view) or request.method in permissions.SAFE_METHODS
 
 
-class IsSuperUser(permissions.BasePermission):
+class CanChangeUser(permissions.BasePermission):
     """
-    Allows access only to super users.
+    Ensures that users have the appropriate permission to change users.
     """
 
     def has_permission(self, request, view):
-        return request.user and request.user.is_superuser
+        return request.user and request.user.has_perm("auth.change_user")
--- a/normandy/base/api/permissions.py
+++ b/normandy/base/api/permissions.py
@@ -19,3 +19,12 @@ class AdminEnabledOrReadOnly(AdminEnabled):
 
     def has_permission(self, request, view):
         return super().has_permission(request, view) or request.method in permissions.SAFE_METHODS
+
+
+class IsSuperUser(permissions.BasePermission):
+    """
+    Allows access only to super users.
+    """
+
+    def has_permission(self, request, view):
+        return request.user and request.user.is_superuser
--- a/normandy/base/api/views.py
+++ b/normandy/base/api/views.py
@@ -19,6 +19,7 @@ from normandy.base.api.permissions import CanChangeUser
 from normandy.base.api.serializers import (
     GroupSerializer,
     ServiceInfoSerializer,
+    UserOnlyNamesSerializer,
     UserSerializer,
     UserWithGroupsSerializer,
 )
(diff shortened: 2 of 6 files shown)


normandy/base/tests/api/test_api.py
```diff
@@ -235,3 +235,35 @@ def test_logged_out(self, settings):
...
+    def test_it_works(self, api_client):
+        user = User.objects.first()
+        res = api_client.get("/api/v3/user/")
+        assert res.status_code == 200
+        assert res.data == {
+            "count": 1,
+            "next": None,
+            "previous": None,
+            "results": [
+                {
```
rehandalal: @mythmon do we care about more than this info when creating users?

normandy/base/api/permissions.py
```diff
@@ -19,3 +19,12 @@ class AdminEnabledOrReadOnly(AdminEnabled):
 
     def has_permission(self, request, view):
         return super().has_permission(request, view) or request.method in permissions.SAFE_METHODS
+
+
+class IsSuperUser(permissions.BasePermission):
```
mythmon: I know we kept talking about only super users having this permission, but I think we should probably use the normal Django permissions for the user object instead of a custom permission. Then we would not grant that permission to anyone. Super users would be able to use the API by virtue of automatically passing all permission checks.

normandy/base/api/views.py
```diff
@@ -79,6 +81,23 @@ def get(self, request):
...
+    """Viewset for managing users."""
+
+    queryset = User.objects.all()
+    serializer_class = UserSerializer
+    permission_classes = (IsSuperUser,)
+
+    def destroy(self, request, *args, **kwargs):
+        instance = self.get_object()
+
+        # Should not be able to delete self
```
mythmon: Good call. :+1:

normandy/base/tests/api/test_api.py
mythmon: I don't care about any other information than this, though we should make sure that the username value is the same as the email. We don't need to expose it though.

mythmon: We'll also need group information somewhere. Maybe here? I'm not sure if we want to publish that though, so we might need to use a clever field that only shows if you have permissions.

rehandalal:
ready for review

peterbe:
I haven't looked at the code but I did (hint, hint) notice that the PR didn't include any changes to the developer documentation or the readme.

normandy/base/tests/api/test_api.py
```diff
@@ -235,3 +236,121 @@ def test_logged_out(self, settings):
             "logout_url": settings.OIDC_LOGOUT_URL,
             "github_url": settings.GITHUB_URL,
         }
+
+
+@pytest.mark.django_db
+class TestUserAPI(object):
```
mythmon: I think that the model viewset also allows creating and deleting users. Can you add tests for that?

normandy/base/tests/api/test_api.py
```diff
@@ -235,3 +236,121 @@ def test_logged_out(self, settings):
...
+        assert res.status_code == 200
+
+    def test_cannot_delete_self(self, api_client):
+        user = User.objects.first()
+        res = api_client.delete(f"/api/v3/user/{user.id}/")
+        assert res.status_code == 403
+
+
+@pytest.mark.django_db
+class TestGroupAPI(object):
```
mythmon: I think that the model viewset also allows creating and deleting groups. Can you add tests for that? Alternatively, do we even want the ability to do that?

normandy/base/api/views.py
```diff
@@ -79,6 +87,75 @@ def get(self, request):
         )
 
 
+class UserViewSet(viewsets.ModelViewSet):
+    """Viewset for managing users."""
+
+    queryset = User.objects.all()
+    serializer_class = UserWithGroupsSerializer
+    permission_classes = (CanChangeUser,)
```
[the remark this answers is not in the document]
rehandalal: Partly for time since I'd have to override it and do a bunch of tweaking and I felt like it was maybe overkill but if we want that level of granularity I can go do it.

normandy/base/tests/api/test_api.py
rehandalal: I figured we wanted that ability but I can go pull it out if not. I did add tests though.

mythmon: It seems only half useful, since even if you can create groups you can't manipulate permissions. However, half useful is better than nothing, so lets keep it.

mythmon:
Thanks for the changes. I think this looks good now.

bors r+

PR #1642, opened 2018-12-10
unbreak docs tests in 3.7

peterbe:
Note, it's still a mystery how we managed to allow bors to land those changes despite the `docs` job failed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,7 +100,7 @@ jobs:
 
   docs:
     docker:
-      - image: circleci/python:3.6
+      - image: circleci/python:3.7
 
     working_directory: ~/repo
 


mythmon:
To figure out why bors hasn't had a problem with this breakage, I dug around a bit in CircleCI.

I specifically looked at [this job](https://circleci.com/gh/mozilla/normandy/6753), which should have failed. Looking closely at it, I see that almost all of the packages came from a cached virtualenv, and so didn't get reinstalled and re-hash-checked. I'm not sure why this cache only works on master and not PRs.

I think we should re-evaluate how we cache Python requirements in CI. We should cache the pypi cache instead of caching the virtualenv.

I'm satisfied with this level of explanation, and will follow up with some changes related to the cacing.

PR #1644, opened 2018-12-10
Use Python 3.7 on readthedocs

mythmon:


mythmon:
Bah. After this merged, I got this error from RTD:

> Error: Problem in your project's configuration. Invalid "python.version": expected one of (2, 2.7, 3, 3.3, 3.4, 3.5, 3.6), got 3.7

So it looks like we're going to have to continue building the docs with Python 3.6. If we have to do that, I'm not sure if we can still build the app with Python 3.7 sanely. We'll have to put some more thought into this.

peterbe:
What about something like:

```
# requirements/docs.txt
-r requirements/default.txt

bcrypt==3.1.4 \
    --hash=sha256:cc2f24dc1c6c88c56248e93f28d439ee4018338567b0bbb490ea26a381a29b1e \
    --hash=sha256:6b662a5669186439f4f583636c8d6ea77cf92f7cfe6aae8d22edf16c36840574 \
    --hash=sha256:67ed1a374c9155ec0840214ce804616de49c3df9c5bc66740687c1c9b1cd9e8d \
    --hash=sha256:d860c7fff18d49e20339fc6dffc2d485635e36d4b2cccf58f45db815b64100b4
...
```
...where the hashes for that `bcrypt==3.1.4` are prepared for Python 3.6. 

Then we can use 3.6 in readthedocs but **overlay** the 3.7 specific hashes for 3.6 use.

mythmon:
That sounds promising. Lets give it a shot. Can you prepare the patch for that?

PR #1645, opened 2018-12-10
Replace NSP with yarn audit

### LINKED ISSUE #1637: Dependency deprecation warning: nsp (npm)
On registry `https://registry.npmjs.org/`, the "latest" version (v3.2.1) of dependency `nsp` has the following deprecation notice:

`The Node Security Platform service is shutting down 9/30 - https://blog.npmjs.org/post/175511531085/the-node-security-platform-service-is-shutting`

Marking the latest version of an npm package as deprecated results in the entire package being considered deprecated, so contact the package author you think this is a mistake.

Please take the actions necessary to rename or substitute this deprecated package and commit to your base branch. If you wish to ignore this deprecation warning and continue using `nsp` as-is, please add it to your [ignoreDeps](https://renovatebot.com/docs/configuration-options/#ignoredeps) array in Renovate config before closing this issue, otherwise another issue will be recreated the next time Renovate runs.

Affected package file(s): `package.json`

 Would you like to disable Renovate's deprecation warning issues? Add the following to your config:

```
"suppressNotifications": ["deprecationWarningIssues"]
```

mythmon:
Fixes #1637

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,12 +12,11 @@
     "test": "karma start",
     "lint": "yarn lint:js && yarn lint:js-security",
     "lint:js": "eslint normandy client karma.conf.js webpack.config.js",
-    "lint:js-security": "nsp check . --threshold 3"
+    "lint:js-security": "yarn audit"
   },
   "license": "MPL-2.0",
   "dependencies": {
     "mozilla-normandy-action-argument-schemas": "0.3.0",
-    "nsp": "3.2.1",
     "uuid": "3.3.2"
   },
   "devDependencies": {
@@ -53,5 +52,8 @@
     "webpack": "4.27.1",
     "webpack-bundle-tracker": "0.3.0",
     "webpack-cli": "3.1.2"
+  },
+  "resolutions": {
+    "deep-extend": ">=0.5.1"
   }
 }
--- a/bin/run.sh
+++ b/bin/run.sh
@@ -28,14 +28,7 @@ case $1 in
     yarn lint:js
 
     echo "Checking JS package security"
-    # Disable automatic failure
-    set +e
-    # nsp uses exit code 1 for security problems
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -53,5 +52,8 @@
     "webpack": "4.27.1",
     "webpack-bundle-tracker": "0.3.0",
     "webpack-cli": "3.1.2"
+  },
+  "resolutions": {
+    "deep-extend": ">=0.5.1"
   }
```
peterbe: What is this? I've never used `resolutions` before.

package.json
```diff
@@ -12,12 +12,11 @@
     "test": "karma start",
     "lint": "yarn lint:js && yarn lint:js-security",
     "lint:js": "eslint normandy client karma.conf.js webpack.config.js",
-    "lint:js-security": "nsp check . --threshold 3"
+    "lint:js-security": "yarn audit"
```
peterbe: nice!

package.json
mythmon: This is [Yarn's "selective dependency resolution" feature](https://yarnpkg.com/en/docs/selective-version-resolutions#search). It allows us to influence the algorithm it uses to resolve dependency versions. It's sort of like a Python constraints file in that regard. It's one of the neat things about Yarn that I don't know how to reproduce in NPM.

We had resolved `deep-extend@0.4.2` for one of our dependencies. That version had a minor security problem that `yarn audit` detected. This line forces Yarn to only install versions `>=0.5.1`, where the security problem was fixed (in practice it resolved it 0.6.0). You can see this change [in the lock file as well](https://github.com/mozilla/normandy/pull/1645/files#diff-8ee2343978836a779dc9f8d6b794c3b2L2080).

PR #1648, opened 2018-12-11
Add a heartbeat check for a working GeoIP DB

### LINKED ISSUE #1647: Heartbeat should fail if maxmind database isn't available
Currently if the Maxmind database isn't available, Normandy just logs an error and ClassifyClient returns None for the country. The dockerflow heartbeat should try to open the maxmind db and return a failure if it can't.

mythmon:
Fixes #1647

### DIFF
--- a/normandy/recipes/checks.py
+++ b/normandy/recipes/checks.py
@@ -8,7 +8,7 @@ from django.db.utils import OperationalError, ProgrammingError
 
 import requests.exceptions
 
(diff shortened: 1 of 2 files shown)


mythmon:
This failed no geoip database is available in CI, though there probably should be. I worked on this in #1649.

mythmon:
#1649 is merged now, so this should be ready.

bors r= @rehandalal

PR #1649, opened 2018-12-11
Pass CI envvar into test containers

mythmon:
I expect this will cause CI to fail on GeoIP tests until providing a maxmind db in CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,6 +46,7 @@ jobs:
           # requires an explicit list so it doesn't try to pull images that
           # have already been built.
           command: |
+            ./bin/download_geolite2.sh
             docker-compose build
             docker-compose pull db
 
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,7 +3,6 @@
 .env*
 .floo*
 docs/_build
-GeoLite2-Country.mmdb
 media
 node_modules
 pip-cache
--- a/README.md
+++ b/README.md
@@ -11,6 +11,9 @@ studies, user surveys and controlled rollout of new features.
 
 Normandy is licensed under the MPLv2. See the `LICENSE` file for details.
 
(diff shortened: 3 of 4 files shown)


mythmon:
I'm not sure if this is the right thing to do. It changes us to put a maxmind DB in our Docker image. @peterbe what do you think about that? @sciurus do you see any problems?

mythmon:
The DB used is licensed with [CC BY-SA 4.0](https://creativecommons.org/licenses/by-sa/4.0/).

sciurus:
From the ops perspective having the BY-SA licensed DB in the image does not matter, since we'll inject and use the licensed DB when we deploy.

From a legal perspective, I'm unsure if this is kosher or not. Someone else could take our image with the BY-SA license and run it, but Normandy wouldn't be providing attribution to Maxmind.

mythmon:
I added an attribution line to our readme file. How do you feel about that, Brian?

mythmon:
Sorry this wasn't very clear.

Our geoip test fixture has a line like `if geoip_db == None: skip_if_not_in_ci()`. The "if in CI" detection is dependent on the `CI` envvar being passed in to the tests correctly. Since that envvar wasn't being passed in, the tests were allowed to be skipped in CI, which is not what we wanted.

Now that I fixed the CI detection in the tests, a bunch of tests failed. Most of the PR is fixing the broken tests.

sciurus:
The note in the README seems to satisfy the requirement at https://dev.maxmind.com/geoip/geoip2/geolite2/ , so LGTM. I'll let wait on peterbe to approve this review since he has questions.

mythmon:
I think that is the best we can do for review here.

@sciurus ?

mythmon:
Thanks!

bors r= @peterbe, @sciurus

PR #1674, opened 2018-12-31
Fix originpyup scheduled update 2018 12 31

peterbe:
I pulled down the `pyup-scheduled-update-2018-12-31` branch from `origin`, started a new branch, made some changes and committed. But I couldn't write to origin/pyup-scheduled-update-2018-12-31` branch. Why is that?

```
▶ git push origin pyup-scheduled-update-2018-12-31
error: src refspec pyup-scheduled-update-2018-12-31 does not match any.
error: failed to push some refs to 'git@github.com:mozilla/normandy.git'
```

(so I pushed to me fork instead and made this PR)


mythmon:
The command `git push origin pyup-scheduled-update-2018-12-31` could be described as "Push the local branch named `pyup-scheduled-update-2018-12-31` to the remote branch named `pyup-scheduled-update-2018-12-31` on the remote named `origin`."

I suspect that when you created a branch, you named it something different than `pyup-scheduled-update-2018-12-31`. If that's the case, the command you probably want is `git push origin my-local-branch-name:pyup-scheduled-update-2018-12-31`.

mythmon:
Your change will be reverted the next time pyup updates requirements. You'll need to add a pyup comment, like here: https://github.com/mozilla/normandy/blob/c34e59e0455ed88d370502c4d82b3fd5b19e4f6d/requirements/constraints.txt#L220

mythmon:
I think we should make these fixes to #1680.

PR #1676, opened 2019-01-04
Exit sync_remote_settings with non-zero with dry-run if unsync

leplatrem:
follow-up #1616 

If the command `sync_remote_settings --dry-run` is meant to be ran in cronjob (or similar) it should exit with non-zero if some recipes are unsync. This will allow us to use the dry-run command as a way to check that everything is kept in sync successfully

### DIFF
--- a/normandy/recipes/management/commands/sync_remote_settings.py
+++ b/normandy/recipes/management/commands/sync_remote_settings.py
@@ -1,3 +1,5 @@
+import sys
+
 from django.core.management.base import BaseCommand
 
 from normandy.recipes.models import Recipe
@@ -79,3 +81,7 @@ class Command(BaseCommand):
             self.stdout.write(f" * {name!r} (id={r.id!r})")
             if not dry_run:
                 remote_settings.unpublish(r)
+
+        # When running with dry-run, we return non-zero if some differences exist.
+        if dry_run and (to_publish or to_update or to_unpublish):
+            sys.exit(1)
--- a/normandy/recipes/tests/test_commands.py
+++ b/normandy/recipes/tests/test_commands.py
@@ -349,7 +349,6 @@ class TestSyncRemoteSettings(object):
             call_command("sync_remote_settings")
 
     def test_it_does_nothing_on_dry_run(self, rs_settings, requestsmock, mocked_remotesettings):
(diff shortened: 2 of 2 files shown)


normandy/recipes/management/commands/sync_remote_settings.py
```diff
@@ -79,3 +81,7 @@ def handle(self, *args, dry_run=False, **options):
             self.stdout.write(f" * {name!r} (id={r.id!r})")
             if not dry_run:
                 remote_settings.unpublish(r)
+
+        # When running with dry-run, we return non-zero if some differences exist.
+        if dry_run and (to_publish or to_update or to_unpublish):
+            sys.exit(1)
```
sciurus: I would use an exit code other than 1, so we can distinguish differences existing from the script exiting due to an unhandled exception.

peterbe: I would raise an exception instead. It's a lot less subtle and it automatically gives us non-zero exit codes for the DevOps purposes. That'd feel more pythonic. 

If raising an exception triggers something to Sentry during this testing then that's unfortunate but it's easily solved with the `ignore_exceptions` tuple config. (I think that's what it's called).

peterbe:
Why would we run this with `--dry-run` in production (cron jobs)? What is the point? I'm sorry if the question sounds "aggressive". I'm just trying to find out. 

You could argue that if there's something to sync, in production, then just sync it. :)

peterbe:
> We'd rather want it to be a way to receive an alert if that happens, and investigate why we ended up with an unsynced recipe.

Then we should definitely raise an exception and allow raven to send this to Sentry. 

Also, I think the code needs a code comment that mentions what you just mentioned. I.e. that the reason we do this is for slow rollout where we're curious to make sure it's always synced. (or something like that)

mythmon:
I don't think this is a good idea. To be `--dry-run` is only useful for interactive sessions, and we shouldn't use it as a cron job check. Further, we don't have any precedent for cron jobs that check a status. Maybe we should have that, but that's a bigger conversation.

If we want a command that compares the state of the DB and Remote Settings, I think we should be explicit about it, and make `./manage.py check-remote-settings`.

PR #1685, opened 2019-01-10
signature checking can fail if signature disappears

### LINKED ISSUE #1683: DoesNotExist (Signature) error during ./manage.py initial_data
https://sentry.prod.mozaws.net/operations/normandy-stage/issues/5108064/

```
KeyError: 'signature'
  File "django/db/models/fields/related_descriptors.py", line 158, in __get__
    rel_obj = self.field.get_cached_value(instance)
  File "django/db/models/fields/mixins.py", line 13, in get_cached_value
    return instance._state.fields_cache[cache_name]

DoesNotExist: Signature matching query does not exist.
(8 additional frame(s) were not displayed)
...
  File "normandy/recipes/checks.py", line 98, in signatures_use_good_certificates
    urls.add(recipe.signature.x5u)
  File "django/db/models/fields/related_descriptors.py", line 164, in __get__
    rel_obj = self.get_object(instance)
  File "django/db/models/fields/related_descriptors.py", line 277, in get_object
    return super().get_object(instance)
  File "django/db/models/fields/related_descriptors.py", line 139, in get_object
    return qs.get(self.field.get_reverse_related_filter(instance))
  File "django/db/models/query.py", line 403, in get
    self.model._meta.object_name

DoesNotExist: Signature matching query does not exist.
```

peterbe:
Fixes #1683

The only explanation I can think of why the checks would fail is if list of recipes-with-signatures is queried and then as it processes the for-loop one of those recipes ceases to have a signature. Feels extremely unlikely. But this PR prevents that from being possible to happen in principle.

### DIFF
--- a/normandy/recipes/checks.py
+++ b/normandy/recipes/checks.py
@@ -99,9 +99,9 @@ def signatures_use_good_certificates(app_configs, **kwargs):
         Action = apps.get_model("recipes", "Action")
 
         urls = set()
-        for recipe in Recipe.objects.exclude(signature=None):
+        for recipe in Recipe.objects.exclude(signature__x5u=None).select_related("signature"):
             urls.add(recipe.signature.x5u)
-        for action in Action.objects.exclude(signature=None):
+        for action in Action.objects.exclude(signature__x5u=None).select_related("signature"):
             urls.add(action.signature.x5u)
     except (ProgrammingError, OperationalError, ImproperlyConfigured) as e:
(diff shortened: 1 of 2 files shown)


peterbe:
That second commit was just because I could :) It hurt my eyes to have to see that it gathered up all the (unique) URLs and then reversed that back to the recipes and actions within the exception handling code.

peterbe:
bors retry

PR #1696, opened 2019-01-16
Validate Heartbeat survey IDs as being unique

### LINKED ISSUE #1655: Validate Heartbeat survey IDs as being unique
It's potentially confusing that you can, in Delivery Console, create two distinct Heartbeat recipes with the exact same survey ID. 

PS. The Normandy client copes with this by simply refusing to run the second recipe since the survey ID has been run before.

peterbe:
Fixes #1655

Seems to work:
screen shot 2019-01-16 at 3 39 59 pm

There *is* a corner case. The checking is only done when you edit a recipe (revision), right? Suppose I do this:

1. Create **recipe A** with surveyId 001
2. Ask for approval, get approval, and publish it (still with surveyId 001). (I guess publishing is unrelated)
3. Now I edit it again and change the surveyId to 002. At this point `recipe.arguments['surveyId'] == '001'` because the approved revision is the "current" one. 
4. Create **recipe B** with surveyId 002 (which I'm allowed to do without an error). Approval or not, this is `recipe.arguments['surveyId'] == '002'` and that's based on its latest_revision.
5. I *go back* to **recipe A** and ask for approval for the latest revision. But I don't edit it. And this will unfortunately allow me to have duplicate surveyId in **recipe A** and **recipe B**.

Options:

1. Ignore this corner case because it's too unrealistic. 
2. Trigger whatever code does that arguments validation when you're changing the recipe's current_revision pointer. 
3. In the uniqueness code (the bulk of this patch) we could search other non-current revisions for that surveyId too. I.e. "You want surveyID 'xxx'? Lemme check *every* revision, approved or not, of every other recipe."

### DIFF
--- a/normandy/recipes/models.py
+++ b/normandy/recipes/models.py
@@ -594,6 +594,7 @@ class Action(DirtyFieldsMixin, models.Model):
         "duplicate_experiment_slug": "Experiment slugs must be globally unique",
         "duplicate_rollout_slug": "Rollout slugs must be globally unique",
         "rollout_slug_not_found": "Rollout slug not found for rollback",
+        "duplicate_survey_id": "Survey ID must be globally unique",
     }
 
     @property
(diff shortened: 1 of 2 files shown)


normandy/recipes/models.py
```diff
@@ -739,6 +740,23 @@ def default():
             if arguments["rolloutSlug"] not in rollout_slugs:
                 errors["slug"] = self.errors["rollout_slug_not_found"]
 
+        elif self.name == "show-heartbeat":
+            # Survey ID should be unique across all recipes
+            other_recipes = Recipe.objects.filter(latest_revision__action=self)
+            if revision.recipe and revision.recipe.id:
+                other_recipes = other_recipes.exclude(id=revision.recipe.id)
+            # Note, `recipe.arguments` refers to `recipe.current_revision.arguments`
```
mythmon: I think the placement of this comment is a bit misleading. This note applies to most of the checks in this function. However putting the comment here makes it seem like it applies uniquely to `show-heartbeat`. Can we move the comment to a more inclusive place?

Overall, I like to see us move away from the helpers defined on Recipe, and instead access revisions directly. This matches well with the sort of changes we made in the v3 API, and I think it will help us think more clearly about things. I'm not sure if it would help here though.

peterbe: > Can we move the comment to a more inclusive place?

Done. 

> ...move away from the helpers defined on Recipe, and instead access revisions directly. 

Yes please! I think I did it the way I did it because of the historical context (the lines above). 

I could easily change it to be:

```python
other_revisions = RecipeRevision.objects.filter(action=self).exclude(
    recipe_id=revision.recipe.id
)
for other_revision in other_revisions:
    if other_revision.arguments["surveyId"] == arguments["surveyId"]:
        errors["surveyId"] = self.errors["duplicate_survey_id"]
        break
```
Roughly. An option is to not do `errors["surveyId"] = self.errors["duplicate_survey_id"] + f' (used in {other_revision.recipe.name})'. 

One caveat is that a surveyId, that I want to use, is lingering inside a revision (in a different recipe) that is long ago replaced with a better revision. Does that make sense?

PR #1697, opened 2019-01-17
Make default GCP storage ACL "publicRead"

mythmon:
This makes URLs used for uploaded files much shorter and never expire.

### DIFF
--- a/normandy/settings.py
+++ b/normandy/settings.py
@@ -374,6 +374,7 @@ class Base(Core, CORS, OIDC):
     AWS_SECRET_ACCESS_KEY = values.Value()
     AWS_STORAGE_BUCKET_NAME = values.Value()
     GS_BUCKET_NAME = values.Value()
+    GS_DEFAULT_ACL = values.Value("publicRead")
 
     GITHUB_URL = values.Value("https://github.com/mozilla/normandy")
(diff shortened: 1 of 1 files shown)


sciurus:
This isn't needed because they are already public by default. E.G. I can download https://storage.googleapis.com/moz-fx-normandy-prod-addons/extensions/normandy-devtools-0.7.0-signed.xpi fine without supplying the GoogleAccessId that normandy is appending to it.

screen shot 2019-01-18 at 11 49 01

I think you probably have code that generates a singed version of the url that you just need to drop.

mythmon:
> I think you probably have code that generates a singed version of the url that you just need to drop.

The code that generates the signed version keys off of this setting, and is in a library. By default it assumes nothing about the buckets, and so generates maximally compatibly URLs. This setting tells it that it should assume that items in the bucket are public, and so it shouldn't sign them.

The code in question is here: [storages/backends/gcloud.py#L256-L267](https://github.com/jschneier/django-storages/blob/efcad7f6b39b79ef2ef5810c38e48065d04b1c8f/storages/backends/gcloud.py#L256-L267).

mythmon:
bors r= @sciurus

PR #1702, opened 2019-01-24
preference rollout/rollback enabled exclusion

### LINKED ISSUE #1400: Don't allow a preference rollout and its rollback to both be enabled at the same time
In [bug 1459920](https://bugzilla.mozilla.org/show_bug.cgi?id=1459920), Adrian brought up that having both a rollout and its rollback targeting the same user will produce strange results and noisy telemetry. Talking to Adrian and @glasserc, we came to the conclusion that this isn't a state that really makes any sense. Since rollbacks can't be undone, after one is created you'll have start over on a brand new recipe anyways.

Since this is a situation that doesn't make sense, and we want to not do, we should make it a rule and enforce it in the API. In the same way that we enforce other inter-recipe rules (like that recipe slugs must be unique).

To be specific, consider two recipes: `r1` is a preference rollout with slug `foo`. `r2` is a preference rollback that references `r1`'s slug. If `r1` is currently enabled, then `r2` cannot be enabled. If `r2` is currently enabled, then `r1` cannot be enabled. (Perhaps if `r2` exists at all?). It is fine to disable both recipes at any time.

peterbe:
Fixes #1400

The unit test is a bit dense but at least it encapsulates all the things. 

Note that this doesn't work great in Delivery Console. If you attempt to publish a rollback/rollout when there's already an enabled rollout/rollback the server will reject it but the UI isn't doing a great job of explaining it. That's what https://github.com/mozilla/delivery-console/issues/729 is about.

### DIFF
--- a/normandy/recipes/tests/test_models.py
+++ b/normandy/recipes/tests/test_models.py
@@ -740,6 +740,40 @@ class TestRecipeRevision(object):
         saved = Recipe.objects.get(id=recipe.id)
         assert saved.approved_revision.enabled
 
+    def test_enable_rollback_enable_rollout_invariance(self):
+        rollout_recipe = RecipeFactory(
+            name="Rollout",
+            approver=UserFactory(),
+            enabler=UserFactory(),
+            action=ActionFactory(name="preference-rollout"),
+            arguments={"slug": "myslug"},
+        )
+        assert rollout_recipe.enabled
+
+        rollback_recipe = RecipeFactory(
+            name="Rollback",
+            action=ActionFactory(name="preference-rollback"),
+            arguments={"rolloutSlug": "myslug"},
+        )
+        approval_request = rollback_recipe.latest_revision.request_approval(creator=UserFactory())
+        approval_request.approve(approver=UserFactory(), comment="r+")
+
+        with pytest.raises(ValidationError) as exc_info:
+            rollback_recipe.approved_revision.enable(user=UserFactory())
+        assert exc_info.value.message == "Rollout recipe 'Rollout' is currently enabled"
+
+        rollout_recipe.approved_revision.disable(user=UserFactory())
+        assert not rollout_recipe.enabled
+        # Now it should be possible to enable the rollback recipe.
+        rollback_recipe.approved_revision.enable(user=UserFactory())
+        assert rollback_recipe.enabled
+
+        # Can't make up your mind. Now try to enable the rollout recipe again even though
+        # the rollback recipe is enabled.
+        with pytest.raises(ValidationError) as exc_info:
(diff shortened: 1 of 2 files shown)


normandy/recipes/tests/test_models.py
```diff
@@ -740,6 +740,40 @@ def test_it_rollbacks_changes_if_error_happens_on_unpublish(self, mocked_remotes
...
+        )
+        assert rollout_recipe.enabled
+
+        rollback_recipe = RecipeFactory(
+            name="Rollback",
+            action=ActionFactory(name="preference-rollback"),
+            arguments={"rolloutSlug": "myslug"},
+        )
+        approval_request = rollback_recipe.latest_revision.request_approval(creator=UserFactory())
+        approval_request.approve(approver=UserFactory(), comment="r+")
```
mythmon: Isn't it possible to do these two lines with `RecipeFactory(..., approver=UserFactory())`?

peterbe: I honestly don't understand what you mean. You can't create a new recipe but you have to create a new approval request. No?

mythmon: Hm. I'm not sure if I'm confused or not. What I meant was this:

```diff
 rollback_recipe = RecipeFactory(
      name="Rollback",
      action=ActionFactory(name="preference-rollback"),
      arguments={"rolloutSlug": "myslug"},
+     approver=UserFactory(),
 )
-approval_request = rollback_recipe.latest_revision.request_approval(creator=UserFactory())
-approval_request.approve(approver=UserFactory(), comment="r+")
```

Does that work? Or am I missing something?

peterbe: Ah! I see. I went down there too. 

You can't create a rollback recipe unless you have a rollout recipe. And you can't create a new rollback recipe (to match a rollout recipe) if there's already another rollback recipe by that `rolloutSlug`. 

Hence, I "revive" that initial `rollback_recipe` instance.

mythmon: I really don't follow anything you just said. I understand that you can't create a rollback without a rollout, but there is already a rollout. I also understand that you can't have a duplicate rollback recipe, but that's ok too because you are only making a single rollout recipe.

I'm trying to figure out why when you make the first recipe, you create and approve it in one step, but then when you make the second recipe, you create and approve it in two steps. I don't think this has anything to do with the intricacies of the particular actions. It just seems like you are being inconsistent.

Feel free to merge this as is though.

mythmon: I also don't understand what you're reviving. There is only one rollback being created here, and it's all being done at once, just in three function calls instead of one.

PR #1703, opened 2019-01-24
Improve addon url rewriting script to adapt to the storage system in use

mythmon:
This changes the script from a simple string replacement to recomputing the URL based on the storage system currently configured. This makes the script work for the current use case, and should make it more reusable in the future, if we need it.

### DIFF
--- a/normandy/recipes/management/commands/update_addon_urls.py
+++ b/normandy/recipes/management/commands/update_addon_urls.py
@@ -1,8 +1,12 @@
-from urllib.parse import urlparse, urlunparse
 from django.core.management.base import BaseCommand
 from django.template.defaultfilters import pluralize
 
 from normandy.recipes.models import RecipeRevision
+from normandy.studies.models import Extension
+
+
+def get_filename_from_url(url):
+    url.split("/")[-1]
 
 
 class Command(BaseCommand):
@@ -11,15 +15,17 @@ class Command(BaseCommand):
     file storage to a new hostname.
     """
 
-    help = "Updates add-on URL in revisions"
+    help = "Updates add-on URL in revisions to match the current storage system"
 
-    def add_arguments(self, parser):
-        parser.add_argument("new_hostname", help="Hostname to convert add-on urls to")
+    def handle(self, *args, **options):
+        extension_by_filename = {}
+        for extension in Extension.objects.all():
+            filename = get_filename_from_url(extension.xpi.url)
+            extension_by_filename[filename] = extension
 
-    def handle(self, *args, new_hostname, **options):
         target_revisions = RecipeRevision.objects.filter(action__name="opt-out-study")
         update_count = 0
-        for rev in target_revisions.iterator():
+        for rev in target_revisions:
             # Pull into a local variable to modify the arguments since
             # `rev.arguments` is actually a property that parses JSON, not a
             # real attribute of the object
@@ -33,17 +39,23 @@ class Command(BaseCommand):
                 )
                 continue
 
-            parsed_url = urlparse(arguments["addonUrl"])
-            if parsed_url.netloc == new_hostname:
+            filename = get_filename_from_url(arguments["addonUrl"])
+
+            if filename not in extension_by_filename:
+                self.stderr.write(
+                    f"Warning: Recipe {rev.recipe.id} revision {rev.id} has an addonUrl that does "
+                    f"not match any in the database.",
+                    ending="\n",
+                )
+                continue
+
+            extension = extension_by_filename[filename]
+            new_url = extension.xpi.url
+            if arguments["addonUrl"] == new_url:
                 # nothing to do
                 continue
 
-            # _replace is not a private method. parsed_url is a named tuple,
-            # and named tuples expose all their public methods with underscores
-            # to avoid colliding with data keys.
-            new_url_parts = parsed_url._replace(netloc=new_hostname)
-            new_url = urlunparse(new_url_parts)
-            arguments["addonUrl"] = new_url
+            arguments["addonUrl"] = extension.xpi.url
             rev.arguments = arguments
             rev.save()
             update_count += 1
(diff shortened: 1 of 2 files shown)


normandy/recipes/management/commands/update_addon_urls.py
```diff
@@ -11,15 +15,17 @@ class Command(BaseCommand):
...
 
-    help = "Updates add-on URL in revisions"
+    help = "Updates add-on URL in revisions to match the current storage system"
 
-    def add_arguments(self, parser):
-        parser.add_argument("new_hostname", help="Hostname to convert add-on urls to")
+    def handle(self, *args, **options):
+        extension_by_filename = {}
+        for extension in Extension.objects.all():
+            filename = get_filename_from_url(extension.xpi.url)
```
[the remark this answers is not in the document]
mythmon: > So extension.xpi.url will point to the URL in whatever the currently configured storage system is, i.e. s3 or gcs?

Yes, that is computed at run time based on the current settings.

> And each recipe revision's addons is one of the Extensions?

In theory, yes. It's possible for that not to be true, which is why there is the check. I expect that everything in prod will match one of the `Extension`s though.

normandy/recipes/management/commands/update_addon_urls.py
```diff
@@ -1,8 +1,12 @@
-from urllib.parse import urlparse, urlunparse
 from django.core.management.base import BaseCommand
 from django.template.defaultfilters import pluralize
 
 from normandy.recipes.models import RecipeRevision
+from normandy.studies.models import Extension
+
+
+def get_filename_from_url(url):
+    url.split("/")[-1]
```
peterbe: Not only is this broken due to the lack of `return` but it also makes me nervous. How come the tests pass when `filename` always becomes `None` from the line `filename = get_filename_from_url(extension.xpi.url)`. 

Also, it makes me nervous. If the URL has `?foo=bar#anchor` on it we probably don't want that for a filename. It should probably be `urlparse(extension.xpi.url).pathname.split('/')[-1]` or something. (The `FileField.url` is unlikely to have that kind of stuff but perhaps we should mention it as a code comment then)

normandy/recipes/management/commands/update_addon_urls.py
```diff
@@ -33,17 +39,23 @@ def handle(self, *args, new_hostname, **options):
...
 
-            parsed_url = urlparse(arguments["addonUrl"])
-            if parsed_url.netloc == new_hostname:
+            filename = get_filename_from_url(arguments["addonUrl"])
+
+            if filename not in extension_by_filename:
+                self.stderr.write(
+                    f"Warning: Recipe {rev.recipe.id} revision {rev.id} has an addonUrl that does "
+                    f"not match any in the database.",
+                    ending="\n",
```
peterbe: I know it was like that before (line 38) but why this extra `ending="\n"`. It's added by default for you. 

> "Note also that you don’t need to end messages with a newline character, it will be added automatically"

https://docs.djangoproject.com/en/2.1/howto/custom-management-commands/

`self.stderr.write("foo")` is almost the same as `sys.stderr("foo\n").

normandy/recipes/management/commands/update_addon_urls.py
```diff
@@ -33,17 +39,23 @@ def handle(self, *args, new_hostname, **options):
...
                 # nothing to do
                 continue
 
-            # _replace is not a private method. parsed_url is a named tuple,
-            # and named tuples expose all their public methods with underscores
-            # to avoid colliding with data keys.
-            new_url_parts = parsed_url._replace(netloc=new_hostname)
-            new_url = urlunparse(new_url_parts)
-            arguments["addonUrl"] = new_url
+            arguments["addonUrl"] = extension.xpi.url
```
peterbe: At this point that precious string `"adonUrl"` has been hardcoded 4 times. Probably best to turn it into a constant before the loop starts.

normandy/recipes/management/commands/update_addon_urls.py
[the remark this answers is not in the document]
peterbe: Is it safe to use the filename? What if there were `https://s3.aws.com/testing/adblocker.xpi` and `https://s3.aws.com/beta/version/123/adblocker.xpi` in there?

Perhaps based on shear insight into what little we have in the database (the `Extension.url` fields) you *know* this isn't a risk.

normandy/recipes/management/commands/update_addon_urls.py
mythmon: This is indeed from writing too much rust lately. I'll improve the tests to catch this kind of issue. I think the better url parsing you proposed is also a good idea, I'll do that.

normandy/recipes/management/commands/update_addon_urls.py
mythmon: I also don't know why this was here, and decided not to think about it, in case there was something complicated. Now that you've thought about it and done the research, I'll remove it.

normandy/recipes/management/commands/update_addon_urls.py
mythmon: I'm very tempted to add pseudo-models to arguments at some point. I wonder how bad that would be...

Amusingly, you typed `"adonUrl"`, which is a typo of the string in the code. That sort of proves your point though. I'll change it.

peterbe: I don't know what a "pseudo-model" is.

mythmon: So in Django, we have models right? Which are bundles of logic and behavior associated with some data from the database.

I'd like to make something similar that we can give the data from the arguments json blob that has methods and property accessors and is generally smarter. But it would **not** be backed by the database. So it is like a model, but not quite. So here pseudo- is used to imply it isn't a true model, but just mimicking one.

PR #1738, opened 2019-02-13
Remove pagination from extensions v1 API

rehandalal:
r?

### DIFF
--- a/normandy/studies/api/v1/views.py
+++ b/normandy/studies/api/v1/views.py
@@ -1,6 +1,5 @@
-from django.db.models import Q
-
-from rest_framework import viewsets
+from rest_framework import mixins, status, viewsets
+from rest_framework.response import Response
 
 from normandy.base.api.filters import AliasedOrderingFilter
 from normandy.base.api.mixins import CachingViewsetMixin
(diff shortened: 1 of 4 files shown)


mythmon:
Is this simply a matter of consistency, or are you using this endpoint for something? I'm almost inclined to remove it entirely, since I don't think any of the v1 consumers will need it.

Is this going to be one of our hot paths? ie, is the client going to hit it regularly? I ask because it is going to grow without bound overtime and not be paginated, which doesn't seem like a good idea.

I guess that's the situation we are in with `/api/v1/recipe/` though, and that endpoint is worse. So maybe it's not a problem.

rehandalal:
yeah i don't actually use the list view so maybe i should get rid of everything but the detail view

normandy/studies/api/v1/views.py
```diff
@@ -12,17 +11,9 @@ class ExtensionOrderingFilter(AliasedOrderingFilter):
...
-    def get_queryset(self):
-        queryset = super().get_queryset()
-
-        if "text" in self.request.GET:
-            text = self.request.GET.get("text")
-            queryset = queryset.filter(Q(name__icontains=text) | Q(xpi__icontains=text))
 
-        return queryset
+    def list(self, *args, **kwargs):
+        return Response(status=status.HTTP_204_NO_CONTENT)
```
mythmon: Oh, that's a lot classier than a 404. Nice thinking.

PR #1755, opened 2019-02-24
Upgrade Django to v2.1.7

rehandalal:
Blocked by #1752.

r?


rehandalal:
Hmm... Looks like some autograph issues...

mythmon:
I think we don't usually use Autograph itself during tests, but instead are doing mocked signing by making the "signature" actually a normal hash of the data. I don't know why that would break in the update though. I skimmed through the release note and didn't find anything that seemed relevant.

peterbe:
@rehandalal can you fix the conflicts and then I can review this.

peterbe:
"review" :)

mythmon:
@rehandalal can you check what you get at `/api/docs/`? It isn't working for me, due to not being able to find jquery, but I also have some other weird stuff happening, so I'm not sure.

Besides that, everything I checked seems to work fine.

rehandalal:
image

Appears to be an inline script CSP issue.

rehandalal:
I hit refresh and it looked okay though...

image

Still seeing the CSP issues in the console however.

mythmon:
The CSP stuff is probably from #1765. You're getting way farther than I am, so I think this is ok.

PR #1768, opened 2019-02-28
Add migration to add extension ID to addon study recipes

rehandalal:
We should not merge this until the delivery console changes have landed and been pushed to production.

r?


normandy/recipes/migrations/0014_auto_20190228_1128.py
```diff
@@ -0,0 +1,85 @@
...
+        return  # Do nothing since there cannot be any recipes using the opt-out-study action
+
+    revisions = RecipeRevision.objects.filter(action_id=action.id)
+
+    for revision in revisions:
+        arguments = json.loads(revision.arguments_json)
+        url = arguments.get("addonUrl")
+        filename = unquote_plus(url.split("/extensions/").pop())
+        extension = Extension.objects.get(xpi=f"extensions/{filename}")
+        arguments["extensionApiId"] = extension.id
```
mythmon: @peterbe pointed out in another PR when I was doing something similar that this method of choosing add-ons is a bit weak. I wanted to share some code I wrote that did the same thing slightly differently. Specifically it parsed the URL and grabbed the path, and then split by slashes. That helps with potential query strings, even though that's unlikely.

https://github.com/mozilla/normandy/blob/5718076dcf8dd79caf0a4d8a2ab0f618276e906c/normandy/recipes/management/commands/update_addon_urls.py

rehandalal:
With the DC changes landed and shipped to prod I believe we can go ahead and land this migration. We should probably also tag a release and deploy this.

mythmon:
Can your remove the signature erasing?

PR #1770, opened 2019-02-28
Add experimenter ID field

### LINKED ISSUE #1767: Add experimenter ID field
When you've done all the planning and prep in experimenter the plan is that you somehow fill all of that into Delivery Console and save it as a Normandy Recipe. What would also be nice to track which experiment the recipe was spawned from. 

Features we can build on top of this. 

1. When looking at a recipe (in Delivery Console) to decide whether to approve it, you'd be able to have a link back to Experimenter to get the back story for the recipe's existence. 

2. Experimenter is going immutable. Once an experiment is "shipped" into Normandy, it can't be edited. (If you late decide you want to change the population you create a *new* experiment) Delivery Console *and* Normandy could follow suit and disallow than an approved (or published?) recipe, that has an Experimenter ID, can't be edited. 

3. It can help keep things in sync and make it possible to look up the recipe from the Experimenter ID even if slugs change. E.g. `GET /api/v3/recipe/?experimenter_id=1234`

peterbe:
Fixes #1767 

CC @jaredkerim 

I just looked at `bug_number` and how was added and filtered and tested etc and extended that as a template. 

CC @rehandalal this PR contains a migration that depends on `0013_auto_20181018_2049`
So if this PR lands first, you're going to have to update https://github.com/mozilla/normandy/pull/1768/files#diff-76d4200d24aa17cf5f37345fac609cc7R81


peterbe:
Ping @jaredkerim 

I'd love to land PRs with migrations in them sooner otherwise it complicates merge conflicts if they linger.

jaredkerim:
@peterbe @mythmon Sorry I took so long to get to this, yes Experimenter does not use integer IDs anywhere, it uses the slug, so that's what should be stored.  Alternatively you could store just the Experimenter URL for the experiment, if the purpose is just to provide a handy link back to Experimenter for more info.  I think you only need the slug if you want to call the Experimenter API for some reason?

peterbe:
I think we should **close** this PR. I hasted into this so much that I, today, can no longer remember the long-term plan for this field in Normandy. 

The ultimate goal is to be able to go from a recipe (in Delivery Console) to an experiment. Is this really something we need?? If you're finding yourself staring at a recipe you can probably copy the recipe *name* and do a [search](https://github.com/mozilla/experimenter/issues/400) or [browse](https://experimenter.services.mozilla.com/?ordering=-latest_change&type=&status=Ship&firefox_channel=&firefox_version=&project=&owner=). Perhaps more realistically if you don't know, you just ping a teammate on Slack. 

If anything, we can restart a PR just like this with the Experimenter navigation **slug** instead.

glasserc:
This got subsumed by #1975.

PR #1785, opened 2019-03-05
Expose last_updated in the v1 serialization

glasserc:
Refs https://bugzilla.mozilla.org/show_bug.cgi?id=1530508. The goal is
to ensure that the recipes being seen by the client are "up to
date".

### DIFF
--- a/normandy/recipes/api/v1/serializers.py
+++ b/normandy/recipes/api/v1/serializers.py
@@ -113,7 +113,8 @@ class MinimalRecipeSerializer(RecipeSerializer):
         # Attributes serialized here are made available to filter expressions via
         # normandy.recipe, and should be documented if they are intended to be
         # used in filter expressions.
-        fields = ["id", "name", "revision_id", "action", "arguments", "filter_expression"]
+        fields = ["id", "name", "revision_id", "action", "arguments", "filter_expression",
+                  "last_updated"]
 
     def get_revision_id(self, recipe):
         # Certain parts of Telemetry expect this to be a string, so coerce it to


glasserc:
This is the approach I took. Looking at this again now, I wonder if maybe I should report `revision_id` instead? What do you think?

rehandalal:
As things currently stand `last_updated` and `revision_id` are probably both fine. As revisions are immutable in practice (the API does not allow revisions to be directly manipulated) it is unlikely the `last_updated` date would ever change for a given revision. 

That being said I've been wondering lately whether we should be doing more with the `last_updated` field to reflect some of the parallel history around enabled state changes.

mythmon:
The code here is fine, though Black wants to reformat it. I expect you'll need to update the tests to account for the new field as well.

I think `revision_id` vs `last_updated` depends on what questions you are trying to ask. We can query our APIs with `revision_id`, and so we can use it to map directly to a specific revision. That makes it easy to correlate this data to other things. On the other hand it means asking questions like "how long has this been running" are harder, since you'd first have to do a join with another data source.

If you use `last_updated`, it quickly gives you something that can be compared to the current date, and so you can determine about how long a recipe has been running. You can't really query the API to find out more details about that time stamp though.

If you are trying to make sure that clients get the latest version of the recipes, I think `revision_id` is the better choice, since it is easier to map that to other data. It's also more discrete, whereas timestamps are more continuous. I think the discrete data would be more helpful here.

glasserc:
I guess I'll close this and use revision_id instead.

PR #1786, opened 2019-03-05
Increase v1 API contract test coverage

chartjes:
Added more contract tests for v1 API commands, fixed some issues detected while reviewing other tests, updated documentation


contract-tests/test_api.py
```diff
@@ -165,3 +165,106 @@ def test_recipe_history(conf, requests_session):
             created = datetime.strptime(revision["date_created"], "%Y-%m-%dT%H:%M:%S.%fZ")
             assert created < last_date
             last_date = created
+
+
+def test_docs(conf, requests_session):
+    r = requests_session.get(conf.getoption("server") + "/api/docs")
```
mythmon: This URL (and all the others in the tests) should end in `/`. As it stands, every test is going to be redirecting to the with-slash version of the URL.

contract-tests/test_api.py
```diff
@@ -165,3 +165,106 @@ def test_recipe_history(conf, requests_session):
...
+
+def test_approval_request(conf, requests_session):
+    r = requests_session.get(conf.getoption("server") + "/api/v1/approval_request")
+    r.raise_for_status()
+    data = r.json()
+    sample_record = data[0]
+    expected_fields = ["approved", "comment", "created", "creator", "id"]
+    expected_approver_fields = ["email", "first_name", "id", "last_name"]
+    expected_creator_fields = ["email", "first_name", "id", "last_name"]
+    assert r.status_code in [200, 302]
```
mythmon: I don't think allowing 30x requests is a good idea. If you include the trailing slashes on the URLs, then this should only return 200.

contract-tests/test_api.py
```diff
@@ -165,3 +165,106 @@ def test_recipe_history(conf, requests_session):
...
+
+    for field in expected_creator_fields:
+        assert field in sample_record["creator"]
+
+
+def test_approval_request_by_record(conf, requests_session):
+    r = requests_session.get(conf.getoption("server") + "/api/v1/approval_request/1")
+    r.raise_for_status()
+    data = r.json()
+    expected_fields = ["approved", "approver", "comment", "created", "creator", "id"]
```
mythmon: This seems like a pretty shallow way of testing the response of the API. Wouldn't it be better to validate the response against a schema instead? I don't think this would have caught any problems we've seen recently that weren't caught by other tests.

chartjes: If there is a downloadable schema available, I'd be happy to use it and rework the tests to focus on validation

mythmon: We don't have any schemas already made. Simply checking for the presence keys doesn't seem like a very interesting test though.

chartjes:
Okay, I will research into some different options for API validation.

contract-tests/test_api.py
```diff
@@ -165,3 +178,58 @@ def test_recipe_history(conf, requests_session):
...
+
+
+def test_recipe_revision_by_id(conf, requests_session):
+    r = requests_session.get(conf.getoption("server") + "/api/v1/recipe_revision/20/")
+    r.raise_for_status()
+    data = r.json()
+
+    assert r.status_code == 200
+    assert_valid_schema(data, 'recipe_revision.schema')
+
```
mythmon: The linter wants to remove this blank line at the end of the file.

```suggestion
```

mythmon:
It seems like your rebase had some issues; several commits were duplicated, and Github was showing a lot of unrelated changed. I used `git rebase -i master` with an updated copy of the master branch to fix it up.

PR #1792, opened 2019-03-07
Addon study `name` argument should be unique

### LINKED ISSUE #1791: The "opt-out-study" study names are not unique
*[Affected Platforms]:*
- All Windows
- All Mac
- All Linux

*[Affected Versions]:*
- Firefox Nightly 67.0a1

*[Prerequisites]:*
- Have access to the [delivery console](http://delivery-console.stage.mozaws.net/).
- Have a "opt-out-study" recipe already created.

*[Steps to reproduce]:*
1. Navigate to http://delivery-console.stage.mozaws.net/.
2. Create an opt-out-study recipe with the same study name as the one from prerequisites.
3. Click the "Save" button and observe the behavior.

*[Expected result]:*
- An error is shown explaining that the "Experiment name" MUST be unique.

*[Actual result]:*
- The recipe is successfully created and can be published without any errors.

*[Notes]:*
- The issue is not reproducible for the "preference-experiment", "show-heartbeat" and "preference-rollout" recipes. 
- Attached a screen recording with the issue:
unique

rehandalal:
Fixes #1791.

r?

### DIFF
--- a/normandy/recipes/models.py
+++ b/normandy/recipes/models.py
@@ -623,6 +623,7 @@ class Action(DirtyFieldsMixin, models.Model):
         "duplicate_rollout_slug": "Rollout slugs must be globally unique",
         "rollout_slug_not_found": "Rollout slug not found for rollback",
         "duplicate_survey_id": "Survey ID must be globally unique",
+        "duplicate_study_name": "Study name must be globally unique",
     }
 
     @property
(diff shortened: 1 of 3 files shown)


rehandalal:
Also do we care about checking for collisions in current existing recipes?

mythmon:
The code and tests look good.

> Also do we care about checking for collisions in current existing recipes?

I checked, and there are a few collisions in existing recipes. There is even a collision in two active recipes, [695](https://delivery-console.prod.mozaws.net/recipe/695/) and [696](https://delivery-console.prod.mozaws.net/recipe/696/).

I think that if we land this now, it would prevent changes to those two recipes, which might be a problem. I also don't think we can change the existing recipes to have unique names. On the other hand, I think this is a good example of why we need this check. I'll follow up with the authors of those recipe and figure out a way forward.

mythmon:
I talked with Rob, who worked on 695 and 696. The collisions were intentional, but it's fine if they aren't allowed. I think we're good to go here.

PR #1796, opened 2019-03-14
Add debugging code to extensions migration

rehandalal:
r?


normandy/recipes/migrations/0014_auto_20190228_1128.py
```diff
@@ -27,10 +30,26 @@ def add_extension_id(apps, schema_editor):
...
+                }
+            )
+        else:
+            arguments["extensionApiId"] = extension.id
+            revision.arguments_json = json.dumps(arguments)
+            revision.save()
+
+    if failures:
+        for failure in failures:
+            sys.stderr.write(f"{failure}")
```
mythmon: I think you need a newline here, since `sys.`stderr.write` isn't as nice as print.

```suggestion
            sys.stderr.write(f"{failure}\n")
```

Alternatively, I think you could do `print(failure, file=sys.stderr)`.

mythmon:
Oops, I typed bors r+ too soon. Thanks automation for saving me from myself.

PR #1798, opened 2019-03-15
XPI files should not allow overwrites of existing files

### LINKED ISSUE #1797: Extensions should not allow files to be overwritten
Currently if a user uploads a new extension with a filename that matches an existing one, we just overwrite the old file. This is pretty dangerous and we should not allow it.

Additionally we may want to rename files to match the standard naming convention:
`{addon-id}-{version}-signed.xpi`.

rehandalal:
Fixes #1797.

`unique` works weird on `FileField` but I kept it in as an extra safeguard.

r?


rehandalal:
Gotta do a little work on this to fix failing tests...

normandy/studies/models.py
```diff
@@ -131,4 +135,11 @@ def save(self, *args, **kwargs):
             if "xpi" in dirty_field_names:
                 self.populate_metadata()
 
+                filename = get_extension_filename(self)
+
+                if Extension.objects.filter(xpi=filename).exclude(pk=self.pk).exists():
+                    # Django just renames the file with a suffix if there is collision and the
+                    # FileField is set to `unique=True`. This forces an integrity error.
```
peterbe: That's horrible that Django does that! Good catch!

It always makes me nervous when you take an exception class from somewhere else and pretend it. I think it exists in `django.db` so you can use it for catching. Have you considered making your own exception class? A custom one?

glasserc: Hey, drive by comment because I saw this in my email. Is this a TOCTOU bug waiting to happen? (It may not be a problem even if it is...)

rehandalal: > Have you considered making your own exception class? A custom one?

I had not considered that. This is the exception I expect to be thrown when there is a constraint violation so it seemed appropriate to raise it here.

rehandalal: > Is this a TOCTOU bug waiting to happen?

That's a good point, it may be. I'm not sure if that really matters but I'll do some digging into a way to avoid that. I've been toying with the idea of changing the storage backend to error on overwrite.

mythmon: > I've been toying with the idea of changing the storage backend to error on overwrite.

I like this option a lot.

normandy/studies/models.py
```diff
@@ -18,9 +18,13 @@
 SIGNING_FILES = {"META-INF/mozilla.rsa", "META-INF/mozilla.sf", "META-INF/manifest.mf"}
 
 
+def get_extension_filename(instance, *args):
+    return f"extensions/{instance.extension_id}-{instance.version}-signed.xpi"
```
mythmon: I don't think we should do this. I think the original filename is useful information we should preserve. If we really want to standardize names, I think we should at least store the original in the API.

normandy/base/storage.py
```diff
@@ -9,6 +9,13 @@
 from storages.backends.gcloud import GoogleCloudStorage
 
 
+class RestrictedOverwriteStorageMixin(object):
+    def get_available_name(self, name, max_length=None):
+        if self.exists(name):
+            raise FileExistsError()
+        return super().get_available_name(name, max_length)
```
mythmon: This is still a TOCTOU error here. I think this is better than the way it was before, but it would be better if we could get an error thrown on over-write. That might just not be possible with Django storage engines though.

normandy/studies/tests/test_models.py
```diff
@@ -30,30 +39,54 @@ def test_recipes_used_by(self, storage):
...
+        extension = ExtensionFactory(xpi__from_func=xpi.open)
+        assert extension.hash == hashed
+
+    @pytest.mark.django_db
+    def test_no_duplicate_files(self, storage):
+        xpi = WebExtensionFileFactory()
+        ExtensionFactory(xpi__from_func=xpi.open)
+        with transaction.atomic():
+            with pytest.raises(FileExistsError):
+                ExtensionFactory(xpi__from_func=xpi.open)
```
mythmon: A minor improvement, you can combine `with`s, like this:

```python
with transaction.atomic(), pytest.raises(FileExistsError):
    ExtensionFactory(xpi__from_func=xpi.open)
```

normandy/studies/tests/__init__.py
```diff
@@ -1,16 +1,136 @@
...
+"""
+
+
+class XPIFileFactory(object):
+    def __init__(self, signed=True):
+        # Generate a unique random path for the new XPI file
+        f, self._path = tempfile.mkstemp(suffix=".xpi")
+
+        # Ensure the path exists
+        os.makedirs(os.path.dirname(self._path), exist_ok=True)
```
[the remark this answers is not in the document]
rehandalal: Good catch! This is a remnant of what I was doing before.

normandy/base/storage.py
rehandalal: Yeah I tried to find a solution that didn't suffer from the TOCTOU issue but there doesn't seem to be one. I'm open to ideas but this seems to be the way it's handled. I'm not sure if S3/GCP storage has options to disallow overwrite by folder or something?

normandy/studies/tests/test_models.py
rehandalal: TIL

rehandalal:
> This got a lot more complicated that the original version.

It did but I think this is also going to be a better way of dealing with fixtures going forward instead of having to maintain a bunch of opaque files.

rehandalal:
bors r=@mythmon

PR #1804, opened 2019-03-18
v3 API tests

chartjes:
* removed debugging message
* added in a set of tests for the v3 API that check schemas and validates SVG file types
* added Python dependency needed for v3 API tests

### DIFF
--- a/contract-tests/test_v3_api.py
+++ b/contract-tests/test_v3_api.py
@@ -1,6 +1,7 @@
 import json
 import jsonschema
 import magic
+import pytest
 
 from os.path import join, dirname
 from random import randint
@@ -24,12 +25,19 @@ def test_v3_actions(conf, requests_session):
     r = requests_session.get(conf.getoption("server") + "/api/v3/action/")
     r.raise_for_status()
     data = r.json()
+
+    if len(data) == 0:
+        pytest.skip("No list of v3 actions found")
+
     key = randint(0, len(data["results"]) - 1)
     action_id = data["results"][key]["id"]
 
     r = requests_session.get(conf.getoption("server") + "/api/v3/action/{}/".format(action_id))
     r.raise_for_status()
     data = r.json()
+
+    if len(data) == 0:
+        pytest.skip("Could not find a v3 action with the ID {}".format(action_id))
 
     assert r.status_code == 200
     assert_valid_schema(data, "v3.action.schema")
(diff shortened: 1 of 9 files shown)


contract-tests/test_v3_api.py
```diff
@@ -0,0 +1,240 @@
...
+
+    key = randint(0, len(data["results"]) - 1)
+    action_id = data["results"][key]["id"]
+
+    r = requests_session.get(conf.getoption("server") + "/api/v3/action/{}/".format(action_id))
+    r.raise_for_status()
+    data = r.json()
+
+    if len(data) == 0:
+        pytest.skip("Could not find a v3 action with the ID {}".format(action_id))
```
mythmon: I think it should be an error for an action to be published in the list, but not to have a detail page.

contract-tests/test_v3_api.py
```diff
@@ -0,0 +1,240 @@
...
+    approval_request_id = data["results"][key]["id"]
+
+    r = requests_session.get(
+        conf.getoption("server") + "/api/v3/approval_request/{}/".format(approval_request_id)
+    )
+    r.raise_for_status()
+    data = r.json()
+
+    if len(data) == 0:
+        pytest.skip(
```
mythmon: I think this should also fail for having a missing detail page.

contract-tests/test_v3_api.py
```diff
@@ -0,0 +1,240 @@
...
+    r.raise_for_status()
+    data = r.json()
+
+    if len(data) == 0 or len(data["results"]) == 0:
+        pytest.skip("No list of v3 approval requests found")
+
+    assert r.status_code == 200
+
+
+def test_v3_approval_request_with_id(conf, requests_session):
```
mythmon: Is there a reason you split this into two tests, but made the action version a single test?

contract-tests/test_v3_api.py
chartjes: I added those in because tests kept failing on CI when they passed every single time on my own laptop when using https://normandy.cdn.mozilla.net as the server. 

I'm open to suggestions on how to solve this chicken-and-egg issue.

contract-tests/test_v3_api.py
chartjes: No reason, I can just split them into two different tests

contract-tests/test_v3_api.py
mythmon: The test skip for "there is no data in the listing" seem good. I think we should keep those. The second check of "there was data in the listing, but there is not detail" were the ones I had a problem with. It seems that you've already removed those though. I think we're good here.

PR #1808, opened 2019-03-20
Prevent updates to extensions that are in use

### LINKED ISSUE #1092: Modifying an uploaded and in-use extension
If I modify an extension that is being used by a recipe, our current behavior is:

- The extension UI accepts the upload and updates the extension entry.
- The recipes using the extension retain their existing extension URLs, which causes the ExtensionSelect widget to display the URL of the old add-on. The recipe can then only be edited to point to another currently-uploaded extension, or left alone to retain the existing value (i.e. if the select is used to select another extension, it cannot be restored to the old value unless you do not save and refresh the page).

I dunno what we want the behavior in this case to _actually_ be, but it feels like the current behavior was not intentional, even if it is the behavior we want. I imagine at a minimum, we should add a warning/special case to ExtensionSelect to inform the user why the value it is set to is not editable without data loss. Other fixes could include making extensions immutable to avoid the problem completely, or blocking changes to extension XPIs if they are in use by any recipes.

rehandalal:
Fixes #1092.

r?


rehandalal:
Failure is #1813

PR #1809, opened 2019-03-20
Update action arguments schemas package to v0.7.0 

rehandalal:
r?

### DIFF
--- a/normandy/settings.py
+++ b/normandy/settings.py
@@ -143,7 +143,7 @@ class Core(Configuration):
     # This node package is expected to have been installed as a pre-build step
     # (i.e. `yarn`) and should always exist.
     ACTIONS_SCHEMA_DIRECTORY = values.Value(
-        os.path.join(BASE_DIR, "node_modules", "mozilla-normandy-action-argument-schemas")
+        os.path.join(BASE_DIR, "node_modules", "@mozilla", "normandy-action-argument-schemas")
     )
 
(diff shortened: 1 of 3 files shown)


peterbe:
Whatever happened to 0.4.0, 0.5.0, 0.6.0??

mythmon:
* v0.4.0 Was making add-on studies a action.
* v0.5.0 Was a simple change to make the schema more correct, but didn't have consequences for the server.
* v0.6.0 Was implementing heartbeat as a local action.
* v0.7.0 Was a retroactive bump for preference-experiments as a local action.

None of these *require* updating the server, but bringing in the new version is a good idea. I think there is a small chance that this has issues when we deploy, but we will catch it in stage. @peterbe do you remember what we did for conflicting schemas between this package and the schemas in the git repo?

mythmon:
Since @peterbe approved this already, I'm going to assume it will be fine, an dif not we'll catch it in stage.

bors r= @peterbe

peterbe:
> do you remember what we did for conflicting schemas between this package and the schemas in the git repo?

Quick answer; no. :)

Long answer; the code comments here: https://github.com/mozilla/normandy/blob/844b048d2be15f0bce2b8e3705d17ae3a21fb315/normandy/recipes/management/commands/update_actions.py#L78-L95 speak for themselves pretty well. Essentially, a warning is displayed on stdout. 

Perhaps that's not what you're referring to but it's a try.

PR #1812, opened 2019-03-22
Replace django-rest-swagger with drf-yasg and fix root views

### LINKED ISSUE #1776: Remove django-rest-swagger for alternative
django-rest-swagger appears to be [unmaintained](https://github.com/marcgibbons/django-rest-swagger#help-wanted).

We should ditch it and move to an alternative like https://github.com/axnsan12/drf-yasg.

### LINKED ISSUE #1788: API definitions not Swagger/OpenAPI compliant
In attempting to add more coverage in testing for the v1 and v3 API's by validating schemas, I have discovered that the API definition supplied by /api/docs is not Swagger/OpenAPI compliant.

Steps to reproduce the error:

1. Go to editor.swagger.io
2. Select File->Import URL and paste in https://normandy.cdn.mozilla.net/api/docs/
3. Observe that it generates the following messages:

Unable to render this definition

The provided definition does not specify a valid version field.

Please indicate a valid Swagger or OpenAPI version field. Supported version fields are swagger: "2.0" and those that match openapi: 3.0.n (for example, openapi: 3.0.0).

rehandalal:
Fixes #1776.
~I think this might also address #1788.~ <-- Confirmed
Fixes #1788.

In addition to swapping `django-rest-swagger` for `drf-yasg` this patch:
- Fixes the `APIRootView` so that it correctly aggregates endpoints without needing them to be in the same router.
- Makes the swagger docs version specific.

r?


rehandalal:
Failures are #1813.

normandy/settings.py
```diff
@@ -148,7 +149,7 @@ def CSP_DEFAULT_SRC(self):
 
     PROD_DETAILS_STORAGE = values.Value("normandy.recipes.storage.ProductDetailsRelationalStorage")
 
-    SWAGGER_SETTINGS = {"DOC_EXPANSION": "list"}
+    # SWAGGER_SETTINGS = {"DOC_EXPANSION": "list"}
```
mythmon: Oops

normandy/settings.py
```diff
@@ -395,8 +396,8 @@ class Development(Base):
     API_CACHE_ENABLED = values.BooleanValue(False)
     API_CACHE_TIME = values.IntegerValue(0)
 
-    SWAGGER_SETTINGS = Base.SWAGGER_SETTINGS
-    SWAGGER_SETTINGS["VALIDATOR_URL"] = None
+    # SWAGGER_SETTINGS = Base.SWAGGER_SETTINGS
+    # SWAGGER_SETTINGS["VALIDATOR_URL"] = None
```
mythmon: Oops here too

normandy/settings.py
rehandalal: Derp! Forgot to clean these up!

rehandalal:
I've removed redoc since there doesn't seem to be a good way of fixing the CSP issue and I'm not sure it adds enough value to warrant the work tracking down a good solution.

mythmon:
@chartjes the reason for your review here was to check out the new docs and schemas and review them. I am guessing you don't have a good set up for testing an in-progress Normandy PR, so I'm going to skip your review here. I definitely would still like your opinion on the results of this new system. I think it will be easier for you to do that on Normandy's dev or stage servers though.

bors r+

PR #1815, opened 2019-03-25
Publish signature in Remote Settings records (fixes #1811)

### LINKED ISSUE #1811: Sign records before publishing them to Remote Settings
In order to verify authenticity of the records that we publish on Remote Settings, we'd have to add a signature from the Normandy signer.

leplatrem:
fixes #1811

### DIFF
--- a/normandy/recipes/exports.py
+++ b/normandy/recipes/exports.py
@@ -21,11 +21,14 @@ def recipe_as_record(recipe):
     """
     from normandy.recipes.api.v1.serializers import (
         MinimalRecipeSerializer,
+        SignatureSerializer,
     )  # avoid circular imports
 
-    serializer = MinimalRecipeSerializer(recipe)
-    record = serializer.data
-    record["id"] = str(recipe.id)
+    record = {
+        "id": str(recipe.id),
+        "recipe": MinimalRecipeSerializer(recipe).data,
+        "signature": SignatureSerializer(recipe.signature).data,
+    }
     return record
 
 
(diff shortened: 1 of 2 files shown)


leplatrem:
Tests fail for unrelated reasons

rehandalal:
> Tests fail for unrelated reasons
#1813

normandy/recipes/tests/test_exports.py
```diff
@@ -174,6 +174,30 @@ def test_publish_and_unpublish_are_noop_if_not_enabled(self, requestsmock):
...
 
+    def test_recipe_as_remotesettings_record(self, mocked_autograph):
+        """Test that recipes are serialized as expected by our clients."""
+
+        recipe = RecipeFactory(
+            name="Test", approver=UserFactory(), enabler=UserFactory(), signed=True
+        )
+
+        record = exports.recipe_as_record(recipe)
+        assert sorted(record.keys()) == ["id", "recipe", "signature"]
```
mythmon: Instead of checking only that the keys exist, how about checking the keys against some sort of schema, or using something like `Whatever` to analyze the contents?

An example of using `Whatever`: https://github.com/mozilla/normandy/blob/2eaf362c061ecc40840d8a61377d2b9851c0ff94/normandy/recipes/tests/api/v3/test_serializers.py#L210-L222

An example of directly checking against a schema: https://github.com/mozilla/normandy/blob/2eaf362c061ecc40840d8a61377d2b9851c0ff94/contract-tests/test_v3_api.py#L39-L56

normandy/recipes/exports.py
```diff
@@ -21,14 +21,30 @@ def recipe_as_record(recipe):
...
 
 
+def records_equal(a, b):
+    """Compare records, ignoring timestamps.
+
+    :param a: one record
+    :param b: another record
+    :returns: True if both are equal.
+    """
+    ignored_fields = ("last_modified", "schema")
```
mythmon: The docstring says this ignores timestamps. Is `schema` a timestamp?

PR #1817, opened 2019-03-27
Replace django-sslserver with django-extensions, which provides an ssl option

mythmon:
`django-sslserver` appears to be "lightly maintained" and broken with Django 2.1.4. `django-extensions` seems much more robust. It also has some other interesting things.

This PR updates the `./bin/runsslserver.sh` script, adds a `requirements/dev.txt`, updates the docs, and includes some configuration changes for the new system.


mythmon:
Yeah, that's it. I missed it because I didn't actually uninstall `django-sslserver` locally.

mythmon:
bors r= @rehandalal

PR #1819, opened 2019-03-28
Integrate django_graphene

rehandalal:
r?


mythmon:
I think that GraphQL could make our API easier to work with for many use cases. When working in [edi](https://github.com/mythmon/normandy-edi), for example I often pull down way more data than I need and then filter it client side in order to answer a question.

I'll pull this down and play with it. I've also added a couple more items to the checklist in the first comment.

My biggest concern is that we currently have some logic that is enforced in API views and serializers, and not enforced by the models. Sometimes this is correct because the limitation is something we want to enforce externally but not internally. Other times I think it may be a mistake and we should move that logic into the models, where it can be shared. I think we need to review these cases before landing any CRUD queries. Perhaps we should land a readonly version first before trying to tackle this?

mythmon:
Regarding GraphiQL, I like it but I don't think we really need to host it on the Normandy servers. I'm a bit sad to see more JS being added to this repo, since we have been removing a lot of it lately.

Perhaps we could recommend external GraphQL clients instead for debugging? For example [Altair](https://altair.sirmuel.design/) can be pointed at arbitrary GraphQL servers (that support CORS, which we should). That might simplify things a bit.

rehandalal:
+1 to landing a read-only version. i'll update the list accordingly. 

i also like ditching all this extra JS that's pretty gross for a client. it's nice to have that built in but maintaining this extra JS is overhead I don't think any of us want. altair looks way nicer too.

normandy/base/schema.py
```diff
@@ -17,10 +17,42 @@ class Meta:
...
     all_groups = graphene.List(GroupType)
+    group = graphene.Field(GroupType, id=graphene.Int(), name=graphene.String())
 
     def resolve_all_users(self, info, **kwargs):
         return User.objects.all()
 
+    def resolve_user(self, info, **kwargs):
+        id = kwargs.get("id")
+        username = kwargs.get("username")
+        email = kwargs.get("email")
```
mythmon: Could these be passed explicitly instead?

```
def resolve_user(self, info, *, id=None, username=None, email=None, **kwargs):
```

That seems like a cleaner way to do this. My personal preference would be to not even include `**kwargs`.

normandy/base/schema.py
```diff
@@ -17,10 +17,42 @@ class Meta:
...
+
+        if id is not None:
+            return User.objects.get(id=id)
```
mythmon: Is there some way to indicate that passing more than one of `id`, `username`, and `email` doesn't make sense?

normandy/conftest.py
```diff
@@ -4,8 +4,10 @@
 
 import pytest
 import requests_mock
+from graphene.test import Client
```
mythmon: Maybe `import Client as GrapheneClient`? It's not obvious what this is a client *for* where it's actually used.

PR #1820, opened 2019-03-28
Re-enable signature health checks with a faster library

mythmon:
As a reminder, we disabled these checks because they were too slow. The [library](https://pypi.org/project/ecdsa/) we were using to validate signatures is a pure Python library, and is not focused on speed. I added [a new library](https://pypi.org/project/fastecdsa/) that is more focused on speed, though it includes C dependencies, which might make our life harder.

The new library is significantly less friendly, as well. I had to jump through some more hoops to actually use it, and I actually still use the slow library to decode the signature into the format expected by the new one. I think that's probably fine for now, but I could work on removing it entirely if it's important.

To test the speed increase, I created 100 signed recipes, re-enabled the test, and timed the signature verification check before and after this change.

Before:

```
Checked signatures for 100 recipes in 0:00:20.166276 (4.96 checks/second)
```

After:

```
Checked signatures for 100 recipes in 0:00:00.913976 (109 checks/second)
```

I think an almost 22x speed up should make it ok to re-enable these checks. :smile: 	

I haven't tested the Docker build yet. Given that the new library has a C component, I suspect it might fail to install on some systems. On my laptop, I had to install `gmp-devel` to build the library. I expect CI will catch this issue though, so I'm not worried.


mythmon:
CC @milescrabill @jvehent @sunahsuh

rehandalal:
> I expect CI will catch this issue though, so I'm not worried.

Yup!

mythmon:
It seems like these CI failures are from the tests introduced in #1804. I'm not really sure why I'm only seeing it here though. I'm going to try and isolate that problem.

mythmon:
@rehandalal I think this would work now, if you want to review it.

PR #1838, opened 2019-04-03
Disable CSRF checks

### LINKED ISSUE #1835: Disable CSRF checks for Normandy
Since we're no longer using cookie-based authentication we should disable CSRF checks.

We checked this with ulfr here: https://bugzilla.mozilla.org/show_bug.cgi?id=1539936

The security guidelines are being updated accordingly here: https://github.com/mozilla-services/websec-check/pull/8

rehandalal:
Fixes #1835 

r?

### DIFF
--- a/normandy/settings.py
+++ b/normandy/settings.py
@@ -197,9 +197,6 @@ class Base(Core, CORS, OIDC):
             # We have disabled CSRF as we do not use cookie-based authentication.
             # As a result the CSRF middleware check is not required.
             "security.W003",
-            # Check CSRF cookie http only. disabled because we read the
-            # CSRF cookie in JS for forms in React.
-            "security.W017",
         ]
     )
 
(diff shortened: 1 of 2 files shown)


normandy/settings.py
```diff
@@ -195,6 +194,9 @@ class Base(Core, CORS, OIDC):
             # We've subclassed Django's security middleware, so Django's
             # checks can't tell we are using the middleware.
             "security.W001",
+            # We have disabled CSRF as we do not use cookie-based authentication.
+            # As a result the CSRF middleware check is not required.
+            "security.W003",
             # Check CSRF cookie http only. disabled because we read the
             # CSRF cookie in JS for forms in React.
             "security.W017",
```
mythmon: Can we stop overriding the HTTP only checks, and then remove W017 from this list?

rehandalal:
bors r=@mythmon

PR #1844, opened 2019-04-12
Restrict fields that are queryable for users type in GraphQL

rehandalal:
r?

### DIFF
--- a/normandy/base/tests/test_schema.py
+++ b/normandy/base/tests/test_schema.py
@@ -1,6 +1,6 @@
 import pytest
 
-from normandy.base.tests import GQ, GroupFactory, UserFactory
+from normandy.base.tests import GQ, GroupFactory, UserFactory, Whatever
 
 
 class TestGQ(object):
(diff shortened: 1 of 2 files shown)


normandy/base/tests/test_schema.py
```diff
@@ -34,6 +34,18 @@ def test_resolve_user_by_id(self, gql_client):
         res = gql_client.execute(GQ().query.user(id=u.id).fields("username"))
         assert res == {"data": {"user": {"username": u.username}}}
 
+    def test_password_not_queryable(self, gql_client):
+        u = UserFactory()
+        res = gql_client.execute(GQ().query.user(id=u.id).fields("password"))
+        assert res == {
+            "errors": [
+                {
+                    "locations": [{"column": 23, "line": 1}],
```
mythmon: This seems a bit fragile since we are generating the queries. Can we change the `[{"column": 23, "line": 1}]` part to something more flexible? Maybe as simple as `"locations": Whatever()`.

PR #1856, opened 2019-04-15
Add response and signing metrics

mythmon:
This adds metrics for how long requests take, and for how many recipes and actions we're signing.


mythmon:
@sciurus Can you take a look at the metric names I used here? Also any other review comments you have.

normandy/base/middleware.py
```diff
@@ -117,3 +120,25 @@ def process_request(self, request):
...
+    def middleware(request):
+        start_time = time.time()
+        response = get_response(request)
+        delta = time.time() - start_time
+
+        view = request.resolver_match.func
+        view_name = f"{view.__module__}.{view.__name__}"
+        metrics.timing(
+            "response",
+            value=delta * 1000.0,
```
sciurus: Why a float? I thought markus wanted an int.

sciurus:
Our general deployment strategy is to replace the "canary" pod of normandy first, make sure it looks healthy, then replace the rest.

It would be nice if there was a way to apply a special tag to the canary instances of normandy. That way we could evaluate its performance separately and pause the deployment if there's a concerning change.

I could imagine a special purpose IS_CANARY env var that when present adds a `canary:true` (or some similar keyvalue) tag to the request metrics.

Alternately, I could imagine a more flexible EXTRA_TAGS env var that contains a string representation of tags to add that we would then set to `EXTRA_TAGS='canary:true'` but could customize further in the future if needed.

normandy/base/middleware.py
```diff
@@ -17,6 +19,7 @@
 
 
 logger = logging.getLogger(__name__)
+metrics = markus.get_metrics("normandy")
```
sciurus: Perhaps this should be the value of `statsd_namespace`, and it should default to "normandy"? I think I'd be surprised if i changed METRICS_STATSD_NAMESPACE and that value was prepended to "normandy" instead of replacing it.

I'm not sure we'd ever need to change the namespace, so you could just drop that option too if you like.

mythmon: I wasn't actually sure what `statsd_namespace` did. Is it just a prefix to the metric name? I didn't see any info about that in the Markus docs. If so I think making it default to `normandy` would work well.

normandy/base/middleware.py
mythmon: This code is mostly [copied from Markus](https://github.com/willkg/markus/blob/master/markus/main.py#L344-L347), so I think it's fine. I've also tested this a bit with the debug stats logger, and it seems to work fine.

rehandalal:
test failures look related though

mythmon:
> Alternately, I could imagine a more flexible EXTRA_TAGS env var that contains a string representation of tags to add that we would then set to EXTRA_TAGS='canary:true' but could customize further in the future if needed.

I took a look at doing this, and it's pretty messy. As far as I can tell, markus doesn't provide any hook points where we could do this. We could make a custom backend, but that's pretty heavy and hard to test. I'm going to see if this is something we could add to markus.

mythmon:
bors r= @rehandalal, @sciurus

PR #1858, opened 2019-04-16
Use therapist for linting in CI

rehandalal:
r?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,47 @@
 #
 version: 2
 jobs:
-  build_test_publish:
+  lint:
+    docker:
+      # Image with python 3.6 and node 8
+      - image: circleci/python:3.6-node-browsers
+
+    working_directory: ~/repo
+
+    steps:
+      - checkout
+
+      # Install python dependencies
+      - run:
+          name: Install python dependencies
+          command: pip install -U -r requirements/default.txt -c requirements/constraints.txt --require-hashes
+
+      - run:
+          name: Install therapist
+          command: pip install -U therapist
+
+      # Download and cache node dependencies
+      - restore_cache:
+          keys:
+          - v2-dependencies-{{ checksum "yarn.lock" }}
+          # fallback to using the latest cache if no exact match is found
+          - v2-dependencies-
+
+      - run:
+          name: Install node dependencies
+          command: yarn install --frozen-lockfile
+
+      - save_cache:
+          paths:
+            - "node_modules"
+          key: v2-dependencies-{{ checksum "yarn.lock" }}
+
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -6,7 +6,53 @@
...
+          paths:
+            - "node_modules"
+          key: v2-dependencies-{{ checksum "yarn.lock" }}
+
+      # Run lint suite using therapist
+      - run:
+          name: Lint
+          command: therapist run --use-tracked-files
+
+  build-test-publish:
```
mythmon: Thanks for updating this to hyphens. We'll need to update it in in `bors.toml` as well.

.circleci/config.yml
```diff
@@ -50,10 +96,6 @@ jobs:
             docker-compose build
             docker-compose pull db
 
-      - run:
-          name: Linting
-          command: docker-compose run web lint
```
mythmon: Since we aren't doing this anymore, I think we can also remove the lint task from `run.sh`.

.circleci/config.yml
```diff
@@ -150,7 +192,8 @@ workflows:
   version: 2
   main:
     jobs:
-      - build_test_publish:
+      - lint
+      - build-test-publish:
```
mythmon: With this change, it's possible for the `build-test-publish` job to publish a docker image even if lints fail. Since The `lint` job contains the package security checks, I think we need to fix this.

We could either make `lint` a dependency of `build-test-publish`, or we could split out `publish` into a separate job and make that depend on both `lint` and `build-test`.

.circleci/config.yml
```diff
@@ -6,7 +6,53 @@
 #
 version: 2
 jobs:
-  build_test_publish:
+  lint:
+    docker:
+      # Image with python 3.6 and node 8
+      - image: circleci/python:3.6-node-browsers
```
mythmon: This is fine for now, but I think we ought to look into making an image in [mozilla/ci-docker-bases](https://github.com/mozilla/ci-docker-bases/) that provides Python 3.7 and Node 10 or 11. That image could also include tools like Therapist.

.circleci/config.yml
```diff
@@ -6,7 +6,53 @@
 #
 version: 2
 jobs:
-  build_test_publish:
+  lint:
+    docker:
+      # Image with python 3.6 and node 8
```
mythmon: I checked this image, and it has Node 10, not 8.

.circleci/config.yml
rehandalal: Filed: https://github.com/mozilla/ci-docker-bases/pull/45

PR #1860, opened 2019-04-16
Add dev dockerfile and dev docker-compose override yml.

jrbenny35:
This adds an additional dockerfile for building normandy with docker in a development context. The ```docker-compose.development.yml``` will override the default normandy config while allowing easy setup and database connections.


jrbenny35:
This will be used on the integration tests on delivery-console.

jrbenny35:
@mythmon I was thinking to put the docs under the Installation section of the Development page. Would this be okay? If not, where would you like them?

mythmon:
I think putting them in a new file in the Developer section would be better. Mostly I think having both instructions on the same page might be confusing, especially if you are trying to search for something specific. It might make sense to also include a link at the top of the existing installation section, like "For an alternate installation method, see [this guide on how to use Docker]"

docs/dev/docker.rst
```diff
@@ -0,0 +1,45 @@
...
+   .. code-block:: bash
+
+      git clone https://github.com/mozilla/normandy.git
+      cd normandy
+
+2. Start the docker images via ``docker-compose``:
+
+   .. code-block:: bash
+
+    docker-compose -f docker-compose.yml -f docker-compose.development up -d --build
```
mythmon: Why specify `-f` twice?

docs/dev/docker.rst
```diff
@@ -0,0 +1,45 @@
+Docker Setup for local development
+==================================
+The following describes how to set up an instance of the site on your
+computer for development using the provided docker and docker-compose
```
mythmon: This says "for development", but the docs below don't explain below how to use the started docker-compose environment to develop the code. Can you add some details about how to make code changes, set up the database, configure, and configure the environment?

Alternatively, if the above things aren't what you meant this to be used for, this paragraph could change to better reflect the goals of this docker-compose set up.

docs/dev/install.rst
```diff
@@ -149,6 +149,13 @@ the endpoint ``/api/v3/``. No user interface is provided.
 
 .. _fork: http://help.github.com/fork-a-repo/
 
+Docker
+______
+
+For local Docker setup please reference the docker_ section.
```
mythmon: Can you change this to make it clearer that this is an alternative setup method?

docs/dev/docker.rst
jrbenny35: I'd rather have someone use this command so they also know they are overriding the non ```.development``` docker-compose file.

mythmon: So saying `-f` twice overlays the two configs? That isn't clear from your docs or from `docker-compose`'s docs. Can you add a note about it here?

docs/dev/docker.rst
jrbenny35: Well we could have them build each time, or have a volume mount to the docker image that is built. I think the volume would be better as building could take a while and use a lot of resources.

docs/dev/docker.rst
mythmon: FYI, I'm about to remove `ci/docker-compose.yml`, since I just refactored all of our CI to not use docker-compose. It probably makes sense to merge this two compose files.

docs/dev/docker.rst
jrbenny35: Looking at this again, this is the reason I wanted to use docker-compose and I would suggest others do too. I don't need to have the user do any additional setup. They can literally start the compose file and get to work with their changes. I don't feel any other documentation about the database or environments is needed at this point as the docker-compose.yml file does all the setup

docs/dev/docker.rst
mythmon: The files here refer to the docker-compose files as if they are in the current directory. However, they are in the `ci` directory, which isn't mentioned anywhere in these docs.

It also seems weird that we are using CI resources for local development.

docs/dev/docker.rst
mythmon: I expected more set up would be needed to get a working site, like creating a user account. I see now that `first-start` handles the bulk of that. I think these docs leave a pretty big "what now?" question about where to go if you aren't already familiar with docker-compose or Normandy, but that might be ok.

jrbenny35: I could point to steps 6-11 here: https://mozilla.github.io/normandy/dev/install.html#installation and say they need to be done within docker using a command

PR #1872, opened 2019-04-24
Remove all extra docker-compose services, and update development dock…

jrbenny35:
Changes are starting to catch-up to me...


Dockerfile.development
```diff
@@ -34,6 +34,8 @@ ENV DJANGO_SETTINGS_MODULE=normandy.settings \
     NEW_RELIC_CONFIG_FILE=newrelic.ini
 EXPOSE $PORT
 
-ENTRYPOINT ["/bin/bash", "/app/bin/run.sh"]
```
rehandalal: why did you opt not to use the run script?

mythmon: That's my fault. The run script was made specifically to support the previous CI setup. In my PR to update to the new system, I removed the run script, and I encouraged @jrbenny35 to not rely on it anymore.

PR #1874, opened 2019-04-25
Check cert root and subject in heartbeat

mythmon:
This would have caught some of our recent signature problems.

@g-k can you take a look at how I validated the certificates here?

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,2 @@
 [flake8]
-max-line-length=99
-ignore=W503
+ignore=W503,E501
--- a/docs/ops/config.rst
+++ b/docs/ops/config.rst
@@ -330,10 +330,40 @@ in other Django projects.
 
    This defaults to ``False`` for local developer instances.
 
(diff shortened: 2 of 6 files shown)


mythmon:
@g-k in that case, can you share the certificate that we were using that caused the problem? I'd like to understand what went wrong, and what Firefox is actually checking that wasn't right.

g-k:
@mythmon of course! https://gist.github.com/g-k/520ab1300267afd645cc67a179a8a056

mythmon:
@g-k Luckily, Gist's have raw URLs :smile:. I was able to use that to test it.

So the issue here is that the issuer and the subject were the same? That sounds like it might be more than what I want to check without a full fledged x509 validator. That sounds a bit out of scope, even more so than this PR already is.

g-k:
@mythmon Yeah, agreed it's probably more than you want to check. We're planning to fully automate the contentsignature rotation soon, so that should rule out the self-signed cert errors (and we'll check for those from the autograph monitor).

mythmon:
bors r= @rehandalal, @g-k

PR #1878, opened 2019-04-29
Speed up docker step in CI

mythmon:
Building the docker image currently takes twice as long as any other step in our set up. This should help it go a bit faster.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@
 # DOCKER_PASS
 #
 
-version: 2
+version: 2.1
 
 jobs:
   lint:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -5,7 +5,7 @@
 # DOCKER_PASS
 #
 
-version: 2
```
mythmon: This is needed for job parameterization

.circleci/config.yml
```diff
@@ -50,12 +50,16 @@ jobs:
...
+      saveToWorkspace:
+        type: boolean
+        default: true
     docker:
       - image: mozilla/cidockerbases:docker-latest
     steps:
       - checkout
       - setup_remote_docker:
-          version: 17.09.0-ce
+          docker_layer_caching: true
```
mythmon: @sciurus told me this should work now.

PR #1879, opened 2019-05-02
Stop requiring explicit approval for bors merges

mythmon:
After talking about it with @glasserc, I'm convinced that requiring both approval and a comment in order to trigger a bors merge is more pain than gain at this point.


glasserc:
I'm OK with this because the only people who can issue `r+` are those who have write permission, so `r+` automatically grants permission.

bors r+

PR #1902, opened 2019-06-12
Print a message when signing commands finish

mythmon:
This makes it much easier to tell that this commands finish succesfully when ops is looking at the logs.


mythmon:
bors delegate= @milescrabill

miles: When you are happy with this PR, you can say `bors r+` on a line alone to merge it safely.

PR #1936, opened 2019-08-02
added docker-compose file and changes to docker.dev

tiftran:
Currently left out a make command for therapist.... realized therapist.yml is used for pre-commit hooks, so have yet to figure out how to optimize efficiently....

was also unsure who should be reviewing PR as well...


rehandalal:
> so have yet to figure out how to optimize efficiently

I think something along the lines of:
```
therapist run --disable-git ./!(node_modules|assets|docs)
```
is probably what you're looking for.

When I tested locally it was only slightly slower than the git-aware version.

rehandalal:
After pulling down this PR I ran into a couple issues:
- I already have postgres installed on my machine and port 5432 is in use. I was able to solve this by just changing to `5454:5432` in `docker-compose.yml`. But I suspect this will be an issue for a number of people.
- After resolving the issue above I was able to get it to build but when I tried to run `make migrate` I got `Could not connect to database: FATAL:  database "normandy" does not exist`. After some fiddling I was able to determine this was not a result of changing the port and was some other kind of issue. @tiftran confirmed she is now experiencing this too.

I would love to see some kind of test that would fail on this PR (and future ones) to let us know that the dev docker setup is broken.

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+    ports:
+      - "8000:8000"
+
+  db:
+    restart: always
+    image: postgres:9.6
+    volumes:
+      - db_volume:/var/lib/postgresql
+    ports:
+      - "5432:5432"
```
rehandalal: to avoid the port in use issue i think you could also just do something like:
```
  db:
    restart: always
    image: postgres:9.6
    volumes:
      - db_volume:/var/lib/postgresql
    networks:
      - private_nw
```

rehandalal: You would also need to put app in the network `private_nw`.

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+      dockerfile: ./Dockerfile.development
+    command: bash -c "/app/bin/wait-for-it.sh db:5432 -- ./bin/download_geolite2.sh;./bin/runsslserver.sh"
+    links:
+      - db
+    networks:
+      - public_nw
+    volumes:
+      - .:/app
+    environment:
+      - DATABASE_URL=postgres://postgres:postgres@db/normandy
```
rehandalal: FYI: changing this to `postgres://postgres:postgres@db/postgres` fixed it for me

tiftran: I think 189c849 might've fixed it too?

rehandalal: yup

Makefile
```diff
@@ -0,0 +1,45 @@
...
+	docker-compose build
+
+createuser: build
+	docker-compose run app python manage.py createsuperuser
+
+check_migrations: build
+	docker-compose run app python manage.py makemigrations --check --dry-run --noinput
+
+makemigrations: build
+	docker-compose run app python manage.py makemigrations
```
rehandalal: nit: i just noticed these are tabs and we use spaces everywhere else.

rehandalal: ignore this nit. make only accepts tabs. 🤮

jaredkerim: YYYYYyeeeeepppppp

docker-compose.yml
jaredkerim: Yep this is the same pattern we use on experimenter to prevent collisions, we should do the same here.

jaredkerim:
Looks really good so far!!  I just tried running `make test` on my machine for the first time and got the following test failures:

============================================================================================== FAILURES ===============================================================================================
_____________________________________________________________________________________ test_run_checks_happy_path ______________________________________________________________________________________

    @pytest.mark.django_db
    def test_run_checks_happy_path():
        errors = run_checks()
>       assert errors == []
E       AssertionError: assert [<Error: leve...ecipes.E006'>] == []
E         Left contains one more item: <Error: level=40, msg='GeoIP DB not available', hint=None, obj=None, id='normandy.recipes.E006'>
E         Use -v to get the full diff

normandy/base/tests/test_checks.py:11: AssertionError
---------------------------------------------------------------------------------------- Captured stderr call -----------------------------------------------------------------------------------------
WARNING 2019-08-08 15:32:59,226 normandy.recipes.geolocation Geolocation is disabled: Cannot load database.
___________________________________________________________________________________ TestUpdateActions.test_it_works ___________________________________________________________________________________

self = <normandy.recipes.tests.test_commands.TestUpdateActions object at 0x7fe8bf408d90>

    def test_it_works(self):
        """
        Verify that the update_actions command doesn't throw an error.
        """
>       call_command("update_actions")

normandy/recipes/tests/test_commands.py:68:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/usr/local/lib/python3.7/site-packages/django/core/management/__init__.py:148: in call_command
    return command.execute(*args, **defaults)
/usr/local/lib/python3.7/site-packages/django/core/management/base.py:364: in execute
    output = self.handle(*args, **options)
/usr/local/lib/python3.7/contextlib.py:74: in inner
    return func(*args, **kwds)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <normandy.recipes.management.commands.update_actions.Command object at 0x7fe8bf3b25d0>, args = ()
options = {'action_name': [], 'configuration': None, 'force_color': False, 'no_color': False, ...}
action_names = {'console-log': '/app/client/actions/console-log', 'opt-out-study': '/app/client/actions/opt-out-study', 'preference-experiment': '/app/client/actions/preference-experiment', 'show-heartbeat': '/app/client/actions/show-heartbeat'}
name = 'utils.js', directory = '/app/client/actions/utils.js', action_schemas_fp = '/app/node_modules/@mozilla/normandy-action-argument-schemas/schemas.json'

    @transaction.atomic
    def handle(self, *args, **options):
        # Generate a list of all action names first. This is so it becomes
        # possible to filter on them by name.
        action_names = {}
        # First get the list of "legacy" actions by listing all directory
        for name in os.listdir(settings.ACTIONS_ROOT_DIRECTORY):
            if name == "tests":
                continue
            directory = os.path.join(settings.ACTIONS_ROOT_DIRECTORY, name)
            if os.path.isdir(directory):
                action_names[name] = directory

        # Next get all the keys from the @mozilla/normandy-action-argument-schemas package.
        action_schemas_fp = os.path.join(settings.ACTIONS_SCHEMA_DIRECTORY, "schemas.json")
>       with open(action_schemas_fp) as f:
E       FileNotFoundError: [Errno 2] No such file or directory: '/app/node_modules/@mozilla/normandy-action-argument-schemas/schemas.json'

normandy/recipes/management/commands/update_actions.py:35: FileNotFoundError
========================================================================================== warnings summary ===========================================================================================


Is there anything else we need to add so that a fresh checkout passes?

rehandalal:
@jaredkerim i think running `make load_geo_location` should fix that issue for you

jaredkerim:
- use -d for useradd to set /app to the homedir
- Setup the geoipd db either as part of the Dockerfile.development or in the Makefile
- Switch everything but the python listener to the private_nw
- Put node_modules in its own volume (similar to the way the db stores its data)

docker-compose.yml
```diff
@@ -0,0 +1,40 @@
...
+services:
+  app:
+    build:
+      context: ./
+      dockerfile: ./Dockerfile.development
+    command: bash -c "/app/bin/wait-for-it.sh db:5432 -- ./bin/download_geolite2.sh;./bin/runsslserver.sh"
+    links:
+      - db
+    networks:
+      - public_nw
```
jaredkerim: need 

    - private_nw

here for to talk to the db

jaredkerim:
bors try

Makefile
```diff
@@ -0,0 +1,59 @@
...
+
+migrate: build
+	docker-compose run app python manage.py migrate
+
+load_geo_location: build
+	docker-compose run app ./bin/download_geolite2.sh
```
jaredkerim: You can take this out can't you and just refer to the other geo one above?

jaredkerim:
bors try

PR #1937, opened 2019-08-02
Update @mozilla/normandy-action-argument-schemas to 0.8.1

glasserc:


mythmon:
Note that the primary [difference](https://diff.intrinsic.com/@mozilla/normandy-action-argument-schemas/0.8.0/0.8.1#file-index.js) here is renaming `multiple-preference-experiment` to `multi-preference-experiment`.

If this had made it to prod, I'd insist we should add a migration, but as it is I think we can simply deploy it and update or delete the existing stage recipes.

PR #1943, opened 2019-08-08
Remove simple filters

### LINKED ISSUE #1524: Remove RecipeRevision.[countries|locales|channels] manytomany fields

mythmon:
I thought this would be a quick and easy bit of cleanup. Unfortunately, this PR currently has an issue that is related to migrations. Pushing it up here to get more eyes.

The problem seems to be that the complicated migrations `recipes 0004` and `0005` don't play nice with the new auto-generated migration I added here. I believe that the new one is not properly cleaning up after itself, and so reversing it doesn't totally undo the forward migration. That makes the 0005 reverse fail.

This is a problem because we have tests that go forwards and backward through the migrations. Once 0015 (the new migration) is applied, then 0005 cannot be reversed.

Fixes #1524


jaredkerim:
bors try

mythmon:
> Error: Request failed "503 Service Unavailable

bors retry

mythmon:
This PR is a low priority. I mostly put it up to share the problem I was running into.

mythmon:
I'm not planning to work on this. It's still something we should do, but it's not worth the hassle that these migration errors are causing. Maybe someday we can squash our migrations and revisit this.

PR #1960, opened 2019-08-30
Don't publish recipes that require advanced capabilities to the general collection

### LINKED ISSUE #1935: Publish recipes that require "non-baseline" capabilities to a separate collection
Once Normandy can identify the capabilities a recipe needs (#1932), we should distinguish recipes that only requires "baseline capabilities" from those that require more. Recipes that only require baseline capabilities should be published in the existing Remote Settings collection (`normandy-recipes`), and those that don't should only be published in a separate collection (`normandy-recipes-capabilities`).

Additionally, this separation should apply to recipes published in the v1 `/api/v1/recipe/signed/` endpoint that older Firefox clients read from. Recipes that require more than the baseline should not be published in that endpoint. Optionally we can enable showing non-baseline recipes in this endpoint with a special query string, such as `/api/v1/recipe/signed/?capabilities=v1`

The baseline capabilities should be defined as the capabilities that are supported by versions of Firefox that don't check capabilities. Versions of Firefox that support capabilities should read from the new collection.

Although older versions of Firefox do not check capabilities, if they did most versions would support these capabilities, and so I propose we use this set as the baseline.

* `action.addon-study`
* `action.console-log`
* `action.preference-rollback`
* `action.preference-rollout`
* `action.show-heartbeat`
* `action.single-preference-experiment`
* `action.opt-out-study`
* `action.preference-experiment`
* `jexl.transform.date`
* `jexl.transform.stableSample`
* `jexl.transform.bucketSample`
* `jexl.transform.preferenceValue`
* `jexl.transform.preferenceIsUserSet`
* `jexl.transform.preferenceExists`
* `jexl.transform.keys`

Notably this leaves out the following JEXL transforms that are implemented in Firefox 68.

* `jexl.transform.length` - Added in Firefox 63
* `jexl.transform.mapToProperty` - Added in Firefox 63
* `jexl.transform.regExpMatch` - Added in Firefox 68
* `jexl.transform.versionCompare` - Added in Firefox 68

Since none of these are present in ESR 60, I suggest that we do not include them in the baseline. This means we would not be taking full advantage of the available resource of Firefox 63-68. We have not used these filters significantly so far due to backwards compatibility concerns. (This is the motivating point behind Normandy capabilities). I think this is the right balance to strike.

mythmon:
This ended up being way more complicated than I thought. I'm not really happy with the amount of duplication and complexity it has introduced, but I also don't see a great way to improve it. Any ideas?

Fixes #1935

### DIFF
--- a/normandy/recipes/api/filters.py
+++ b/normandy/recipes/api/filters.py
@@ -1,8 +1,10 @@
 import django_filters
 
+from normandy.recipes.models import Recipe
+
 
 class EnabledStateFilter(django_filters.Filter):
-    # A special case filter for filtering recipes by their enabled state
+    """A special case filter for filtering recipes by their enabled state"""
 
     def filter(self, qs, value):
         if value is not None:
@@ -11,6 +13,34 @@ class EnabledStateFilter(django_filters.Filter):
                 return qs.only_enabled()
             elif lc_value in ["false", "0"]:
                 return qs.only_disabled()
+        return qs
+
+
+class BaselineCapabilitiesFilter(django_filters.Filter):
+    """Filters recipe by whether they use only baseline capabilities, defaulting to only baseline."""
+
+    def __init__(self, *args, default_only_baseline=False, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.default_only_baseline = default_only_baseline
+
+    def filter(self, qs, value):
+        baseline_only = self.default_only_baseline
+
+        if value is not None:
+            lc_value = value.lower()
+            if lc_value in ["false", "0"]:
+                baseline_only = False
+
+        if baseline_only:
+            recipes = list(qs)
+            if not all(isinstance(recipe, Recipe) for recipe in recipes):
+                raise TypeError("BaselineCapabilitiesFilter can only be used to filter recipes")
+            match_ids = []
+            for recipe in recipes:
+                if recipe.uses_only_baseline_capabilities():
+                    match_ids.append(recipe.id)
+            return Recipe.objects.filter(id__in=match_ids)
+
         return qs
 
 
--- a/normandy/recipes/tests/test_exports.py
+++ b/normandy/recipes/tests/test_exports.py
@@ -1,5 +1,5 @@
 import base64
-import json
+from unittest.mock import call
 
 import pytest
 import kinto_http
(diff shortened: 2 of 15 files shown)


mythmon:
@jaredkerim @leplatrem Can you both look at this? I think it's broad enough that it would benefit from both of your reviews. I know that Mat is on PTO next week, which is fine. This can just wait a while.

normandy/recipes/api/filters.py
```diff
@@ -14,6 +16,33 @@ def filter(self, qs, value):
...
+        baseline_only = self.default_only_baseline
+
+        if value is not None:
+            lc_value = value.lower()
+            if lc_value in ["false", "0"]:
+                baseline_only = False
+
+        if baseline_only:
+            recipes = list(qs)
+            assert all(isinstance(recipe, Recipe) for recipe in recipes)
```
leplatrem: note: assertions errors are not raised if the server runs with `python -O` (optimized)

normandy/recipes/tests/test_exports.py
```diff
@@ -287,113 +436,250 @@ def test_publish_raises_an_error_if_request_fails(self, rs_settings, requestsmoc
...
+        # Expect delete calls
+        requestsmock.delete(ws_urls["capabilities"]["record"].format(recipe.id), json={"data": {}})
+        # Baseline is expected, just in case the recipe changed from baseline to not baseline
+        requestsmock.delete(ws_urls["baseline"]["record"].format(recipe.id), status_code=404)
+
+        remotesettings.unpublish(recipe)
+
+        requests = requestsmock.request_history
+        for i, req in enumerate(requests):
+            print(i, req.method, req.url)
```
leplatrem: leftover?

normandy/recipes/tests/test_exports.py
```diff
@@ -287,113 +436,250 @@ def test_publish_raises_an_error_if_request_fails(self, rs_settings, requestsmoc
...
-        assert requestsmock.request_history[1].url == collection_url
-        assert requestsmock.request_history[1].method == "PATCH"
-        assert requestsmock.request_history[2].url == records_url
-        assert requestsmock.request_history[3].url == records_prod_url
-        assert requestsmock.request_history[4].url == record_url
-        assert requestsmock.request_history[4].method == "PUT"
-        submitted = requestsmock.request_history[4].json()
+        requests = requestsmock.request_history
+        for i, req in enumerate(requests):
+            print(i, req.method, req.url)
```
leplatrem: leftover?

normandy/recipes/tests/test_exports.py
```diff
@@ -287,113 +436,250 @@ def test_publish_raises_an_error_if_request_fails(self, rs_settings, requestsmoc
...
-            f"/buckets/{exports.RemoteSettings.MAIN_BUCKET_ID}/",
-        )
+
+        capabilities_record_url = rs_urls["workspace"]["capabilities"]["record"].format(recipe.id)
         # Creating the record works.
-        requestsmock.request("put", record_url, content=b'{"data": {}}', status_code=201)
+        requestsmock.request("put", capabilities_record_url, json={"data": {}}, status_code=201)
         # Approving fails.
-        requestsmock.request("patch", collection_url, status_code=403)
+        print("expecting patch to:", rs_urls["workspace"]["capabilities"]["collection"])
```
leplatrem: leftover?

normandy/recipes/api/filters.py
mythmon: Good point. I'll change this to an exception.

mythmon:
> Note: We released a «rollback» feature on Remote Settings (yet to be deployed) that will simplify greatly the handling of publishing errors.

I look forward to this! You should file an issue here when that is something we can use here.

bors r= @leplatrem

PR #1972, opened 2019-09-04
When checking signatures, fail gracefully in the face of more db errors

### LINKED ISSUE #1971: Modify signature check to gracefully fail on db schema failure
If the signature check attempts to access the database and fails, potentially because the db is unvailable or because some schema is not available, the check should warn and pass, and a separate db check will fail.  This prevents a circular dependency between migrations running and the checks depending on them.

mythmon:
Fixes #1971

### DIFF
--- a/normandy/recipes/checks.py
+++ b/normandy/recipes/checks.py
@@ -3,7 +3,7 @@ from datetime import timedelta
 
 from django.apps import apps
 from django.conf import settings
-from django.core.checks import Error, Info, register as register_check
+from django.core.checks import Error, Info, Warning, register as register_check
 from django.core.exceptions import ImproperlyConfigured
 from django.db.utils import OperationalError, ProgrammingError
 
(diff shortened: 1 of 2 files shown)


normandy/recipes/checks.py
```diff
@@ -50,8 +51,11 @@ def recipe_signatures_are_correct(app_configs, **kwargs):
         signed_recipes = list(Recipe.objects.exclude(signature=None).select_related("signature"))
     except (ProgrammingError, OperationalError, ImproperlyConfigured) as e:
         errors.append(Info(f"Could not retrieve recipes: {e}", id=INFO_COULD_NOT_RETRIEVE_RECIPES))
-    else:
+        return errors
+
+    try:
         for recipe in signed_recipes:
+            print("@@@", recipe.canonical_json)
```
rehandalal: did you intend to have this in here?

jaredkerim:
I was able to recreate the problem locally and was able to apply migrations with this patch, so lgtm 👍

normandy/recipes/checks.py
jaredkerim: I'll wait for @mythmon to respond to this before we merge.

jaredkerim:
bors try

normandy/recipes/checks.py
mythmon: Oops. Fixed.

mythmon:
bors r= @jaredkerim

PR #1978, opened 2019-09-11
v3 API Tests

chartjes:
This is the first set of tests for the v3 API. They only use API calls that do not require authentication


chartjes:
I can move those around into constraints.txt, no problem.

I used these instructions https://mozilla.github.io/normandy/dev/install.html. I also updated pip once I was inside the virtualenv.

chartjes:
I also used pip when testing adding those dependencies in, so I was very surprised to see all those hash errors later on.

chartjes:
Redoing this work because Tavern requires reverting Pytest to an older version. Will investigate other solutions

PR #1980, opened 2019-09-12
Tweak docs

glasserc:
I've handled two different users lately who have had trouble with Remote Settings and Normandy so it seemed like it would be worth mentioning it in the documentation. I noticed also that these docs still assumed Remote Settings as a tentative feature, whereas it has been on since 68.

Some of the things I was going to write as notes for developing Normandy client seemed like they would fit better in this project's documentation, in particular the history of actions as external to the client. On the other hand, maybe it was being omitted to keep things simple. I'm happy to drop that commit if you don't think we need it.

### DIFF
--- a/docs/dev/remote-settings.rst
+++ b/docs/dev/remote-settings.rst
@@ -6,14 +6,8 @@ Remote Settings
 Remote Settings is a Mozilla service that makes it easy to manage evergreen
 settings data in Firefox.
 
-When enabled on the Normandy server, the enabled recipes are published as
-Remote Settings records.
-
-When enabled on the client side, instead of polling regularly the Normandy
-API to obtain the list of recipes, the recipes are read from the Remote
-Settings synchronized data locally.
-
-Delivering Normandy recipes via Remote Settings allows us:
+As of Firefox 68, Normandy uses Remote Settings to serve/update its
+recipes. This allows us to:
 
 - Unify the way we obtain fresh data from our services in Firefox
 - Get rid of the specific data channel of Normandy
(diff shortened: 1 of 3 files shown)


docs/dev/remote-settings.rst
```diff
@@ -53,25 +47,9 @@ Use ``--dry-run`` to only print out the result of the synchronization.
...
-    * `Client setup with a local Remote Settings server
-      <https://remote-settings.readthedocs.io/en/latest/tutorial-local-server.html#prepare-the-client>`_
-
-The Normandy recipe runner will now be able to read the list of recipes to
-execute from the local Remote Settings database.
+The Normandy client integration was added in Firefox 66 (`Bug 1506175
+<https://bugzilla.mozilla.org/show_bug.cgi?id=1506175>`_), refined through
+Firefox 67 and 68 (`Bug 1538248
+<https://bugzilla.mozilla.org/show_bug.cgi?id=1538248>`_) and turned on by
+default in Firefox 68 (`bug 1513854
```
leplatrem: Above you wrote 69

glasserc: Good catch, thanks! I thought it was 69 based on other discussions in the Normandy chat room, but @sciurus pointed out that it seems to have landed in 68.

mythmon:
That's a weird error in CI. I'm going to try this again, because I expect it is intermittent. Lets try this again

bors retry

mythmon:
The problem is #2043. It's not related to this PR, but it blocks CI so we can't merge. I've put up a fix for that PR that should be pretty easy to get merged soon.

mythmon:
The CI issue should be fixed now.

bors r= @jaredkerim, @mythmon

PR #1990, opened 2019-09-25
Wrap approve/reject/close methods for ApprovalRequest in a transaction

### LINKED ISSUE #1989: Approving a revision should happen in a transaction
We recently had an issue where an approval request was approved but updating the recipe signature failed for some reason causing the newly approved revision not to be enabled. This resulted in the recipe getting disabled and users being unenrolled from the related study. To prevent this in the future we should wrap the `approve` method of the `ApprovalRequest` model in a transaction.

rehandalal:
Fixes #1989 

r?

### DIFF
--- a/normandy/recipes/models.py
+++ b/normandy/recipes/models.py
@@ -590,6 +590,7 @@ class ApprovalRequest(models.Model):
                     },
                 )
 
+    @transaction.atomic
     def approve(self, approver, comment):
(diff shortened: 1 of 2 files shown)


jaredkerim:
@rehandalal Is there a way we can write a unit test that mocks out an autograph failure and ensures that we wind up in the right state?

PR #1995, opened 2019-09-26
Don't require write permission to the main remote settings bucket

### LINKED ISSUE #1994: Remote settings exporter demands write access to main bucket, but it is not needed
The remote settings recipe exporter demands that it can write to the `normandy-recipes` collection on the `main` bucket. That isn't actually needed. We write the `main-workspace` bucket, and use Kinto's approval process to migrate changes to `main`. This was a regression introduced in #1960.

mythmon:
Fixes #1994


jaredkerim:
@mythmon Ah looks like a test needs to be updated

mythmon:
Oops. Lets try this.

PR #1997, opened 2019-09-26
Only include 'capabilities-v1' in non-baseline recipes

### LINKED ISSUE #1996: The capability "v1-capabilities" is unconditionally included in all recipes
The capability "v1-capabilities" is not a baseline capability, and is supposed to mark that a recipe requires capability filtering to process correctly. It should only be automatically added to recipes that contain other non-baseline capabilities.

mythmon:
Fixes #1996

### DIFF
--- a/normandy/recipes/tests/test_models.py
+++ b/normandy/recipes/tests/test_models.py
@@ -819,8 +819,15 @@ class TestRecipeRevision(object):
 
     @pytest.mark.django_db
     class TestCapabilities:
-        def test_v1_marker_included_if_non_baseline_capabilities_are_present(self, settings):
-            recipe = RecipeFactory(extra_capabilities=["non-baseline"])
+        def test_v1_marker_included_only_if_non_baseline_capabilities_are_present(self, settings):
+            action = ActionFactory()
+            settings.BASELINE_CAPABILITIES |= action.capabilities
+
+            recipe = RecipeFactory(extra_capabilities=[], action=action)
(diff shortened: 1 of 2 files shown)


normandy/recipes/tests/test_models.py
```diff
@@ -819,8 +819,15 @@ def test_enable_rollback_enable_rollout_invariance(self):
...
     class TestCapabilities:
-        def test_v1_marker_included_if_non_baseline_capabilities_are_present(self, settings):
-            recipe = RecipeFactory(extra_capabilities=["non-baseline"])
+        def test_v1_marker_included_only_if_non_baseline_capabilities_are_present(self, settings):
+            action = ActionFactory()
+            settings.BASELINE_CAPABILITIES |= action.capabilities
+
+            recipe = RecipeFactory(extra_capabilities=[], action=action)
+            assert recipe.capabilities <= settings.BASELINE_CAPABILITIES
+            assert "capabilities-v1" not in recipe.capabilities
```
jaredkerim: In general I would consider this two separate tests (if they have different factory setups), but that's just a preferential thing, this looks good 👍

mythmon: Yeah, I was on the fence about it. On the one hand they share some things (like the action set up) and both test independently can't fully verify the feature. On the other hand, there are two set ups, two assertions, so it should be two tests. If they were bigger I'd split them, but as is I think I prefer them as one.

mythmon:
bors r= @jaredkerim

PR #2010, opened 2019-10-04
Allow experimenter_slug field to be set to null via API

glasserc:
I'm not 100% sure if this is correct. The [Django models documentation](https://docs.djangoproject.com/en/2.2/ref/models/fields/#null) suggests that we should rely on empty strings for string fields, but we already have `null=True` on the model field itself (presumably because this field didn't exist previously). However, we need to have some way to clear this field in order to support https://github.com/mozilla/delivery-console/issues/808.


glasserc:
Another approach could be to set `blank=True`. I wasn't sure if there was a canonical way of dealing with the null/empty string situation on this project.

jaredkerim:
@glasserc Yeah I would make this change to the serializer but also add `blank=True` to the model https://github.com/mozilla/normandy/blob/master/normandy/recipes/models.py#L349

It's important to be able to distinguish between an empty string and None, one can imply that no value was ever set and the other can imply that a value was set but if for instance it's an empty string that's not what we expect for an experiment slug so it may indicate a failure at some point in the system.  

But it's important to also remember that `experimenter_slug` is completely optional for a recipe so we should never block creating/modifying/saving recipes on that information not being present.  @mythmon is that a reasonable assumption?

Also @glasserc it would be helpful to accompany this with tests against the serializers to check the following cases

- if the field is not provided at all in the API call JSON body, it should still successfully save the recipe with `experiment_slug == None`
- if the field is provided with a string value, it is successfully saved to the `experiment_slug` field

Thanks for taking the time to look at this! :D

mythmon:
> But it's important to also remember that `experimenter_slug` is completely optional for a recipe so we should never block creating/modifying/saving recipes on that information not being present. @mythmon is that a reasonable assumption?

thumbs up

glasserc:
> also add blank=True to the model

Are you sure? I'm happy to go along with whatever you like here but it seems to me that having two different ways to have data absent is a recipe for hard-to-identify bugs. It sounds like you're saying allowing blank strings could allow incorrect data to creep in -- in that case, shouldn't we not allow that incorrect data and generate errors when we get it?

> tests against the serializers

That's a good idea, thanks.

glasserc:
> if the field is provided with a string value, it is successfully saved to the experiment_slug field

It seems like this is provided already by https://github.com/mozilla/normandy/blob/master/normandy/recipes/tests/api/v3/test_serializers.py#L176.

jaredkerim:
> > if the field is provided with a string value, it is successfully saved to the experiment_slug field
> 
> It seems like this is provided already by https://github.com/mozilla/normandy/blob/master/normandy/recipes/tests/api/v3/test_serializers.py#L176.

It looks to me like that is testing that the field will be correctly validated by the serializer, not that it is saved to the model and accessible via the model field after saving?

jaredkerim:
> > also add blank=True to the model
> 
> Are you sure? I'm happy to go along with whatever you like here but it seems to me that having two different ways to have data absent is a recipe for hard-to-identify bugs. It sounds like you're saying allowing blank strings could allow incorrect data to creep in -- in that case, shouldn't we not allow that incorrect data and generate errors when we get it?


Yeah that's what I'm saying we want the opposite of that, because the data is totally optional we DON'T want to strongly validate it at any stage or ever prevent the creation or modification of a recipe record because of any None, empty, or non-empty value in this field.

glasserc:
> It looks to me like that is testing that the field will be correctly validated by the serializer, not that it is saved to the model and accessible via the model field after saving?

OK fair enough. How about https://github.com/mozilla/normandy/blob/master/normandy/recipes/tests/api/v3/test_api.py#L390 ?

I'll add `blank=True` and add tests for {creation, editing} with `experimenter_slug` {null, ""}. I can add checks for `experimenter_slug=None` in existing tests or add new tests for that, what do you think?

glasserc:
Just to observe that the DRF docs also discourage this:

> The allow_null option is also available for string fields, although its usage is discouraged in favor of allow_blank. It is valid to set both allow_blank=True and allow_null=True, but doing so means that there will be two differing types of empty value permissible for string representations, which can lead to data inconsistencies and subtle application bugs.

jaredkerim:
> Just to observe that the DRF docs also discourage this:
> 
> > The allow_null option is also available for string fields, although its usage is discouraged in favor of allow_blank. It is valid to set both allow_blank=True and allow_null=True, but doing so means that there will be two differing types of empty value permissible for string representations, which can lead to data inconsistencies and subtle application bugs.

Well if it comes back to bite us you can say 'I told you so' ;)

jaredkerim:
@glasserc I appreciate the attention to detail but in general I try to avoid this kind of combinatoric testing, otherwise it just gets out of hand testing every possibility for every field.  I still think we only need specifically the two cases I mentioned before, I don't think this change is big enough to warrant this many tests.

PR #2016, opened 2019-10-08
Version gh-pages in package.json like everything else

### LINKED ISSUE #1942: Docs aren't being published
In CircleCI our `docs-publish` job is failing with this error:

```
$ gh-pages --dotfiles --message "[skip ci] Docs updates" --dist docs/_build/html

The "file" argument must be of type string. Received type undefined
Exited with code 1
```

Here is an example CI job: https://circleci.com/gh/mozilla/normandy/12787

This job only runs on master, so bors hasn't noticed it. This problem seems to have started recently. I'd guess it is because we don't actually pin a version of the `gh-pages` tool, and install it via `npm install gh-pages@2`. We should probably move this to package.json and manage it like all the other Node dependencies.

glasserc:
Fixes #1942.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -144,7 +144,7 @@ jobs:
       - run:
           name: Install and configure dependencies
           command: |
-            npm install -g gh-pages@2
+            yarn install --frozen-lockfile
             git config user.email "ci-build@normandy.mozilla.org"
             git config user.name "ci-build"
       - run:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -144,7 +144,7 @@ jobs:
       - run:
           name: Install and configure dependencies
           command: |
-            npm install -g gh-pages@2
+            yarn install
```
jaredkerim: Does switching from npm to yarn introduce any kind of whackiness?

glasserc: It's hard to test because this only runs on master. But we use `yarn` elsewhere so even if it does produce that kind of wackiness then I think it's worth it.

Issue #1890, opened 2019-05-23
Stop using public key sent from Autograph


mythmon:
Currently, Normandy stores and validates the `public_key` field returned by autograph when signing objects. This is not used by Normandy Client when verifying recipe signatures. Instead the x5u is used to retrieve a certificate chain, which is a more sound verification process.

The server on the other hand *is* attempting to validate these signatures, which isn't good. To quote @jvehent on slack:

> ok so the public key field returned by autograph here isn't good
> the way to verify the signature is to use the chain from the x5u
> we'll need to remove it from autograph
> it should stop using the public key returned by autograph, and autograph should stop returning that value

We should remove the public key field from our database, and ignore it if Autograph sends it. We should stop validating it, and stop sending it to Normandy Client. None of this is a security problem, it is just not an optimal way of handling certificates and signatures.

shell1:
we query it - so if they remove from their API we break.

glasserc:
I'm not sure what to do with this issue. We currently now verify signatures using `x5u`. However, this signature verification is kind of slapdash -- we don't actually verify the cert chain. There are other places around Mozilla where we do this kind of verification -- see mozilla-services/poucave#106 -- and hopefully we should be able to standardize on a single solution. @g-k's comment on that issue links to three libraries claiming to implement this algorithm, but two are clearly marked as "work in progress" and the third doesn't inspire confidence either (`new` bugs older than a year).

Besides this, there's one other outstanding concern with Autograph in dev, which is that certs are served as `file://` URLs, which we don't support. See https://github.com/mozilla-services/autograph/issues/350. In the meantime, I've been developing against `autograph` 2.7.0 using `docker run -p 8765:8000 -it mozilla/autograph:2.7.0`.

The stated purpose of this issue is to stop using the public-key field of the Autograph response, but it seems to me that for old recipes, that may be the only way we have available to validate them (since `public_key` predates `x5u`). Absent any compelling certificate-chain-validation library to use, I don't think there's anything more that can be done on the Normandy side of this issue, so I'm closing it.

PR #2017, opened 2019-10-09
Fix contract tests

glasserc:
Refs #1890. Update the contract tests to pass x5us instead of public keys.

### DIFF
--- a/contract-tests/test_api.py
+++ b/contract-tests/test_api.py
@@ -100,17 +100,11 @@ def test_recipe_signatures(conf, requests_session):
     if len(data) == 0:
         pytest.skip("No signed recipes")
 
-    cert_urls = set()
-
     for item in data:
         canonical_recipe = canonical_json(item["recipe"])
         signature = item["signature"]["signature"]
-        pubkey = item["signature"]["public_key"]
-        cert_urls.add(item["signature"]["x5u"])
-        assert signing.verify_signature(canonical_recipe, signature, pubkey)
-
-    for url in cert_urls:
-        signing.verify_x5u(url)
+        x5u = item["signature"]["x5u"]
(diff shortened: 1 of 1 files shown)


contract-tests/test_api.py
```diff
@@ -121,17 +115,11 @@ def test_action_signatures(conf, requests_session):
...
         canonical_action = canonical_json(item["action"])
         signature = item["signature"]["signature"]
-        pubkey = item["signature"]["public_key"]
-        cert_urls.add(item["signature"]["x5u"])
-        assert signing.verify_signature(canonical_action, signature, pubkey)
-
-    for url in cert_urls:
-        signing.verify_x5u(url)
+        x5u = item["signature"]["x5u"]
+        assert signing.verify_signature_x5u(canonical_action, signature, x5u)
```
mythmon: This is going to cause a lot more outbound network traffic. Of the dozens of signed objects we have, there are only a few unique `x5u` urls. I wonder if it is possible for `verify_signature_x5u` to cache the response from the server, or to use the etag information to do a "if not modified" request?

```
$ http GET 'https://normandy.cdn.mozilla.net/api/v1/recipe/signed/?enabled=1' | jq '.[].recipe.id' | sort | uniq | wc -l
40

$ http GET 'https://normandy.cdn.mozilla.net/api/v1/recipe/signed/?enabled=1' | jq '.[].signature.x5u' | sort | uniq | wc -l
1
```

glasserc: While this is true that this is going to produce more requests and load, it seems like a relatively low priority for a set of contract tests. These aren't on the hot path..

mythmon: Apologies, I thought that these were the heartbeat tests. This is fine.

mythmon:
This looks like an intermittent that crops up occasionally. It should pass this time.

bors retry

Issue #1863, opened 2019-04-18
When validating recipe arguments, "name" may not be included when expected


mythmon:
https://sentry.prod.mozaws.net/operations/normandy-stage/issues/5562863/

```
KeyError: 'name'
(17 additional frame(s) were not displayed)
...
  File "rest_framework/views.py", line 492, in dispatch
    response = handler(request, *args, **kwargs)
  File "rest_framework/mixins.py", line 84, in partial_update
    return self.update(request, *args, **kwargs)
  File "rest_framework/mixins.py", line 70, in update
    self.perform_update(serializer)
  File "rest_framework/mixins.py", line 80, in perform_update
    serializer.save()
  File "rest_framework/serializers.py", line 209, in save
    self.instance = self.update(self.instance, validated_data)

KeyError: 'name'
```

mythmon:
The recipe that triggered this is https://stage-admin.normandy.nonprod.cloudops.mozgcp.net/api/v3/recipe/741/

Notably, the field `recipe.arguments['name']` is `!@#*()`, which is likely what is triggering this.

mythmon:
I found the root cause of this. Recipe 741 was once a preference experiment, and is now an add-on experiment. The latest version of the recipe is an add-on experiment, but the latest *approved* version is a preference experiment.

The code that does does the `argument.name` uniqueness check confuses latest and latest approved. It gets all recipes where the latest revision is the right action type, but then access the names from the "current revision" which is the latest approved if the recipe has been approved.

In order to unblock QA, I've approved a newer version of the recipe that is an opt-out study. This fixes the bug for me on stage. We should also change the code to prevent this in the future.

jaredkerim:
Is this related to changing action types? Should we disallow this entirely as far down as the model?

mythmon:
Yes, was the problem. Specifically, the problem was changing to a different action type that had incompatible argument schemas. This combined with another problem where some code confuses `Recipe.latest_revision` and `Recipe.approved_revision`. Because of this it was seeing the recipe as both types of actions depending on what property it accessed. I think that it got the action type from one revision and the arguments from another revision.

This confusion is a problem that can hurt us in other ways. Disallowing action type changes would fix this case of the problem, but that confusion is still a problem.

glasserc:
[...]

PR #2018, opened 2019-10-09
Remove aliases

glasserc:
Refs #1863. This PR removes a bunch of aliases for recipe fields that actually come from revisions. The ones removed in this PR are the easy ones that aren't really used anywhere but it's a good first step and helps clarify that these aliases are exceptions rather than the rule.

@mythmon Could I have like a 10% review? Does this seem directionally accurate? I wasn't sure how much leeway I had in terms of altering the API and in particular dropping fields from serializers. The contract-tests seem to pass (modulo #2017) but maybe they aren't extensive enough?

### DIFF
--- a/contract-tests/test_api.py
+++ b/contract-tests/test_api.py
@@ -100,17 +100,11 @@ def test_recipe_signatures(conf, requests_session):
     if len(data) == 0:
         pytest.skip("No signed recipes")
 
-    cert_urls = set()
-
     for item in data:
         canonical_recipe = canonical_json(item["recipe"])
         signature = item["signature"]["signature"]
-        pubkey = item["signature"]["public_key"]
-        cert_urls.add(item["signature"]["x5u"])
-        assert signing.verify_signature(canonical_recipe, signature, pubkey)
-
-    for url in cert_urls:
-        signing.verify_x5u(url)
+        x5u = item["signature"]["x5u"]
+        assert signing.verify_signature_x5u(canonical_recipe, signature, x5u)
 
 
 def test_action_signatures(conf, requests_session):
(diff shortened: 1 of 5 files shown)


mythmon:
> The contract-tests seem to pass (modulo #2017) but maybe they aren't extensive enough?

The contract-tests don't validate that the API serves the needs of any client that I know of, beyond the level of "the API exists". The Python unit tests that we have do a better job of expressing many client uses cases that exist.

mythmon:
Regarding changing the serializers: I don't expect that any of these changes will affect the v3 API very much, because I don't think it uses any of the aliases. The v1 API is the one we need to worry about. Technically, we shouldn't make any changes to the v1 API, though in practice we've added fields without worry.

Specifically, the [`SignedRecipeSerializer`](https://github.com/mozilla/normandy/blob/3a6e789612d41a6fa5fd33ad23e6d8fc16fa2f6c/normandy/recipes/api/v1/serializers.py#L156-L166) (and by extension `MinimalRecipeSerializer` and `SignatureSerializer` are the ones we need to be very careful changing. That collection of serializers is what generates the data that Firefox clients read to run recipes (both for legacy clients accessing the API directly and modern Remote Settings-based clients).

I don't think any of the fields you listed here affect those serializers (the tests would have broken if they did). I also don't think anyone is still using the v1 APIs except the use cases I outlined above. Once you get to the point where you're removing aliases for fields that the minimal serializer is using, you'll have to do something to keep the serializer compatible.

In short: I think this is a good direction, though it will likely get harder.

mythmon:
@glasserc Do you plan to do anymore work on this PR?

mythmon:
In the interest of cleaning up old PRs, I'd like to close this PR. It seems like these changes are reasonable changes that make the code better, so I'm going to merge them. We definitely have more work to do in this direction in the future.

bors merge

PR #2019, opened 2019-10-10
Leverage Remote Settings rollback feature on errors

leplatrem:
* [x] Adjust tests
* [x] Add test for signer version

### DIFF
--- a/normandy/recipes/tests/test_exports.py
+++ b/normandy/recipes/tests/test_exports.py
@@ -135,12 +135,14 @@ class TestRemoteSettings:
             exports.RemoteSettings().check_config()
         assert "Invalid Remote Settings credentials" in str(exc.value)
 
+        # Collections writable.
         requestsmock.get(
             f"{rs_settings.REMOTE_SETTINGS_URL}/",
             json={
                 "user": {"id": f"account:{rs_settings.REMOTE_SETTINGS_USERNAME}"},
                 "capabilities": {
                     "signer": {
+                        "version": "5.2.0",
                         "to_review_enabled": True,
                         "group_check_enabled": True,
                         "resources": [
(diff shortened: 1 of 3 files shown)


leplatrem:
FYI running lint from Docker does not seem to work apparently...  (BTW I find it really overkill for that kind of simple tasks :))

Screenshot from 2019-10-10 13-23-26

I had issues running `make test` too. Fortunately I had kept my simple setup, without docker-compose, and I could run `py.test` on the single test file that I had changed. The patch was not high on my priority list, so I guess that without that, I would have given up contributing it :)

leplatrem:
Also, note that `make test` takes 3min40 on my machine, and `py.test normandy/recipes/tests/test_exports.py` 6 sec, which is already a lot for what the tests do (due to migrations apparently which are applied individually even if it's an empty test db...)

normandy/recipes/tests/test_exports.py
```diff
@@ -190,7 +192,26 @@ def test_check_connection(self, rs_settings, requestsmock):
             # restore write permissions
             requestsmock.get(collection_url, json=allow_write_payload)
 
+        # Signer version should be > 5.1.0.
```
mythmon: Should this be the below instead?

```suggestion
        # Signer version should be >= 5.1.0.
```

mythmon:
> due to migrations apparently which are applied individually even if it's an empty test db...

This is done by default for correctness reasons, but you can generally pass `--no-migrations` to pytest to make it either re-used the DB directly, or just start at the finalized state instead of running through all the migrations. Both of these strategies save time.

For me `pytest normandy/recipes/tests/test_exports.py` takes about 3.5 seconds, and adding `--no-migrations` takes about 1.5 seconds.

leplatrem:
Awesome @mythmon thanks! 

`py.test normandy/recipes/tests/test_exports.py -s -vv --no-migrations` takes ~2sec (~4sec according to `time`)

leplatrem:
bors r+

(hope it's all good :))

mythmon:
Ah, I thought you also wanted @glasserc to look at this, since you flagged him for review.

Also, in the future you can say `bors r= @mythmon`, to mark the pull request as approved by another user.

PR #2020, opened 2019-10-10
V3 API Contract Tests

chartjes:
Added in tests for all the V3 API calls that are GET calls and do not require authentication


chartjes:
@jaredkerim Made the requested change to use urljoin and pushed those changes (forgot to run the linter on the first push)

PR #2035, opened 2019-10-22
Ask yarn to run gh-pages to fix path problems

mythmon:
After #2016 we still aren't building docs correctly. I think this is just a path problem, which this patch should fix. Unfortunately we can't really tell until this gets to master. :man_shrugging:


mythmon:
And with this, our CI builds on master are working, and our docs are being updated again.

PR #2039, opened 2019-11-01
(docs) Fix link to Firefox Source Tree docs

mythmon:
Without this change, the link goes to the relative url, `./fx-source-docs`, which doesn't work.


uhlissuh:
bors merge

PR #2071, opened 2020-01-08
Use licence to download the GeoLite2 Country DB (fixes #2069)

### LINKED ISSUE #2069: GeoLite2 database is no longer available publicly
An account is required on the maxmind.com website, and an API required for download.

Shall we do that? Shall we remove the GeoIP parts of Normandy instead?

leplatrem:
Fixes #2069 

An account should be created on maxmind.com and the licence key be put in `MAXMIND_LICENCE_KEY` env var on CircleCI.

@sciurus Do we want a BZ ticket for that?

### DIFF
--- a/bin/download_geolite2.sh
+++ b/bin/download_geolite2.sh
@@ -1,11 +1,11 @@
 #!/usr/bin/env bash
 # This script downloads and extracts the free GeoLite2 country database
 # from MaxMind for use in development.
+# The licence key is read from environment variable $MAXMIND_LICENCE_KEY, and can be obtained
+# from https://www.maxmind.com
 BASE_DIR="$(dirname "$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )")"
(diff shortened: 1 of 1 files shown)


mythmon:
@tiftran Does the Docker workflow you set up use this script as well?

tiftran:
@mythmon  yes, docker workflow uses download_geolite2.sh

mythmon:
We'll need to make sure that the Docker workflow is updated too, either to not use GeoIP, or to guide the user into configuring authentication for this.

If we can get CI working before all of the above nice-to-haves, I'm fine landing this without updating docs/Docker to unblock things.

tiftran:
@mythmon  sounds good, I'll create a separate issue for updating the docker workflow?

mythmon:
Sure, that sounds like a good plan. They are pretty independent.

leplatrem:
Since in prod, there is already a different workflow, with auth and the update tool, I don't think it makes sense to modify the dev/CI script this way.

PR #2074, opened 2020-01-08
Add version range filter object

### LINKED ISSUE #2073: Filter object for version ranges
Currently, we use a filter object for version filtering that can only check major versions, and cannot distinguish between different point releases or different beta versions. It will also break when we hit a 3 digit version number of Firefox (closer than you might think :fearful:).

In [bug 1543484](https://bugzilla.mozilla.org/show_bug.cgi?id=1543484) Rehan added a proper version comparison function that fixes all of these problems. We couldn't use it though, since we didn't have a way to deal with backwards compatibility.

Now we have capabilities, so we can start using this functionality. Lets add a filter object for it.

mythmon:
Fixes #2073

### DIFF
--- a/normandy/recipes/filters.py
+++ b/normandy/recipes/filters.py
@@ -312,6 +312,47 @@ class VersionFilter(BaseFilter):
         return set()
 
 
+class VersionRangeFilter(BaseFilter):
+    """
+    Match a user running a version in the given range. Uses a version compare
+    filter instead of simple string comparison like VersionFilter.
+
+    The version range is half-open, like Python ranges: If min is 72 and max
+    is 75, 72.0 will be include, 75.0 will not be. `min <= version < max`.
+
+    ..attribute:: type
+
+        ``version_range``
+
+    .. attribute:: min_version
+
+        :example: ``72.0b5``
+
+    .. attribute:: max_version
+
+        :example: ``75.0.1``
+    """
+
+    type = "version_range"
+    min_version = serializers.CharField()
+    max_version = serializers.CharField()
+
+    def to_jexl(self):
+        min_version = self.initial_data["min_version"]
+        max_version = self.initial_data["max_version"]
+
+        return "&&".join(
+            [
+                f'(env.version|versionCompare("{min_version}")>=0)',  # browser version >= min_version
+                f'(env.version|versionCompare("{max_version}")<0)',  # browser version < max_version
+            ]
+        )
+
+    @property
+    def capabilities(self):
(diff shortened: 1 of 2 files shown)


normandy/recipes/filters.py
```diff
@@ -312,6 +312,47 @@ def capabilities(self):
...
+        return "&&".join(
+            [
+                f'(env.version|versionCompare("{min_version}")>=0)',  # browser version >= min_version
+                f'(env.version|versionCompare("{max_version}")<0)',  # browser version < max_version
+            ]
+        )
+
+    @property
+    def capabilities(self):
+        return {"jexl.context.env.version", "jexl.transform.versionCompare"}
```
jaredlockhart: So the idea is that we'll use `VersionRangeFilter` *instead* of `VersionFilter` for clients that support the capability, and fall back to `VersionFilter` for those that don't?

normandy/recipes/tests/test_filters.py
```diff
@@ -39,7 +41,35 @@ def test_uses_only_baseline_capabilities(self, settings):
...
+        if versions is None:
+            versions = [72, 74]
+        return VersionFilter.create(versions=versions)
+
+    def test_generates_jexl(self):
+        filter = self.create_basic_filter(versions=[72, 74])
+        assert set(filter.to_jexl().split("||")) == {
+            '(normandy.version>="72"&&normandy.version<"73")',
+            '(normandy.version>="74"&&normandy.version<"75")',
+        }
```
jaredlockhart: The filter already existed, but it was just missing tests so you're adding them here?

normandy/recipes/filters.py
mythmon: Yes. Importantly, the plan I have here is that Normandy wouldn't actually be checking for that, and it would be up to the people writing recipes to manually pick the right one. It might be nice to have some explicit checks. I think that's mostly something we'd do in Normandy Devtools though.

I happen to know that Firefox 72 and above can safely use this, and that Firefox 71 and below cannot safely use this. Something like [bug 1563564](https://bugzilla.mozilla.org/show_bug.cgi?id=1563564) should make it easier to deal with that in tooling in the future.

normandy/recipes/tests/test_filters.py
mythmon: Yes exactly. Thanks for reading my mind :smile:.

mythmon:
bors r= @jaredlockhart

PR #2075, opened 2020-01-09
Don't fetch a geoip db during CI

mythmon:
Ref #2069.


mythmon:
I think that this should unblock Normandy's CI.

uhlissuh:
bors merge

Issue #1713, opened 2019-01-30
Add filter object for dates


mythmon:
It would be useful to have a filter object that can filter by date. This is already possible in JEXL, and we've used it in practice. It can help future experiments be more declarative about their schedules, especially in relation to tools like Experimenter.

In JEXL this looks something like this:

```jexl
normandy.request_time > '2018-09-25T07:15:00'|date
```

I propose two new filter objects: `notBefore` and `notAfter`, both of which take a single argument: an ISO8601 formatted datetime. These filter objects would specify the time range the recipe is valid. Examples:

```
{
    "type": "notBefore",
    "datetime": "2019-02-01T00:00:00Z"
}
```

```
{
    "type": "notAfter",
    "datetime": "2019-03-01T00:00:00Z"
}
```

These names are borrowed from TLS certificate validity metadata. Alternatively, we could go with other names. Perhaps simply `before` and `after` (which have the opposite meanings than the above) or "startDatetime" and "endDatetime".

Another option would be to combine these into a single filter object, such as:

```
{
    "type": "datetime",
    "notBefore": "2019-02-01T00:00:00Z",
    "notAfter": "2019-03-01T00:00:00Z"
}
```

We would probably want to make each field optional, but require that at least one is set. This might make tools like Delivery Console easier to write.

@peterbe what do you think about the above?

peterbe:
Gut-reaction, on the topic of what to call it, is that I immediately understood what  `notBefore` means. Same with `notAfter`. For `before` and `after` it's less clear. 
Another gut-reaction was that `startDatetime` makes me think "Enable this recipe on $datetime" which is probably not what it means. It just means any time `>= $datetime`. 
So I think that's a good name. 

I'd have to (already) refresh my memory how Filter Objects work in DC and Normandy but I have a feeling that... whatever we do we're going to need to do invariance validation. 

If we keep it as a bundle of:
```json
{
    "type": "datetime",
    "notBefore": "2019-02-01T00:00:00Z",
    "notAfter": "2019-03-01T00:00:00Z"
}
```
it'd be dead-easy to validate that particular one. Just make each date is a sane format and that one is greater than the other if both are present. Then we can concentrate the validation on one individual filter object. However, we might still need to check each one of these against each other e.g. 

```json
[{
    "type": "datetime",
    "notBefore": "2019-02-01T00:00:00Z",
    "notAfter": "2019-04-01T00:00:00Z"
}, {
    "type": "datetime",
[...]

PR #2087, opened 2020-01-30
Add date range filter object

uhlissuh:
ref #1713 

So... would love some feedback on this...


requirements/dev.txt
```diff
@@ -30,3 +30,5 @@ watchdog==0.9.0 \
 Werkzeug==0.16.0 \
     --hash=sha256:7280924747b5733b246fe23972186c6b348f9ae29724135a6dfc1e53cea433e7 \
     --hash=sha256:e5f4a1f98b52b18a93da705a7458e55afb26f32bff83ff5d19189f92462d65c4
+appnope==0.1.0 \
+   --hash sha256:5b26757dc6f79a3b7dc9fab95359328d5747fcb2409d331ea66d0272b90ab2a0
```
uhlissuh: Wasn't sure if you actually wanted to add this to the dependencies... this is what I had to add manually to get the dependencies installed on my machine..

requirements/dev.txt
```diff
@@ -1,5 +1,7 @@
 -r default.txt
 
+appnope==0.1.0 \
+   --hash sha256:5b26757dc6f79a3b7dc9fab95359328d5747fcb2409d331ea66d0272b90ab2a0
```
uhlissuh: Wasn't sure if you actually wanted to add this to the dependencies... this is what I had to add manually to get the dependencies installed on my machine..

normandy/recipes/filters.py
```diff
@@ -353,6 +353,29 @@ def capabilities(self):
...
+            [
+                f'(normandy.request_time>"{not_before}"|date)',
+                f'(normandy.request_time<"{not_after}"|date)',
+            ]
+        )
+
+    @property
+    def capabilities(self):
+        # do we need to make sure 'date' is.. a capability? What is 'date'?
+        return set()
```
mythmon: The word for the thing that `date` is, is a JEXL transform. There is a capability that indicates that a client has the `date` transform, and that capability is `jexl.transform.date` (there isn't really a good way for you to have known that, sorry).

normandy/recipes/filters.py
```diff
@@ -353,6 +353,29 @@ def capabilities(self):
...
+    not_before = serializers.DateTimeField()
+    not_after = serializers.DateTimeField()
+
+    def to_jexl(self):
+        not_before = self.initial_data["not_before"]
+        not_after = self.initial_data["not_after"]
+
+        return "&&".join(
+            [
+                f'(normandy.request_time>"{not_before}"|date)',
```
mythmon: This should probably be `>=` instead of `>`, so that ranges align exactly. This doesn't really matter because time uses such small increments, but for these kind of range filters I think we should be consistent.

requirements/dev.txt
mythmon: This makes sense. Our CI is all Linux, and I think everyone working on Normandy so far is using Linux either directly or via Docker. This looks like an OSX specific thing.

That being said, I *suspect* this line should probably be in `constraints.txt` instead of `dev.txt`. I don't think we actually want to ask pip to install it (for example, on Linux or CI systems), but if it does get installed we want to pin the version and hash.

That of course depends on the error you're getting. If moving it to the constraints file breaks things, lets keep it as is.

uhlissuh: Yes I assumed this was an OSX thing from #2 in the troubleshooting tips in the docs that suggested the platform difference could be the cause! Yay docs. It looks like I can move this to `constraints.txt` just fine.

uhlissuh:
Hey @mythmon, I think I've fixed the things you mentioned, but let me know if I misunderstood. Thanks for reviewing this even while at all hands! I didn't expect it, so thank you.

normandy/recipes/filters.py
mythmon: Sorry, I wasn't quite clear here before. I think that just the "left" side should be `>=`, the right side should stay as `<`. That makes it so that the ranges line up nicely.

Consider a simplified example with real numbers instead of dates: With `>=` and `<` the two ranges `[0, 5)` and `[5, 10)` add up to exactly the range `[0, 10)`, and the first doesn't contain any values the second has, and vice versa. With `>`/`<`, there is a small gap in the middle (5 falls into neither), and with `>=`/`<=` there is one that is duplicated (5 falls into both).

Again, this doesn't really matter for dates, but it matters for some other range filters we might do: It doesn't really matter if the last second in a range is included, but it definitely matters if the last version is included or not. I think it would be good to be consistent anywhere we have ranges in bucket filters.

normandy/recipes/filters.py
```diff
@@ -353,6 +353,28 @@ def capabilities(self):
         return {"jexl.context.env.version", "jexl.transform.versionCompare"}
 
 
+class DateRangeFilter(BaseFilter):
```
mythmon: Sorry, I forgot this in the previous review. Can you add a doc string to this class, like the others have? It is used to automatically generated documentation that gets published here: https://mozilla.github.io/normandy/user/filters.html#module-normandy.recipes.filters

normandy/recipes/filters.py
uhlissuh: ok I see what you are saying. 👍

normandy/recipes/filters.py
uhlissuh: Ok I added one, not sure the language is the best it could be... Can you double check?

mythmon: This looks good to me. I'm a little unsure about the syntax of the RST in the comment, but that's fine for now. It turns out that this isn't being automatically included in the docs anyways. I'm going to look into that in another PR (unless you beat me to it).

PR #2094, opened 2020-02-03
Add profile creation date filter object

### LINKED ISSUE #1714: Add filter object for profile creation date
One of the types of targeting that is needed occasionally is to distinguish between new and existing users. The way this is usually done is to pick a particular date, and declare that any profile created before this date is an existing user, and any profile created after this date is a new user.

This data is available in Telemetry via a variable known as `profile.creationDate`, often abbreviated PCD. Because this value is in telemetry,  it is affected by [bug 1472781](https://bugzilla.mozilla.org/show_bug.cgi?id=1472781): It isn't available during the Telemetry session. Because of this, the JEXL for this type of targeting is more complex than usual. Here's an example:

```jexl
!normandy.telemetry.main ||
normandy.telemetry.main.environment.profile.creationDate > 17483
```

Of note is the units of `creationDate`: it is specified in *days* since January 1st 1970. The reduced precision is to limit the finger-printability of this metric. As such, the value `17483` represents November 13th 2017 (the release of Firefox 57). This particular value was used several times to target users that started using Firefox after the Quantum release.

I propose this structure for the filter object:

```json
{
    "type": "profileCreationDate",
    "direction": "olderThan",
    "date": "2019-02-01",
}
```

The key `direction` can have either the key "olderThan" or "youngerThan". The key `date` is an ISO8601 formatted **date**. Since PCD is given in terms of days, there is no use in being more precise.

The filter expression renderer will need to convert from the date given to the units of "days since epoch".

Alternatives include having separate filter objects for "profile older than" and "profile newer than", or allowing one or two dates to be specified, such as 

```json
{
    "type": "profileCreationDate",
    "notBefore": "2019-02-01",
    "notAfter": "2019-03-01",
}
```

Each key would be optional, but at least one would be required. This kind of ranged PCD filter is less common, but may still be useful.

uhlissuh:
fixes #1714

### DIFF
--- a/normandy/recipes/filters.py
+++ b/normandy/recipes/filters.py
@@ -396,21 +396,21 @@ class DateRangeFilter(BaseFilter):
 
 class ProfileCreateDateFilter(BaseFilter):
     """
-        This filter is meant to distinguish between new and existing users.
-        Target users who have a profile creation date older than or newer than
-        a given date. This date must be specified in days since January 1st 1970.
+    This filter is meant to distinguish between new and existing users.
+    Target users who have a profile creation date older than or newer than
+    a given date.
 
-        .. attribute:: type
+    .. attribute:: type
 
-            ``profile_creation_date``
+        ``profile_creation_date``
 
-        .. attribute:: direction
+    .. attribute:: direction
 
-           :example: ``newer_than``
+       :example: ``newer_than``
 
-       .. attribute:: date
+   .. attribute:: date
 
-          :example: ``18657``
+      :example: ``2020-02-01``
     """
 
     type = "profile_creation_date"
--- a/normandy/recipes/filters.py
+++ b/normandy/recipes/filters.py
@@ -395,6 +395,23 @@ class DateRangeFilter(BaseFilter):
 
 
 class ProfileCreateDateFilter(BaseFilter):
+    """
+        This filter is meant to distinguish between new and existing users.
+        Target users who have a profile creation date older than or newer than
+        a given date. This date must be specified in days since January 1st 1970.
+
+        .. attribute:: type
+
+            ``profile_creation_date``
+
+        .. attribute:: direction
+
+           :example: ``newer_than``
+
+       .. attribute:: date
+
+          :example: ``18657``
+    """
 
     type = "profile_creation_date"
     direction = serializers.CharField()
--- a/normandy/recipes/filters.py
+++ b/normandy/recipes/filters.py
@@ -21,6 +21,7 @@ field, so the final JSON would look something like this:
 """
 
 from rest_framework import serializers
+from datetime import datetime
 
 # If you add a new filter to this file, remember to update the docs too!
 
@@ -393,6 +394,37 @@ class DateRangeFilter(BaseFilter):
         return {"jexl.transform.date"}
 
 
+class ProfileCreateDateFilter(BaseFilter):
+
+    type = "profile_creation_date"
+    direction = serializers.CharField()
+    date = serializers.DateField()
+
+    def to_jexl(self):
+        direction = self.initial_data["direction"]
+        date = self.initial_data["date"]
+
+        days = (datetime.strptime(date, "%Y-%M-%d") - datetime(1970, 1, 1)).days
+
+        if direction == "olderThan":
+            symbol = "<="
+        elif direction == "newerThan":
+            symbol = ">"
+        else:
+            raise serializers.ValidationError(f"Unrecognized direction {direction!r}")
+
+        return "||".join(
+            [
+                "(!normandy.telemetry.main)",
+                f"(normandy.telemetry.main.environment.profile.creationDate{symbol}{days})",
+            ]
+        )
+
+    @property
+    def capabilities(self):
+        return set()
+
+
 by_type = {
     f.type: f
     for f in [
@@ -404,6 +436,7 @@ by_type = {
         VersionFilter,
         VersionRangeFilter,
         DateRangeFilter,
+        ProfileCreateDateFilter,
     ]
 }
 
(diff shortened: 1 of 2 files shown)


normandy/recipes/filters.py
```diff
@@ -353,6 +354,35 @@ def capabilities(self):
...
+    def to_jexl(self):
+        direction = self.initial_data["direction"]
+        date = self.initial_data["date"]
+
+        days = (datetime.strptime(date, "%Y-%M-%d") - datetime(1970, 1, 1)).days
+
+        if direction == "olderThan":
+            symbol = "<"
+        else:
+            symbol = ">"
```
mythmon: I'm a little worried about this logic. It would be pretty easy to have a mistake like `direction: "older-than"` which would surprisingly result in a filter that targets newer users (since it doesn't match exactly `olderThan`).

What would you think about checking for exactly `olderThan` or `newerThan`, and throwing an error if that's not the case?

normandy/recipes/filters.py
```diff
@@ -353,6 +354,35 @@ def capabilities(self):
...
+    date = serializers.DateField()
+
+    def to_jexl(self):
+        direction = self.initial_data["direction"]
+        date = self.initial_data["date"]
+
+        days = (datetime.strptime(date, "%Y-%M-%d") - datetime(1970, 1, 1)).days
+
+        if direction == "olderThan":
+            symbol = "<"
```
mythmon: I think it would be good if these were `<=`/`>` or `<`/`>=`, so that the two opposing filters meet up exactly. We often use this kind of filter in a pair of recipes that we expect to be complements.

normandy/recipes/tests/test_filters.py
```diff
@@ -44,6 +45,18 @@ def test_uses_only_baseline_capabilities(self, settings):
...
 
+class TestProfileCreationDateFilter(FilterTestsBase):
+    def create_basic_filter(self, direction="olderThan", date="2020-02-01"):
+        return ProfileCreateDateFilter.create(direction=direction, date=date)
+
+    def test_generates_jexl(self):
+        filter = self.create_basic_filter()
+        assert set(filter.to_jexl().split("||")) == {
+            "(!normandy.telemetry.main)",
+            "(normandy.telemetry.main.environment.profile.creationDate<18262)",
```
mythmon: Since direction is "interesting", it would probably be good to have test for both values, and maybe some invalid values as well?

normandy/recipes/filters.py
uhlissuh: That makes sense.

normandy/recipes/tests/test_filters.py
uhlissuh: Yes, right.

normandy/recipes/filters.py
uhlissuh: is validation error the right error to be throwing? I see that error being thrown in other filters.

uhlissuh: see code right below this comment...

uhlissuh:
@mythmon I need to add a docstring to this, woops.

uhlissuh:
OK, docstring added.

normandy/recipes/filters.py
```diff
@@ -393,6 +394,54 @@ def capabilities(self):
...
+
+            ``profile_creation_date``
+
+        .. attribute:: direction
+
+           :example: ``newer_than``
+
+       .. attribute:: date
+
+          :example: ``18657``
```
mythmon: Sorry for not seeing this before. I'm looking at this with fresh eyes, and it seems like this should probably be a date. And in fact it seems that's how you are using it in the code? The docs don't seem to match the code.

normandy/recipes/filters.py
```diff
@@ -393,6 +394,54 @@ def capabilities(self):
         return {"jexl.transform.date"}
 
 
+class ProfileCreateDateFilter(BaseFilter):
+    """
+        This filter is meant to distinguish between new and existing users.
```
mythmon: Usually doc strings aren't indented here. More like

```python
def foo():
    """
    docstring here
    """
	...
```

uhlissuh: oh yea.

normandy/recipes/filters.py
uhlissuh: Yea.. I think I just hurriedly added this docstring after the fact without thinking about it enough, sorry about that.

uhlissuh:
I guess I can't be for sure that this filter will take a date formatted like I indicate.

mythmon:
> I guess I can't be for sure that this filter will take a date formatted like I assumed it would be in the filter code.

What happens when someone provides a date in an invalid format? What should happen? It would probably be good to have a test for this error case, and also the case where the user provides a bad value for direction.

uhlissuh:
When the provided date is incorrectly formatted, `strptime` throws a `ValueError`... this occurs in the `create` method on the `Basefilter` class.. which then throws an `assertionError`. I added two tests that test the cases where errors are thrown. Is this what you meant @mythmon?

mythmon:
Looks good to me. I was also curious how the errors that were thrown would convert to HTTP responses. Looks like they work just fine:

```shell
$ https --verify=no -pb POST :8000/api/v3/recipe/ Authorization:'Insecure mcooper@mozilla.com' action_id=4 name="Te
st bad filter object" arguments:='{"message": "it works?"}' filter_object:='[{"type": "profile_creation_date", "direction": "newer_than", "date": "tuesday"}]'

{
    "filter_object": {
        "0": {
            "date": [
                "Date has wrong format. Use one of these formats instead: YYYY-MM-DD."
            ]
        }
    }
}
```

mythmon:
bors merge

uhlissuh:
@mythmon thanks for showing me that. Good to know!

PR #2111, opened 2020-02-18
Add POSTGRES_HOST_AUTH_METHOD: trust to docker-compose file and CI

glasserc:
I think this was previously assumed, but recent Postgres containers
now refuse to start unless given this reassurance. For CI and local
dev, this is totally fine.


glasserc:
https://github.com/docker-library/postgres/pull/658 is the upstream change.

Arguably the right change for CI is to move to a hard-coded password, but I'm not sure it's worth it?

PR #2116, opened 2020-02-27
Remove more aliases

glasserc:
Refs #1863. 

This PR is mostly structured as individual commits which remove one property at a time, converting existing uses to be either `.approved_revision` or `.latest_revision` based on my best guess about what is appropriate in each case.

The main point of contention here is the serializers, which use these properties more-or-less explicitly. The v1 serializers are mostly used for signed/enabled recipes, so the cleanest way to fix this would be to restrict their use to that. However, the tests indicate that the v1 API can be/still is used with non-enabled recipes. I'm not expert enough in Django/DRF to figure out how to remove that stuff surgically, so instead this PR takes the approach of moving the existing `current_revision` logic (i.e. `approved_revision or latest_revision`) into the RecipeSerializer on the assumption that it's only used in read-only circumstances so it's probably fine. The `MinimalRecipeSerializer` is converted to only use `approved_revision` as it is mostly intended for signed recipes which must be approved. These changes are perhaps not the cleanest possible approach but I thought changing the v1 API was outside the scope of the changes I wanted to make (these changes already being substantial).


normandy/capabilities/api/v3/views.py
```diff
@@ -17,7 +17,7 @@ def get(self, request):
             capabilities[cap] = {"is_baseline": True}
 
         for recipe in Recipe.objects.all():
-            for cap in recipe.capabilities:
+            for cap in recipe.latest_revision.capabilities:
```
glasserc: I wasn't 100% sure about the implications for this API.

normandy/recipes/api/filters.py
```diff
@@ -36,7 +36,7 @@ def filter(self, qs, value):
                 raise TypeError("BaselineCapabilitiesFilter can only be used to filter recipes")
             match_ids = []
             for recipe in recipes:
-                if recipe.uses_only_baseline_capabilities():
+                if recipe.latest_revision.uses_only_baseline_capabilities():
```
glasserc: I'm not sure about this one either. I think it probably doesn't matter much what it does since we don't use it when e.g. publishing to Remote Settings, but maybe you have a better idea.

glasserc:
I'm not sure who to ask for review but here's what I got.

normandy/capabilities/api/v3/views.py
mythmon: This API isn't really used much, so I think it is fine. I think the more correct thing would be to iterate over all revisions, but I don't expect capabilities to vary much over time so it shouldn't make any difference.

That being said, this whole API isn't that useful, because what we should be doing is pulling this from the build artifacts generated by Firefox now. That's future work though. I think that this is good as it is.

normandy/recipes/api/filters.py
mythmon: I think that this should be `approved_revision`. This is primarily used so that the `/api/v1/recipes/signed/` endpoint stays compatible with old clients. Since that endpoint primarily deals with approved revisions, I think this filter should as well.

normandy/recipes/models.py
```diff
@@ -108,10 +108,6 @@ def __repr__(self):
     def __str__(self):
         return self.latest_revision.name
 
-    @property
-    def current_revision(self):
-        return self.approved_revision or self.latest_revision
```
mythmon: :tada:

glasserc:
I changed it to check for recipes that have approved revisions only. I wasn't sure of the best way to change the test; it was trying to export to Remote Settings so I threw a mock for that in. I know you gave me authority to merge it myself but I'd like another review.

glasserc:
Also, what do I do about the yarn-audit failures? Merge in master?

mythmon:
> Also, what do I do about the yarn-audit failures? Merge in master?

Bors will run its tests on the result of merging this to master, so we don't need to fix it in the branch. Merging master into this PR just makes history messier.

PR #2123, opened 2020-03-03
Add OS filter object

### LINKED ISSUE #2101: Filter Object: OS
Let's add a filter object for filtering OS, I guess the options look something like:

Windows
OSX
Linux

We need to be able to filter for multiple OS's but it's not clear to me whether a single instance of this filter object takes a single OS parameter and then we can OR multiple of them togethether? @mythmon ?

uhlissuh:
fixes #2101 

I named this PlatformFilter because I thought using OS in the name seemed a bit clunky but I can change it if others want. We use the word "platform" on experimenter a lot to refer to this same thing.

I still am not clear on how to know whether or not this requires a capabilities check. I can see from the capabilities build artifact that there is something referencing OS... again I probably don't have enough information on understanding when a thing I'm doing inside a filter object requires a capabilities check. 

I am assuming the jexl we want this filter object to spit out looks like
`(normandy.os.isMac || normandy.os.isLinux)`

### DIFF
--- a/docs/user/filters.rst
+++ b/docs/user/filters.rst
@@ -22,6 +22,7 @@ Filter Objects
 .. autoclass:: BucketSampleFilter()
 .. autoclass:: StableSampleFilter()
 .. autoclass:: VersionFilter()
+.. autoclass:: PlatformFilter()
 
 Filter Expressions
 ------------------
--- a/normandy/recipes/filters.py
+++ b/normandy/recipes/filters.py
@@ -157,6 +157,41 @@ class CountryFilter(BaseFilter):
         return set()
 
 
+class PlatformFilter(BaseFilter):
+    """Match a user based on what operating system they are using.
+
+    .. attribute:: type
+
+        ``platform``
+
+    .. attribute:: platforms
+
+        List of platforms to filter against. The choices are `all_linux`,
+        `all_windows`, and `all_mac`.
+
+        :example: ``["all_windows", "all_linux"]``
+    """
+
+    type = "platform"
+    platforms = serializers.ListField(child=serializers.CharField(), min_length=1)
+
+    def to_jexl(self):
+        platforms_jexl = []
+        for platform in self.initial_data["platforms"]:
+            if platform == "all_mac":
+                platforms_jexl.append("normandy.os.isMac")
+            elif platform == "all_windows":
+                platforms_jexl.append("normandy.os.isWindows")
+            elif platform == "all_linux":
+                platforms_jexl.append("normandy.os.isLinux")
+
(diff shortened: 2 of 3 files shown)


normandy/recipes/filters.py
```diff
@@ -157,6 +157,41 @@ def capabilities(self):
...
+
+    def to_jexl(self):
+        platforms_jexl = []
+        for platform in self.initial_data["platforms"]:
+            if platform == "all_mac":
+                platforms_jexl.append("normandy.os.isMac")
+            elif platform == "all_windows":
+                platforms_jexl.append("normandy.os.isWindows")
+            elif platform == "all_linux":
+                platforms_jexl.append("normandy.os.isLinux")
```
mythmon: Shouldn't there be an error if there is an unrecognized platform in the list?

normandy/recipes/filters.py
```diff
@@ -157,6 +157,41 @@ def capabilities(self):
...
+
+        return "||".join((f"{p}" for p in platforms_jexl))
```
mythmon: Couldn't you replace `f"{p}"` with just `p`, since `p` is a string already?

normandy/recipes/filters.py
uhlissuh: I thought about that... but I also thought that this was probably going to be coming in from a multiselect... I thought the possibility of something else getting entered was slim. I can add it though.

mythmon: One of the rules for Normandy server is to always double check the client. We cannot assume that this is coming from a multi-select, or even from a client that we control. In the worst case, this could be coming from an attacker that has slipped through other defenses. In the more realistic case, the data could be coming from a mythmon who is using a CLI because I couldn't the normal UI to do something. (I actually did this today to debug a problem).

Generally, we should consider what this API looks like independent of an UI. In fact, this API *is* our interface. The user of the interface are tools like NDT, Edi, Firefox, and the Data Pipeline. Within this hierarchy, NDT doesn't get special privilege. It's just another user.

uhlissuh: OK I understand and will add it to this one, as well as future ones. Thanks.

uhlissuh:
@mythmon thank you for explaining more about the capabilities check. I think it makes sense and will try to approach future filter objects with this explanation in mind.

docs/user/filters.rst
```diff
@@ -22,6 +22,7 @@ Filter Objects
 .. autoclass:: BucketSampleFilter()
 .. autoclass:: StableSampleFilter()
 .. autoclass:: VersionFilter()
+.. autoclass:: PlatformFilter()
```
mythmon: Thanks for adding this!

Issue #2124, opened 2020-03-03
Make make check usable for users who run normandy with docker and without


tiftran:
I run normandy locally without docker but during releases would run `make check`to ensure everything passes. As a result `make check` goes through the venv folder for lint checking and fails.

PR #2125, opened 2020-03-03
add exclusions for venv for make check and code_format

tiftran:
Using docker workflow and not using it, has caused some tiny problems.
`make check` and `make code_format` goes into the venv and checks/lints, change so that it excludes the venv file.

ref: #2124


tiftran:
bors merge

PR #2127, opened 2020-03-05
Add pref filter object

### LINKED ISSUE #2103: Filter Object: Pref
We need a filter object that takes a pref and optionally a value and supports the following options:

- eq
- neq
- gt
- lt
- is_user_set (this can be either true or false)

uhlissuh:
fixes #2103 

update: this is now 3 filter objects. 
__________

So, about the design of this:

you pass three things into the filter... the `pref`, `comparison` and `value`. 

The options for what I call `comparison` are either going to be values that translate to symbols... like "greater_than" or "equal" or they will be the other two things... "is_user_set" and "preference_exists".

then `value` will be a string representation of either characters or a number or bool. So for the `comparison` of "is_user_set" you are passing a value of "true" or "false".

### DIFF
--- a/normandy/recipes/filters.py
+++ b/normandy/recipes/filters.py
@@ -223,24 +223,24 @@ class PrefCompareFilter(BaseFilter):
         value = self.initial_data["value"]
         pref = self.initial_data["pref"]
 
+        if comparison == "contains":
+            return f"{json.dumps(value)} in '{pref}'|preferenceValue"
         if comparison == "equal":
             symbol = "=="
         elif comparison == "not_equal":
             symbol = "!="
         elif comparison == "greater_than":
-            symbol = "<"
-        elif comparison == "greater_than_equal":
-            symbol = "<="
-        elif comparison == "less_than":
             symbol = ">"
-        elif comparison == "less_than_equal":
+        elif comparison == "greater_than_equal":
             symbol = ">="
-        elif comparison == "in":
-            symbol = "in"
+        elif comparison == "less_than":
+            symbol = "<"
+        elif comparison == "less_than_equal":
+            symbol = "<="
         else:
             raise serializers.ValidationError(f"Unrecognized comparison {comparison!r}")
 
-        return f"{json.dumps(value)} {symbol} '{pref}'|preferenceValue"
+        return f"'{pref}'|preferenceValue {symbol} {json.dumps(value)}"
 
     @property
     def capabilities(self):
--- a/normandy/recipes/filters.py
+++ b/normandy/recipes/filters.py
@@ -194,6 +194,76 @@ class PlatformFilter(BaseFilter):
         return set()
 
 
+class PrefFilter(BaseFilter):
+    """Match a user based on a preference having some sort of equality with a value.
+        Also, this filter can match based on if a pref is user set.
+
+    .. attribute:: type
+
+        ``pref``
+
+    .. attribute:: value
+
+        Can be a string representation of a number, characters, or boolean value.
+
+        :example: ``true`` or ``10`` or ``default``
+
+    .. attribute:: comparison
+
+        Options are ``equal``, ``not_equal``, ``greater_than``,
+        ``less_than``, ``is_user_set``, or ``profile_exists``.
+    """
+
+    type = "pref"
+    pref = serializers.CharField()
+    value = serializers.CharField()
+    comparison = serializers.CharField()
+
+    def to_jexl(self):
+        comparison = self.initial_data["comparison"]
+        value = self.initial_data["value"]
+        pref = self.initial_data["pref"]
+
+        if comparison == "is_user_set":
+            if value == "true":
+                return f"'{pref}'|preferenceIsUserSet"
+            elif value == "false":
+                return f"!('{pref}'|preferenceIsUserSet)"
+            else:
+                raise serializers.ValidationError(
+                    f"Unrecognized value {value!r} for comparison {comparison!r}"
+                )
+        elif comparison == "preference_exists":
+            if value == "true":
+                return f"'{pref}'|preferenceExists"
+            elif value == "false":
+                return f"!('{pref}'|preferenceExists)"
+            else:
+                raise serializers.ValidationError(
+                    f"Unrecognized value {value!r} for comparison {comparison!r}"
+                )
+        if comparison == "equal":
+            symbol = "=="
+        elif comparison == "not_equal":
+            symbol = "!="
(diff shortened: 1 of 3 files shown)


normandy/recipes/tests/test_filters.py
```diff
@@ -175,6 +176,47 @@ def test_throws_error_on_bad_platform(self):
...
+        assert filter.to_jexl() == "'browser.urlbar.maxRichResults'|preferenceExists"
+
+    def test_generates_jexl_pref_exists_false(self):
+        filter = self.create_basic_filter(comparison="preference_exists", value="false")
+        assert filter.to_jexl() == "!('browser.urlbar.maxRichResults'|preferenceExists)"
+
+    def test_generates_jexl_is_pref_exists_error(self):
+        filter = self.create_basic_filter(comparison="preference_exists", value="dog")
+        with pytest.raises(serializers.ValidationError):
+            filter.to_jexl()
```
mythmon: What's the error here? Can you make it more specific, either in a comment or (preferably) by making the test check that the expected error is the one that was thrown?

normandy/recipes/filters.py
```diff
@@ -194,6 +194,76 @@ def capabilities(self):
...
+
+    .. attribute:: type
+
+        ``pref``
+
+    .. attribute:: value
+
+        Can be a string representation of a number, characters, or boolean value.
+
+        :example: ``true`` or ``10`` or ``default``
```
mythmon: I'm not quite sure how to do this, but I think we should use the JSON type system instead of making everything a string. So instead of `{"value": "10"}`, `{"value": "true"}` and `{"value": "'default'"}`, we would have `{"value":10}`, `{"value": true}`, and `{"value": "default"}`. This makes it clearer what type is intended, and I think it will also avoid errors where we forget the quotes in the case the value is a string.

One important reason for this is that sometimes we can't tell if something should be a string or a number easily by looking at it. In [this rollout](https://experimenter.services.mozilla.com/experiments/doh-us-staged-rollout-to-all-us-desktop-users/) we needed to set a preference value to the *string* `"1896163212345"`. This is because that number is too large to fit in a preference integer without overflowing.

Again, I'm not sure how we can get the serializer to accept this, but I'm sure there is a way.

normandy/recipes/tests/test_filters.py
```diff
@@ -175,6 +178,72 @@ def test_throws_error_on_bad_platform(self):
...
+        filter = self.create_basic_filter(comparison="greater_than_equal")
+        assert filter.to_jexl() == "10 <= 'browser.urlbar.maxRichResults'|preferenceValue"
+
+    def test_generates_jexl_less_than(self):
+        filter = self.create_basic_filter(comparison="less_than")
+        assert filter.to_jexl() == "10 > 'browser.urlbar.maxRichResults'|preferenceValue"
+
+    def test_generates_jexl_less_than_equal(self):
+        filter = self.create_basic_filter(comparison="less_than_equal")
+        assert filter.to_jexl() == "10 >= 'browser.urlbar.maxRichResults'|preferenceValue"
```
mythmon: These tests are fine. I'd like to show you a trick we could use to make them more concise. I'll leave it up to you which version you'd prefer though.


```suggestion
    @pytest.mark.parametrize("comparison,symbol", [
      ("greater_than", ">"),
      ("greater_than_equal", ">="),
      ("less_than", "<"),
      ("less_than_equal", "<="),
    ])
    def test_generates_jexl_number_ops(self, comparison, symbol):
        filter = self.create_basic_filter(comparison=comparison)
        assert filter.to_jexl() == f"10 {symbol} 'browser.urlbar.maxRichResults'|preferenceValue"
```

normandy/recipes/filters.py
```diff
@@ -194,6 +194,122 @@ def capabilities(self):
...
+        comparison = self.initial_data["comparison"]
+        value = self.initial_data["value"]
+        pref = self.initial_data["pref"]
+
+        if comparison == "equal":
+            symbol = "=="
+        elif comparison == "not_equal":
+            symbol = "!="
+        elif comparison == "greater_than":
+            symbol = "<"
```
mythmon: Isn't `<` the symbol for less than, not greater? Why'd you flip them?

normandy/recipes/tests/test_filters.py
rehandalal: 😮 TIL: this trick as well as suggested change!

normandy/recipes/filters.py
rehandalal: I suspect this was because of my suggestion that value should be on the left of the operator, thus looking at how `value` compares to `preferenceValue` instead of the other way around. This change was largely to allow the `in` to make sense. But maybe we should flip that back and `in` should be a special case that is renamed to `contains`?

rehandalal: i'm thinking about the UI now and realizing you probably want to enter the pref name first, then the comparison type and then the value you're comparing to. so maybe the original order of comparing `preferenceValue` to `value` makes more sense.

i.e. for `>`, `>=`, `<`, `<=`, `==`:
```py
return f"'{pref}'|preferenceValue {symbol} {json.dumps(value)}"
```

but for `contains`:
```py
return f"{json.dumps(value)} in '{pref}'|preferenceValue"
```

mythmon: Oh 🤦‍♂️. I totally missed the swapped arguments below. I think the way it is now, where comparison words and the comparison operators are flipped is going to cause a lot of confusion (see: me an hour ago). This works, I suppose, but it seems very strange to me. At the very least, it should have a comment explaining what's going on.

normandy/recipes/tests/test_filters.py
uhlissuh: I made this change. I decided to not use Github's automatic "commit suggestion" button though, so I could change the comparison direction back.

normandy/recipes/filters.py
uhlissuh: change made.

rehandalal: > Oh 🤦‍♂️. I totally missed the swapped arguments below. I think the way it is now, where comparison words and the comparison operators are flipped is going to cause a lot of confusion (see: me an hour ago). This works, I suppose, but it seems very strange to me. At the very least, it should have a comment explaining what's going on.

yeah i also did a bad job explaining that i didn't think the operators should be flipped. but i think where we've landed is the best solution. thanks @uhlissuh!

uhlissuh:
bors merge

PR #2132, opened 2020-03-09
v3 API contract tests requiring authentication

chartjes:
This PR contains v3 API contract tests that required authentication.

* documentation now updated to require Docker as a dependency to run the tests as this is how it would be done in CI
* tests create a superuser inside the `normandy_app` container for use by the v3 API tests
* v1 and v3 API tests now in their own directories
* updated requirements/constraints.txt and requirements/dev.txt for new dependencies

Given how authentication works in the application, I am unsure of a solution that allows these tests to be against non-Dockerized versions of Normandy.


contract-tests/conftest.py
```diff
@@ -1,11 +1,16 @@
...
 import pytest
 import requests
+import sys
 import urllib3
 
 # Disable any warnings about SSL connections
 urllib3.disable_warnings()
 
+# Fix our path so our tests can see modules inside Normandy
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```
jaredlockhart: Isn't there some better way to do this outside of python space?  I fear this will be inflexible to changes.  Let me try this locally and see if I can find something simpler.

contract-tests/v3_api/conftest.py
```diff
@@ -1,4 +1,26 @@
...
+    # Add them as a superuser to the system running in Docker
+    docker_compose = (
+        subprocess.check_output("which docker-compose", shell=True).decode("ascii").strip("\n")
+    )
+    subprocess.call(
+        "{} run app python manage.py createsuperuser --noinput --email={} --user={}".format(
+            docker_compose, email, user
+        ),
+        shell=True,
+    )
```
jaredlockhart: I think we can probably do this inside the circleci config rather than here?

jaredlockhart:
bors try

contract-tests/conftest.py
chartjes: The reason I did this was to make sure the tests that involve signing things could see the libraries. My knowledge of Python internals is very shallow and these seemed to be a solution that worked. Definitely open to a better way of doing it

contract-tests/v3_api/conftest.py
chartjes: Yeah, it is just a command to create the super-user the test needs. Unless you want to create a specific superuser just for CI and have it always available?

jaredlockhart: Like can't you just move these steps here:

https://github.com/mozilla/normandy/pull/2132/files#diff-1d37e48f9ceff6d8030570cd36286a61R257-R261

.circleci/config.yml
```diff
@@ -254,20 +252,13 @@ jobs:
...
-          name: Waiting for web server to be available
+          name: Build Docker image for test
           command: |
-            ./ci/wait-for-it.sh localhost:8000 \
-              --timeout=30 --strict \
-              -- echo "Done waiting. It should work now."
+            docker-compose build
+            docker-compose run app sh -c "/app/bin/wait-for-it.sh db:5432 -- python manage.py migrate"
+            docker-compose run app python manage.py update_actions
+            docker-compose run app python manage.py initial_data
```
jaredlockhart: You should also be able to replace these with `make refresh` and it'll do all of those

jaredlockhart:
bors try

contract-tests/v3_api/conftest.py
```diff
@@ -1,4 +1,26 @@
...
+
+from faker import Faker
+
+
+@pytest.fixture(scope="session")
+def headers():
+    # Create a test user
+    fake = Faker()
+    email = fake.company_email()
+    user = fake.user_name()
```
jaredlockhart: I think you can just drop the faker stuff out altogether and just use a hard coded test name like

`test-user@example.com`

jaredlockhart:
Also I think if you merge the latest master it'll get rid of that js test stuff for you

contract-tests/v3_api/conftest.py
chartjes: Done

jaredlockhart:
bors try

PR #2135, opened 2020-03-09
Add windows build number filter object

### LINKED ISSUE #2102: Filter Object: Windows Build Number
We need to be able to filter for specific OS version and I believe this is only applicable for Windows.  It's not clear to me whether this will allow for some concept of a version range? @mythmon ?

uhlissuh:
fixes #2102

### DIFF
--- a/normandy/recipes/filters.py
+++ b/normandy/recipes/filters.py
@@ -48,6 +48,22 @@ class BaseFilter(serializers.Serializer):
         """Render this filter to a JEXL expression"""
         raise NotImplementedError
 
+    def to_operator(self, comparison):
+        if comparison == "equal":
+            return "=="
+        elif comparison == "not_equal":
+            return "!="
+        elif comparison == "greater_than":
+            return ">"
+        elif comparison == "greater_than_equal":
+            return ">="
+        elif comparison == "less_than":
+            return "<"
+        elif comparison == "less_than_equal":
+            return "<="
+        else:
+            raise serializers.ValidationError(f"Unrecognized comparison {comparison!r}")
+
 
 class ChannelFilter(BaseFilter):
     """
@@ -545,6 +561,24 @@ class DateRangeFilter(BaseFilter):
     @property
     def capabilities(self):
         return {"jexl.transform.date"}
+
+
+class WindowsBuildNumberFilter(BaseFilter):
+    type = "windows_build_number"
+    build_number = serializers.IntegerField()
+    comparison = serializers.CharField()
+
+    def to_jexl(self):
+        comparison = self.initial_data["comparison"]
+        build_number = self.initial_data["build_number"]
+
+        symbol = self.to_operator(comparison)
+
+        return f"(normandy.os.isWindows && normandy.os.windowsBuildNumber {symbol} {build_number})"
+
+    @property
+    def capabilities(self):
+        return set()
 
 
 class ProfileCreateDateFilter(BaseFilter):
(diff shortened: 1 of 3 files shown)


normandy/recipes/filters.py
```diff
@@ -48,6 +48,22 @@ def to_jexl(self):
...
+            return ">"
+        elif comparison == "greater_than_equal":
+            return ">="
+        elif comparison == "less_than":
+            return "<"
+        elif comparison == "less_than_equal":
+            return "<="
+        else:
+            raise serializers.ValidationError(f"Unrecognized comparison {comparison!r}")
+
```
uhlissuh: What do you think about putting a method like this on the base filter class and using when needed?

rehandalal: I don't think that's the right direction. even if we were going with this kind of solution there's a couple things that seem wrong:
- the name is confusing since you're not converting the instance of the filter to an operator.
- this should likely have been a static method since it's not instance specific and is largely a utility function.

That being said I think the better approach is to extend `BaseFilter` into `BaseComparisonFilter`, something along the lines of:
```py
class BaseComparisonFilter(BaseFilter):
    value = serializers.IntegerField()
    comparison = serializers.CharField()

    @property
    def left_of_operator(self):
        raise NotImplementedError("Not correctly implemented.")

    def to_jexl(self):
        comparison = self.initial_data["comparison"]
        value = self.initial_data["value"]

        if comparison == "equal":
            operator = "=="
        elif comparison == "not_equal":
            operator = "!="
        elif comparison == "greater_than":
            operator = ">"
        elif comparison == "greater_than_equal":
            operator = ">="
        elif comparison == "less_than":
            operator = "<"
        elif comparison == "less_than_equal":
            operator = "<="
        else:
            raise serializers.ValidationError(f"Unrecognized comparison {comparison!r}")

        return f"{self.left_of_operator} {operator} {value}"
```

And then `WindowsBuildNumberFilter` would be:
```py
class WindowsBuildNumberFilter(BaseComparisonFilter):
    type = "windows_build_number"

    @property 
    def left_of_operator(self):
        return "normandy.os.windowsBuildNumber"

    @property
    def capabilities(self):
        return set()

    def to_jexl(self):
        return f"normandy.os.isWindows && ${super().to_jexl()}"
```

uhlissuh: ok, sounds good, thanks for writing out so much example code.

normandy/recipes/filters.py
```diff
@@ -547,6 +577,21 @@ def capabilities(self):
...
+    def left_of_operator(self):
+        return "normandy.os.windowsBuildNumber"
+
+    @property
+    def capabilities(self):
+        return set()
+
+    def to_jexl(self):
+        return f"(normandy.os.isWindows && {super().to_jexl()})"
+
```
uhlissuh: @rehandalal I put parentheses around this because that seems to be how it's often represented in the filter expressions... but I am guessing they don't actually matter in this case because it's an `&&` operation. Is that correct?

rehandalal: yeah given that it's an `&&` it should not matter either way

uhlissuh: @rehandalal So in the case where the parentheses don't matter, even if it often appears that way in the filter expressions of past recipes, would you prefer I left the parentheses out or use them just for a bit of extra clarity?

mythmon: It doesn't matter today, since everything is `&&`, but it might matter in the future if we add an OR filter. Lets keep the parens.

uhlissuh: OK, well I inserted Rehan's example above, and kept the parens, so let me know what y'all think of it now.

PR #2136, opened 2020-03-10
Add addon filter objects

### LINKED ISSUE #2104: Filter Object: Addon
We need a filter object for filtering for the presence of an addon by its addon id.  I think if we want to exclude an addon we will have to combine this with https://github.com/mozilla/normandy/issues/2105

uhlissuh:
fixes #2104 

So, splitting these into two filter objects came from Rehan, which I thought was a good idea. He mentioned seeing about having them share some code, because there is duplication here, and I kinda looked at that for a while and arrived at wondering if we will really gain much here trying to have these two share code? There's only two that are shaped like this right now. There isn't a clear thing I could pull out, maybe someone sees something. Perhaps that's an argument for making them one filter?

### DIFF
--- a/normandy/recipes/filters.py
+++ b/normandy/recipes/filters.py
@@ -835,6 +835,8 @@ by_type = {
         WindowsVersionFilter,
         WindowsBuildNumberFilter,
         NegateFilter,
+        AddonActiveFilter,
+        AddonInstalledFilter,
     ]
 }
 
--- a/normandy/recipes/filters.py
+++ b/normandy/recipes/filters.py
@@ -224,6 +224,98 @@ class PlatformFilter(BaseFilter):
         return set()
 
 
+class AddonActiveFilter(BaseFilter):
+    """Match a user based on if a particular addon is active.
+
+    .. attribute:: type
+
+        ``addon_active``
+
+    .. attribute:: addons
+        List of addon ids to filter against.
+
+        :example: ``["uBlock0@raymondhill.net", "pioneer-opt-in@mozilla.org"]``
+
+    .. attribute:: any_or_all
+        This will determine whether the addons are connected with an "&&" operator,
+        meaning all the addons must be active for the filter to evaluate to true,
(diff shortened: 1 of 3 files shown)


uhlissuh:
I added the Jared review and then removed it because he was doing a trial of his slackbot. :)

normandy/recipes/filters.py
```diff
@@ -224,6 +224,98 @@ def capabilities(self):
...
+                (f'normandy.addons["{addon}"].isActive' for addon in self.initial_data["addons"])
+            )
+        elif any_or_all == "any":
+            return "||".join(
+                (f'normandy.addons["{addon}"].isActive' for addon in self.initial_data["addons"])
+            )
+        else:
+            raise serializers.ValidationError(
+                f"Unrecognized string for any_or_all: {any_or_all!r}"
+            )
```
rehandalal: probably a cleaner way to do this is:
```py
if any_or_all == "all":
    glue = "&&"
elif any_or_all == "any":
    glue = "||"
else:
    raise # Exception here

return glue.join(f'...')
```

uhlissuh: yes you're right

jaredlockhart: Not to butt in but generally for this kind of mapping pattern you can use a dictionary

normandy/recipes/tests/test_filters.py
```diff
@@ -206,6 +208,54 @@ def test_throws_error_on_bad_platform(self):
             filter.to_jexl()
 
 
+class TestAddonInstalledFilter(FilterTestsBase):
+    def create_basic_filter(self, addons=["@abcdef", "ghijk@lmnop"], any_or_all="any"):
+        return AddonInstalledFilter.create(addons=addons, any_or_all=any_or_all)
+
+    def test_generates_jexl_installed_not_active(self):
```
mythmon: I the the names of these tests are out of date? It looks like they may have been written before the filter was split into two.

uhlissuh:
bors merge

PR #2141, opened 2020-03-16
Remove double publication on Remote Settings (fixes #2047)

### LINKED ISSUE #2047: Only publish to Remote Settings recipes-capabilities collection
With the current implementation #1960, recipes are published to two Remote Settings collections.

When a recipe does not leverage advanced client capabilities, it is published to two collections: one fetched by latest clients, and one for the legacy clients. This is done synchronously within the HTTP request lifecycle on the Normandy server, when the recipe is created/modified.

This means that, on the Remote Settings side, we receive two set of HTTP requests (to create and sign both recipes collections), within a few milliseconds. Recipes are published or changed quite often (several times a day), so we see these double hits on Remote Settings quite often. Each publication leads to two consecutive requests to Autograph and the Push service.

I don't really like this situation where two Push notifications are sent to clients with a few milliseconds of gap. I think it would be better if we could avoid it.

The solution I would suggest consists in publishing in the exhaustive collection (with capabilities) only.  And then, have a scheduled job to backport records in the legacy collection every hour or so for Firefox < 70.

We already have some backport jobs for the onecrl blocklist etc., so this would be trivial to add.
This is would be a bit particular since it would filter based on the listed capabilities, so instead of having some Normandy specific code in the Remote Settings lambdas repo, we could have an extra field on recipes (eg. `is_baseline: true/false`) and just filter on this...

What do you think?

leplatrem:
Fixes #2047

### DIFF
--- a/docs/ops/config.rst
+++ b/docs/ops/config.rst
@@ -195,7 +195,7 @@ in other Django projects.
 
     The account password to authenticate as DJANGO_REMOTE_SETTINGS_USERNAME.
 
-.. envvar:: REMOTE_SETTINGS_CAPABILITIES_COLLECTION_ID
+.. envvar:: DJANGO_REMOTE_SETTINGS_CAPABILITIES_COLLECTION_ID
 
     :default: ``normandy-recipes-capabilities``
 
--- a/docs/ops/config.rst
+++ b/docs/ops/config.rst
@@ -195,9 +195,9 @@ in other Django projects.
 
     The account password to authenticate as DJANGO_REMOTE_SETTINGS_USERNAME.
 
-.. envvar:: DJANGO_REMOTE_SETTINGS_COLLECTION_ID
+.. envvar:: REMOTE_SETTINGS_CAPABILITIES_COLLECTION_ID
 
-    :default: ``normandy-recipes``
+    :default: ``normandy-recipes-capabilities``
 
     The name of the Remote Settings collection where the recipes will be published.
 
--- a/normandy/recipes/tests/test_commands.py
+++ b/normandy/recipes/tests/test_commands.py
@@ -495,13 +495,13 @@ class TestSyncRemoteSettings(object):
         call_command("sync_remote_settings")
 
         requests = requestsmock.request_history
-        # The first two requests should be to get the existing records
+        # First request should be to get the existing records
         assert requests[0].method == "GET"
         assert requests[0].url.endswith(self.capabilities_published_records_url)
-        # The next two should be to PUT the missing recipe2
+        # The next should be to PUT the missing recipe2
         assert requests[1].method == "PUT"
         assert requests[1].url.endswith(r2_capabilities_url)
-        # The final two should be to approve the changes to the two collections
+        # The final one should be to approve the changes
         assert requests[2].method == "PATCH"
         assert requests[2].url.endswith(self.capabilities_workspace_collection_url)
         # And there are no extra requests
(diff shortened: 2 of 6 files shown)


leplatrem:
```
===============================================================================
FAILED: eslint
===============================================================================
yarn run v1.22.4
$ /root/repo/node_modules/.bin/eslint client/actions/tests/index.js client/actions/opt-out-study/index.js client/actions/tests/test_console-log.js client/actions/tests/test_show-heartbeat.js client/tests/utils.js client/actions/console-log/index.js client/utils/absolute-path.js client/actions/tests/test_opt-out-study.js client/utils/closest.js client/actions/tests/test_preference-experiment.js client/actions/show-heartbeat/index.js karma.conf.js client/actions/utils.js normandy/base/management/commands/data/console-log.js webpack.config.js babel.config.js client/actions/tests/utils.js client/actions/preference-experiment/index.js
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

-------------------------------------------------------------------------------
Additional error output:
-------------------------------------------------------------------------------
TypeError: Cannot read property 'range' of null
Occurred while linting /root/repo/client/actions/opt-out-study/index.js:42
    at SourceCode.getTokenBefore (/root/repo/node_modules/eslint/lib/source-code/token-store/index.js:298:18)
    at checkSpacingBefore (/root/repo/node_modules/eslint/lib/rules/template-curly-spacing.js:60:42)
    at TemplateElement (/root/repo/node_modules/eslint/lib/rules/template-curly-spacing.js:119:17)

```

docs/ops/config.rst
```diff
@@ -195,9 +195,9 @@ in other Django projects.
 
     The account password to authenticate as DJANGO_REMOTE_SETTINGS_USERNAME.
 
-.. envvar:: DJANGO_REMOTE_SETTINGS_COLLECTION_ID
+.. envvar:: REMOTE_SETTINGS_CAPABILITIES_COLLECTION_ID
```
mythmon: The `DJANGO_` prefix is still required. Almost of the our settings that use `django-configurations` use that prefix by default.

normandy/recipes/tests/test_exports.py
```diff
@@ -25,35 +25,24 @@ def rs_urls(self, rs_settings):
             "workspace": {"capabilities": {}, "baseline": {}},
```
mythmon: Since we don't have to vary based on capabilities vs baseline, we should remove the second layer of this hierarchy. So instead of `rs_urls["workspace"]["capabilities"]["records"]` we'd have `rs_urls["workspace"]["records"]`.

normandy/recipes/tests/test_commands.py
```diff
@@ -522,48 +498,32 @@ def test_publishes_missing_recipes(self, rs_settings, requestsmock):
         # The first two requests should be to get the existing records
```
mythmon: The comments in this section refer to pairs of requests ("the first two") which was true when we were interacting with both collections, but should be updated now to only talk about one request.

normandy/recipes/tests/test_commands.py
```diff
@@ -574,44 +534,31 @@ def test_republishes_outdated_recipes(self, rs_settings, requestsmock):
...
```
mythmon: "two" to "one" here too.

normandy/recipes/tests/test_commands.py
```diff
@@ -622,17 +569,11 @@ def test_unpublishes_extra_recipes(self, rs_settings, requestsmock):
...
```
mythmon: "two" to "one" here too.

mythmon:
bors try

PR #2142, opened 2020-03-17
Add windows version filter object

### LINKED ISSUE #2129: Filter Object: Windows Version NT
"The first is the Windows Version, which has values like 10.0, 6.2, and 6.1. These are the NT versions listed in the "Release version" column of this table on Wikipedia. Note that these are historically a little weird, for example the product released as "Windows 8.1" has an NT version of 6.3. For these we should probably have a list of versions, like we do for locales and countries. This is normandy.os.windowsVersion in JEXL. In UIs we should probably present this as just a list of checkboxes for Windows 7, 8, 8.1, and 10." -@mythmon

uhlissuh:
fixes #2129 

From a slack conversation on this filter object, I took that Mythmon thought the way we currently do the channel (making a database table and populating it with the possible channels) was a good example for doing the windows versions.


normandy/recipes/tests/test_filters.py
```diff
@@ -122,7 +128,7 @@ def test_generates_jexl(self):
         }
 
 
-class TestWindowsBuildNumberFiter(FilterTestsBase):
+class TestWindowsBuildNumberFilter(FilterTestsBase):
     def create_basic_filter(self, value=12345, comparison="equal"):
```
uhlissuh: fixing a typo here.

normandy/recipes/filters.py
```diff
@@ -592,6 +592,50 @@ def to_jexl(self):
...
+    """
+    Match a user based on what windows version they are running. This filter
+    creates jexl that compares the windows NT version.
+
+    .. attribute:: type
+
+        ``windows_version``
+
+    .. attribute:: value
+        string, must be one of the following: 6.1, 6.2, 6.3, 10.0
```
mythmon: Shouldn't this be a number, not a string?

normandy/recipes/models.py
```diff
@@ -45,6 +45,17 @@ def __repr__(self):
         return "<Channel {}>".format(self.slug)
 
 
+class WindowsVersion(models.Model):
+    slug = models.CharField(max_length=255, unique=True)
```
mythmon: To quote [the Django docs](https://docs.djangoproject.com/en/3.0/ref/models/fields/#slugfield):

> Slug is a newspaper term. A slug is a short label for something, containing only letters, numbers, underscores or hyphens. They’re generally used in URLs.

I don't think this field should be named slug. Perhaps `nt_version`, and make it a numeric type instead?

normandy/recipes/management/commands/initial_data.py
```diff
@@ -41,3 +42,16 @@ def add_countries(self):
...
+        self.stdout.write("Adding Windows Versions...", ending="")
+        versions = {
+            "6.1": "Windows 7",
+            "6.2": "Windows 8",
+            "6.3": "Windows 8.1",
+            "10.0": "Windows 10.0",
+        }
+
+        for slug, name in versions.items():
+            WindowsVersion.objects.update_or_create(slug=slug, defaults={"name": name})
```
mythmon: As a stylistic note, not a functional one: I'd use a tuple here instead of a key/value pair, since this isn't really being used as a lookup:

```suggestion
        versions = [
            ("6.1", "Windows 7"),
            ("6.2", "Windows 8"),
            ("6.3", "Windows 8.1"),
            ("10.0", "Windows 10"),
        ]

        for slug, name in versions:
            WindowsVersion.objects.update_or_create(slug=slug, defaults={"name": name})
```

normandy/recipes/tests/test_filters.py
```diff
@@ -149,6 +155,39 @@ def test_generates_jexl_error_on_bad_comparison(self):
...
+            == f"(normandy.os.isWindows && normandy.os.windowsVersion {symbol} 6.1)"
+        )
+
+    def test_generates_jexl_error_on_bad_comparison(self):
+        filter = self.create_basic_filter(comparison="typo")
+        with pytest.raises(serializers.ValidationError):
+            filter.to_jexl()
+
+    def test_generates_jexl_error_on_bad_version(self):
+        with pytest.raises(AssertionError):
```
mythmon: I haven't been able to find it. What is raising this AssertionError? I can only find `ValidationError`s that should be raised from here.

normandy/recipes/models.py
uhlissuh: ok

normandy/recipes/filters.py
uhlissuh: ok 👍

normandy/recipes/tests/test_filters.py
uhlissuh: It's coming from this place in the BaseFilter: https://github.com/mozilla/normandy/blob/master/normandy/recipes/filters.py#L35

the `validate_value` method is causing `ob.is_valid( )` to be false there.

mythmon: Ah, ok. Thanks for showing me that. I also verified that this comes through in HTTP requests nicely:

```shell
$ $ edi recipes create
    -A 'Insecure mcooper@mozilla.com'
    --server local
    --cert ~/src/normandy/etc/ssl/normandy_dev.crt 
	'{"action_id":4,"arguments":"{\"message\":\"test\"}","name":"test","
filter_object":[{"type":"windows_version","value":"asdf","comparison":"less"}]}'

Error: HTTP Error Code 400: {"filter_object":{"0":{"value":["Unrecognized windows version slug 'asdf'"]}}}
```

mythmon:
I noticed that you forgot to add the filter to the `by_type` list in `normandy/recipes/filters.py`. I think we may need to modify our tests to make sure that we are exercising that part of the code as well.

uhlissuh:
bors merge

PR #2143, opened 2020-03-17
Add negate filter object

### LINKED ISSUE #2105: Filter Object: Negate
I think this is the first of a concept of a meta filter object, ie a filter object that takes another filter object as a parameter.  This filter object will take another filter object and just negate it.

uhlissuh:
fixes #2105

### DIFF
--- a/normandy/recipes/filters.py
+++ b/normandy/recipes/filters.py
@@ -636,6 +636,37 @@ class WindowsVersionFilter(BaseComparisonFilter):
         return set()
 
 
+class NegateFilter(BaseFilter):
+    """
+    This filter takes the JSON needed for a filter as an
+    argument,  and this filter will "negate" it.
+
+
+    .. attribute:: type
+
+        ``negate``
+
+    .. attribute:: data_for_filter
+
+        This is a JSON object that would be passed to the filter
+        you want to negate.
+
+       :example: `{ "type": "channel", "channels": ["release", "beta"]}`
+    """
+
+    type = "negate"
+    filter_to_negate = serializers.JSONField()
+
+    def to_jexl(self):
+        filter = from_data(self.initial_data["filter_to_negate"])
+        jexl = filter.to_jexl()
+        return f"!({jexl})"
+
+    @property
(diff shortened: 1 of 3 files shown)


normandy/recipes/tests/test_filters.py
```diff
@@ -206,6 +207,16 @@ def test_throws_error_on_bad_platform(self):
             filter.to_jexl()
 
 
+class TestNegateFilter(FilterTestsBase):
+    def create_basic_filter(self):
+        pref_filter = PrefExistsFilter.create(pref="browser.urlbar.maxRichResults", value=True)
+        return NegateFilter.create(filter=pref_filter)
```
mythmon: This works great as a unit test, but how could users pass this in via JSON? The data for a recipe's filter object needs to be a simple dict, and there isn't anything that is doing the translation from that to an inflated filter.

I expect that the negate filter will need to do some work using (or at least something like) [`filters.from_data`](https://github.com/mozilla/normandy/blob/8849a4ab066c76adc9ff660daeff5cb2effa068c/normandy/recipes/filters.py#L663-L668).

uhlissuh: Oh, yea, thanks, of course the child filter would need to be passed in as a JSON object.

normandy/recipes/filters.py
```diff
@@ -636,6 +636,37 @@ def capabilities(self):
...
+    .. attribute:: type
+
+        ``negate``
+
+    .. attribute:: data_for_filter
+
+        This is a JSON object that would be passed to the filter
+        you want to negate.
+
+       :example: `{ "type": "channel", "channels": ["release", "beta"]}`
```
mythmon: These docs don't match the actual arguments passed. For what it's worth, I prefer `filter_to_negate` over `data_for_filter`.

The phrasing "that would be passed to the filter" is a bit odd too, since the object is a filter, not something that is passed to a filter. Filters aren't really functions or things that take in data. They are descriptions of behavior.

Issue #2145, opened 2020-03-17
Eslint failing in make check


jaredlockhart:
I just got a failure on eslint on master @ 680e2221f8251f66a3bce503c99d4a509f532a9d

```
===============================================================================
FAILED: eslint
===============================================================================
yarn run v1.22.4
$ /app/node_modules/.bin/eslint client/actions/tests/test_show-heartbeat.js client/utils/closest.js client/actions/tests/utils.js client/actions/console-log/index.js client/actions/tests/test_preference-experiment.js client/actions/show-heartbeat/index.js client/actions/opt-out-study/index.js karma.conf.js normandy/base/management/commands/data/console-log.js client/actions/tests/index.js webpack.config.js client/actions/utils.js client/actions/preference-experiment/index.js babel.config.js client/actions/tests/test_opt-out-study.js client/actions/tests/test_console-log.js client/tests/utils.js client/utils/absolute-path.js
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

-------------------------------------------------------------------------------
Additional error output:
-------------------------------------------------------------------------------
TypeError: Cannot read property 'range' of null
Occurred while linting /app/client/actions/tests/test_show-heartbeat.js:395
    at SourceCode.getTokenBefore (/app/node_modules/eslint/lib/source-code/token-store/index.js:298:18)
    at checkSpacingBefore (/app/node_modules/eslint/lib/rules/template-curly-spacing.js:60:42)
    at TemplateElement (/app/node_modules/eslint/lib/rules/template-curly-spacing.js:119:17)
    at listeners.(anonymous function).forEach.listener (/app/node_modules/eslint/lib/linter/safe-emitter.js:45:58)
    at Array.forEach (<anonymous>)
    at Object.emit (/app/node_modules/eslint/lib/linter/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (/app/node_modules/eslint/lib/linter/node-event-generator.js:254:26)
    at NodeEventGenerator.applySelectors (/app/node_modules/eslint/lib/linter/node-event-generator.js:283:22)
    at NodeEventGenerator.enterNode (/app/node_modules/eslint/lib/linter/node-event-generator.js:297:14)
    at CodePathAnalyzer.enterNode (/app/node_modules/eslint/lib/linter/code-path-analysis/code-path-analyzer.js:634:23)
error Command failed with exit code 2.

-------------------------------------------------------------------------------
Completed in: 14.09s
[...]

PR #2148, opened 2020-03-23
Split python and node dependency commands in dev docker

mythmon:
I don't understand why, but this seems to help #2145. I'd like to figure out why, but I'd be happy to merge this without understanding why it helps.


mythmon:
Oh no, somehow I made it worse. Now CI is breaking with the same error.

jaredlockhart:
@mythmon Good idea but I just tried locally and still got the same failure, this version lock might help?

https://github.com/babel/babel-eslint/issues/681#issuecomment-597359735

mythmon:
This isn't the right solution. Going to close this to reduce the noise.

PR #2153, opened 2020-03-25
Simplify JS Infrastructure

### LINKED ISSUE #2149: Simplify JS infrastructure
It used to be that we actively developed Normandy action code in this repo (in JS), and we had a React front end. Because of those we built up decent amount of JS tooling, including tests, linting, and dependencies.

These things are mostly slowly us down at this point. No one is changing the code, so the tests don't matter. No one is working on the code, so the linters are just getting in the way. Every extra dependency is one more that we keep up to date and watch for security problems.

We can't completely remove the JS from this repository, for two reasons. First, we current use NPM to fetch action schemas that are published out of mozilla-central. Second, we still have to serve the built action files for legacy clients that rely on server-hosted actions, at least partially. As of now this is still a decent sized population. Importantly, removing these actions from the repo would severely impact the range of Firefox versions that Normandy is compatible with.

We can do some things though. I propose that we

* Remove the JS tests, and all tooling for them
* Remove the JS linting, and all related tooling
* Remove any unneeded NPM packages
* Remove any usage of Babel. We can write our code to work on the target browsers

mythmon:
This PR removes a lot of JS tooling that we don't need anymore.

Before, we had 2705 NPM packages, totaling 211 MB. After, we have 1274 NPM packages, totaling 39 MB. Considering our Docker image is 1.09 GB, 172 MB is a nice amount to save.

Fixes #2149.


jaredlockhart:
bye bye eslint! 👋 

bors r+

PR #2154, opened 2020-03-25
Move SSL to container volume fixes #2150

### LINKED ISSUE #2150: Store SSL keys for local development inside docker container
Rather than writing the SSL keys back out to the host filesystem (which can cause file permission collisions when we run the tests in circle) we can just store them inside a docker volume.

jaredlockhart:
Small change to the compose file so the SSL files stay in a volume, I don't think it should affect non docker workflows in any way.  And I updated the `make kill` command to a more ergonomic pattern I figured out in experimenter.  Both of these things should make it easier for the new API contract tests to run in circle.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,8 +41,17 @@ code_format:
 	docker-compose run app therapist run --fix --disable-git ./!(node_modules|assets|docs|venv)
 check: check_migrations lint test
 
-kill:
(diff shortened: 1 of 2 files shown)


jaredlockhart:
bors r=tiftran

mythmon:
An intermittent.

> E   AttributeError: 'Locale' object has no attribute '_as_dict'


bors retry

PR #2164, opened 2020-03-31
Add caching and a timeout to x5u fetching for signature verification

### LINKED ISSUE #2161: Add caching and timout for requests in verify_x5u
The heartbeat calls recipe_signatures_are_correct for each signed recipe. This results in multiple calls to verify_x5u which download the PEM file from https://content-signature-2.cdn.mozilla.net. These calls currently have no timeout. This can lead to an outage if that CDN is slow to respond and all app processes become tied up waiting on it. To reduce this risk, we can

1) Cache responses so that we download the PEM once per heartbeat request instead of once per signed recipe
2) Add a timeout to the request for the PEM

mythmon:
CC @sciurus 

Fixes #2161


mythmon:
@sciurus This approach uses Django's built-in cache, which we currently have set up as an in-memory per process cache. That will save us from requesting often in a hot loop, but isn't shared between workers, and is flushed on deploy. Does that sound ok to you, or should we use one of the more robust cache mechanisms?

normandy/recipes/signing.py
```diff
@@ -190,12 +191,18 @@ def verify_x5u(url, expire_early=None):
...
-    req.raise_for_status()
-    pem = req.content.decode()
 
-    der_encoded_certs = extract_certs_from_pem(pem)
-    decoded_certs = [parse_cert_from_der(der) for der in der_encoded_certs]
+    cache_key = f"fetch-x5u-certs::{url}"
+
+    decoded_certs = cache.get(cache_key)
+    if decoded_certs is None:
+        req = requests.get(url, timeout=10)
```
sciurus: Looks like this is called from `recipe_signatures_are_correct`, `action_signatures_are_correct`, and `signatures_use_good_certificates`. Only the latter will actually catch the exception if this times out. The others don't catch any exceptions from requests at all now. This needs fixed so that the heartbeat will still return a response when this fails.

Additionally,  I wonder if you should short-circuit further requests to the same x5u url within a heartbeat request once one has failed. Otherwise in the case of an empty cache during a significant cloudfront outage we risk the heartbeat taking `(signed_recipe + signed_action) * request timeout` , which even if we dropped the timeout from 10 to 1 seconds would cause the same problem we saw yesterday. I don't think there's any benefit to repeating the x5u request for additional recipes or actions, since after the first one fails you want the heartbeat to return a 500, right?

You could probably accomplish this by caching a marker for an error in the request and bailing out of this function if that's present. Ideally this marker would persist only for the life of the hearbeat request, but you could probably accomplish the same effect using django's cache but setting a shorter TTL like 5 seconds.

normandy/recipes/signing.py
```diff
@@ -190,9 +191,15 @@ def verify_x5u(url, expire_early=None):
...
     """
-    req = requests.get(url)
-    req.raise_for_status()
-    pem = req.content.decode()
+
+    cache_key = f"fetch-x5u-pem::{url}"
+
+    pem = cache.get(cache_key)
+    if pem is None:
+        req = requests.get(url, timeout=10)
```
leplatrem: Isn't the returned object a response? (`resp = requests.get()`?)

Also, you don't want the timeout as constant or a setting?

mythmon: Ah, good point.

I thought about making the timeout a setting, but I don't think it's really going to change. A constant would be a good idea.

normandy/recipes/signing.py
sciurus: This timeout is really generous. Under normal conditions this will return in under 100ms. I think it could be okay if we short-circuit for failed requests, but otherwise we should make it lower.

sciurus:
An in-memory, per-process cache sounds fine for this.

normandy/recipes/signing.py
```diff
@@ -190,9 +193,26 @@ def verify_x5u(url, expire_early=None):
...
+    if pem is None:
+        try:
+            resp = requests.get(url, timeout=X5U_REQUEST_TIMEOUT)
+            resp.raise_for_status()
+        except requests.RequestException as exc:
+            cache.set(cache_key, exc, settings.X5U_CACHE_TIME)
+            raise
+
+        pem = resp.content.decode()
+        if settings.X5U_CACHE_TIME:
```
sciurus, line 214: This is to guard against bothering with caching when TTL is 0? If that is helpful here, would it also be helpful at line 210?

normandy/recipes/signing.py
```diff
@@ -190,9 +193,26 @@ def verify_x5u(url, expire_early=None):
...
+
+    if isinstance(pem, Exception):
+        raise CachedRequestException(url=url)
+
+    if pem is None:
+        try:
+            resp = requests.get(url, timeout=X5U_REQUEST_TIMEOUT)
+            resp.raise_for_status()
+        except requests.RequestException as exc:
+            cache.set(cache_key, exc, settings.X5U_CACHE_TIME)
```
sciurus, line 210: I'm nervous about caching failures for as long as we cache successes. This means a transient error or latency spike could cause all heartbeats requests to this process to fail for 10 minutes. It's common to have separate knobs for successes vs errors (e.g. [cloudfront does](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/custom-error-pages-expiration.html)). I'd lean towards making the error's TTL closer to the duration of the heartbeat request; that's currently averaging ~4 seconds.

normandy/recipes/signing.py
mythmon: This was specifically so that tests could totally disable caching (we can't cache mock objects), but yes, it would be helpful on line 210 as well.

normandy/recipes/signing.py
```diff
@@ -190,9 +191,29 @@ def verify_x5u(url, expire_early=None):
...
+        # make a new one.
+        raise CachedRequestException(url=url)
+
+    if pem is None:
+        try:
+            resp = requests.get(url, timeout=settings.X5U_REQUEST_TIMEOUT)
+            resp.raise_for_status()
+        except requests.RequestException as exc:
+            if settings.X5U_CACHE_TIME:
+                cache.set(cache_key, exc, settings.X5U_CACHE_TIME)
```
sciurus, line 211:
```suggestion
            if settings.X5U_ERROR_CACHE_TIME:
                cache.set(cache_key, exc, settings.X5U_ERROR_CACHE_TIME)
```

normandy/recipes/signing.py
```diff
@@ -190,9 +191,29 @@ def verify_x5u(url, expire_early=None):
...
+            raise
+
+        pem = resp.content.decode()
+        if settings.X5U_ERROR_CACHE_TIME:
+            cache.set(cache_key, pem, settings.X5U_ERROR_CACHE_TIME)
```
sciurus, line 216:
```suggestion
        if settings.X5U_CACHE_TIME:
            cache.set(cache_key, pem, settings.X5U_CACHE_TIME)
```

mythmon:
Thanks for the diligence here, I'm glad one of us has it right now.

bors delegate @sciurus

PR #2168, opened 2020-04-02
API contract tests update

chartjes:
* tests can now be run using two Docker containers (one for 'test' and the other for 'app')
* added CircleCI job to run them

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -257,6 +257,27 @@ jobs:
               --junitxml=test-reports/junit.xml \
               --server http://localhost:8000
 
+  containerized-tests:
+    machine:
+      docker_layer_caching: true
+    steps:
+      - run:
+          name: Install Docker ComposeCreate virtualenv
+          command: |
+            set -x
+            sudo bash -c "curl -L https://github.com/docker/compose/releases/download/1.11.2/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose"
+            sudo chmod +x /usr/local/bin/docker-compose
+      - run:
+          name: Docker info
+          command: docker -v
+      - run:
+          name: Docker compose info
+          command: docker-compose -v
+      - checkout
+      - run:
+          name: Run contract tests in containers
+          command: make containerized_tests
+
 workflows:
   version: 2
   main:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -262,11 +262,11 @@ jobs:
       docker_layer_caching: true
     steps:
       - run:
-          name: Install Docker ComposeCreate virtualenv
+          name: Install Docker Compose
           command: |
             set -x
-            sudo bash -c "curl -L https://github.com/docker/compose/releases/download/1.11.2/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose"
-            sudo chmod +x /usr/local/bin/docker-compose
+            bash -c "curl -L https://github.com/docker/compose/releases/download/1.11.2/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose"
+            chmod +x /usr/local/bin/docker-compose
       - run:
           name: Docker info
           command: docker -v
--- a/Makefile
+++ b/Makefile
@@ -72,16 +72,8 @@ generate_deploy_bug: build
 	docker-compose run app python bin/generate_deploy_bug.py $(FROM) $(TO)
 
 # Used for running the contract tests in two containers instead of running locally
-contract_update_actions: build
-	docker-compose run app sh -c "/app/bin/wait-for-it.sh db:5432 -- python manage.py update_actions"
-
-contract_load_initial_data:
-	docker-compose run app sh -c "/app/bin/wait-for-it.sh db:5432 -- python manage.py initial_data"
-
-contract_user:
+create_superuser:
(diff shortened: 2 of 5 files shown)


.circleci/config.yml
```diff
@@ -257,6 +257,27 @@ jobs:
               --junitxml=test-reports/junit.xml \
               --server http://localhost:8000
 
+  containerized-tests:
+    machine:
+      docker_layer_caching: true
+    steps:
+      - run:
+          name: Install Docker ComposeCreate virtualenv
```
mythmon, line 265: Looks like two names got mashed together here.

.circleci/config.yml
```diff
@@ -257,6 +257,27 @@ jobs:
...
+          command: |
+            set -x
+            sudo bash -c "curl -L https://github.com/docker/compose/releases/download/1.11.2/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose"
+            sudo chmod +x /usr/local/bin/docker-compose
```
mythmon, line 269: Is sudo needed here? We should already be running as root.

Makefile
```diff
@@ -70,3 +70,18 @@ refresh: kill migrate load_data
 # Usage: make generate_deploy_bug FROM=fromtag TO=totag
 generate_deploy_bug: build
 	docker-compose run app python bin/generate_deploy_bug.py $(FROM) $(TO)
+
+# Used for running the contract tests in two containers instead of running locally
+contract_update_actions: build
+	docker-compose run app sh -c "/app/bin/wait-for-it.sh db:5432 -- python manage.py update_actions"
+
+contract_load_initial_data:
+	docker-compose run app sh -c "/app/bin/wait-for-it.sh db:5432 -- python manage.py initial_data"
```
mythmon, line 79: These two rules seem very similar to `update_actions` and `load_initial_data`. I expect that this definition would work fine for both rules. Can they be merged?

.circleci/config.yml
```diff
@@ -257,6 +257,27 @@ jobs:
               --junitxml=test-reports/junit.xml \
               --server http://localhost:8000
 
+  containerized-tests:
+    machine:
+      docker_layer_caching: true
```
mythmon, line 262: Since the tests always start with `make kill`, does Docker layer caching actually help us here? It has a non-trivial cost, so I'd prefer to not use if it isn't needed.

Makefile
mythmon, line 79: Can these two rules be merged with the non-contract ones? It seems like they could be made compatible with eachother.

Makefile
```diff
@@ -70,3 +70,18 @@ refresh: kill migrate load_data
...
+
+contract_user:
+	docker-compose run app sh -c "/app/bin/wait-for-it.sh db:5432 -- python manage.py createsuperuser --noinput --email=test-user@example.com --user=testuser"
```
mythmon, line 82: Can you make it clear that this use as a super user?

Makefile
```diff
@@ -70,3 +70,18 @@ refresh: kill migrate load_data
...
+
+run_container_contract_tests:
+	docker-compose run test sh -c "/app/bin/wait-for-it.sh db:5432 -- pytest contract-tests/ --server https://app:8000"
+
+containerized_tests: kill build migrate contract_update_actions contract_load_initial_data contract_user run_container_contract_tests
```
mythmon, line 87: Why have both `run_container_contract_tests` and `containerized_tests`. It seems like this would be simpler if all of the dependencies were listed on `run_container_contract_tests`, or the command from `run_container_contract_tests` was moved here. Less indirection seems helpful in this case.

contract-tests/conftest.py
```diff
@@ -44,6 +44,10 @@ def requests_session(conf):
 def only_readonly(requests_session, conf):
     """Check if the current server is a readonly server, skip if it is not"""
     r = requests_session.get(conf.getoption("server") + "/__version__")
+
+    if r.status_code == 404:
```
mythmon, line 48: The version endpoint is an important part of Normandy, we should make it work in the environments that are under test instead of quietly ignoring it.

.circleci/config.yml
chartjes: Fixed

Makefile
chartjes: Fixed

Makefile
chartjes: Combined them into one

contract-tests/conftest.py
chartjes: Well, those tests always failed when I ran them locally so something needs to change with the current setup to make that __version__ end point work.

.circleci/config.yml
chartjes: Fixed

Makefile
chartjes: Removed and using the existing steps

.circleci/config.yml
chartjes: Set to false

Makefile
chartjes: Replaced with calls to existing steps

.circleci/config.yml
jaredlockhart: @mythmon Yeah, it's two different things.  `make kill` doesn't remove the cached image layers, it just removes the running instances and their volumes.  `docker system prune` removes the cached image layers.  So caching would still help here, but only if you want it to be really fast.

PR #2206, opened 2020-04-14
Fix filter names and make them more consistent.

mythmon:
There are two problems that I found while working on a new filter object.

1. Some of the filters shared a type, which made them impossible to use. This was the main thing I wanted to fix.
2. Some of the types users `camelCase` naming, and others used `snake_case` naming. All of the "original" filter objects were camel case, so I standardized them all that way, and added a test.


mythmon:
bors r= @uhlissuh

PR #2208, opened 2020-04-14
Automatically set needed envvars during tests

mythmon:
This is another one of those "how did this even work" moments. We clearly set up these environment variables for tests at some point, but it doesn't work now, and I can't find a point in recent history where it did work.

Without these changes, a lot of the tests fail for me, especially when I'm using VSCode's Pytest integration. With this change, everything works great.


mythmon:
@tiftran have you had a chance to look at this? I can have someone else review it if it's helpful.

tiftran:
@mythmon  eeek! lemme do get this done today

PR #2210, opened 2020-04-15
Add JEXL filter object

mythmon:
This could be used by NDT to combine with other filter objects, so that we can nest a JEXL filter inside a sticky or OR expression, without having to bail out entirely. It also adds the ability to comment the reasons behind a JEXL expression, and note any additional capabilities that may be needed.

### DIFF
--- a/normandy/recipes/filters.py
+++ b/normandy/recipes/filters.py
@@ -20,9 +20,11 @@ field, so the final JSON would look something like this:
     }
 """
 
-from rest_framework import serializers
-from datetime import datetime
 import json
+from datetime import datetime
+
+from pyjexl import JEXL
+from rest_framework import serializers
 
 
 # If you add a new filter to this file, remember to update the docs too!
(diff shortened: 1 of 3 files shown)


normandy/recipes/filters.py
```diff
@@ -816,6 +818,71 @@ def capabilities(self):
...
+
+        # Add mock transforms for validation. See
+        # https://mozilla.github.io/normandy/user/filters.html#transforms
+        # for a list of what transforms we expect to be available.
+        jexl.add_transform("date", lambda x: x)
+        jexl.add_transform("stableSample", lambda x: x)
+        jexl.add_transform("bucketSample", lambda x: x)
+        jexl.add_transform("preferenceValue", lambda x: x)
+        jexl.add_transform("preferenceIsUserSet", lambda x: x)
+        jexl.add_transform("preferenceExists", lambda x: x)
```
rehandalal, line 873: i guess there's no way to grab the list of transforms and iterate to add these in?

mythmon: This is how we validate JEXL elsewhere as well, I just copy/pasted it. I couldn't really think of a better place to put it, and the other place is the `extra_filter_expression` evaluation, which we'll get rid of eventually. The server doesn't actually know what any of these transforms do, we just have to keep this in sync with the client.

A better thing would be to get information about the available transforms from the Normandy metadata that Firefox is publishing now, but maybe that's for another patch.

rehandalal: > A better thing would be to get information about the available transforms from the Normandy metadata that Firefox is publishing now, but maybe that's for another patch.

+1 would love to not have to manually maintain this.

PR #2233, opened 2020-06-09
Remove JS build step and tooling

### LINKED ISSUE #2232: Remove webpack build step
We currently use webpack to build the action code that lives in this repo for use with older Normandy clients. As this code likely does not need to change I think we can compile these to static files and then remove the build infrastructure. 

Currently our linting is failing over a package flagging yarn audit and maintaining these dependencies seems more work than it's worth.

rehandalal:
Fixes #2232 

r?


mythmon:
Looks like a transient network failure:

> requests.exceptions.ConnectionError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
 
bors retry

Issue #2236, opened 2020-06-15
The Normandy recipe cannot be created if the string preference value is empty


SoftVision-CosminMuntean:
[Affected versions]:
- Chrome 83.0.4103.97
- Firefox release 77;
- Firefox Beta 78.0b7;
- Nightly 79.0a1;

[Affected Platforms]:
- All Windows;
- All macOS;

[Prerequisites]:
- Have access to https://delivery-console.stage.mozaws.net

[Steps to reproduce]:
1. Open the browser and navigate to "https://delivery-console.stage.mozaws.net"
2. Create a multi preference experiment.
3. Fill all the mandatory fields with valid data.
4. Create a branch with a string preference and leave the "Preference Value" field empty.
5. Click the "Save" button.
6. Observe the behavior.

[Expected result]:
- The recipe is successfully saved.

[Actual result]:
- The recipe cannot be saved and the "preferenceValue This field may not be blank." error is displayed.

[Notes]:
- Here is the request as a HAR file: [link](https://drive.google.com/open?id=1AiBuCOgmG68cgOjXDPrZuGgNS9LlLpVs).
- Attached a screen recording with the issue:
![string preference value empty](https://user-images.githubusercontent.com/19226534/84676353-15611480-af36-11ea-92a1-5890cf1d5b24.gif)

mythmon:
As a side note, @SoftVision-CosminMuntean I would suggest using [Normandy  Devtools](https://mozilla-extensions.github.io/normandy-devtools/#/) for future testing. It supports all environments, and it is has a better chance of being  updated than Delivery Console going forward. It is still lacking history support,  but most everything else still works.

shell1:
@mythmon how does this issue impact entering "" in control branch.  Does it mean it doesn't work?  Or just that we can't simulate the change with existing tools?

mythmon:
It means it does not work, we cannot instruct Normandy to set that as a branch value. I've recommended using a dummy branch for this study. Note that Experiment also does not support this.

shell1:
Is a dummy branch what you said where Normandy can group the people and NOT DO ANYTHING to them.  which isn't a true control - in that we aren't enforcing that they are in the state we want.

But in this case it sounds like the best option for gathering a set amount of random users NOT in the treatment to compare against.

mythmon:
> Is a dummy branch what you said where Normandy can group the people and NOT DO ANYTHING to them

[...]

PR #2238, opened 2020-06-16
Enforce argument schemas on model creation

mythmon:
I realized that since we weren't checking schemas at the model level, none of the tests for argument verification checked their schemas. I needed this assurance for the tests I'm writing for #2236, but it's interesting enough to include as a separate PR.


mythmon:
The lint error is an outdated NPM package. I think it was removed in #2233.

bors try

rehandalal:
yarn audit shouldn't be a thing anymore

mythmon:
I guess that the other commit just hadn't merged yet.

bors try

rehandalal:
also you've got some python formatting issues

PR #2259, opened 2020-08-18
Make recipe enable and disable operations non-atomic

mythmon:
This was originally added so that if there was a Remote Settings API error, the database changes would be rolled back. However, this gives too much power to the sync process. Instead we should record what the state should be in the database, and if it doesn't sync immediately, then the background job that periodically re-sync RS and the database will catch it.

This avoid confusion such as the API call to Remote Settings seeming to fail, but Remote Settings actually committed the change. Having the potential for the Remote Settings collection to be "ahead" of the database is more damaging then having to wait for eventual consistency.

### DIFF
--- a/normandy/recipes/models.py
+++ b/normandy/recipes/models.py
@@ -431,7 +431,6 @@ class RecipeRevision(DirtyFieldsMixin, models.Model):
         self.recipe.update_signature()
         self.recipe.save()
 
-    @transaction.atomic
     def enable(self, user, carryover_from=None):
         if self.enabled:
             raise EnabledState.NotActionable("This revision is already enabled.")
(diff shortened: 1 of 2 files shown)


normandy/recipes/models.py
```diff
@@ -431,7 +431,6 @@ def _create_new_enabled_state(self, **kwargs):
         self.recipe.update_signature()
         self.recipe.save()
 
-    @transaction.atomic
```
leplatrem, line 434: So that is enough? 
Shouldn't we catch and log Remote Settings client exceptions?
Django does not rollback DB operations if an error is raised and a 5XX is returned?

mythmon: Exceptions should be caught and logged by our normal error handling mechanisms (I don't know why that didn't happen in the case we saw this morning). I don't think we need any additional error handling besides what we already have.

Django's default behavior is to auto-commit every database call, so there is no request-level transaction that can be rolled back.

leplatrem (approved):

mythmon:
bors r= @leplatrem

PR #2279, opened 2020-10-26
Update flake8 to 3.8.*, and apply new lints

mythmon:
Poetry will allow any versions with semver specifers to flex during package updates, which is good for libraries that follow semver. Apparently flake8 does not. As well as updating flake8 to a new minor version, this adapts our code to two apparently new lints in flake8 3.8:

- `l` is no longer an allowable variable name, because it is ambiguous with I and 1
- f-strings must have placeholders


rehandalal (approved):
bors r+

rehandalal:
bors retry

PR #2288, opened 2020-12-03
Fix that recipes could not be filtered by FO type

mythmon:
I found this while trying to set up automatic namespace allocation. We could filter by all the other filter object fields, but `type` in particular wasn't available. Considering that's one of the most important ones, lets fix that.

### DIFF
--- a/normandy/recipes/tests/api/v3/test_api.py
+++ b/normandy/recipes/tests/api/v3/test_api.py
@@ -227,6 +227,46 @@ class TestRecipeAPI(object):
                 "filter_object": "Filters must be of the format `key1:val1,key2:val2,..."
             }
 
+        def test_list_can_filter_by_filter_object_type(self, api_client):
+            locale = LocaleFactory()
+            recipe1 = RecipeFactory(
+                filter_object=[
+                    filter_objects.BucketSampleFilter.create(
+                        start=100,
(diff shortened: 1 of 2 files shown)


normandy/recipes/tests/api/v3/test_api.py
```diff
@@ -218,6 +218,45 @@ def test_list_can_filter_by_filter_object_fields(self, api_client):
...
 
+        def test_list_can_filter_by_filter_object_type(self, api_client):
+            locale = LocaleFactory()
+            recipe1 = RecipeFactory(
+                filter_object=[
+                    filter_objects.BucketSampleFilter.create(
+                        start=100,
+                        count=200,
+                        total=10_000,
+                        input=["normandy.userId", '"global-v4'],
```
tiftran, line 229: super tiny, there's a `"` missing after v4?

tiftran (approved):
Looks good to me

normandy/recipes/tests/api/v3/test_api.py
mythmon: Ah, good catch. I'll fix that.

mythmon:
bors r= @tiftran

mythmon:
bors r= @tiftran

PR #2289, opened 2020-12-03
Fix that passing an invalid value to the filter_object filter would crash the list view

mythmon:
Requests like `/api/v3/recipe/?filter_object=abc` would crash, because the
expected format is `/api/v3/recipe/?filter_object=abc:def`. The code assumed a
colon would be present without handling the errors.


leplatrem (approved):
Looks good!

mythmon:
bors r= @leplatrem

PR #2293, opened 2020-12-11
Remove unused newrelic dependency

### LINKED ISSUE #2292: Remove newrelic dependency
In a similar vein to #2291, we have a dependency on the `newrelic` package which was originally used to monitor and scale Normandy. It was never active by default, and required activation in deployment by changing the application launch command. To my knowledge we don't use New Relic anymore, and either way I haven't used it in a long time. We should remove that dependency.

@sciurus Does that sound right?

mythmon:
Fixes #2292

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,8 +40,7 @@ USER app
 ENV DJANGO_SETTINGS_MODULE=normandy.settings \
     DJANGO_CONFIGURATION=Production \
     PORT=8000 \
-    CMD_PREFIX="" \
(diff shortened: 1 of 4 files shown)


sciurus (changes requested):
While you're doing this, you'll also want to

* delete newrelic.ini
* remove NEW_RELIC_CONFIG_FILE from the Dockerfiles

sciurus (approved):
bors r+

mythmon:
bors retry

PR #2300, opened 2021-02-01
Add revision metadata to model and API

### LINKED ISSUE #2299: Add revision metadata
Several of the proposed ideas for managing Normandy operations have involved being able to record some extra structured data for revisions. This data is useful for tooling like NDT and reports on Normandy health, but won't be used for recipe execution. It would be valuable to have a place to store data related to revisions without having to add a database field for each one.

I propose that we add a new field to revisions, `metadata` that will be a JSONField, and can be modified by authenticated clients. Notably, this field will be considered on a revision, and can be changed later (unlike most fields which are immutable). Metadata will be optional for all revisions, and will default to an empty object `{}`. Only JSON objects are valid at the top level (not arrays, numbers, or strings).

The motivating use case here is to categorize revisions to mark specific kinds of edits such as fixing an error or a simple pause. This data can then feed into dashboards that can automatically report the health of Normandy operations, like the [Experiments One Pager](https://docs.google.com/document/d/1kDjBp46p6TtL31us-olWLayCIk963KKcQQIEKgGkszM/edit#).

It should be possible to provide metadata when creating recipes, as well as when revising them. Additionally, it should be possible to issue a `PATCH` to revision resource that includes a metadata field that edits the metadata without removing any existing metadata. For example:

Before:
```js
revision = {
  id: 456,
  metadata: {"is_pause": true}
}
```

```
curl -I PATCH https://normandy-admin/api/v3/recipe_revision/456/ \
	-d '{"metadata": {"errors": ["filter expression has a typo"]}}'
```

After:
```js
revision = {
  id: 456,
  metadata: {
	"is_pause": true,
    "errors": ["filter expression has a typo"]
  }
}
```

Finally, it should be possible to filter metadata, similarly to how we can filter on the `filter_object` field:

* Search for pauses
  * `/api/v3/recipe_revision/?metadata=is_pause:true`
* Search for any revision with listed errors
  * `/api/v3/recipe_revision/?metadata=errors:`
* Search for deprecated revisions that are blue (whatever that means)
  * `/api/v3/recipe_revision/?metadata=color:blue,deprecated:true`

mythmon:
This can be used to store unstructured data for tools like NDT. This will let us move faster with new features, like automated operational metrics.

Fixes #2299.


rehandalal (approved):
bors r+

PR #2305, opened 2021-02-12
Update Stage Content Signature root key hash in documentation

mythmon:
The certificate described by this hash has changed, and so we need to update it in our docs.


mythmon:
superceded by #2306

PR #2313, opened 2021-03-24
Use a method instead of a property to determine capabilities for filter objects

mythmon:
The property name "capabilities" collided with the field "capabilities" for the JEXLFilterObject. Coincidentally this was ok, but it certainly was not the intended path. Changing to "get_capabilities" stops the collision, and it is unlikely to collide in the future. However, this points out a serious flaw in this way of definining filter objects.


rehandalal (approved):
bors r+

PR #2333, opened 2021-12-02
Remove check about Remote Settings version

leplatrem:


tiftran (approved):
Looks good to be, but could we get a version for signer(I can create a github issue/bugzilla etc for a formal request)? If so, we could keep this check around, once it's available again

tiftran:
bors merge

tiftran:
bors retry

leplatrem:
> Looks good to be, but could we get a version for signer(I can create a github issue/bugzilla etc for a formal request)? If so, we could keep this check around, if version will be available again

We will probably restore that version field, but the string comparison was dumb (my fault :grimacing:) anyway. 

I think the best way would be to rely no the server version directly, and not just the version of a plugin.

PR #2335, opened 2022-01-17
Remove check about Remote Settings group_check_enabled

leplatrem:
From kinto-dist >= 27, we don't offer the possibility to disable the group check on a particular collection. Hence, there's no need to check it anymore (the field will always be `true`).

Plus, the check in this code was wrong, it was raising `Review was not disabled ...` when the `group_check_enabled` field was `True`


tiftran (approved):
looks good to me

tiftran:
bors merge

mythmon:
bors retry

PR #2342, opened 2022-12-01
Cache the API root endpoint for longer.

smarnach:
The information at the API root endpoints virtually never changes; it's just a list of endpoints in the API  with no dynamic information. At the same time, it's the endpoint we get by far the most requests for. We can reduce the load on the origin servers considerably if we cache this endpoint for a whole day instead of just 30 seconds.

### DIFF
--- a/docs/ops/config.rst
+++ b/docs/ops/config.rst
@@ -243,6 +243,13 @@ in other Django projects.
     set to 0 in non-production environments to ease testing. In production
     environments, setting this value too low can be a denial-of-service risk.
 
+.. envvar:: DJANGO_API_ROOT_CACHE_TIME
+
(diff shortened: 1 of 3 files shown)


smarnach:
Looks like we need to start using a different Poetry installer; sigh. I guess I should do that in a separate PR, but I'm waiting for feedback on this PR first.

tiftran (approved):
lgtm, makes sense since the v1/ endpoint returns the same thing and doesn't change.

I tried to fix up the poetry issue in #2343, contract tests are failing though, since it requires maxmindb and we can't simply just download it via url since it requires licenses now? Was wondering whether it's something we request with SRE?

tiftran:
bors merge

PR #2348, opened 2023-06-16
add middleware to disable introspection query

tiftran:
used https://docs.graphene-python.org/en/v2.1.8/execution/middleware/ as reference

### DIFF
--- a/normandy/schema.py
+++ b/normandy/schema.py
@@ -11,12 +11,21 @@ class NormandyQuery(BaseQuery, RecipesQuery, StudiesQuery, graphene.ObjectType):
 
 class DisableIntrospectionMiddleware:
     """
-    This class hides the introspection.
+    This class hides the introspection. As it is best practice to not allow introspection queries
+    in production. ref: https://docs.graphene-python.org/en/latest/execution/queryvalidation/#disable-introspection
     """
 
     def resolve(self, next, root, info, **kwargs):
-
-        if info.field_name.lower() in ['__schema', '_introspection']:
+        # introspection fields taken from https://graphql.org/learn/introspection/
+        if info.field_name.lower() in [
+            "__Schema",
+            "__Type",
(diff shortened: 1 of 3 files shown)


normandy/schema.py
```diff
@@ -9,4 +9,16 @@ class NormandyQuery(BaseQuery, RecipesQuery, StudiesQuery, graphene.ObjectType):
     pass
 
 
-schema = graphene.Schema(query=NormandyQuery)
+class DisableIntrospectionMiddleware:
+    """
+    This class hides the introspection.
```
leplatrem, line 14: Could you make it explicit why we are doing that? 
Link to bug? A few words maybe?

normandy/schema.py
```diff
@@ -9,4 +9,16 @@ class NormandyQuery(BaseQuery, RecipesQuery, StudiesQuery, graphene.ObjectType):
...
+    """
+
+    def resolve(self, next, root, info, **kwargs):
+
+        if info.field_name.lower() in ['__schema', '_introspection']:
```
leplatrem, line 19: We are looking for private fields here, where do they come from? A link maybe?

leplatrem (approved):
Please add a bit more info about context etc. ;)

normandy/schema.py
tiftran: since there's no bug, i added a link that states its best practice to disable introspection queries. Though, I could great a git issue and reference it?

normandy/schema.py
tiftran: I changed up the list, with a list that is listed from https://graphql.org/learn/introspection/

leplatrem (approved):

normandy/schema.py
leplatrem: It's ok ;)

tiftran:
bors merge

PR #2353, opened 2024-02-15
dependency updates

tiftran:


jaredlockhart (commented):
After discussion with @tiftran and @brennie, we've decided that the safest thing to do is leave Normandy server in its current state while we continue to migrate the last remaining consumers over to Nimbus.  We will be conducting diagnostic health checks on it during this time, and making only the absolute minimum necessary changes to keep it alive before it is fully decomissioned.  Feel free to contact me with any questions or followup.

PR #2355, opened 2024-03-22
add key id for autograph

tiftran:


tiftran:
we would need to update the version for `requests` for the circle ci job to pass....

normandy/settings.py
```diff
@@ -371,6 +371,7 @@ def RAVEN_CONFIG(self):
     AUTOGRAPH_HAWK_SECRET_KEY = values.Value()
     AUTOGRAPH_SIGNATURE_MAX_AGE = values.IntegerValue(60 * 60 * 24 * 7)
     AUTOGRAPH_X5U_CACHE_BUST = values.Value(None)
+    AUTOGRAPH_KEYID = values.Value(None)
```
leplatrem, line 374: nit: I think could safely add `"normandy"` as the default key id value

jaredlockhart (approved):
This all looks correct to me, thank you @tiftran .  Why do we have to update requests?  Your instincts are correct here, we should be changing as little as possible.

normandy/settings.py
leplatrem: makes sense 👌

PR #2358, opened 2024-08-20
Fix CI status by moving docs deployment to GHA

janbrasna:
Currently the CircleCI secrets only have read-only access scope for the GitHub token, being unable to write the built docs to `gh-pages` branch. It's the only failure in the pipeline, and instead of providing write permissions it may be easier to just move the actual docs deployment to GitHub Actions, avoiding any commits to the pages' branch and publishing straight to the deployment env without any underlying git tree being written etc.

This PR keeps the `docs-build` step in CI for consistency in testing, only removing the actual `docs-publish` and creating a workflow here instead.

> [!NOTE]
> To enable Pages publishing via Actions, you may need to to change the [publishing source](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site) in repo settings for this workflow to take over, and also make sure if there are any [env protections](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-deployments/managing-environments-for-deployment) defined they also include "master" and not "gh-pages" only.)

Once you're happy with the trigger-based deploys from master, you can safely remove the `gh-pages` branch completely (which hasn't received updates for more than two years) as it won't be used anymore, and also remove the `tschaub/gh-pages` yarn dev dependency and relevant scripts as all docs will be published from master branch deploys automatically from that point on.


janbrasna:
You can check the sample [run for `b05d1a6`](https://github.com/janbrasna/normandy/actions/runs/10376468566/job/28728493801#step:5:20) ^^

